PR #9, opened 2020-02-19
Torus 4 core editor

darrensiegel:
PR to add the core editor component with floating and fixed toolbar support and one (Image) editing component.

The state of the editor component and the individual editing components isn't complete - for sure - but this will at least serve as a solid starting point.


lib/oli_web/templates/editor_test/index.html.eex
```diff
@@ -0,0 +1,7 @@
+<script type="text/javascript" src="<%= Routes.static_path(@conn, "/js/components.js") %>"></script>
+
+<div id="app"/>
```
bitbldr: We might want to call this id something less generic than app, like `<div id="testEditor"/>` since there could be other react component mounts in the rendered html in the future

assets/src/components/editor/Toolbars.tsx
```diff
@@ -0,0 +1,183 @@
...
+    right: -gutterWidth + 'px',
+    borderRadius: '4px',
+    transition: 'opacity 0.75s',
+  } as any;
+
+  const buttons = collapsed
+    ? []
+    : toolbarItems.map((t) => {
+      if (t.type === 'CommandDesc') {
+        return <ToolbarButton icon={t.icon} command={t.command}/>;
```
bitbldr: toolbarItems.map needs keys. Consider putting <ToolbarButton and <Spacer inside of a <React.Fragment key={...}

lib/oli_web/templates/editor_test/index.html.eex
darrensiegel: Agreed, done.

PR #92, opened 2020-04-13
Enhanced editor

### LINKED ISSUE #40: Basic page editor
An author can edit a page. 

This is for basic page editing - content sections only and the ability to edit the title of the page.

### LINKED ISSUE #50: Author can attach learning objectives to a page
The pages that can be selected should be scoped to only those objectives that are present in the parent module.

### LINKED ISSUE #54: Author can reorder top level items within a page
This has to be accessible.  It cannot simply be drag and drop.

darrensiegel:
This PR completes the 'enhanced resource editor'.  It adds in:

* Ability to reorder resource content items via drag and drop
* Ability to reorder resource content items via accessible keyboard manipulation (SHIFT + arrow keys)
* Ability to attach and detach learning objectives to a resource

Run `mix ecto.reset` after syncing.  The test project with the default resource will then contain three learning objectives that you can test adding / removing. 

Closes #40 
Closes #50 
Closes #54

### DIFF
--- a/assets/css/app.css
+++ b/assets/css/app.css
@@ -155,3 +155,222 @@ nav.navbar .user img {
   overflow: hidden;
   clip: rect(0 0 0 0);
 }
+
+.objectives-editor {
+  color: #666666;
+  padding-bottom: 20px;
+  margin-bottom: 20px;
+  border-bottom: 1px solid #cccccc;
+}
+
+.objectives-editor .learning-objectives-label {
+  margin-top: 10px;
+  margin-bottom: 4px;
+  font-size: .8em;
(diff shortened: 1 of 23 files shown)


LEDThereBeLight:
This looks really great. Some comments - not sure if we want to fix these now or wait:
* Cannot redo a title edit once you undo the change
* Saving an empty title triggers a 500 but does not show an error on the page. all subsequent page edits fail silently
* We might want to change the content block outline to always show on the lefthand side even if there's only one content block - it took me awhile to figure out that I could edit the page without first adding a block

PR #93, opened 2020-04-14
use splitChunks optimization

darrensiegel:
This PR adjusts the webpack plugin to use the `splitChunks` optimization.  This is the current way that webpack supports splitting out vendor (react, slate, etc) and common and entry point code.

This is step one of N for activity integration. This needs to be in place so that our natively implemented activities can be bundled separately from other application code, thus unifying the approach for delivery and authoring for both 'native' and 'third-party' activities.


darrensiegel:
I also cleaned up the two index HTML pages to remove a couple of unused dependencies.

PR #96, opened 2020-04-15
Context refactor

LEDThereBeLight:
I split out the context refactoring from the curriculum changes.

This PR changes flat model list in `oli` app into contexts. Not all functions have been moved to the correct modules, and the interfaces between the contexts hasn't been cleaned up, but I believe this setup offers an improvement over what we had originally.

Primary changes:
- Moved code into five main contexts: Accounts, Authoring, Delivery, Publishing, and Utils
- Remove some generated code not in use to make it easier to see what code we're actually using and what public functions are available in each module
- Refactor a lot of the functions that use multiple contexts to work the same way (`with` macro, provide {:ok} {:error} statuses, etc)

Contexts
image
image
image
image
image

Tests
image


lib/oli/accounts/accounts.ex
```diff
@@ -1,116 +1,34 @@
 defmodule Oli.Accounts do
-  @moduledoc """
```
bitbldr, line 2: Why would we removing all the documentation here? It's very convenient when hovering over the functions for quick reference in other parts of the codebase

LEDThereBeLight: Personally, I feel that the generated docs just make the file harder to read. they all use the same conventions, so the docs don't really tell you anything.

PR #102, opened 2020-04-20
Dashboard

darrensiegel:
Enables Phoenix LiveDashboard via an admin only route at:

/admin/dashboard

This represents the first component in a future suite of admin specific troubleshooting and maintenance tools.


config/config.exs
```diff
@@ -12,6 +12,7 @@ config :oli,
 
 # Configures the endpoint
 config :oli, OliWeb.Endpoint,
+  live_view: [signing_salt: "SECRET_SALT"],
```
bitbldr, line 15: I wonder if we should add a System.get_env("LIVE_VIEW_SALT") || raise to prod.secret.exs config to make sure that we don't forget to actually set this when we take it to production

PR #118, opened 2020-04-29
Course outline

### LINKED ISSUE #64: Student can access course section outline view
## Student Experience

A student LTI launching into a course should land on a page that displays the top-level view of the course contents, including links to the individual pages. 

This top-level view should include a page numbering. 

##  Accessibility

This impl should be accessible. This task should include a deep dive on accessibility and a relay of information found to the rest of the team.

## Mobile First

Ensure that this impl is mobile first.

darrensiegel:
This PR adds support for displaying course outline for instructors and students

It also:
1) Adds a new `DeliveryResolver`
2) Introduces "enrollments" as a relation between users and sections.  The LTI launch logic is adjusted to create / update enrollments on each launch.
3) Splits some logic present into delivery controller into a student and instructor delivery controller, leaving the DeliveryController still there to handle some general functionality. 
4) Introduces `Oli.Timing` module as a way to instrument code for timing and reporting of telemetry.  I use that in both the authoring and delivery resolvers. 
5) I noticed that the logic of publishing a course project wasn't pushing forward all course sections to the new publication - I made a fix here to do that. 

Closes #64


lib/oli_web/controllers/student_delivery_controller.ex
```diff
@@ -0,0 +1,32 @@
...
+          context_id: context_id, pages: overview.pages, title: overview.title, description: overview.description)
+        {:error, _} -> render(conn, "error.html")
+      end
+    else
+      render(conn, "not_authorized.html")
+    end
+
+  end
+
+  def page(_conn, %{"context_id" => _context_id, "revision_slug" => _revision_slug}) do
```
bitbldr, line 27: is this function used anywhere?

darrensiegel: That is the soon to be used handler for viewing a page.  I have the route in place in the router, so I left this in here as a stub.

PR #134, opened 2020-05-01
Activity attempts

rgachuhi:
This PR adds learner activity attempts recording models to torus delivery mode.
Summary:
An "activity-access" record is created for each learner for each activity they launch; and activity here means any page, each embedded activity within a page, non-embedded activity etc. For "graded" activities, an "activity-attempt" record is created automatically or when a learner issues a "start attempt" action. An activity-access can have one or more activity-attempts. Each activity-attempt keeps track of one or more "problem-attempt". Note that problem-attempts are hierarchical. Each problem-attempt tracks one or more "interactions" (interactions here represent any learner input type; checkbox, text field, hot-spot, text-area, dnd location etc.). Interactions track one or more learner "response". Scoring and feedback happens at the "response" level. Those get rolled up all the way to the activity-access grade.


darrensiegel:
Closed, landed portion of this in another PR

PR #145, opened 2020-05-06
fix LiveView session support, update other 1.5 LV facilities

darrensiegel:
I finally tracked down a problem with our LiveView configuration that had prevented us from using session support in LiveViews - the browser pipeline has the protect from forgery plug disabled - which was interfering with CSRF LV session setup. 

I also upgraded our LV setup to match how it ships now in phoenix 1.5:
1. Cleaned up our layouts to reflect the new approach
2. Added in a dependency for showing progress on LV form submits.

### DIFF
--- a/assets/js/app.js
+++ b/assets/js/app.js
@@ -3,30 +3,6 @@
 // its own CSS file.
 import css from "../css/app.css"
 
-// webpack automatically bundles all modules in your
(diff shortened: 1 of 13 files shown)


bitbldr:
this breaks sign-in from LTI, we will need to find another fix for the LTI stuff if we reenable CSRF in dev environment

PR #147, opened 2020-05-07
model spec and parsing, evaluation framework

darrensiegel:
This PR is step one towards activity attempt implementation.  It introduces:

1. Simplified schemas for access, attempt, snapshot and other related entities
2. Implementation of a structured representation of an activity model, including parsing
3. Beginning of an evaluation framework


lib/oli/delivery/evaluation/regex.ex
```diff
@@ -0,0 +1,43 @@
+defmodule Oli.Delivery.Evaluation.Regex do
+
+  alias Oli.Delivery.Attempts.{StudentInput, Result}
+  alias Oli.Activities.Model.{Part, Response}
+
+  @behaviour Oli.Delivery.Evaluation.Evaluator
+
```
bitbldr, line 7: numeric has `@impl Evaluator`, is this necessary here as well?

darrensiegel: added it

PR #164, opened 2020-05-18
Grades export

bitbldr:
Adds the ability to export the gradebook to csv file for a course section. When logged in as an instructor or administrator through LTI, a button will be available to download the current state of the gradebook.

Screen Shot 2020-05-18 at 10 50 30 AM


lib/oli/grading.ex
```diff
@@ -0,0 +1,124 @@
...
+      |> Enum.map(fn {_resource, revision} -> revision end)
+
+    # get students enrolled in the section, filter by role: student
+    students = Sections.list_enrollments(section.context_id)
+      |> Enum.filter(fn e -> e.section_role_id == SectionRoles.get_by_type("student").id end)
+      |> Enum.map(fn e -> e.user end)
+
+    # create a map of all resource accesses, keyed off resource id
+    resource_accesses = Enum.reduce(graded_pages, %{}, fn revision, acc ->
+      user_resource_access_map = Attempts.get_resource_access_for_context(revision.resource_id, section.context_id)
```
darrensiegel, line 89: This is a potential performance issue. Here we are doing a query for resource access records for all users in a context for a given resource - INSIDE of a loop over all graded resources in a course.  For LearnLab this list of graded pages in a course will likely be of length one or two - but eventually large courses could have dozens of graded assessments.  We may want to rework this so that we do all of this in one query (a query for all published resource ids joined with access records).

bitbldr: consolidated to a single database call with multiple joins

PR #168, opened 2020-05-19
Analytics

LEDThereBeLight:
The analytics feature became more of an investigation and design project than I initially expected going into it. There are several places I'd love to improve, but this should provide a solid foundation for improving the analytics over time. It will be easier to see if we're approaching analytics the right way once people start using it.

image

The insights page works by selecting a "view" into the analytics. Currently, all analytics results are performed by activity. I originally set it up to "group by" pages and objectives to show "at a glance" performance for, say, each objective or page. But I found that the analytics calculations we're using now (relative difficulty, eventually correct, first try correct) don't map well to anything that's not aggregated by an activity (or part). 

I believe it would still be valuable to find a way to show "at a glance" performance for each objective and page, but we would need to come up with an appropriate way to aggregate the data. Maybe averaging the data across all of the activities on that page, for example.

Beyond the way that the data is displayed, there are other areas of improvement:
- Add the insights page actions into routing to support back/forward button, URL query parameters with sorting and search filters (we could do this by implementing Insights as a live-routed page in the Router, but when I tried this I couldn't figure out how to get the project plugs we use to authorize project users to work)
- Improve efficiency of queries (there is unnecessary redundancy I'm not super proud of, but I wanted to get the queries working the simplest way I could until we commit to a set of analytics tools we know will be valuable)
- Resource/objective analytics - provide those "at a glance" performance stats across each resource or objective
- Heuristics/suggestions for resources that need the most attention
- More advanced charting (bar charts ranking activities to make it easy to see which ones to focus on)

Issues seen:
- Sorting doesn't work on numbers (only applies to resource/objective views)
- Links in the "activity" analytics section are not set up. Ideally, clicking an activity would bring them somewhere related (like to the resource with the activity in edit mode)


LEDThereBeLight:
There is one unit test that is failing after merging master. I would appreciate someone taking a look to help me figure out what could be causing the issue.

PR #174, opened 2020-05-22
Media Library integration

### LINKED ISSUE #55: Author can include media items from a media library in a page
This work involves:

1. Porting over from Echo the relevant media library UI code
2. Creating a new media library backend that stores media items in cloud storage in a way that they end up being resolvable as https URLs from Torus front end. 
3. It also has to account for browsing.  The media library needs to be able to access and display all media library items for a project in an efficient manner.  As a last resort this should involve the Torus database. We should try using cloud storage APIs for this.
4. Must account for determining references.  For any given media item Torus should be able to inform the user where in the project that media item is being used.

darrensiegel:
This PR adds media library support. This includes the new Torus backend support, S3 storage and a porting of the Echo media library front-end code.

Three top-level S3 buckets now exist `torus-media`, `torus-media-test` and `torus-media-dev` to store media in production, test and dev environments.  Under each bucket a folder for the project slug gets created to scope media files per project. This happens on demand upon the first item upload. 

It introduces the `Oli.Authoring.MediaLibrary` context which allows adding items and accessing existing items using the same set of paging and filtering options that exist in Echo. 

New backend dependencies that it introduces are libraries for MIME type detection and for uploading files into S3. Front end dependencies added are for the processing of `.scss` files. 

Note: this implementation does include support for media item references. 

To run this code you need to:
1. Reset the database to pickup the new `media_items` table
2. Include the AWS access and secret keys to your `oli.env` file. I'll post that in Slack. 

Closes #55


lib/oli/authoring/media_library.ex
```diff
@@ -0,0 +1,269 @@
...
+
+  end
+
+  @doc """
+  Adds an item to the media library for a specific project slug.
+
+  The other supplied parameters are the file name, including extension
+  and the contents of the file as a bitstring.
+
+  An example of the format of the `file_name` param: `untitlted.jpg`
```
bitbldr, line 39: untitled.jpg

lib/oli/authoring/media_library/item_options.ex
```diff
@@ -0,0 +1,45 @@
+defmodule Oli.Authoring.MediaLibrary.ItemOptions do
```
bitbldr, line 1: Really like this pattern for typing and creating structs

darrensiegel:
Addressed review comments - and filed follow up issue for the suggestions for ER2 work

lib/oli/authoring/media_library/item_options.ex
darrensiegel: yeah, i agree

lib/oli/authoring/media_library.ex
darrensiegel: Done

PR #180, opened 2020-05-27
Objective part sync

rgachuhi:
PR to track and updated objective lists tied to activity parts. Implements both front and backend sync. Deletes those objectives who's corresponding parts have been deleted in when authoring activities.


lib/oli/authoring/editing/activity_editor.ex
```diff
@@ -157,10 +158,25 @@ defmodule Oli.Authoring.Editing.ActivityEditor do
...
   # Applies the update to the revision, converting any objective slugs back to ids
   defp update_revision(revision, update) do
+    objectives = if Map.has_key?(update, "objectives"), do: Map.get(update, "objectives"), else: revision.objectives
+    parts = update["content"]["authoring"]["parts"]
+    update = sync_objectives_to_parts(objectives, update, parts)
     {:ok, updated} = Resources.update_revision(revision, update)
     updated
   end
 
+  defp sync_objectives_to_parts(_objectives, update, nil), do: update
```
darrensiegel, line 168: nice usage of a pattern match to catch this possible case of no parts!

PR #187, opened 2020-06-02
Qa

LEDThereBeLight:
Adds QA reviews and warnings to the publish page. 

I tried to prioritize ease of adding new reviews over efficiency, so I took a top-down approach by segmenting reviews into separate "types" which each have a list of "checks" that they run and create warnings from. The reason for the separate reviews table is to be able to determine when a review has been started and whether it is still in progress or completed (when all rows in the reviews table for a project are marked done).

**Reviews types and checks**:
- Pedagogy
  - No attached objectives
  - No attached activities
- Content
  - Broken URIs (images, links)
- Accessibility
  - No alt text (images, youtube)

**Todo**:
- UI
  - Migrate to liveview
  - Add review filtering functionality
  - Warning count badges should match warning icon colors
  - Show when review was last run even if no warnings are generated
- Expand testing
- Efficiency
  - Split reviews out into parallel tasks
  - TSV support for jsonb path queries
- Refactor
  - The `Reviewer` modules all use a common implicit pattern that should be made explicit somehow


lib/oli/publishing.ex
```diff
@@ -6,18 +6,38 @@ defmodule Oli.Publishing do
   alias Oli.Authoring.Course.Project
   alias Oli.Accounts.Author
   alias Oli.Delivery.Sections
+  alias Oli.Resources.{Revision, ResourceType}
   alias Oli.Publishing.{Publication, PublishedResource}
 
+  def query_unpublished_revisions_by_type(project_slug, type) do
```
darrensiegel, line 12: Add comments here describing this function

lib/oli/resources.ex
```diff
@@ -104,6 +104,13 @@ defmodule Oli.Resources do
 
   end
 
+  # returns a list of resource ids that refer to activity references in a page
```
darrensiegel, line 107: This context (Oli.Resources) does not seem to be the correct place for "activity aware" code

PR #253, opened 2020-06-22
(Authoring) Keyboard navigation

LEDThereBeLight:
1) Add a "skip link" that appears when you hit tab after first loading the page. This link allows you to skip past the navigation links on the lefthand side and jump straight to the authoring page content.
2) Fix curriculum view to support tab-navigation
- Tabbing through page items allows you to focus them and press enter to select them and change the settings for that item
- Pressing enter when a page link is focused will navigate to that page
- Focusing a page item allows you to reorder it with shift+up / shift+down
3) Fix objectives view to support tab-navigation
- Similar changes


assets/src/phoenix/app.ts
```diff
@@ -17,7 +17,18 @@ import { initActivityBridge } from './activity_bridge';
...
 const liveSocket
-  = new LiveSocket('/live', Socket, { hooks: Hooks, params: { _csrf_token: csrfToken } });
+  = new LiveSocket('/live', Socket, {
+    hooks: Hooks,
+    params: { _csrf_token: csrfToken },
+    metadata: {
+      keydown: (e: any, el: any) => {
+        return {
+          key: e.key,
+          shiftKey: e.shiftKey,
```
LEDThereBeLight, line 27: This is necessary to trap the "shiftKey" pressed modifier from the javascript event and pass it to liveview actions

PR #265, opened 2020-06-26
Publication diffing

### LINKED ISSUE #221: Publishing feature doesn't detect a change when only a page setting changes
If one simply toggles the type of a resource from "Ungraded" to "Graded" or change the number of attempts, the Publish feature does not detect this as a change and will not allow publication. 

We need to expand its check to include the grading attribute, the scoring strategy and the number of attempts

darrensiegel:
This PR fixes a problem with the CurriculumEditor's approach to editing settings on a resource.  It was not generating a new revision, which allowed it to edit a *published* revision - which is why it wasn't showing up in the Publication diffing. 

I added features to the revision history tool to track publications, as well as showing a revision's author's email instead of just the author id. 

Closes #221


bitbldr:
Publication diffing appears to be fixed. However, when viewing the admin revision history, after a resource has been published, I would expect to see a Currently Published badge. Currently this badge only seems to appear when a new, unpublished revision is created for a resource. If a resource has ever been published, this badge should always be shown to indicate the current state to an admin

PR #274, opened 2020-06-29
Cleanup delivery part 1

bitbldr:
This PR cleans up some of the delivery presentation and adds some functionality for completion tracking for students.

- [x] show a left hand outline of location within course in every page, single next button. Content should come first, then navigation functionality and branding
- [x] see my student progress on index page of delivery
- [x] refactor views so prologue and finalize have outline showing
- [x] clean up Purpose type styles and colors on activities
- [x] remove authoring sidebar from course preview, should be same look and feel as delivery
- [x] fix oli logo font
- [x] tables should have reasonable default formatting, including headers showing up with special styling
- [x] show if a page has been previously viewed

Remaining Tasks for another PR:
- [ ] Show summary of objectives on each page
- [ ] Code block should have some sort of syntax highlighting or special formatting
- [ ] missing "review this attempt" for summative
- [ ] Add outline collapse button
- [ ] Add content outline on right for pages
- [ ] Submitting "multiple times" within an assessment feels like an anti pattern. Formative activities should only show a submit for each activity and not the entire page (this is currently the behavior), summative activities should only show a single submit on a page, not on every activity. Specifically submit and compare activities
- [ ] preview Logo link should not try to redirect to a delivery path, same with logo in default layout


lib/oli_web/templates/layout/page.html.eex
```diff
@@ -0,0 +1,37 @@
...
+        <%= @inner_content %>
+
+        <%= render OliWeb.PageDeliveryView, "_previous_next_nav.html",
+          conn: @conn, context_id: @context_id, summary: @summary, previous_page: @previous_page, next_page: @next_page %>
+
+      </div>
+  </div>
+
+  <script>
+    OLI.initActivityBridge('eventIntercept');
```
darrensiegel, line 33: It seems wrong to init the bridge here, given that this template isn't responsible for the rendering of the 'eventIntercept' DOM element.  Can we move this init back to `lib/oli_web/templates/page_delivery/page.html.eex` where that DOM element is rendered?

bitbldr: Yes, agreed. Fixed in latest commit

PR #325, opened 2020-07-08
UI Rework without styling

LEDThereBeLight:
I'm splitting the authoring styling PR into two so that we can make sure we at least get some of these changes in for the alpha test tomorrow. 

Most of the styling changes will be in a PR with a new "clean" theme so that we can evaluate whether or not we want to use it for Torus authoring. The set of changes in this PR includes most of the stuff that applies across any theme.

- Add more spacing in activity editor
- Fix bug with "project description" being longer than 255 characters: change to "text" field in table
- Add minimal accounts page with ability to change author name
- Add table/card display mode toggle to projects view. The styling is not included in this PR.


darrensiegel:
I saw a couple of problems here that I pushed a commit to address.

First, the buttons for toggling the displaymode weren't actually buttons, which then didn't give the end user the correct experience when hovering over them or clicking them.  I implemented the same approach used in revision history. 

The positioning of the buttons and title were off, and the title redundant in the case of cards. 

Also, the history nav wasn't working correctly - all three params should be handled together for this to work correctly.

That being said, i still saw some weirdness when using the browswer back and forth buttons to navigate.  In certain situations (and this isn't limited to the ProjectsLive view as I saw it in other live views) certain history transitions trigger a full mount-params-render cycle - which was causing a brief pink flash background to appear.  I tracked this down to teh phoexnix.scssd where it applies that style to the `phx-error` class.  I do not understand why this error class exists during a view transition/reload.  This might be a LiveView bug (a similar one was fixed in 0.14.0).  I pulled in the latest LivewDashboard and LviewView release.

PR #388, opened 2020-07-10
Undo redo activities

### LINKED ISSUE #349: Undo/Redo in activities is broken
The action buttons update their enabled state, and when clicked trigger saves - but no changes are actually seen in the content and no changes seem to get saved.

darrensiegel:
Closes #349 

There were two problems that were breaking undo/redo in the activity editors, one minor, and one significant:

1. The objective/part synchronization found in `ActivityEditor.tsx` was introducing extra states in the undo/redo stacks. I reworked this to make it inline to the current update. 

2. The Immer Reducer and React useEffect based approach found in our authoring components for managing state and applying updates is fundamentally flawed in that it cannot easily take into account the fact that the `ActivityEditor` is designed to push down to the authoring components *new* models on Undos and Redos.  These authoring components (short answer and mcq) look at the `props` only once - on their initial render - and then use the state within the reducer as the source of truth for applying changes in dispatched actions.  So they basically never respond to pushed down new states due to undo / redos. 

I racked my brain for a good half hour trying to use a combination of `useEffect` and a special action that could update the state within the reducer when new props were pushed down.  None of the solutions worked as they would fix one aspect but cause other problems, and none of the solutions were elegant.  So I pulled the plug and ripped out the entire Immer Reducer dependency.  It turned out to be fairly straightforward to map the Immer Reducer classes into Action creators that I could then inject an Immer draftState into - and then run the same code that was in the reducer.  

All the unit tests still pass, and I've tested it interactively and everything seems to be working well.


LEDThereBeLight:
:(

PR #414, opened 2020-07-14
Recapture

rgachuhi:
PR adds support for reCAPTCHA to the email user creation form


rgachuhi:
> Looks great and works locally.
> 
> Can you remove the Site Key and private key from the PR?
> 
> All other places where we have keys or secrets we have keep these out of the repository.



> Looks great and works locally.
> 
> Can you remove the Site Key and private key from the PR?
> 
> All other places where we have keys or secrets we have keep these out of the repository.

Actually those keys and secrets are public supplied by google for testing purposes, so no harm in maintaining them on repo. They certainly need to be supplanted with actual key and secret on testing and production.

PR #427, opened 2020-07-17
Display captions in delivery/preview mode

LEDThereBeLight:
* Audio
* Image
* Code
* Table
* Youtube

The accessible solution was to wrap the elements in a `<figure>` and use `<figcaption>` to display the caption. Tables are a little different in that they use a `<caption>` element within the table itself.


LEDThereBeLight:
Address darren's comments

Although now the youtube video defaults to a small size in preview and we don't have a way to change the size

darrensiegel:
It looks like the display of captions for images has been lost.  

Also there is a failing unit tes.t

PR #430, opened 2020-07-17
Expand page preview

bitbldr:
This PR implements activity rendering in page preview and also changes the preview to show inline with the page instead of opening in a new window


bitbldr:
Addressed review comments. Preview now shows a message indicating preview mode. Also adds back in the ability to open preview in another tab using Meta (CMD,Win) modifier key for power users

PR #431, opened 2020-07-20
Cascade page delete activities

rgachuhi:
This PR extends page deletion processing to include marking child activities within a page as deleted whenever the parent page is marked deleted.

### DIFF
--- a/lib/oli/authoring/editing/page_editor.ex
+++ b/lib/oli/authoring/editing/page_editor.ex
@@ -243,7 +243,28 @@ defmodule Oli.Authoring.Editing.PageEditor do
     # Handle the case where this change does not include content
     case Map.get(change, "content") do
 
-      nil -> {revision, []}
+      nil ->
+
+        if Map.get(change, :deleted) do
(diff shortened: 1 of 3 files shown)


lib/oli/authoring/editing/page_editor.ex
```diff
@@ -240,38 +240,46 @@ defmodule Oli.Authoring.Editing.PageEditor do
   # deleted state of the activity revision correct.
   defp resurrect_or_delete_activity_references(revision, change, project_slug) do
 
-    # Handle the case where this change does not include content
-    case Map.get(change, "content") do
-
-      nil -> {revision, []}
+    if Map.get(change, :deleted) do
```
darrensiegel, line 243: `If Map.get(change, :deleted, false) do`  will make this more robust.  This guarantees that the deletion code that follows only runs when the page is being edited to set its "deleted" field to "true".

rgachuhi:
> Can you expand the page_editor_test.exs to add a test that verifies this functionality.

Test verifying the activity has been added

PR #432, opened 2020-07-20
Rearrange last drop area should be bigger

### LINKED ISSUE #401: When rearranging content blocks, the last drop area should be bigger
Just like the last AddResourceContent area in a page, when in drag mode the last Drop area should be bigger as well since users might be inclined to simply drop the item in the lower margin of the page to move it to be last

bitbldr:
This PR expands the last drop area in content drag and drop mode to enable easier rearranging. It also addresses another point made in the alpha test by adding a pseudo button increasing the discoverability of how to add content and activities to a page.

Closes #401

### DIFF
--- a/assets/src/components/content/AddResourceContent.scss
+++ b/assets/src/components/content/AddResourceContent.scss
@@ -1,6 +1,8 @@
 @import 'authoring/variables';
+@import 'authoring/mixins';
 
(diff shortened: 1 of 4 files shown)


darrensiegel:
This looks fantastic.

Issue #179, opened 2020-05-26
Improve media library experience


darrensiegel:
There are other enhancements we should probably consider tackling in ER2 such as:

- [x] Finally adding other Media type filters to the left sidebar under "All Media" or removing it
- [x] Restyling to fix inconsistencies and swap Insert and Cancel buttons for consistency
- [ ] Show upload progress bar(s)
- [ ] Maintain sort by preference when media window is closed/reopened
- [ ] Allow multiple selection of media items
- [ ] Consider consolidating media insertion to single button that encompasses images, videos, audio instead of having separate buttons

gjbunyea:
Suggestion: Ability to remove media from the library. 
If someone uploads the wrong image (for example), there's no way to remove that from the project.

czerwinskie:
Revisited on 8-7.
First step is to access the media library regardless of actions needed (just be able to browse).
project-based media library at the moment; access the project-wide media library and being able to manage media there.

tanvidomadia:
From slack thread: good to have alt text for an image that lives in the media library but that can be overridden in a particular page.

PR #435, opened 2020-07-21
Media Library Improvements

### LINKED ISSUE #382: Media library - submitting unsupported filetypes gives no indication of failure

### LINKED ISSUE #384: Media library - inserting with no selection causes crash

LEDThereBeLight:
Address a series of bugs and UX comments related to the media library:
- Fix a bug where the media controller was being given an invalid pattern on error conditions
- Fix a bug where a media query parameter was named "urlFilter" on the server and "pathFilter" on the client. This was leading to general weirdness with media queries
- Fix a bug where opening the media manager to change the audio file caused a crash
- Handle errors and show banner to user
- Disable the insertion button when no media item is selected. This implementation was a bit of a hack-y workaround but was necessary get working to prevent several other bugs.
- Correctly handle "all" mimetype when fetching all media on server
- Expand allowed audio filetypes
- Enable filters to display media types. This is to alleviate feedback from the alpha test that users were uploading files and not knowing what happened to them (they were uploading but didn't show up in the filtered media manager view). I also added a "video" section ("managed by youtube") to suggest to users that videos exist outside of the media library workflow, but could go either way on this.
- Improve some of the styling (it can still use work, but it's a step in the right direction)
- Add `controls` property to audio html editor to make it display in the page

Closes #384 
Closes #382
Makes progress toward #179


lib/oli/rendering/content/html.ex
```diff
@@ -67,7 +67,7 @@ defmodule Oli.Rendering.Content.Html do
   end
 
   def audio(%Context{} = _context, _, %{"src" => src} = attrs) do
-    wrap_with_figure(attrs, ["<audio src=\"", src, "\">Your browser does not support the
+    wrap_with_figure(attrs, ["<audio controls src=\"", src, "\">Your browser does not support the
```
LEDThereBeLight, line 70: Make audio element visible

darrensiegel:
Wow, nice work. There are some really nice improvements here.

PR #441, opened 2020-07-21
Locked page message

rgachuhi:
PR to post banner on top of editor indicating which author has current edit lock on the page. The banner only shows up to other users attempting/viewing the same page


rgachuhi:
Redoing this on a separate branch applying suggested feedback

PR #444, opened 2020-07-22
hide submit button

rgachuhi:
On delivery mode; disables the "Submit Assessment" button if the assessment page has zero activities


lib/oli_web/controllers/page_delivery_controller.ex
```diff
@@ -107,6 +107,7 @@ defmodule OliWeb.PageDeliveryController do
       next_page: context.next_page,
       title: context.page.title,
       graded: context.page.graded,
+      activity_count: context.activity_count,
```
darrensiegel, line 110: Replace this with `activity_count: map_size(context.activities),`

rgachuhi: Replacement done

PR #535, opened 2020-07-28
Refactor themes

### LINKED ISSUE #532: Themes should be refactored for easier extensibility
Right now a theme requires you to define all styles for every theme. We should have a base theme for both authoring and delivery and custom themes should override those. Themes should also be compiled automatically by webpack and official themes be stored in source control

bitbldr:
This PR addresses a couple of issues related to theming:
- Refactors theming approach to extend from base authoring and delivery themes. This was always the original intent but the previous implementation didn't work quite right so this addresses that issue. From now on when developing a UI component, styles can be declared under assets/styles/authoring or as a sibling .scss file to react components within assets/src/** and will propagate to all themes
- "official" themes are now stored in source control with the rest of the codebase
- Sass file imports are no longer required inside of .tsx files that use the styles. All scss files under src will be compiled with the base theme. This is a required change in order to allow these files to use the proper scss variables for the active theme
- Dark theme is now automatically used when dark mode is enabled on the user agent in both authoring and delivery
- Color scheme was cleaned up for a cleaner more professional look. These colors were chosen based on some previous conversations with the dev team. There were also some theme-specific custom styles added to fix any remaining style issues and oddities

**NOTICE**: Requires a `mix ecto.reset` to re-seed the database

Risk: Moderate. Though this change is mostly stylesheet related, it covers a lot of code
Stability: Tested on Chrome, Safari and Firefox and appears to be stable without issues

Closes #532


bitbldr:
requested changes have been addressed. Authoring now has a default Automatic option for using system setting as well as Light and Dark options. I will create the wiki page requested outlining the theming approach

PR #541, opened 2020-07-29
Add MathJax latex rendering to preview

LEDThereBeLight:
**Issue**: Latex was not properly rendering in page preview mode.
**Cause**: When the `ResourceEditor` swaps to preview mode, the MathJax delivery script is re-initialized with a new window variable but not executed to compile the Latex and render it. It works when opening the preview in a new tab, but not when it swaps out the current editor view.
**Solution**: This is more of a workaround than anything -- I would like to find a way to get the script to execute when loading the delivery file in preview mode -- but I added the MathJax script to the authoring template so that it can run in the author. I wrapped the authoring HTML in a div with an `ignore` class name, and then white listed the `delivery` preview mode HTML for MathJax to render it.
**Risk**: Moderate. This change is to the high level authoring/delivery templates, and could cause a problem. I haven't seen any issues, but the templates are used everywhere.

### DIFF
--- a/lib/oli_web/templates/layout/authoring.html.leex
+++ b/lib/oli_web/templates/layout/authoring.html.leex
@@ -44,6 +44,10 @@
         ignoreHtmlClass: 'tex2jax_ignore',
         processHtmlClass: 'tex2jax_process'
       },
+      // Disable MathJax by default. Startup must be called explicitly to run MathJax transformations.
+      startup: {
+        ready: () => {}
+      },
       loader: {
         load: ['[tex]/noerrors']
       }
@@ -56,9 +60,7 @@
   </head>
   <body>
 
-    <div class="tex2jax_ignore">
-      <%= Map.get(assigns, :inner_layout) || @inner_content %>
-    </div>
+    <%= Map.get(assigns, :inner_layout) || @inner_content %>
 
     <script>
     window.$ = $;
--- a/lib/oli_web/templates/layout/authoring.html.leex
+++ b/lib/oli_web/templates/layout/authoring.html.leex
@@ -32,12 +32,33 @@
     <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/10.1.1/styles/atom-one-light.min.css" integrity="sha512-11xYl5MU0/AMaYnuBOXDDQdZnl5WGtLVidxqa0XUAXYf6cGnJShNpgtbmB/0MW6ypeev+9Bwj7I/J8wexX8ePw==" crossorigin="anonymous" />
     <script src="//cdnjs.cloudflare.com/ajax/libs/highlight.js/10.1.1/highlight.min.js"></script>
 
+    <script>
+    window.MathJax = {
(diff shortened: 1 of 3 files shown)


lib/oli_web/templates/layout/authoring.html.leex
```diff
@@ -32,12 +32,33 @@
     <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/10.1.1/styles/atom-one-light.min.css" integrity="sha512-11xYl5MU0/AMaYnuBOXDDQdZnl5WGtLVidxqa0XUAXYf6cGnJShNpgtbmB/0MW6ypeev+9Bwj7I/J8wexX8ePw==" crossorigin="anonymous" />
     <script src="//cdnjs.cloudflare.com/ajax/libs/highlight.js/10.1.1/highlight.min.js"></script>
 
+    <script>
```
LEDThereBeLight, line 35: Add MathJax to authoring template for use anywhere in torus authoring mode

lib/oli_web/templates/layout/authoring.html.leex
```diff
@@ -32,12 +32,33 @@
...
+
     <%= csrf_meta_tag() %>
     <script type="text/javascript" src="<%= Routes.static_path(@conn, "/js/app.js") %>"></script>
   </head>
   <body>
 
-    <%= Map.get(assigns, :inner_layout) || @inner_content %>
+    <div class="tex2jax_ignore">
+      <%= Map.get(assigns, :inner_layout) || @inner_content %>
+    </div>
```
LEDThereBeLight, line 61: Add `ignore` class name to tell MathJax not to compile authoring files by default

lib/oli_web/templates/layout/delivery.html.eex
```diff
@@ -31,7 +31,7 @@
 
     <script src="https://polyfill.io/v3/polyfill.min.js?features=es6"></script>
     <script>
-    window.MathJax = {
+    window.MathJax = window.MathJax || {
```
LEDThereBeLight, line 34: Tell MathJax not to re-initialize itself (which happens if swapping from authoring to delivery in preview mode)

lib/oli_web/templates/resource/page_preview.html.eex
```diff
@@ -7,10 +7,11 @@
 
 <%= render OliWeb.PageDeliveryView, "_objectives.html", objectives: @objectives %>
 
-<div id="eventIntercept">
+<div id="eventIntercept" class="tex2jax_process">
```
LEDThereBeLight, line 10: Tell mathjax to process this element and initialize the typesetting script (this is required for dynamic content since MathJax runs on page render)

lib/oli_web/templates/layout/authoring.html.leex
bitbldr: This is a bit concerning wrapping everything in authoring in a new div. Is there any way we can disable everywhere except where explicitly set?

PR #571, opened 2020-08-11
fix datashop input parsing logic

LEDThereBeLight:
This PR fixes one of two issues with datashop exports: improperly parsing hints / student input when logging attempts.

### DIFF
--- a/lib/oli/analytics/datashop/elements/event_descriptor.ex
+++ b/lib/oli/analytics/datashop/elements/event_descriptor.ex
@@ -44,26 +44,20 @@ defmodule Oli.Analytics.Datashop.Elements.EventDescriptor do
 
           case activity_type(part_attempt) do
             # for short answer questions, the input is the text the student entered in the field
-            "oli_short_answer" -> input
+            "oli_short_answer" ->
+              input
               |> Utils.parse_content
             # for multiple choice questions, the input is a string id that refers to the selected choice
             "oli_multiple_choice" ->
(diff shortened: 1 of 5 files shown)


lib/oli/analytics/datashop/elements/skills.ex
```diff
@@ -14,19 +14,20 @@ defmodule Oli.Analytics.Datashop.Elements.Skills do
   require Logger
 
   def setup(%{ publication: publication, skill_ids: skill_ids }) do
-    skill_ids
-    |> Enum.map(& element(:skill, [make_skill_element(publication, &1)]))
-    |> Enum.filter(& &1 != nil)
-  end
+    objective_revs = Publishing.get_published_revisions(publication.id, skill_ids)
```
LEDThereBeLight, line 17: Get all skill revisions in a single query for efficiency

lib/oli/analytics/datashop/elements/event_descriptor.ex
```diff
@@ -44,26 +44,20 @@ defmodule Oli.Analytics.Datashop.Elements.EventDescriptor do
 
           case activity_type(part_attempt) do
             # for short answer questions, the input is the text the student entered in the field
-            "oli_short_answer" -> input
+            "oli_short_answer" ->
```
darrensiegel, line 47: Platform code should not make any assumptions about the details of an activity's implementation.  We should provide a mechanism in the activity to specify what this datashop logic needs - rather than have this logic embed assumptions about the activity impl

PR #579, opened 2020-09-01
Page editor changes

LEDThereBeLight:
Hovering (formatting) toolbar
- Move some insertion buttons (such as lists, titles, quotes) into the formatting toolbar, allowing toggling
- Show active formatting on the current selection when the toolbar is open
- Add "precondition" support to enable/disable buttons when the selection does not allow that content type
- Fix bug with multiple formatting options not being supported (e.g. bold, italic on same text)
- Fix bug with toolbar not showing when selection spans multiple formatting options
- Fix bug with toolbar remaining open when selecting another content block

Fixed (insertion) toolbar
- Move toolbar to align with cursor when inside an empty paragraph instead of sitting above the content block
- Fix bug with losing selection when toggling text to title

Editing
- Wrap content editing blocks inside a "card" to allow them to be dragged
- Change slate content model to allow less combinations - most blocks now do not support nesting other block-level items. For example, you cannot add a codeblock inside a list item
- Add markdown support for titles, lists, blockquotes, and code blocks
- Add max-width to content editor/preview
- Move caption editing from options menu to below the content item

Images
- Add rendering support for alt text
- Add "float" option to align images

Links
- Change link editor to automatically add protocol

Code block
- Fix crash when deleting code block

Blockquote
- Fix crash when deleting blockquote

Lists
- Fix bug with not being able to delete lists at beginning of page
- Fix crash when toggling nested lists

Tables
- Fix bug with tables resizing when editing content in table cells
- Fix bug with table cell options resizing the cell when shown
- Move "remove table" button to table cell options
- Change "size picker" styling/size when inserting a table

Youtube
- Add query parameters to simplify what's shown in the embedded youtube video
- Make youtube video not playable in edit mode, opens options instead

Other
- Fix a variety of console warnings (outdated packages and missing react keys)
- Remove `package.lock`, preferring yarn

Known issues:
- Caption editing in activities loses focus (this seems to be a problem with how activities are updated, updates cause a page refresh)


bitbldr (changes requested):
Style changes break main page and login page

Screen Shot 2020-09-03 at 9 15 32 AM

bitbldr (commented):
These two insert arrows look too identical, users might get them confused even though they serve two different purposes

Screen Shot 2020-09-03 at 9 20 12 AM

bitbldr (changes requested):
First off, these editor changes are really slick. Overall, I think this is a HUGE improvement to the editing experience. The functionality appears to be solid after testing, I just have a couple of recommendations, most of which are stylistic and only two of which need to be addressed (# 1, maybe # 2). I'll let you decide which of the remaining ones you want to fix/file a follow up issue for:

1. The "Section" terminology might be a little misleading and collide with the organization hierarchy vocabulary like we discussed. Also, numbering these 1,2,3... etc doesn't seem to make sense to me since there are other blocks dispersed, the can be rearranged and ordinality of these doesn't really add any value to author or user, and in-fact might be a bit confusing.

An interesting way around this issue could be to simply put here whatever block type is selected on the right. i.e., leave it blank for "None", Example when "Example", or Learn by doing when "Learn by doing". Also, Learn by doing is an important OLI term, which we should probably keep in place instead of Learn more, which is often used to provide related-software information and doesn't accurately depict the intent of the learning material type.
Screen Shot 2020-09-08 at 2 18 15 PM

2. Does this new functionality retain the level of keyboard accessibility that the old one provided? I wasnt able to access the new insertion menu with only the keyboard

3. Different insertion buttons should probably be styled a bit differently to indicate that they provide different functionality. I think the item insertion could be smaller, or a inverted color scheme (white on black) but something to differentiate
Screen Shot 2020-09-08 at 2 13 44 PM

4. Markdown support is AWESOME. But I wasn't able to tell which markdown was supported, it seems like maybe only headers using # as the first character are supported? I think we should provide some indication to users that it exists- maybe tooltips that describes the function and markdown shortcut key (e.g. ) when hovered over -or- simply a markdown icon in the lower corner like github has which when clicked on shows the markdown keyboard shortcuts
Screen Shot 2020-09-08 at 2 31 56 PM

5. I encountered a strange issue when resizing the window where the insert button overlapped text. Maybe a relative positioning issue?
Screen Shot 2020-09-08 at 2 01 11 PM

6. It would be nice if caption placeholder text was easier to differentiate than actual set captions. Maybe a lighter font color or some sort of background color styling? Also, placeholder text should disappear as soon as the cursor is placed in the text box
Screen Shot 2020-09-08 at 2 21 00 PM

LEDThereBeLight:
> 1. The "Section" terminology might be a little misleading and collide with the organization hierarchy vocabulary like we discussed. Also, numbering these 1,2,3... etc doesn't seem to make sense to me since there are other blocks dispersed, the can be rearranged and ordinality of these doesn't really add any value to author or user, and in-fact might be a bit confusing.

I removed the section label and just moved the purpose type dropdown to the left. I didn't see a reason to label it twice.

> 2. I had trouble deleting a table once I created one, even selecting surrounding paragraphs and pressing delete would leave me with a single table cell.

This is a slate problem with tables, I "hid" the issue by preventing deletion of tables except by using the cell dropdown menu. I found a solution that makes deletion work if you highlight across an entire table (from the paragraph before to the paragraph after)

> 3. Does this new functionality retain the level of keyboard accessibility that the old one provided? I wasn't able to access the new insertion menu with only the keyboard

What I realized after testing the editor is that we actually didn't have keyboard accessibility for the old editor either. You can tab into the fixed toolbar, but then you lose your editor selection and you can't actually use any of the buttons. The formatting/hovering toolbar also cannot be tabbed into.

This PR doesn't add keyboard accessibility, but it's something we should make a ticket for.

> 4. Different insertion buttons should probably be styled a bit differently to indicate that they provide different functionality. I think the item insertion could be smaller, or a inverted color scheme (white on black) but something to differentiate

I agree, but I'm not sure how to handle it exactly, so I didn't make this change.

> 5. Markdown support is AWESOME. But I wasn't able to tell which markdown was supported, it seems like maybe only headers using # as the first character are supported? I think we should provide some indication to users that it exists- maybe tooltips that describes the function and markdown shortcut key (e.g. ) when hovered over -or- simply a markdown icon in the lower corner like github has which when clicked on shows the markdown keyboard shortcuts

I added bare-bones tooltips on the toolbar buttons, there is room for improvement though. In short, the shortcuts are:
```
  #: h1
  ##: h2
  *: ul
  -: ul
  +: ul
  1): ol
  1.: ol
  >: blockquote
  ```: code
```

> 6. I encountered a strange issue when resizing the window where the insert button overlapped text. Maybe a relative positioning issue?

Fixed, it was relative to the wrong div.

> 7. It would be nice if caption placeholder text was easier to differentiate than actual set captions. Maybe a lighter font color or some sort of background color styling? Also, placeholder text should disappear as soon as the cursor is placed in the text box

Changed styling.

bitbldr (approved):
Great work. This represents a huge improvement to the editor experience.

PR #600, opened 2020-10-09
Curriculum refactor

LEDThereBeLight:
Change curriculum page based on https://docs.google.com/presentation/d/1bbADcItf9IPWi1OgKv3u-sNKADZwZOI5RpKskuHg5iY/edit

Changes:
* Add "current editor" to resources locked by an author (view a page in another tab while looking at the curriculum to trigger it)
* Break dependencies on `root_container`
* Remove settings from right-hand side
* Add components for two views: Simple and Learning Summary
* Add ability to create new container
* Add container-specific routing
* Change `Numbering` calculations a bit to fit the view's use cases

Notes:
* I started with the table layout from the ppt but my mind kept wanting to sort and reorder the columns, but this isn't possible because the curriculum has a fixed ordering. So I took some creative liberty and made the resource entries into cards
* I refactored the `broadcast` module somewhat -- this totally wasn't necessary but came about as I was trying to understand how it all worked
* Fixed a separate bug from the original view with reordering curriculum items not refreshing the page


lib/oli_web/live/curriculum/entries/entry.ex
```diff
@@ -0,0 +1,135 @@
...
+                  <%= live_component @socket, LearningSummaryEntry, assigns %>
+                <% end %>
+              </div>
+            </div>
+          </div>
+        </div>
+      </div>
+    </div>
+
+    <%= if @modal_shown do %>
```
bitbldr, line 81: I think using phx to control these modals is the wrong approach here. We should instead simply render the html for any modal on a page and leverage bootstrap javascript to show/hide them whether thats using the data-toggle attribute or javascript, just as you would on a static html page. We can still use a reusable modal live component if we want, but remove the `if @model_shown` guard and `show` class and add the data-toggle and data-target to the button

bitbldr (changes requested):
This is a big improvement, there's just a couple of issues I see which need addressed

PR #619, opened 2020-10-20
Curriculum refactor + Breadcrumbs

LEDThereBeLight:
This PR adds curriculum changes and breadcrumb trails based on [this slide deck](https://docs.google.com/presentation/d/1bbADcItf9IPWi1OgKv3u-sNKADZwZOI5RpKskuHg5iY/edit) and [this implementation strategy](https://docs.google.com/document/d/12c57lT9NbrMv3uMIaD2hT1F-FmwlQVWtgRyWrAamTCk/edit#). The container dropdown menu is not supported - there's some discussion needed around that first.

**Note**: requires a `mix deps.get`

**Curriculum**: see [this pull request](https://github.com/Simon-Initiative/oli-torus/pull/600). Comments are addressed.

**Breadcrumb trail**

_How to review_:
- Review code for:
  - `lib/oli_web/live/breadcrumb/`
  - `lib/oli_web/live/curriculum/entries/`
  - `lib/oli_web/live/curriculum/settings/`
  - `lib/oli/resources/numbering.ex`
  - `lib/oli/utils/breadcrumb.ex`
  - Other relevant code is in `lib/oli_web/live/curriculum/container/`, but much of it is the same from the `Curriculum Refactor` PR
- Go to curriculum
  - Add new graded page
    - Click into page, make sure it works
    - Go back to curriculum using the breadcrumb trail at the top
    - Edit settings for page using the curriculum view
    - Reload page with settings menu open, make sure it persists
  - Add new ungraded page
  - Add new container
    - Go into container, make sure the container name displays in the breadcrumb
    - Add a second-level container and then a third-level container
      - Go into third-level container, make sure the breadcrumb trail names at the top are "shortened"
  - Change container name using settings menu
  - Make sure deletion works from settings menu for both pages and containers


darrensiegel:
Here are some comments from first strictly looking at the implementation and doing some testing:

The “Simple” and “Learning Summary” options buttons are not visually consistent with other option toggling UI approaches in Torus.  Change these to match the UI approach for the different views found int he “Projects” page (i.e. make them small buttons with icons and with tooltips that display on hover detailing what they do)

The prototype specified three views, “Simple” “Details” and “learning summary”, but it looks like this UI has combined those three distinct views into just “simple” and “learning summary”.  I believe strongly that the “Simple” view - one that only displays the resource title and icon has to exist, and has to be the default.   The “Details” view should show the title, icon and the creation/update information.  The “Learning Summary” view should show the title, icon and the objective and activity information (but not the creation/update information)

Currently the “Unit” or “Module” and numbering only displays in the breadcrumb. It is not displaying in the actual curriculum view. It should display in both places. 

There seems to be a problem with Drag and Drop reordering.  I created a child group (a unit) and then within that new unit I created a page and a group (module).  I cannot reorder these two items.  

After I created a small hierarchy (a unit with a module and page in it) and failed to reorder the module and page, I went back to the root of my curriculum and now I see the module appearing at the root alongside the unit. This only happens when I attempt to reorder - so something related to reordering is still pinned to the root_resource

LEDThereBeLight:
Re. above:

1. I reworked the views to split them out into "Simple," "Details," and "Learning summary." However, I cannot get the tooltips to work for the life of me. I spent a fair amount of time trying to get them to show up originally, and then switched over to the text labels because I could not trigger them.

2. I added the numbering prefixes to the curriculum entries to match what shows up in the breadcrumb trail.

3. You are right, there was a reference to the root container in the reordering.

darrensiegel (changes requested):
Overall this is a fantastic implementation.  

I have just a few observations and requests for changes:

1. When I visit the root curriculum, the route that I see in the URL bar is: 
http://localhost:4000/project/example_open_and_free_course/curriculum/curriculum
Can we eliminate that extra “curriculum”?

2. If I type in a non-existent route (http://localhost:4000/project/example_open_and_free_course/curriculum/fake) I get a nil.children/0 is undefined error in the container_editor.  Can we gracefully handle non-existent containers and route that to a generic "Resource not found" page?


Oli.Utils.Breadcrumb
———————————
This is a nice and efficient implementation. 

“Oli.Utils” seems like the wrong place for this module. A module in Oli.Utils should not depend on Oli.Publishing or Oli.Resources like this currently does. 

Can you add comments to the public methods in this module describing what they do, what inputs they take and what is the output?


The live_modal construct
—————————
This is really great.  A really clean way to orchestrate a modal in live view.

darrensiegel (approved):

PR #626, opened 2020-10-26
Course ingest

darrensiegel:
This PR adds baseline support for course digest archive ingestion. 

It adds an admin facing navigation item called "Ingest" that takes the user to a view where they can upload a course digest archive file to turn that into a Torus package.

Right now the functionality is limited to processing only a project hierarchy and practice pages. 

More work is needed in the construction of the digest files to actually yield usable courses.  But for now, one can import the `course.zip` file found in `test/oli/authoring` and Torus will create a navigable course from it.  Many of the pages, however, won't render correctly.  Again, that is a limitation of digest construction, not ingestion.

Also, this PR moves to Elixir 1.11. It changes all the Elixir, Erlang and Phoenix dependencies appropriately, including in the GitHub actions and workflows.  There were a handful of new compiler warnings that the migration to 1.11 introduced.  This PR fixes those as well. The only other change that is needed I believe is a rebuild and push of the Docker image that the phoenix-builder action uses to push that image forward to use Elixir 1.11 and OTP 23.1


lib/oli/authoring/ingest.ex
```diff
@@ -0,0 +1,199 @@
...
+  alias Oli.Repo
+  alias Oli.Publishing.ChangeTracker
+
+  @project_key "_project"
+  @hierarchy_key "_hierarchy"
+  @media_key "_media-manifest"
+
+  @doc """
+  Ingest a course digest archive that is sitting on the file system
+  and turn it into a course project.  Gives the author specied access
```
bitbldr, line 12: typo "specified"

lib/oli_web/controllers/ingest_controller.ex
```diff
@@ -0,0 +1,23 @@
...
+
+  def upload(conn, %{"upload" => upload}) do
+
+    author = conn.assigns[:current_author]
+
+    path_upload = upload["digest"]
+
+    case Oli.Authoring.Ingest.ingest(path_upload.path, author) do
+      {:ok, project} -> redirect(conn, to: Routes.project_path(conn, :overview, project))
+      e -> render(conn, "error.html", title: "Ingest", error: e)
```
bitbldr, line 17: based on the documentation for ingest, I would expect this error case to be {:error, e}. If that is true, then error.html template will render that struct instead of the inner error message

bitbldr (changes requested):

bitbldr (approved):

PR #675, opened 2020-11-18
Full grade sync

darrensiegel:
Adds a "Grades Management" UI that instructors can use to update line items, sync grades, and export gradebook.

The UI is LiveView driven and allows the serial execution of a collection of tasks related to creating, updating line items and posting scores.  After completion of each "task" a progress bar in the UI is updated.

A key change that was also made was to introduce a prefix to the resourceId attribute that we are populating in the LMS created line item.  This helps to guarantee that another, non-Torus tool populating line items and setting resource_ids does not conflict with Torus. 

This feature uses LTI Names and Roles Provisioning service to determine all currently enrolled students so that we do not attempt to post scores for students that have been removed from the LMS section. 

A key limitation of the grade sync feature is that there is no way for it to know if a grade has been overridden in the LMS by an instructor (at least based on Canvas and the LTI_AGS testing I did).  Therefore, I changed this feature page specific and to include a warning regarding loss of custom grade adjustments. 

To access this new "Grades Management" UI, launch in from the LMS as an instructor.  On the Overview page at the top-right you will see a "Grades Management" link.

### DIFF
--- a/lib/oli/lti/lti_nrps.ex
+++ b/lib/oli/lti/lti_nrps.ex
@@ -1,0 +1,74 @@
+defmodule Oli.Lti.LTI_NRPS do
+
+  @moduledoc """
+  Implementation of LTI Names and Roles Provisioning Service
+
+  For information on the standard, see:
+  https://www.imsglobal.org/spec/lti-nrps/v2p0
+
+  This module contains no dependencies on attempts, resources, or any other delivery specific construct.
+  """
+
+  @lti_nrps_claim_url "https://purl.imsglobal.org/spec/lti-nrps/claim/namesroleservice"
+  @context_memberships_url_key "context_memberships_url"
+
+  alias Oli.Lti.Membership
+  alias Oli.Lti_1p3.AccessToken
+
+  require Logger
+
+  defp to_membership(raw) do
+    %Membership{
+      status: Map.get(raw, "status"),
+      name: Map.get(raw, "name"),
+      picture: Map.get(raw, "picture"),
+      given_name: Map.get(raw, "given_name"),
+      middle_name: Map.get(raw, "middle_name"),
+      family_name: Map.get(raw, "family_name"),
+      email: Map.get(raw, "email"),
+      user_id: Map.get(raw, "user_id"),
+      roles: Map.get(raw, "roles"),
+    }
+  end
+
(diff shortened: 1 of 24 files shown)


bitbldr:
Looks really great, I had a few things I wanted to bring up but may not necessarily be issues.

1. Trying to submit an assessment as an administrator/instructor results in a Not Authorized error which could be confusing
Screen Shot 2020-11-20 at 3 00 45 PM

2. When I took an assessment as a student and got a "0%" or all incorrect, the score still shows up as a dash "-" as if I never took the assessment. I'm not sure if this is a canvas thing or a bug on our side
Screen Shot 2020-11-22 at 4 48 57 PM
 
Screen Shot 2020-11-22 at 4 53 48 PM

3. I accidentally left/reloaded the page after initiating a Grade Sync (then subsequently noticing the "do not leave this page" message). It left the gradebook in a strange state. Is it possible to store all the LTI details we need for grade passback an kick this process off in process outside of the live view so that if someone leaves the page this doesn't happen?
Screen Shot 2020-11-22 at 5 16 45 PM

lib/oli/lti/lti_nrps.ex
```diff
@@ -0,0 +1,74 @@
+defmodule Oli.Lti.LTI_NRPS do
```
bitbldr, line 1: Should this service live in the existing Oli.Lti_1p3 context? i.e. Oli.Lti_1p3.Nrps

darrensiegel: My take on this was the `Oli.Lti_1p3` is a context for LTI v1.3 launching.  The NRPS and AGS aren't 1.3 specific.

darrensiegel:
Addressed the review comments:
1) I removed the restriction that was preventing instructors from submitting graded pages
2) I tracked down the bug that was causing 0% assessments from not showing up in the gradebook. This was a bug in the "best" scoring strategy implementation wrt how it calculates the "out_of" value.  I added a unit test assertion that exposes the bug and made the fix that allows the updated test to succeed.

darrensiegel:
Also adjusted the `submission_test.exs` since the change in the scoring strategy resulted in a different score being selected here as "the best".  Since in this test both parts were 100%, i adjusted it to avoid that state making the part 2 answer not 100%.

bitbldr (approved):

PR #677, opened 2020-11-19
Author invite

rgachuhi:
This PR adds support for new author invite (to a particular course package) and fires an email to notify the user about the invite, including a link to allow the user to accept the invite. For already existing users, an invite to collaborate on a project will now also trigger an email to notify them that they have been invited as authors into a course project.


lib/oli_web/templates/pow_invitation/invitation/edit.html.eex
```diff
@@ -0,0 +1,51 @@
...
+        placeholder: "Confirm New Password", required: true, autofocus: focusHelper(f, :password_confirmation) %>
+    <%= label f, :password_confirmation, "Confirm Password", class: "control-label" %>
+    <%= error_tag f, :password_confirmation %>
+</div>
+
+<%= if assigns[:link_account] do %>
+<%= hidden_input f, :link_account, value: @link_account %>
+    <% end %>
+
+<%= submit "Submit", class: "btn btn-md btn-primary btn-block" %>
```
bitbldr, line 48: I think it would make it clearer about the difference between Submit and Sign in if we rename Submit to something like Create Account

lib/oli_web/templates/pow_invitation/mailer/invitation_new_user.html.eex
```diff
@@ -0,0 +1,3 @@
+<%= content_tag(:h3, "Hi,") %>
```
bitbldr, line 1: maybe consider changing the heading to something like "Collaborator Invitation" so that a user can see a preview of what the email is related to in their inbox preview, similar to the Confirm Password email

bitbldr (changes requested):

bitbldr:
Implementation looks great, I had just a few comments. This also raises the concern that we need to at some point offer the ability for authors to opt-out of these types of notifications. AWS SES makes it clear that this can have a huge impact on our email domain and we dont want to get flagged as a spamming account or lose our good email rating.

config/test.exs
```diff
@@ -17,6 +17,9 @@ config :oli, Oli.Repo,
 config :oli, Oli.Mailer,
   adapter: Bamboo.TestAdapter
 
+config :oli, OliWeb.Pow.Mailer,
+  adapter: Bamboo.LocalAdapter
```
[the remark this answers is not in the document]
rgachuhi: LocalAdapter switched to TestAdapter

lib/oli/authoring/collaborators.ex
```diff
@@ -51,22 +92,54 @@ defmodule Oli.Authoring.Collaborators do
...
+    invited_by = Pow.Plug.current_user(conn)
+
+    url = case view do
+      :invitation_new_user -> token = PowInvitation.Plug.sign_invitation_token(conn, user)
+                              Routes.pow_invitation_invitation_path(conn, :edit, token)
+      :invitation_existing_user -> Routes.project_path(conn, :overview, project.slug)
+    end
+
+    invited_by_user_id = Map.get(invited_by, invited_by.__struct__.pow_user_id_field())
+    email = Mail.new(
```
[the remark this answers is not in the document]
rgachuhi: Function moved. That resolve the layout issue. Thanks

lib/oli_web/templates/pow_invitation/invitation/edit.html.eex
rgachuhi: Button text switched to "Create Account"

lib/oli_web/templates/pow_invitation/mailer/invitation_new_user.html.eex
rgachuhi: Done

bitbldr (approved):
Looks great

PR #681, opened 2020-11-24
Check All That Apply

LEDThereBeLight:
Check all that apply activity type.

Goals:
- Simple mode (one correct response and one incorrect response)
- Targeted mode to support imported courses (simple mode + a list of targeted responses that match specific answer choices)

A CATA is either a SimpleCATA or a TargetedCATA.

**Model**:
image

Responses are evaluated according to rules. I went through a couple different implementations using the idea of "combinations" from Echo's CATA question type, but the code ended up becoming too complicated, so I landed on the idea of:
 - One correct response
 - A list of targeted responses that match specific choice combinations.
 - One (catch-all) incorrect response, which is the negation of the correct response and the targeted responses

**Rules**:
image

Opportunities:
- Partial credit support (this would not be too much more, but I didn't want to add partial credit only to a single question type)
- Custom scoring (specific scores for targeted feedback combinations)
- Styling (I'm not super happy with how similar it looks to the multiple choice question, it's not very clear that you can select multiple)


bitbldr (approved):
Great work, this is a really exciting addition. I submitted a separate PR to fix the issue with attempt submission missing lti_params that is broken on master as well.

PR #722, opened 2021-01-06
Ordering Activity

LEDThereBeLight:
Ordering activity type

This was from before the break, for some reason I hadn't made the PR for it. I think I was fixing a bug related to activity tests in another branch and missed making the PR for this once it was completed. But I tested it again today and it still seems to work like it did before after merging new changes from master.


assets/src/components/activities/ordering/OrderingDelivery.tsx
```diff
@@ -0,0 +1,210 @@
...
+      <button
+        className="btn btn-primary mt-2 float-right"
+        disabled={isEvaluated || selected.length !== choices.length}
+        onClick={onSubmit}>
+        Submit
+      </button>
+    );
+
+  return (
+    <div className={`activity multiple-choice-activity ${isEvaluated ? 'evaluated' : ''}`}>
```
darrensiegel, line 183: multiple-choice-activity reference here should be changed to be ordering activity specific

assets/src/components/activities/ordering/utils.ts
```diff
@@ -0,0 +1,149 @@
...
+        ],
+        hints: [
+          fromText(''),
+          fromText(''),
+          fromText(''),
+        ],
+      }],
+      correct: [[choice1.id, choice2.id], correctResponse.id],
+      // FIX THIS CHOICE ID LIST
+      incorrect: [[], incorrectResponse.id],
```
darrensiegel, line 117: I noticed that there is no entry in the schema for this "incorrect" attribute.

darrensiegel (changes requested):
Two comments, see inline.  Minor things I thing, but can you address them?

Nice work!

assets/src/components/activities/ordering/utils.ts
darrensiegel: Further investigation: it looks like "incorrect" attribute isn't used anywhere

darrensiegel (approved):

PR #723, opened 2021-01-06
Add external media by url

### LINKED ISSUE #197: Add the ability to add images by URL rather than only through media library
- What are the tradeoffs?

LEDThereBeLight:
Two parts:

- [x] Add the ability to paste images into structured content editors
- [x] Add the ability to embed images directly by URL in the media manager modal

Closes #197

### DIFF
--- a/assets/src/components/editing/editor/handlers/paste.ts
+++ b/assets/src/components/editing/editor/handlers/paste.ts
@@ -18,4 +18,4 @@ export const onPaste = (editor: SlateEditor, e: React.ClipboardEvent<HTMLDivElem
     .map(URL.createObjectURL)
     .forEach(url =>
       Transforms.insertNodes(editor, ContentModel.image(url)));
-}
+};
--- a/assets/src/components/editing/editor/handlers/paste.ts
+++ b/assets/src/components/editing/editor/handlers/paste.ts
@@ -1,21 +1,33 @@
 import { Transforms, Editor as SlateEditor } from 'slate';
 import * as ContentModel from 'data/content/model';
+import { uploadFiles } from 'components/media/manager/upload';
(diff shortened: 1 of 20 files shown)


assets/src/components/editing/editor/handlers/paste.ts
```diff
@@ -0,0 +1,21 @@
...
+  // The clipboard item 'type' attr is a mime-type. look for image/xxx.
+  // 'Rich' images e.g. from google docs do not work.
+  const images = [...e.clipboardData.items].filter(({ type }) => type.includes('image/'));
+  if (images.length === 0) {
+    return;
+  }
+
+  return images
+    .map(image => image.getAsFile())
+    .map(URL.createObjectURL)
```
darrensiegel, line 18: This doesn't seem to be working as you intended.  Here is what I tried:

1. I copied into my clipboard a file that I have on my local machine (from Finder, selected the file and right-clicked "Copy")
2. I pasted into the content of a page, the image rendered correctly
3. I left the page and returned and I see a "broken image" placeholder.  

It looks like this is what it serialized in the content:

```%{"caption" => "x", "children" => [%{"text" => ""}], "display" => "block", "id" => "3200919815", "src" => "blob:https://localhost/cfb87685-f843-49a3-8ae9-ed3fa185e45b", "type" => "img"}```

darrensiegel: For this to work correctly, I believe a step will need to be added that uploads the pasted image into the media library, and then puts that URL in for the "src"

darrensiegel (changes requested):
See my comment regarding pasting

darrensiegel (approved):

PR #749, opened 2021-01-07
Freshdesk tickets

rgachuhi:
This PR adds "help request" support feature to the system.

- Implements a UI to capture user help requests in both authoring and delivery mode
- It default to sending help request to pre-configured help email but also allows swapping out this default implementation with custom help-desk support.
- Includes a custom implementation Freshdesk; default for Torus production.
- Uses Freshdesk API to generate ticket from users "help requests"

Note: Configuring this for production requires and API key from Freshdesk

Pending:

- Unify delivery and authoring mode view implementations
- Accommodate LTI view when LTI login fails


lib/oli/help/providers/freshdesk_help.ex
```diff
@@ -0,0 +1,53 @@
...
+    {"Content-type", "application/json"},
+    {"Accept", "application/json"},
+    {"Authorization", "Basic " <> Base.encode64(System.get_env("FRESHDESK_API_KEY", "examplekey"))}
+  ]
+
+  @impl Oli.Help.Dispatcher
+  def dispatch(%HelpContent{} = contents) do
+    url = System.get_env("FRESHDESK_API_URL", "example.edu")
+
+    {:ok, body} =
```
darrensiegel, line 16: It would be more robust here to handle {:error, error} clause and return an {:error, message}, as opposed to letting this process die.

lib/oli_web/controllers/help_delivery_controller.ex
```diff
@@ -0,0 +1,91 @@
...
+  end
+
+  def create(conn, params) do
+    with {:ok, :true} <- validate_recapture(Map.get(params, "g-recaptcha-response")),
+         {:ok, content_params} <- additional_help_context(conn, params),
+         {:ok, help_content} <- HelpContent.parse(content_params),
+         {:ok, _} <- Oli.Help.Dispatcher.dispatch(Application.fetch_env!(:oli, :help)[:dispatcher], help_content)
+      do
+      conn
+      |> put_flash(:ok, "Your help request has been successfully submitted")
```
darrensiegel, line 24: Should be `put_flash(:info, ` instead of `put_flash(:ok `

darrensiegel (changes requested):
Fantastic work, just a couple of small things that I saw to address.

darrensiegel (approved):

darrensiegel (approved):

PR #768, opened 2021-01-13
remove institution_id fkey

### LINKED ISSUE #767: fkey links to institution from users should be eliminated
There's no need for this association and it add the opportunity for multiple sources of truth. Links to institution should simply be maintained by the user ⟶ enrollment ⟶ section ⟶ institution relationship.

bitbldr:
This PR removes the unnecessary link from user to institution.

Closes #767


bitbldr:
Before the section is created, the user institution key is actually used to create the section. Closing

PR #827, opened 2021-03-10
KTH image coding

andersweinstein:
Initial working implementation of image coding activity used in KTH programming course, with authoring, delivery and client-side evaluation of graded exercises.


darrensiegel (approved):
This looks great.  We need to do actual Delivery mode testing of this activity, but for now we can land this checkpoint.

PR #840, opened 2021-03-15
Activity scope

rgachuhi:
This PR adds the following features

- New "globally_available" field in the activity_registration table that helps track whether an activity is by default available for use in all course packages or whether it is private
- New table tracking package to activity mapping
- New UI in package overview that allows authors to opt into "private" activities
- New admin only UI that allows admins to promote or demote activities to global status

### DIFF
--- a/lib/oli/activities.ex
+++ b/lib/oli/activities.ex
@@ -134,6 +134,17 @@ defmodule Oli.Activities do
     Enum.sort_by(activities_enabled, & &1.global, :desc)
   end
 
+  def set_global_status(activity_slug, status) do
+    with {:ok, activity_registration} <-
+           get_registration_by_slug(activity_slug)
+           |> trap_nil("An activity with that slug was not found.")
+      do
+      update_registration(activity_registration, %{globally_available: status})
+    else
+      {:error, {message}} -> {:error, message}
+    end
+  end
+
   def get_registration_by_slug(slug) do
     Repo.one(from p in ActivityRegistration, where: p.slug == ^slug)
   end
--- a/lib/oli/activities.ex
+++ b/lib/oli/activities.ex
@@ -10,6 +10,7 @@ defmodule Oli.Activities do
   alias Oli.Activities.ActivityRegistration
   alias Oli.Authoring.Course
   alias Oli.Activities.ActivityRegistrationProject
+  alias Oli.Activities.ActivityMapEntry
   import Oli.Utils
 
   def register_activity(%Manifest{} = manifest) do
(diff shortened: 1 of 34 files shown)


lib/oli_web/controllers/project_controller.ex
```diff
@@ -102,22 +121,27 @@ defmodule OliWeb.ProjectController do
   end
 
   def insights(conn, _project_params) do
-    render conn, "insights.html", breadcrumbs: [Breadcrumb.new(%{full_title: "Insights"})], active: :insights
+    render(conn, "insights.html",
```
darrensiegel, line 124: There seems to be a lot of "auto formatted" changes here. We should synchronize across all of the engineering team the standard that we are formatting to and the tools that we are using to do that formatting, otherwise we will face a lot of these types of diffs during PR review.

lib/oli/activities.ex
```diff
@@ -7,37 +7,146 @@ defmodule Oli.Activities do
...
-
   end
 
   def create_registered_activity_map() do
     list_activity_registrations()
+    |> Enum.map(&Oli.Activities.ActivityMapEntry.from_registration/1)
+    |> Enum.reduce(%{}, fn e, m -> Map.put(m, e.slug, e) end)
+  end
+
+  def create_registered_activity_map(project_slug) do
```
darrensiegel, line 41: The new functions in this module should have documentation added.

darrensiegel (changes requested):
This looks really nice, simple and understandable UIs.  

I made some comments inline that need to be addressed.  The auto-formatting of the source in this PR is going to be problematic until we can all synchronize on the same tool and approach. @eliknebel Can you track down what the recommended Elixir source formatting style and tooling is and share that out in the Slack so we can all get aligned on one approach.  

Also, there is a functional problem in that once I had created an activity of a particular type in my course, if I then remove access to that activity type I see an "Unsupported Activity" in the page editor.  Which activities are available should only control the popup list of activities  - it should not affect the ability to preview and edit existing activities. See screen shot below:

Screen Shot 2021-03-15 at 2 14 11 PM

rgachuhi:
> This looks really nice, simple and understandable UIs.
> 
> I made some comments inline that need to be addressed. The auto-formatting of the source in this PR is going to be problematic until we can all synchronize on the same tool and approach. @eliknebel Can you track down what the recommended Elixir source formatting style and tooling is and share that out in the Slack so we can all get aligned on one approach.
> 
> Also, there is a functional problem in that once I had created an activity of a particular type in my course, if I then remove access to that activity type I see an "Unsupported Activity" in the page editor. Which activities are available should only control the popup list of activities - it should not affect the ability to preview and edit existing activities. See screen shot below:
> 
> Screen Shot 2021-03-15 at 2 14 11 PM

rgachuhi:
> This looks really nice, simple and understandable UIs.
> 
> I made some comments inline that need to be addressed. The auto-formatting of the source in this PR is going to be problematic until we can all synchronize on the same tool and approach. @eliknebel Can you track down what the recommended Elixir source formatting style and tooling is and share that out in the Slack so we can all get aligned on one approach.
> 
> Also, there is a functional problem in that once I had created an activity of a particular type in my course, if I then remove access to that activity type I see an "Unsupported Activity" in the page editor. Which activities are available should only control the popup list of activities - it should not affect the ability to preview and edit existing activities. See screen shot below:
> 
> Screen Shot 2021-03-15 at 2 14 11 PM

Interesting I thought I tested for this. Will dig into it.

darrensiegel (approved):

PR #841, opened 2021-03-16
KTH table problem support

andersweinstein:
This adds support for the table programming problems using csv files with grading by matching text output to an author-specified regular expression. With this addition, the activity should be able to support every problem in the KTH course. 

This makes effectively two types of coding problems: image problems, which are graded by comparing solution-code generated image to the user-code generated image with a specified tolerance; and all others (including but not limited to table problems), which are graded by matching a regular expression to the student code's text output. Any problem that uploads an image when authoring is considered an image problem. The authoring allows uploading either images or spreadsheets (csv) but not both.

### DIFF
--- a/assets/src/components/activities/image_coding/actions.ts
+++ b/assets/src/components/activities/image_coding/actions.ts
@@ -39,9 +39,17 @@ export class ICActions {
     };
   }
 
-  static addImageURL(value: string) {
+  static addResourceURL(value: string) {
     return (draftState: ImageCodingModelSchema) => {
-      draftState.resourceURLs.push(value);
+      if (draftState.resourceURLs.indexOf(value) === -1) {
+        draftState.resourceURLs.push(value);
+      }
+    };
+  }
+
+  static removeResourceURL(value: string) {
+    return (draftState: ImageCodingModelSchema) => {
+      draftState.resourceURLs = draftState.resourceURLs.filter(url => url !== value);
     };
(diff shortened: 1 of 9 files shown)


assets/src/components/activities/image_coding/actions.ts
```diff
@@ -41,7 +41,7 @@ export class ICActions {
 
   static addImageURL(value: string) {
     return (draftState: ImageCodingModelSchema) => {
-      draftState.imageURLs.push(value);
+      draftState.resourceURLs.push(value);
```
darrensiegel, line 44: While are you representing the collection of resources as an array, we actually want set semantics here and want to prevent duplicates.

darrensiegel (changes requested):
See me inline comment about preventing duplications.

Also, for the list of Resources that a user adds to their image coding activity, we need to expose a way to delete each entry.  A simple UI approach here is to render an "X" button next to the end of each line.  A cleaner approach is to place the list of resources in a Bootstrap ListGroup, with an "X" button right aligned.  

https://getbootstrap.com/docs/5.0/components/list-group/

assets/src/components/activities/image_coding/actions.ts
andersweinstein: OK, these should be addressed now.

darrensiegel (approved):

PR #846, opened 2021-03-22
Clone project

LEDThereBeLight:
Add the ability to duplicate/clone a project on the project overview page. The user will be redirected to the duplicated project.


lib/oli_web/controllers/project_controller.ex
```diff
@@ -173,4 +174,18 @@ defmodule OliWeb.ProjectController do
     )
   end
 
+  def clone_project(conn, _project_params) do
+    case Clone.clone_project(conn.assigns.project.slug, conn.assigns.current_author) do
+      {:ok, project} ->
+        IO.inspect(project)
```
eliknebel, line 180: Remove IO.inspect

eliknebel (changes requested):

lib/oli_web/controllers/project_controller.ex
```diff
@@ -173,4 +174,18 @@ defmodule OliWeb.ProjectController do
...
+        conn
+        |> put_flash(:info, "Project copied. You've been redirected to your new project.")
```
eliknebel, line 182: The language here should probably be consistent with the operation name on the button that was clicked. For example, "Project duplicated. You are now viewing your new project.". Project copied terminology might indicate it has been copied to the clipboard or something and further steps are required

eliknebel:
Acquiring a lock on a resource in the old course shows up in the duplicated course, but it should probably be a separate lock now
Screen Shot 2021-03-22 at 1 24 10 PM

eliknebel:
Using the admin Revision History tool, it appears a page resource in the original project seems to still be tracking changes in the duplicated project as well

Screen Shot 2021-03-22 at 1 33 19 PM

eliknebel (approved):

PR #868, opened 2021-03-26
Edit page warning after publish/duplication

### LINKED ISSUE #849: Trying to edit a page while the project is concurrently published results in save error
**Describe the bug**
While editing a page, if another user publishes the project an error is displayed.

**To Reproduce**
1. Go to a page as user A
2. In another browser, Publish the project as user B
3. Try to make some more changes as user A, error message is shown

**Expected behavior**
A user should be able to continue to work on a page and a new revision should transparently be created.

**Screenshots**
Screen Shot 2021-03-22 at 1 09 40 PM

 - OS: macOS 11.2.3
 - Browser Chrome
 - Version 89

LEDThereBeLight:
The main bug fix is to show a warning to the user to refresh the page after losing the lock on a page/activity after a project is published or duplicated. 

closes #849 

**Main fixes**
- Fix return types and specs for locking functions in `Locks.ex`, `ActivityEditor.ex`, `PageEditor.ex` (was not handling `:lock_not_acquired` properly)
- Properly handle `:lock_not_acquired` in `ResourceController`, and show new banner messages depending on the type of edit failure
- Wrap `Publishing.publish_project()` in a `Repo.transaction` and change the methods to fail on a transaction failure. Previously, records could fail silently and the transaction would continue
- Fix the `RevisionHistory` module to sort publications from newest to oldest. Previously, the **"currently published"** publication was incorrect because the sorting was backwards
- Add more `publishing_test` tests to cover more edit sequence combinations

**Smaller Fixes**
- Change all `resource_mappings` to the new name of `published_resources` in the codebase
- Fix several other lint issues


lib/oli/authoring/locks.ex
```diff
@@ -159,27 +169,27 @@ defmodule Oli.Authoring.Locks do
     from(pr in PublishedResource,
       where: pr.publication_id == ^publication_id and not is_nil(pr.locked_by_id),
       select: pr)
-    |> Repo.update_all(set: [locked_by_id: nil, lock_updated_at: nil])
+    |> Repo.update_all(set: [locked_by_id: nil])
```
eliknebel, line 172: I'm not sure if its useful to the system, but it would seem more fitting to me to actually set the `lock_updated_at` here to either the current datetime (since technically setting to null is updating it) or setting it to null if it really is not useful information. Retaining the last `lock_updated_at` value though seems a bit inconsistent to me

lib/oli/authoring/course.ex
```diff
@@ -96,20 +101,14 @@ defmodule Oli.Authoring.Course do
           resource: resource,
           resource_revision: resource_revision,
           publication: publication,
-          resource_mapping: resource_mapping,
+          published_resource: published_resource,
```
eliknebel, line 104: Nice, glad we got arounds to renaming mapping variables to published_resource

eliknebel:
Does this close #849?

eliknebel (approved):

PR #870, opened 2021-03-29
Fix editor media item deletion / webpage embed

### LINKED ISSUE #832: Delete key does not delete currently selected media in authoring
**Describe the bug**
For a currently selected YouTube video or image, in the page editor pressing the "Delete" key should remove that element from the page. 

**To Reproduce**
1. Go to the page editor and insert an image
2. Click on the image to make it "selected".  It should have a light blue border
3. Click the "delete" key to attempt to remove the image
4. It does not remove

LEDThereBeLight:
Fix deleting media items in editors when using the backspace/delete keys. This ended up being a simple fix for as long as it took to figure out what the problem was. Slate void nodes must have a `content-editable=false` node INSIDE them to work properly. The `content-editable` attribute apparently can't be on the void node itself.

Closes #832 

**Fixes**
- Handle keyboard deletion for `image`, `youtube`, `audio`, `webpage` elements
- Fix several errors on webpage/iframe embeds (normalizer failing, deletion, extra newline not being inserted after iframes) that had to do with `iframe` not being a `schema` type recognizable to slate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 ### Bug fixes
   - Fix an issue where slug creation allowed some non-alphanumeric chars
   - Fix an issue where publishing or duplicating courses would cause save errors in page and activity editors
+  - Fix keyboard deletion with media items
+  - Add extra newline after an iframe/webpage is inserted into an editor
 
 ## 0.7.0 (2021-3-23)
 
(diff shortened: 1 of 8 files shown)


eliknebel (approved):
Looks good. Nice work

PR #883, opened 2021-04-05
Open and free

### LINKED ISSUE #810: Open and Free section support
Full details, requirements and approach TBD, but essentials are:
1. Admin determines a course is to be "Open and Free" enabled.
2. A single, canonical hyperlink is then made available to allow direct access (not an LTI launch) to the course, in delivery mode.
3. An anonymous user is created to track that usage, persisting with a cookie, and a unique URL (accessible from the UI to the anon user)

**Implementation Approach**:
- [x] Create admin-only interface in course authoring that can be used to create and manage open and free sections. Show a section specific URL that can be accessed publicly (outside of an LMS) by students to enroll in the Open and Free section anonymously.
      - O&F sections will automatically track the latest publication, same as how regular sections do
- [x] Create routes for O&F section access/enrollment
- [x] Create simple student facing O&F delivery interface which can be accessed by a user outside of LTI context.
- [x] Generate and display an access URL with user-specific token which can be used to authenticate a user and set the session cookie
- [x] Remove all uses of LTI params except for in certain necessary modules (lti_controller and delivery_controller) and replace with section in assigns
- [x] Follow-up on migration `add_if_not_exists`
- [x] Add captcha for accessing open and free
- [x] Start/End date enforcement
- [x] Add ability to toggle a section open_registration so that new enrollments can be enabled/disabled, implement logic to abide
- [x] Improve Project selection drop down. Currently only descriptions, not titles are searchable
- [x] ~~Investigate long term cookie, persistent through browser restarts for user session~~
- [x] ~~Use separate login_token field different from sub~~
- [x] Discuss changing login token approach to email magic link/email & password

### LINKED ISSUE #888: Rewrite internal links on delivery rendering
**Describe the bug**
Internal links need to be rewritten when rendering for delivery so they can load the correct section.

eliknebel:
This PR adds support for open and free sections. It also includes an overall refactoring of core delivery routes, plugs and the way the system leverages lti_params. Lti params are no longer the main driver of data for delivery, instead any necessary data given by an LMS via LTI params are stored in the relevant section/enrollment records. This helps to streamline the implementation for open and free support, as well as simply the system by removing an overall dependency on lti parameters.

Open and free sections can be created/managed in the admin workspace using the Open and Free link in the sidebar.
Screen Shot 2021-04-05 at 7 11 58 PM

Users can access all concurrent open and free courses by clicking the OLI logo (delivery index path). This will be useful for future Open and Free work where users can track their progress and return to courses later.
Screen Shot 2021-04-05 at 7 11 48 PM

Registration open/closed, start and end date gating is also implemented and enforced for new enrollments. User's who have previously enrolled can always return (as long as their session is still active)
Screen Shot 2021-04-05 at 7 17 25 PM

Closes #810

This PR also includes a fix for course link resolution by providing client activities with a context that contains the section slug. A section slug may be undefined in editing and preview modes, rendering logic will handle accordingly (for internal links <a, the href is simply set to '#' to do nothing)

Closes #888


darrensiegel (changes requested):
Amazing work, works well once I got help in setting up my local recaptcha.

Only thing I noticed was that internal hyperlinks in pages (a link from one page to another) do not work in O+F sections. This is due to these links which are formulated as "course/link/:slug" routing through the "course" scope that expects LTI users.  We may need to peel out this specific endpoint and handle it specially - to determine either an LTI user or an O+F user, then resolve the page from the link.

darrensiegel (approved):

PR #911, opened 2021-04-07
fix timezone support in certain schemas

### LINKED ISSUE #779: Some schemas/table don't have timezone support
There are some schemas in torus that do not store timezone information, when they probably should to ensure consistency  with other datetimes in the system.

- enrollment.ex
- user_identitiy.ex

eliknebel:
This PR fixes timezone support in some schemas that didn't specify. It also sets a configuration that all migrations should use `type: :timestamptz` by default for `timestamps()`. It appears that no migration is necessary for existing timestamps that didnt explicitly specify `:timestamptz` since they utilize the same underlying storage type and the timezone conversion is simply done at the DB layer, but the configuration was added for consistency and documentation purposes.

Related Posts:
- https://elixirforum.com/t/why-cant-timestamptz-be-set-up-as-default-timestamp-for-migrations-in-config/25778/2
- http://www.creativedeletion.com/2019/06/17/utc-timestamps-in-ecto.html

Closes #779

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -40,6 +40,10 @@ config :oli,
   world_universities_and_domains_json: world_universities_and_domains_json,
   countries_json: countries_json
 
+# Configure database
+config :oli, Oli.Repo,
+  migration_timestamps: [type: :timestamptz]
+
 # Configures the endpoint
 config :oli, OliWeb.Endpoint,
   live_view: [signing_salt: System.get_env("LIVE_VIEW_SALT", "LIVE_VIEW_SALT")],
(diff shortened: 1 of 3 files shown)


LEDThereBeLight:
Do these tables need this change?
`enrollments`
`user_identities`
`pending_registrations`

E.g.
```
create table(:enrollments) do
      timestamps()
      add :user_id, references(:users), primary_key: true
      add :section_id, references(:sections), primary_key: true
      add :section_role_id, references(:section_roles)
    end
```

eliknebel:
Based on the some research it appears that no migration is necessary for these, so unless you've discovered any issues in testing, I think they are okay to remain the same. I would prefer to not change migrations if we dont have to, and the new config in config/config.exs should cover it

LEDThereBeLight (approved):

PR #919, opened 2021-04-08
Research consent

rgachuhi:
This PR adds support for research and cookie consent forms. The research consent form is delivery mode only. It is triggered whenever a student or instructor attempts to access a course via an LTI launch. The cookie consent request is set up to launch upon access to any torus page. Note that the cookie consent modal dialog allows users to dismiss it without taking any action. It then delays for an hour before re-prompting the user again.

This PR also fixes a minor bug in high-stakes attempts order by enforcing an order by date-evaluated. The PR also adds minor error logging to the help_controller.


eliknebel (changes requested):
It looks like when the browser is narrow, the Cookie Preferences link and Submit button overlap.

Screen Shot 2021-04-08 at 11 42 39 AM

eliknebel (changes requested):
We probably should make the consent form be configurable with a specific name until we add support for multi-configurable forms

eliknebel (approved):

PR #927, opened 2021-04-12
ImageCode authoring UI tweaks; delete unneeded scripts

andersweinstein:
Image code authoring: always show solution code box, disabled with prompt if no image. Remove original scripts from source tree


eliknebel (approved):

eliknebel (changes requested):
Can you add an entry to changelog please?

eliknebel (approved):

Issue #641, opened 2020-11-04
Adaptive Page integration support (ongoing): CMU-66


darrensiegel:
Overview of work to be done:

- [x] Change the existing server-side evaluation endpoints to return a collection of "actions", one type of which will be the existing "show feedback and score" action that currently exists.
- [ ] DEFER UNTIL NECESSARY Implement GET endpoints for retrieving state for an activity attempt (this never was created since the basic delivery inlines this state into the activity web component)
- [x] Implement a POST bulk endpoint for retrieving state for a collection of activity attempts. 
- [x] Change the URLs for all endpoints to be `/api/v1/state/` prefixed, and `/api/v1/state/course/:course` prefixed when they are course specific
- [x] Introduce the Extrinsic state endpoints.  
- [x] Implement ability to maintain arbitrary state at the resource attempt level. 
- [x] Inline the activity to attempt guid mapping into the server rendered portion of the adaptive delivery page, so that it can be wired in to the adaptive React app as a prop
- [x] Inline the page attempt (resource attempt) custom state into the server rendered portion of the adaptive page, so that it can be wired in to the adaptive React app as a prop
- [x] Elevate the `outcomes` attribute from being a child of the part of an activity to being a sibling of `parts`, rename it to `rules`. 
- [x] Allow for parts in an activity model that have no responses.  Enforce that either all parts have at least one response or that at least one "rule" exists. 
- [x] Fix the OpenAPI docs for the bulk delivery storage service endpoint.
- [x] Make available to the Adaptive Delivery React app the activity type meta data, so that it can render instances of any activity type. 
- [x] Check ingestion, copy, rendering to allow for “group” with “children”
- [x] Server side rendering of custom CSS script overrides
- [x] Provide capability for "chrome-less" page rendering
- [ ] Ensure graded page submission API is invokable from non server rendered page contexts
- [ ] Restructure part attempt record to store the rule that triggered evaluation/finalization

PR #928, opened 2021-04-12
Extrinsic user state

darrensiegel:
This is part 1 of work towards realizing the full set of user state changes in #641 

This PR adds support for the global and section-level "extrinsic state" endpoints.  These endpoints are accessible from `/api/v1/state` and `/api/v1/state/course/:section_slug` URLs, respectively, and allow an activity to read-write arbitrary key-value pairs from state tied to those contexts.  

In addition to basic CRUD operations, there is support for WebSocket based subscription to the state. This is currently implemented in a way that fires on every top-level key change.  We may want to adjust this in the future to have it be tied to a specific top-level key in a state context. 

To test this out interactively, create an instance of an Adaptive Activity in a regular page and use the temporary UIs there to exercise these endpoints. 

The next part of work here is to change the URL endpoints of the rest of the user state endpoints, and add OpenAPI docs to all of them (including these new extrinsic state endpoints).

### DIFF
--- a/assets/src/data/persistence/extrinsic.ts
+++ b/assets/src/data/persistence/extrinsic.ts
@@ -9,7 +9,7 @@ export type ExtrinsicDelete = {
   result: 'success',
 };
 
-export function read_global(keys: string[] | null = null) {
+export function readGlobal(keys: string[] | null = null) {
 
   const params = {
     method: 'GET',
@@ -19,7 +19,7 @@ export function read_global(keys: string[] | null = null) {
   return makeRequest<ExtrinsicRead>(params);
 }
 
-export function delete_global(keys: string[]) {
+export function deleteGlobal(keys: string[]) {
 
   const params = {
     method: 'DELETE',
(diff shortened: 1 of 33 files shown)


assets/src/data/persistence/extrinsic.ts
```diff
@@ -0,0 +1,87 @@
...
+
+export type ExtrinsicRead = Object;
+export type ExtrinsicUpsert = {
+  result: 'success',
+};
+export type ExtrinsicDelete = {
+  result: 'success',
+};
+
+export function read_global(keys: string[] | null = null) {
```
eliknebel, line 12: I'm surprised the linter didnt complain here, but our ts coding standards say to use camelCase instead of snake_case for functions https://github.com/airbnb/javascript#naming--camelCase

eliknebel (changes requested):

lib/oli/accounts/schemas/user.ex
```diff
@@ -26,6 +26,7 @@ defmodule Oli.Accounts.User do
     field :phone_number_verified, :boolean
     field :address, :string
     field :research_opt_out, :boolean
+    field :state, :map, default: %{}
```
eliknebel, line 29: User state will be accessible even across different courses and sections. Is this desired? Or should user state be scoped to a course section view enrollment

eliknebel: Oh, I see there is one on enrollment as well to cover this case.

eliknebel (approved):

PR #953, opened 2021-04-15
Properly load a user and their platform roles in lti pipeline

eliknebel:
This PR fixes an issue where an existing user session will cause the page to crash when accessing the Register Institution page. This bug was produced as a result of changing an existing registration to demo the registration process.

The fix is to load the full current user in the LTI pipeline if a user is logged in to prevent accessing unloaded user platform_roles association in the deliver header template.


LEDThereBeLight:
This looks good. Added changelog and approved.

LEDThereBeLight (approved):

PR #969, opened 2021-04-15
Disallow the rendering of raw html in image captions

### LINKED ISSUE #951: Captions allow rendering of arbitrary HTML
**Describe the bug**
Caption contents render raw HTML into the page. 

**To Reproduce**
1. Create an image, edit the caption and insert `<b>Hey</b>` as the caption. 

**Expected behavior**
The literal string `<b>Hey</b>` should render, but the page renders this as actual HTML

eliknebel:
This PR disallows the rendering of raw html in image captions.

Closes #951

### DIFF
--- a/lib/oli/rendering/content/html.ex
+++ b/lib/oli/rendering/content/html.ex
@@ -288,7 +288,7 @@ defmodule Oli.Rendering.Content.Html do
           end
         }>"
       ] ++
-      [caption] ++
+      [escape_xml!(caption)] ++
       ["</figcaption>"] ++
       ["</figure>"] ++
       ["</div>"]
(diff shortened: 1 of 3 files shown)


LEDThereBeLight:
This looks good. Should we add `escape_xml` for all attributes?

That would be lines:
- 158 (for `language`)
- 112 (for `caption`)
- 104 (for `src`)
- 97 (for `src`)
- 76 (for `src`)
- 59, 64 (for `alt` and `src`)

eliknebel:
Yes, good catch. We should be escaping anything that comes from user input

LEDThereBeLight:
I added one more escape and approved

PR #983, opened 2021-04-19
image coding: use ACE syntax highlighting code editor

andersweinstein:
Adds a syntax coloring, auto-indenting code editor for image code activity and authoring, using the react-ace version of the ACE code editor. It's configured to use a 14 point fixed width font and the "xcode" theme;  available themes can be viewed at [https://manubb.github.io/react-ace-builds/](https://manubb.github.io/react-ace-builds/) . Other options have been set to emulate a plain text box: no line numbering, gutter, current line highlighting, or auto completion. The background is set to grey when disabled like our other other text input boxes. However it does not get the focus highlight border our other text boxes do.


andersweinstein:
Replaced by newer PR based on latest master.

Issue #210, opened 2020-06-02
Instructor preview mode: CMU-206


darrensiegel:
Allow an instructor to view any page in a course, and all details of all activities (all feedbacks, all hints) without having to exercise the activities.

PR #984, opened 2021-04-20
Editor sequential resource navigation in editor and course preview

### LINKED ISSUE #960: Add sequential navigation to authoring and preview
Users have expressed difficulty when trying to review authored materials.

- Page preview should be expanded to support an entire course project with prev/next buttons like delivery
- Authoring should have prev/next buttons that takes a user to the prev/next sequential page in a course

eliknebel:
This PR adds sequential page navigation to resource editor and page preview.

Screen Shot 2021-04-19 at 8 00 45 PM

This effectively makes page preview a full course preview feature, which is still accessible on a specific page but now also from the Review (QA) page.

Screen Shot 2021-04-19 at 8 00 18 PM

Closes #960, and maybe #210


darrensiegel (approved):
Wow this is really great and a major improvement over the way it worked previously.

PR #997, opened 2021-04-22
[FEATURE] Parts Components

bsparks:
sets things up for "part components" similar to how activities are registered but representing a different set of custom elements that are used in adaptive (and future) activities

### DIFF
--- a/priv/repo/migrations/20210421162709_create_part_component_registrations.exs
+++ b/priv/repo/migrations/20210421162709_create_part_component_registrations.exs
@@ -1,0 +1,51 @@
+defmodule Oli.Repo.Migrations.CreatePartComponentRegistrations do
+  use Ecto.Migration
+
+  def change do
+    create table(:part_component_registrations) do
+      add :slug, :string
+      add :title, :string
+      add :icon, :string
+      add :description, :text
+      add :delivery_element, :string
+      add :authoring_element, :string
+      add :delivery_script, :text
+      add :authoring_script, :text
+      add :allow_client_evaluation, :boolean
+      add :globally_available, :boolean, default: false, null: false
+
+      timestamps(type: :timestamptz)
+    end
+
+    create unique_index(:part_component_registrations, [:slug], name: :index_part_component_slug_registrations)
+
+    create unique_index(:part_component_registrations, [:delivery_element],
+             name: :index_part_component_delivery_element_registrations
+           )
+
+    create unique_index(:part_component_registrations, [:authoring_element],
+             name: :index_part_component_authoring_element_registrations
+           )
+
+    create unique_index(:part_component_registrations, [:delivery_script],
+             name: :index_part_component_delivery_script_registrations
+           )
+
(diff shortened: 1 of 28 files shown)


priv/repo/migrations/20210421162709_create_part_component_registrations.exs
```diff
@@ -0,0 +1,51 @@
...
+    create table(:part_component_registrations) do
+      add :slug, :string
+      add :title, :string
+      add :icon, :string
+      add :description, :text
+      add :delivery_element, :string
+      add :authoring_element, :string
+      add :delivery_script, :text
+      add :authoring_script, :text
+      add :allow_client_evaluation, :boolean
```
darrensiegel, line 14: Allow client evaluation is probably unique to Activities and doesn't apply to Part Components

darrensiegel (changes requested):
An existential question: Do we want the term "janus" to live on in the oli-torus codebase? I see it in this PR as a directory/file name, component names, etc. I don't think that we do.  This is probably the opportunity to remove it - but I don't know what impact that might have on sparrow conversion.

There are three aspects of the "Registered Activities" that you didn't replicate in the PartComponent infrastructure: 

1. There are a few unit tests in `Oli.ActivitiesTest`.
2. There is a user interface to allow the global management of registered activities. This is driven from `OliWeb.ActivityManageController`. 
3. There is a user interface to allow authors to opt-in to using non-global activities. 
 This is driven from `OliWeb.ProjectActivityController`

It makes sense to replicate those unit tests in this PR, but we probably don't need (or even want) to replicate those user interfaces, yet.

priv/repo/migrations/20210421162709_create_part_component_registrations.exs
bsparks: yeah I almost removed that, but then thought maybe it would be needed for something... I'll remove it though, better to add something *when* it is needed :)

bsparks:
> An existential question: Do we want the term "janus" to live on in the oli-torus codebase? I see it in this PR as a directory/file name, component names, etc. I don't think that we do. This is probably the opportunity to remove it - but I don't know what impact that might have on sparrow conversion.
> 
> There are three aspects of the "Registered Activities" that you didn't replicate in the PartComponent infrastructure:
> 
> 1. There are a few unit tests in `Oli.ActivitiesTest`.
> 2. There is a user interface to allow the global management of registered activities. This is driven from `OliWeb.ActivityManageController`.
> 3. There is a user interface to allow authors to opt-in to using non-global activities.
>    This is driven from `OliWeb.ProjectActivityController`
> 
> It makes sense to replicate those unit tests in this PR, but we probably don't need (or even want) to replicate those user interfaces, yet.

hmm yeah, well I can change the folders and the id of the part, the element tags are what are currently tied to the conversion. I suppose they too could be changed and re-converted... just need to figure out what to call them. it does seem to me like a namespace is still appropriate `torus-image` ??

darrensiegel (approved):

PR #1002, opened 2021-04-23
[BUG FIX] Add missing database indexes, infra for flagging problematic queries 

darrensiegel:
This PR does two things:

1. It addresses the problem of a large amount of sequential scans being performed on the `published_resources` table by adding some necessary database indexes and by optimizing the queries in `DeliveryResolver` and `AuthoringResolver`.
2. It adds developer-oriented support for flagging as a log warning "problematic queries".   

The problematic query detection impl works by intercepting all queries flowing through `Oli.Repo` via the `prepare_query` override.  If the `:problematic_query_detection` flag is set to `:enabled` (and you are not running in production), the query will first be explained and analyzed using the `"EXPLAIN ANALYZE true (...sql)"` sql. After explained, that JSON structure is traversed to find two different problems: 
1. Steps that involve sequential table scans
2. Any query whose total compute cost exceeds some threshold value (currently set to 150).

The guts of the explain impl was lifted from `revelrylabs/ecto_explain`, after I was unable to get it to work in its intended fashion (via `use Ecto.Explain` macro)

Running Torus with problematic query detection enabled, one can see many queries that are performing full scans.  We should follow up this work with work to address some of those.

### DIFF
--- a/config/dev.exs
+++ b/config/dev.exs
@@ -2,7 +2,12 @@ use Mix.Config
 
 config :oli,
   env: :dev,
-  problematic_query_detection: :disabled,
+  problematic_query_detection: System.get_env("DEV_PROBLEMATIC_QUERY_DETECTION_ENABLED", "false")
+    |> String.downcase()
+    |> (case do
+      "true" -> :enabled
+      _ -> :disabled
+    end),
   s3_media_bucket_name: "torus-media-dev",
   media_url: "torus-media-dev.s3.amazonaws.com",
   slack_webhook_url: System.get_env("SLACK_WEBHOOK_URL")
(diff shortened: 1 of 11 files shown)


config/dev.exs
```diff
@@ -2,6 +2,7 @@ use Mix.Config
 
 config :oli,
   env: :dev,
+  problematic_query_detection: :disabled,
```
eliknebel, line 5: We might want to set this up to read from `oli.env` environment configs, e.g. `System.get_env("DEV_PROBLEMATIC_QUERY_DETECTION_ENABLED")`, or simply always enable this in development so that someone doesn't accidentally check in a change here

eliknebel (approved):

PR #1003, opened 2021-04-25
Chrome free pages

darrensiegel:
Some Adaptive Pages need to exercise full control over the CSS of the page, and to have nothing else rendered as "application chrome" by Torus.  These types of pages may provide their own page-to-page navigation. 

This PR introduces a top-level page content attribute `displayApplicationChrome`, that if set to `false` will allow the adaptive page to render without any Torus application chrome around it.  This attribute defaults to `true` if not set. 

Another top-level attribute `additionalStylesheets` is a collection of arbitrary style sheet URLs, that if set, will get injected into the `<head>` portion of the rendered page.

### DIFF
--- a/lib/oli_web/views/layout_view.ex
+++ b/lib/oli_web/views/layout_view.ex
@@ -41,6 +41,7 @@ defmodule OliWeb.LayoutView do
   """
   def additional_stylesheets(assigns) do
     Map.get(assigns, :additional_stylesheets, [])
+    |> URI.encode()
     |> Enum.map(fn url -> "\n<link rel=\"stylesheet\" href=\"#{url}\">" end)
     |> raw()
   end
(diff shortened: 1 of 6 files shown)


lib/oli_web/views/layout_view.ex
```diff
@@ -35,6 +35,16 @@ defmodule OliWeb.LayoutView do
     )
   end
 
+  @doc """
+  Allows a delivery content template to specify any number of additional stylesheets, via URLs,
+  to be included in the head portion of the document.
+  """
+  def additional_stylesheets(assigns) do
+    Map.get(assigns, :additional_stylesheets, [])
+    |> Enum.map(fn url -> "\n<link rel=\"stylesheet\" href=\"#{url}\">" end)
```
eliknebel, line 44: This could be a risk, if an author injects a dangerous html string in for url. Is this url protected from user input? To be safe, we could surround with escape_xml we use in rendering

eliknebel (changes requested):

eliknebel (approved):

PR #1010, opened 2021-04-29
[FEATURE] Image resizing

LEDThereBeLight:
Implement a `Resizer` component and add it to the `Image` editor to handle resizing. Image dimension changes are saved to the model's `height` and `width`, which are used in layout CSS rules to determine the rendered image size. The actual size of the uploaded image is not changed.


darrensiegel (changes requested):
I made one code change to remove the usage of a type in non-testing code that was being imported from from a dependency declared to be a "dev" dependency. (I simply replicated the `Nullish` type in `types/nullish.ts`

I am seeing one bug that we do need to track down and fix.  Here is how to replicate it:
1. Create a page that has a single content block, which contains a single image in it.
2. Reload the authoring view for that page.
3. Create a second content block, and reorder that second content block - moving it above the original one.  You now have two content blocks:  the top most on the page is an empty one, and the bottom content block is the one with the image.
4. Resize the image - you should see a problem where  the image get duplciated in the topmost content block.

^ The above seems to perhaps be driven by content blocks holding on to the *index* that were originally positioned within in the model array.

darrensiegel (approved):

PR #1013, opened 2021-04-30
[FEATURE] Custom branding

### LINKED ISSUE #977: Add support for custom branding
Branding is the ability for the system, at run-time, to display different organization names, logos, etc, at delivery time. This allows an institution to substitute their own logo and name for OLI, CMU, etc. 

What needs to be done:

1. Add a new `brands` table in the database.  This table will define all of the brandable attributes, and will at least include logo, and maybe institution name (I can't find anything right now in delivery beyond just the logo that needs to be brandable at this point).
2. This "brands" table should be bootstrapped to include an initial brand (the OLI brand). 
3. Build a new, admin-facing only, UI for viewing, editing, and creating new brands.  This UI *must* show the number of course sections and the number of institutions that are using each brand. This is to safeguard cases where an admin is about to "edit" a brand and doesn't realize how widely it is being used. 
3. Add to both of the `registrations` and `sections` schemas a `brand_id` attribute that references the brand to use. 
4. In the existing, admin-facing Institution/registration editor, allow a brand to be selected/changed at the registration level.
5. In the Open and Free section creation UI, allow a brand to be selected/changed for open and free sections. 
6. Change the logic for section initialization at time of LTI launch to set the brand on the newly created section to match the brand that exists at the institution's registration level that is performing the launch.  
7. Change the relevant delivery code to always look at the brand that exists pinned to the section, to determine which brand logo and other text to display.
8. Build a database migration to not only introduce the new brand schema and branding relationships, but to set the brand_id of all existing registrations and all existing sections to be the default OLI brand_id. 

What the above accomplishes:

1. Institutions can pick a brand that all future course sections that get created will use.  Since this is at the registration level, an institution can have multiple brands (aka "CMU College of Engineering" vs "CMU SCS")
2. Once created, a section, and only the section, determines what brand to use. If an institution wants to change their brand (and all brands of existing sections) they would *edit* that existing brand as oppposed to creating an new one and using that new one at the registration level.

eliknebel:
This PR introduces the ability to apply custom branding to torus at a compile time as well as dynamically at runtime.

Torus compile-time branding is controlled by setting the BRANDING_ environment variables (see oli.example.env). Custom branding assets can be placed in `assets/static/custom` which will ensure they are available from the web server without them being accidentally committed to source control.

Brands can be created and managed by an administrator using the Brands page from the admin sidebar. A brand can be associated with a particular Institution which will limit the visibility of that brand to registrations belonging to that institution (this functionality will be more useful in the future when branding becomes available to institution administrators). Once a brand is created it can then be applied to a registration or open and free section which will brand all sections belonging to a registration or open and free section.

Due to the nature of the work here, light/dark theming had to be refactored. Torus "themes" now treat light and dark mode as first class concepts, instead of totally separate themes. If a new them is to be created, it will need to have a light and dark support. As part of this refactoring, the ability to change authoring "themes" has been removed and there is now only a single authoring theme that supports both light and dark modes. Based on some research, this goes along with latest web trends to remove the ability to toggle dark mode from web pages themselves and instead delegate that customization to the user's operating system using the prefers-color-scheme media selector in css.

Finally, this PR includes some graphic design experimentation with creating a Torus specific brand and logo, since we would rather not have other forks using the OLI brand for their instances. This logo is simply a prototype and can be easily removed from the PR if it is not desirable or appropriate for the overall project's branding. OLI torus instances can be configured during CI/CD compilation to use whichever branding we deem fit for OLI production servers.

Closes #977


darrensiegel (approved):
This looks really good. I tested it locally and it works well.  We should take this as-is, but there are a couple of observations around Brand usage statistics:

1. The Branding UI is potentially misleading in how it uses `Oli.Branding.list_brands_with_stats/0` to show usage counts for each brand. The count of number of sections shows only those sections that specifically are set to that brand, but doesn't reflect the actual count of sections using that brand. This is due to sections that do have a brand specified "falling back" to the registration.  
2. There is no way to see how many sections are using the default brand.

PR #1027, opened 2021-05-05
[FEATURE] Allow previewing for advanced authoring into chromeless view

darrensiegel:
This PR adds the ability for advanced authoring to preview a lesson by launching a chromeless window that bootstraps the Delivery app, with a new `previewMode` attribute present and set to `true`. 

To exercise this functionality, open the authoring app for an adaptive page.  There is now a "Preview" button on it that when clicked will open the preview in a named window tab.  

I fixed a problem with how the `additional_stylesheets` was implemented.

### DIFF
--- a/lib/oli_web/controllers/resource_controller.ex
+++ b/lib/oli_web/controllers/resource_controller.ex
@@ -52,8 +52,20 @@ defmodule OliWeb.ResourceController do
       nil ->
         render_not_found(conn, project_slug)
 
+      %{content: %{"advancedDelivery" => true}} = revision ->
+        put_root_layout(conn, {OliWeb.LayoutView, "chromeless.html"})
+        |> render("advanced_page_preview.html",
+          revision: revision,
+          scripts: Activities.get_activity_scripts(),
+          user: author,
+          project_slug: project_slug,
+          title: revision.title
+        )
+
       %{content: content} ->
-        activity_ids = Oli.Authoring.Editing.Utils.activity_references(content) |> MapSet.to_list()
+        activity_ids =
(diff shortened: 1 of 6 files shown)


lib/oli_web/controllers/resource_controller.ex
```diff
@@ -52,8 +52,20 @@ defmodule OliWeb.ResourceController do
       nil ->
         render_not_found(conn, project_slug)
 
+      %{content: %{"advancedDelivery" => true}} = revision ->
+        put_root_layout(conn, {OliWeb.LayoutView, "chromeless.html"})
```
bsparks, line 56: I suppose that preview mode should also use the "displayApplicationChrome" flag for this?

lib/oli_web/controllers/resource_controller.ex
```diff
@@ -52,8 +52,20 @@ defmodule OliWeb.ResourceController do
...
+        |> render("advanced_page_preview.html",
+          revision: revision,
```
bsparks, line 58: I think we need this too
```
additional_stylesheets: Map.get(context.page.content, "additionalStylesheets", []),
```

lib/oli_web/controllers/resource_controller.ex
```diff
@@ -52,8 +52,20 @@ defmodule OliWeb.ResourceController do
...
+          scripts: Activities.get_activity_scripts(),
```
bsparks, line 59: oh and these will need to be `:delivery_script`
and add
```
part_scripts: PartComponents.get_part_component_scripts(:delivery_script),
```

PR #1028, opened 2021-05-06
[FEATURE] Adaptive Rules Engine & Preview Page

bsparks:


darrensiegel (approved):
I was leary about the upgrade to React 17, but I did some local testing and do not see any issues. 

I did, however, remove the top-level `lodash` dependency.  I can see that `json-rules-engine` uses `lodash`, but no actual Torus/Janus code uses it.

PR #1029, opened 2021-05-06
[FEATURE] Flame Graph generation and viewing

darrensiegel:
This PR integrates flame graph creation and viewing (from a dev environment only). 

To trace code and generate a flame graph, wrap it in a function and pass it into `Oli.Utils.FlameGraph.create/2` along with a string tag name as the second argument. For instance, to generate a flame graph of the body of a controller function:

```
def index(conn, _params) do 
  Oli.Utils.FlameGraph.create(fn ->
    render(conn, "template.html", assign: some_crazy_function())
  end), "controller-index")
end
```

Then exercise that code (interactively or via unit test).  The flame graph raw data will be captured in `./_flame_graph_stacks/#{tag}.out`.  From that raw data file, a command line shell script in the eflame package converts it to an SVG file. This manual step of converting and viewing was painful, so I took the next step:

One can automate the creation of the SVG, and view them in the Torus Web UI by visiting `https://localhost/dev/flame_graphs`.  You will see the list of all generated flame graphs, click on their links to see the SVG.  Upon visiting this page all SVG files will be created for all raw data files that exist in the `/_flame_graph_stacks/` directory.  It is currently up to the developer to clean up both the raw data flies and the SVG files. 

Currently, the impl overwrites the graph every time one exercises the graphed function.  I played around with preserving instances via attaching a timestamp to the filename, but for now went with keeping just one graph per tag. 

I'm not sure why now suddenly my editor reformatted `router.ex`.  I then ran `mix format` locally and see that there are a bunch of files that are out of format.  I've included these reformats in this PR, but have added inline comments to help a reviewer find the non-formatting changes.  Basically they are modules `Oli.Utils.FlameGraph`, The `DevController` and corresponding view, and the `flame_graph.html.eex` template.  Also, a change in `router.ex` to wire in the DevController.


lib/oli/utils/flame_graph.ex
```diff
@@ -0,0 +1,27 @@
+defmodule Oli.Utils.FlameGraph do
```
darrensiegel, line 1: The module for creating flame graphs, converting to SVG

lib/oli_web/controllers/dev_controller.ex
```diff
@@ -0,0 +1,13 @@
+defmodule OliWeb.DevController do
+  use OliWeb, :controller
+
+  alias Oli.Utils.FlameGraph
+
+  def flame_graphs(conn, _params) do
```
darrensiegel, line 6: Displays all SVG urls, and when invoked, rebuilds the actual SVGs from the raw stack data for each file

eliknebel (approved):
This is really awesome being able to visualize the performance and "drill in" on the generated svg at `/dev/flame_graphs`

PR #1046, opened 2021-05-12
Feature/audio video slider part components

schowkekarats360:


assets/src/components/parts/janus-audio/Audio.tsx
```diff
@@ -0,0 +1,121 @@
...
+const handleAudioEnd = () => {
+    if (!enableReplay) {
+        setShowControls(false);
+    }
+};
+const handleAudioPlay = (data: any) => {
+    if (isAudioStarted) return;
+    //Need this otherwise, save state will called on every second
+    isAudioStarted = true;
+    console.log('Audio has started playing');
```
bsparks, line 59: need to remove or comment out console calls

assets/src/components/parts/janus-slider/Slider.tsx
```diff
@@ -0,0 +1,165 @@
...
+  useEffect(() => {
+    if (typeof props?.model === 'string') {
+      setModel(JSON.parse(props.model));
+    }
+    if (typeof props?.state === 'string') {
+      setState(JSON.parse(props.state));
+    }
+  }, [props]);
+
+  const { x, y, z, width, height, src, alt, customCssClass,
```
bsparks, line 19: seems like the formatting is off here, should be one per line

assets/src/components/parts/janus-slider/Slider.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+const handleSliderChange = (e: any) => {
+    setSliderValue(e.target.value);
+    //saveState({ sliderVal: e.target.value, userModified: true });
+};
+
+
+const handleActivityStateChange = (stateData: any) => {
+    const interested = stateData.filter(
+        (stateVar: any) => stateVar.id.indexOf(`stage.${props.id}`) >= 0
```
bsparks, line 76: this one is missing the end `.`

assets/src/components/parts/janus-slider/Slider.tsx
```diff
@@ -0,0 +1,165 @@
...
+};
+
+const [sliderValue, setSliderValue] = useState(value ? value : minimum); // if value is undefined, set default to minimum value;
+const [isSliderEnabled, setIsSliderEnabled] = useState(true);
+
+const inputWidth: any = document
+    .getElementById(`${props.id}`)
+    ?.getBoundingClientRect().width;
+const thumbWidth: any = document
+    .getElementById('slider-thumb')
```
bsparks, line 59: wait, this shouldn't be an id on the whole document

assets/src/components/parts/janus-slider/Slider.tsx
```diff
@@ -0,0 +1,165 @@
...
+                {label}
+            </label>
+        )}
+    </div>
+);
+};
+
+export const tagName = 'janus-slider';
+
+// TODO: redo web component
```
bsparks, line 163: can remove this comment I think

assets/src/components/parts/janus-video/Video.tsx
```diff
@@ -0,0 +1,189 @@
...
+  useEffect(() => {
+    if (typeof props?.model === 'string') {
+      setModel(JSON.parse(props.model));
+    }
+    if (typeof props?.state === 'string') {
+      setState(JSON.parse(props.state));
+    }
+  }, [props]);
+
+  const { x, y, z, width, height, src, alt, customCssClass,autoPlay = false,
```
bsparks, line 19: formatting

bsparks (approved):

assets/src/components/parts/janus-slider/Slider.tsx
```diff
@@ -0,0 +1,141 @@
...
+    width: '100%',
+    display: `flex`,
+    flexDirection: 'row',
+  };
+
+  const [sliderValue, setSliderValue] = useState(value ? value : minimum); // if value is undefined, set default to minimum value;
+  const [isSliderEnabled, setIsSliderEnabled] = useState(true);
+
+  const inputWidth: any = document.getElementById(`${id}`)?.getBoundingClientRect().width;
+  const thumbWidth: any = document.getElementById('slider-thumb')?.getBoundingClientRect().width;
```
darrensiegel, line 65: With the way that this coded, there will be multiple DOM elements that get assigned the id of "slider-thumb" if there are multiple instances of the `Slider` component present in the same view.   In the case where there is more than one DOM element with the same id, a call to `document.getElementById` will return a reference to the *first* element with that identifier.  Is that going to work here?  Can the widths of these 'slider-thumb' components ever differ?  

If this isn't a concern due to invariants that either the widths never differ across instances, or there are never multiple instances, this should be documented as such.  Otherwise, the fix here would be to dynamically set the id.

darrensiegel (changes requested):

bsparks (approved):

darrensiegel (approved):

Issue #914, opened 2021-04-07
Curriculum item settings modal doesn't work after single use


eliknebel:
**Describe the bug**
After opening a settings modal for a curriculum item, a user cannot open the modal again without refreshing the page. Also, when adding new curriculum items, they do not appear until a page refresh. Something is broken with the liveview modal implementation. We should also switch this to bootstrap modal implementation for consistency.

**To Reproduce**
1. Go to curriculum in a course
2. Click on the three dots for an item to open the settings dialog
3. Close the dialog and repeat

**Expected behavior**
Modal should open after and new items created should appear.

**Screenshots**
<img width="776" alt="Screen Shot 2021-04-07 at 3 29 14 PM" src="https://user-images.githubusercontent.com/6248894/113927864-06e13380-97bc-11eb-967e-93363ea2cd8f.png">

PR #1053, opened 2021-05-13
Move curriculum items

### LINKED ISSUE #1007: Add support for moving curriculum items
1. Make the curriculum UI more consistent in the Unit vs Page dropdown locations.
2. Introduce a dropdown menu of various actions for both containers and pages. This should include "Move to", "Delete", and "Settings".  The "Move To" option should display a secondary pop-up/modal hierarchy selection UI that the author uses to pick the location to move to.  The "Delete" option displays a confirmation dialog prior to deletion.  The "Settings" option displays the settings modal. Ideally we would want to remove renaming the item or page from the settings dialog for containers and make that available as a "Rename" option in the drop down menu. 
3. See if we can fix the tooltips, and to also add more description into the tooltips
4. See if we can get a different icon for Assessment that isn’t a checkbox.

eliknebel:
This PR adds functionality to move items around the curriculum as well as the following other improvements:

- Make the curriculum UI more consistent in the Unit vs Page dropdown locations
- Fix tooltips by adding more detailed labels to view
- Persist view selection to author preferences so that the select view is remembered across page requests
- Add a dropdown actions menu with Options, Move to, and Delete actions
- Fix "dialog in modal" issue with delete action
- Change icons
- Fix modals, use bootstrap styles
- Fix a couple of stability issues in the curriculum view by refactoring the data model

Closes #1007, #914


darrensiegel (approved):
This is great - works nicely and has a very slick look and feel to it. The impl is good as well. I tacked on a couple commits.  One to include a broadcast of the change to the destination container (which allows anyone editing that container to "see" this move immediately) and 2) I removed the dropdown on the tail breadcrumb that is now redundant (and was broken).

PR #1062, opened 2021-05-18
[FEATURE] PMP-1160: Support Author Lesson Preview from Torus - Screen Navigation

dsuriano:
Support Author Lesson Preview from Torus such that the Author can get to the view of the Lesson that has the Screen Navigation and maneuver through the lesson as a Student would or by jumping to a particular view.

Trap States will be handed in a separate ticket PMP-1159 as will the Inspector Feature PMP-866. Instructor use of this feature have been split into another story - PMP-1373

### DIFF
--- a/assets/src/apps/delivery/components/PreviewTools.tsx
+++ b/assets/src/apps/delivery/components/PreviewTools.tsx
@@ -1,0 +1,214 @@
+import React, { useState } from 'react';
+import { useDispatch, useSelector } from 'react-redux';
+import { selectCurrentActivity } from '../store/features/activities/slice';
+import { navigateToActivity } from '../store/features/groups/actions/deck';
+
+// Title Component
+interface TitleProps {
+  title: string;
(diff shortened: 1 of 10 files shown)


assets/src/apps/delivery/components/PreviewTools.tsx
```diff
@@ -0,0 +1,214 @@
...
+  const selected = useSelector(selectCurrentActivity);
+  const dispatch = useDispatch();
+  const sequence = props.model[0].children
+    ?.filter((child: any) => !child.custom.isLayer && !child.custom.isBank)
+    .map((s: any) => {
+      return { ...s.custom };
+    });
+
+  // Navigates to Ensemble
+  const navigate = (ensembleId: any) => {
```
bsparks, line 112: should rename to activityId

dsuriano: fixed

PR #1088, opened 2021-05-25
[REFACTOR] Cookie impl refactor

rgachuhi:
This PR moves both cookie consent and cookie preference UI's into react components. All other functionality remains similar to the javascript version


lib/oli_web/templates/layout/_footer.html.eex
```diff
@@ -14,4 +13,7 @@
       </div>
     </div>
   </div>
-</footer>
+  <script>
```
darrensiegel, line 16: I wanted to eliminate this call (and the one for delivery mode above) back to the server since it will happen on every single page render, regardless of whether the user ever opens the "Cookie Preferences" page.

darrensiegel (changes requested):

lib/oli_web/templates/layout/_footer.html.eex
rgachuhi: Server call now blocked if opt-in cookie present in the browser

darrensiegel (approved):

PR #1139, opened 2021-06-10
[BUGFIX] Carousel Part component

dtiwarATS:
-- updated tag name for the Carousel component. It should be 'janus-image-carousel' not 'janus-carousel'

### DIFF
--- a/assets/src/components/parts/janus-Carousel/manifest.json
+++ b/assets/src/components/parts/janus-Carousel/manifest.json
@@ -1,14 +1,14 @@
 {
-  "id": "janus_carousel",
+  "id": "janus_mage_carousel",
   "friendlyName": "Carousel",
-  "description": "Displays a carousel.",
+  "description": "Displays a image carousel.",
   "global": true,
   "delivery": {
-    "element": "janus-carousel",
+    "element": "janus-image-carousel",
     "entry": "./delivery-entry.ts"
   },
   "authoring": {
-    "element": "janus-carousel",
+    "element": "janus-image-carousel",
     "entry": "./authoring-entry.ts"
   }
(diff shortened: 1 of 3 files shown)


assets/src/components/parts/janus-Carousel/manifest.json
```diff
@@ -1,14 +1,14 @@
 {
-  "id": "janus_carousel",
+  "id": "janus_mage_carousel",
```
bsparks, line 2: missing i

dtiwarATS: Ooops... Done

bsparks (approved):

priv/repo/seeds.exs
```diff
@@ -106,7 +106,7 @@ Oli.Registrar.register_local_part_components(
     "janus_audio",
     "janus_video",
     "janus_slider",
-    "janus_carousel",
+    "janus_image_carousel",
```
darrensiegel, line 109: Renaming this part component is not sufficient to allow existing Torus instances to pick up this change.  This renaming works for seeding new databases, but existing ones will end up having the old "janus_carousel' and a new 'janus_image_carousel' both exist as part components due to how `Oli.PartComponents.register_part_component' works.  

A database migration is necessary to do this "renaming" correctly.  The migration can simply delete the 'janus_carousel' record, allowing the seed.ex change to drive the creation of a new one with the updated id.

darrensiegel (changes requested):

dtiwarATS:
Closing this PR for now. Once the migration script is ready and the rendering issues fixed then I will create another PR.

PR #1145, opened 2021-06-11
remove duplicate vendor load from chromeless; remove adv delivery class

bsparks:


bsparks:
incorporating changes as part of another PR

PR #1147, opened 2021-06-14
Build Header Bar with the Components and the Preview

schowkekarats360:


assets/src/apps/authoring/icons.tsx
```diff
@@ -0,0 +1,404 @@
+/* eslint-disable react/no-unknown-property */
+
+import React from 'react';
+
+export const ImageIcon: React.FC<any> = () => {
```
bsparks, line 5: let's create a folder called icons and put one per file pls

assets/src/apps/authoring/icons/icon-multiChoice.svg
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="29px" height="24px" viewBox="0 0 29 24" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <title>icon-multiChoice</title>
+    <g id="Screen-Panel-/-Adaptivity" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
```
bsparks, line 4: I see this same id in another file, are the ids even necessary?

assets/src/apps/authoring/icons/icon-componentList.svg
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
bsparks, line 1: these raw svg assets should be stored in the "static" folder. since the authoring tool is closer to the "server side" we have access to the full hosting power of the server and should use it.

assets/src/apps/authoring/HeaderNav.tsx
```diff
@@ -8,7 +20,53 @@ interface HeaderNavProps {
...
+      className={`aa-header-nav top-panel${
+        isVisible ? ' open' : ''
+      } d-flex justify-content-between`}
+      style={{ alignItems: 'center' }}
+    >
+      <div>{content.title}</div>
+      <div className="btn-toolbar" role="toolbar" aria-label="Toolbar with button groups">
+        <div className="btn-group me-2" role="group" aria-label="First group">
+          <div className="px-2">
+            <TextIcon></TextIcon>
```
dsuriano, line 33: should opt for self closing tags when no children are present: `<TextIcon />`

assets/src/apps/authoring/icons/icon-multiChoice.svg
dsuriano: they look like export artifacts from the design program i used. probably safe to remove, but might want to test.

something i forgot to do with these svgs is run them thru an optimizer like: https://jakearchibald.github.io/svgomg/

schowkekarats360: I checked after removing the id and the things were looking fine.

dsuriano (changes requested):
looks like we have a duplicate set of icons ... the ones in `assets/static/icons` (not being used) and the ones in `assets/src/apps/authoring/icons` ...

bsparks:
I was wondering why we need to have icons embedded in React? perhaps it's because we are just "used to" not having a server that can host static assets?

schowkekarats360:
Messed up changes while merging with another branch

PR #1148, opened 2021-06-14
Delivery user signin

### LINKED ISSUE #1049: Allow open and free students to maintain progress by creating an account
We need to allow students using Open and Free courses to maintain their progress by creating an account that they can use to login. 

Changes here include:

- [x] Refactor authoring and user pow accounts to support multiple domains in the same phoenix app https://powauth.com/guides/2019-09-12-multiple-schemas.html
- [x] Maintain the ability to access and browse an open and free course section without creating an account or logging in with existing account. 
- [x] Change the language on `proton.oli.cmu.edu` landing page from "Welcome to OLI Authoring" to "Welcome to OLI".  This is necessary since now students can land here and login to access their sections. 
- [x] Create a "My Sections" view that a student will be redirected to once they log in. 
- [x] Ensure that a student with an account and logged in gets "enrolled" properly in any Open and Free section that they end up clicking into. Students will get URLs for open and free sections from some other means, not from some dedicated view/feature in Torus (at least not yet). 
- [x] For Open and free sections, when a student is accessing material WITHOUT having logged in, display a banner or message that informs the student that their progress is lost once they close their browser and that they can create an account to maintain that progress (include a link to the account creation / sign in view right from within that message).
- [x] Student who is already enrolled should not be able to access the enroll page to enroll again
Screen Shot 2021-06-04 at 12 25 28 PM

- [x] Make sure email verification and other extensions work properly
    
### Deployment Tasks:
- [ ] Update Github OAuth routes to use `/authoring` when feature is deployed, remove (2) unused Google OAuth routes

eliknebel:
This PR adds the ability for a delivery user to create an account and persist course progress.

Important Notes: The sign in button in the top right is now dedicated to learner login, whereas the "Create an Account" button has been changed to the course author login button. Also, anything authoring related has now been scoped under "/authoring" path which could break existing user's bookmarks.

**Required Deployment Tasks:**
- [ ] Update Github OAuth routes to use `/authoring` when feature is deployed, remove (2) unused Google OAuth routes

Closes #1049


assets/src/components/messages/CreateAccountPopup.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+  useEffect(() => {
+    if (firstRender) {
+      const showPref = sessionStorage.getItem('createAccountPrompt') != 'hide'
+        && localStorage.getItem('createAccountPrompt') != 'hide'
+      setShow(showPref);
+      window.addEventListener('click', onBlurClick);
+    }
+    setFirstRender(false);
+  });
```
darrensiegel, line 37: I only recently learned of this myself, but you can give `useEffect` an empty array as a second argument which causes the effect to run only one time after the initial render. So the pattern of introducing a boolean state variable is necessary to get the "run just once" semantics.

darrensiegel: See here: 

https://stackoverflow.com/questions/53120972/how-to-call-loading-function-with-react-useeffect-only-once

darrensiegel (changes requested):
Customize the error message displayed when a user fails to log in with user credentials to let them know if they are "looking for authoring" where to go.

assets/src/components/messages/CreateAccountPopup.tsx
eliknebel: Great tip, thanks!

darrensiegel (approved):

PR #1149, opened 2021-06-14
[Bugfix] phases of moon bugfix

dtiwarATS:
-- contains a fix for Janus-dropdown part component
-- phases of Moon bug fix
-- timeout issue when rendering an activity without any part component.
-- overriding configData value with Init trap state configuration

### DIFF
--- a/assets/src/components/activities/adaptive/AdaptiveDelivery.tsx
+++ b/assets/src/components/activities/adaptive/AdaptiveDelivery.tsx
@@ -68,12 +68,12 @@ const Adaptive = (props: DeliveryElementProps<AdaptiveModelSchema>) => {
     let resolve: any;
     let reject;
 
-    if (parts.length == 0 && resolve) {
-      //if an activity does not have any parts, send the resole promise
+    if (!parts.length) {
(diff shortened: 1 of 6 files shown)


assets/src/components/activities/adaptive/AdaptiveDelivery.tsx
```diff
@@ -89,7 +89,13 @@ const Adaptive = (props: DeliveryElementProps<AdaptiveModelSchema>) => {
       }, 2000);
     });
     sharedPromiseMap.set(props.model.id, { promise, resolve, reject });
-
+    if (parts.length == 0 && resolve) {
```
bsparks, line 92: I think we can move this to the top of this method and don't even need to setup the promise stuff if there are no parts
```
if (!parts.length) {
  if (props.onReady) {
    props.onReady(attemptState.attemptGuid);
  }
  return;
 }
```

dtiwarATS: Done

bsparks (approved):

PR #1164, opened 2021-06-16
Feature/styling accordion and tab

schowkekarats360:


assets/src/apps/authoring/Accordion/Accordion.tsx
```diff
@@ -1,37 +1,53 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
 /* eslint-disable react/jsx-key */
 /* eslint-disable @typescript-eslint/no-explicit-any */
 /* eslint-disable react/prop-types */
 
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+const plusIcon = require('../../../../static/images/plusIcon.svg');
```
bsparks, line 7: not seeing where this is used, also I think things in the static folder should be accessible via url as a hosted item, so I don't think you'd need to require them

dsuriano (changes requested):
Getting some issues while testing this locally: 

image

assets/src/apps/authoring/Accordion/Accordion.tsx
bsparks: I have a branch where I am working on bringing the data in for the authoring tool, here I have added a way for React to access the path to the static folder correctly
https://github.com/Simon-Initiative/oli-torus/blob/d28837fc162998a40d8f6d6b48ac236212d71818/lib/oli_web/templates/resource/advanced.html.eex

this is the branch: https://github.com/Simon-Initiative/oli-torus/tree/feature/authoring-bff

so I think what we should do is after bringing that path into React props then just use 
```
<img src={`${imagePath}/plusIcon.svg`}
```
something like that. (the path would be in `props.paths.images`) . 
it's also possible that you might want to create a component like `<Icon basePath={props.paths.images} file="plusIcon.svg" />`  and then internally render the `img` tag with maybe a fallback if the image isn't found or something. Possibly also have a map of icon names and / or accept a url.

You might want to build off of my branch and then we can just merge the changes together tomorrow.

schowkekarats360: This line i forgot to remove. I have removed that now. But what you have explained i can use that for header tab icons.

schowkekarats360 (commented):
I tried to use glyphicons and bootstrap icons (bi) to show plus icons but it was not working for some reason. so i decided to use the svg for that. Later I come across the font awesome icons (fa) which worked , but i forgot to remove the svg reference from the file

assets/src/apps/authoring/Accordion/Accordion.tsx
```diff
@@ -1,37 +1,52 @@
...
+  const [expandedCards, setExpandedCards] = useState<any>([]);
+  useEffect(() => {
+    const tabsId = tabs.map((tab: any) => tab.id);
+    setExpandedCards(tabsId);
+  }, []);
   return (
-    <div className="accordion">
+    <div className='accordion'>
       {tabs.map((tab: any) => (
-        <div className="card">
```
dsuriano, line 11: react will complain in the console if you don't specify a key on the first element of a `.map()`.  refactor to: 
```
{tabs.map((tab: any, index: number) => ( 
   <div key={index} className="card">
   ...
```
and check that you've got that for all `.map()`s ...

dsuriano (changes requested):

PR #1165, opened 2021-06-16
[BUG FIX] Allow course project deletion when title contains special characters

### LINKED ISSUE #1131: Delete Course does not handle special characters
**Describe the bug**
When trying to delete a course you need to type in the name of the course. One such course I have has an ampersand  in the name and the validation does not properly recognize it.

**To Reproduce**
1. Create a new project titled "Variables & Iframes"
2. Click on "Delete Project" in the course details page
3. Attempt to type or copy/paste the name into the box to confirm
4. The button remains greyed out.

**Expected behavior**
I expect to be able to delete the course by following the verification instructions.

**Screenshots**
image

 - Windows 10
 - Not browser specific
 - Version 0.10.0 (4D54FC6) 06/07/2021

darrensiegel:
Closes #1131 

There was a collection of encoding/decoding issues, both in the server code and client, preventing this from working.  The most robust way to allow it to work is to have both input and title strings be base 64 encoded - from the server rendering through to the comparison in TS code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Bug fixes
+
+- Fix an issue preventing deletion of projects whose names contain special characters
+
 ## 0.11.0 (2021-6-15)
 
 ### Enhancements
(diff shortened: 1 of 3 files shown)


bsparks (approved):
I really like the simplicity of this solution

PR #1172, opened 2021-06-17
[BUGFIX] small world issues

dtiwarATS:
-- added support for variables and global variables.
-- changes for applying layer custom classes to its child screens
-- Check if the first screen is a Layer or QB
-- implement reset activity on initial load

### DIFF
--- a/assets/src/apps/delivery/store/features/adaptivity/actions/triggerCheck.ts
+++ b/assets/src/apps/delivery/store/features/adaptivity/actions/triggerCheck.ts
@@ -27,12 +27,13 @@ export const triggerCheck = createAsyncThunk(
     // reset timeStartQuestion (per attempt timer, maybe should wait til resolved)
     // increase attempt number
     const extrinsicState = selectExtrinsicState(rootState);
-    const modifiedExtrinsicState = [extrinsicState]?.reduce((collect: any, entry: any) => {
-      Object.keys(entry).forEach((key) => {
+    const modifiedExtrinsicState = Object.keys(extrinsicState).reduce(
+      (collect: any, key: string) => {
         collect[key] = extrinsicState[key];
-      });
-      return collect;
-    }, {});
+        return collect;
+      },
+      {},
+    );
     const timeStartQuestion = modifiedExtrinsicState['session.timeStartQuestion'];
     const timeOnQuestion = Date.now() - timeStartQuestion;
     modifiedExtrinsicState['session.timeOnQuestion'] = timeOnQuestion;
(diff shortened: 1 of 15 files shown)


assets/src/apps/delivery/store/features/adaptivity/actions/triggerCheck.ts
```diff
@@ -91,7 +90,22 @@ export const triggerCheck = createAsyncThunk(
         }
       });
     });
-    const stateSnapshot = { ...allResponseState, ...modifiedExtrinsicState };
+
+    const spanshot = getEnvState(defaultGlobalEnv);
```
bsparks, line 94: p and n are transposed on "spanshot" vs "snapshot"

assets/src/apps/delivery/store/features/adaptivity/actions/triggerCheck.ts
```diff
@@ -91,7 +90,22 @@ export const triggerCheck = createAsyncThunk(
...
+    const getGlobalSnapshot = [spanshot].reduce((collect: any, variableSnappshot: any) => {
```
bsparks, line 95: you don't need to wrap the object into an array for this reduce, just start with
`const globalSnapshot = Object.keys(snapshot).reduce((collect: any, key: string) => {
  if (key.indexOf('app.') === 0 || key.indexOf('variables.') === 0) {
          collect[key] = snapshot[key];
        }
return collect;
});
```

File: assets/src/apps/delivery/store/features/adaptivity/actions/triggerCheck.ts [hunk 3]
```diff
@@ -91,7 +90,22 @@ export const triggerCheck = createAsyncThunk(
...
+      Object.keys(spanshot).forEach((key) => {
+        if (key.indexOf('app.') === 0 || key.indexOf('variables.') === 0) {
+          collect[key] = spanshot[key];
+        }
+      });
+      return collect;
+    }, {});
+    const updatedExtrinsicState = { ...modifiedExtrinsicState, ...getGlobalSnapshot };
+    await dispatch(setExtrinsicState({ state: updatedExtrinsicState }));
```
bsparks, line 104: I don't think that we want to write variables and app into Extrinsic state, just send the values into the "check" call. the variables are one shot properties of the lesson model, and the app ones are from user state. when we implement user state storage then we can collect the values from that (wherever they are stored), but still we should just include them in the call to check, not write them to the lesson attempt

dtiwarATS:
Closing the PR and will create another PR with more changes

PR #1174, opened 2021-06-17
PMP-1470 - Slider is not of right size

schowkekarats360:


bsparks:
closing for now until part components with external css files strategy is determined

PR #1189, opened 2021-06-23
[Feature] Carousel & FIB Part component

dtiwarATS:
-- contains changes for FIB and Carousel part components.
-- there are changes in the webpack.config that was needed to handle .css files of the part component
-- Contains migration script for removing janus-carousel part component as this is replaced by janus-image-carousel.


dtiwarATS:
Closing this PR as there might not be a need for webpack changes to handle the part component css files.

PR #1193, opened 2021-06-24
Vendor script

darrensiegel:
This restores the generation of `vendor.js`.  Oddly, though, I had to leave in the `shared.js` which bundles just `react` and `react-dom`.  As soon as I removed it I got the React hook error again that complained that multiple copies of React were loaded.


bsparks (approved):

bsparks (changes requested):
im not sure if this is needed anymore, per slack discussion

PR #1196, opened 2021-06-25
Navigation button issue

schowkekarats360:


assets/src/components/parts/janus-navigation-button/NavigationButton.tsx
```diff
@@ -137,7 +137,7 @@ const NavigationButton: React.FC<any> = (props) => {
     if (sSelected === undefined) {
       sSelected = currentStateSnapshot[`stage.${id}.selected`];
     }
-    if (sSelected !== undefined) {
+    if (sSelected !== undefined && typeof sSelected === 'boolean') {
```
bsparks, line 140: should we instead use the `parseBoolean` function in case some value like `"false"` or `1` comes in?

bsparks (changes requested):

assets/src/components/parts/janus-navigation-button/NavigationButton.tsx
```diff
@@ -137,9 +138,7 @@ const NavigationButton: React.FC<any> = (props) => {
     if (sSelected === undefined) {
       sSelected = currentStateSnapshot[`stage.${id}.selected`];
     }
-    if (sSelected !== undefined) {
-      setButtonSelected(sSelected);
-    }
+    setButtonSelected(parseBoolean(sSelected));
```
bsparks, line 141: why are we moving out of the `!== undefined` test for these? it should only be set if it has some value, otherwise it will be reset to false unless something specifically wants it to be true

schowkekarats360: updated.

bsparks (approved):

PR #1202, opened 2021-06-25
[FEATURE] Implement remaining command-driven undos for activities

### LINKED ISSUE #1197: Implement remaining activity undoables
This task is to implement the remaining undo commands across all current activities.

darrensiegel:
Closes #1197 

This PR adds in generation of Undoables for all the rest of the activities, in all the places that they introduce destructive edits. 

Undoing removal of choices in CATA and Ordering is complex due to the post processing that occurs in these action impls. To simplify this, I added a new undo operation type: `ReplaceOperation` that simply replaces a matched node with another.  This allows these undos to simply restore entire sub-trees of the activity model.  There is some potential weirdness that can occur, however, with this approach.  Imagine this scenario, with a CATA:

1. User deletes a choice (A).  An undo is generated that restores the state of choices prior to this deletion. 
2. User deletes another choice (B). An undo is generated that restores the state of choices prior to this deletion. 
3. User clicks "Undo" on the undo from A... which then restores both choice A and B. 

This is a bit of an edge case, and isn't terrible in the grand scheme of things, but it is worth acknowleding. 

We could eliminate this by introducing a "scrubbing" step that the `ResourceEditor` would perform.  It could work like this: any time that a new undo is posted, the editor looks to see if there are any other undos for that same activity.  If there is, and either of these undos (the existing and the new) contain at least one `ReplaceOperation`, then the existing one is removed so that it cannot be invoked "out of order" and lead to unexpected results.  For now I think it makes sense to ship this without this optimization.

### DIFF
--- a/assets/src/components/activities/types.ts
+++ b/assets/src/components/activities/types.ts
@@ -1,7 +1,7 @@
 import { create, ID, Identifiable, ModelElement, Selection } from 'data/content/model';
 import { ResourceContext } from 'data/content/resource';
 import { ResourceId } from 'data/types';
-import { InsertOperation } from 'utils/undo';
+import { UndoOperation } from 'utils/undo';
 import guid from 'utils/guid';
 
 export type PostUndoable = (undoable: Undoable) => void;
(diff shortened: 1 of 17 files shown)


assets/src/components/activities/types.ts
```diff
@@ -1,17 +1,28 @@
...
 
+export function makeUndoable(description: string, operations: UndoOperation[]) : Undoable {
+  return {
+    type: 'Undoable',
+    description,
+    operations
+  };
+}
+
+export function clone(o : any) {
```
bsparks, line 23: I often find myself doing this, should we put it in utils/common or somewhere similar instead?

darrensiegel: Yes, agreed.

PR #1204, opened 2021-06-28
[BUGFIX] Torus player hardening contd...

dtiwarATS:


assets/src/components/parts/janus-mcq/MultipleChoiceQuestion.tsx
```diff
@@ -89,6 +94,7 @@ const MCQItem: React.FC<JanusMultipleChoiceQuestionProperties> = ({
   };
 
   return (
+    <React.Fragment>
```
bsparks, line 97: seems like indent formatting is not there

bsparks (approved):

PR #1209, opened 2021-06-29
[FEATURE] Redesign multiple choice activity

LEDThereBeLight:
Move check all that apply changes over to multiple choice activities.

This will conflict with #1202 

There’s an issue with the bootstrap tabbed navigation now that multiple editors are on the same page, since it routes by URL/unique ID. I’ll need to fix this before we move the activity redesigns to production.


assets/src/components/activities/common/choices/authoring/choiceActions.ts
```diff
@@ -0,0 +1,42 @@
...
+    };
+  },
+
+  removeChoice(id: string) {
+    return (model: HasChoices, post: PostUndoable) => {
+      const choice = getChoice(model, id);
+      const index = getChoices(model).findIndex((c) => c.id === id);
+      model.choices = model.choices.filter((choice) => choice.id !== id);
+
+      post({
```
darrensiegel, line 29: To allow `ChoiceActions` to be reusable in contexts where the removal of a choice also requires the removal of feedback (or execution of other logic), `ChoiceActions.removeChoice` cannot post an undo.  It should instead simply return the choice that has been removed, perhaps also the index that it existed at.  This will then allow the user of `ChoiceActions` to implement and post the correct undoable.

assets/src/components/activities/common/hints/authoring/hintActions.ts
```diff
@@ -0,0 +1,38 @@
...
+  removeHint(id: string) {
+    return (model: HasHints, post: PostUndoable) => {
+      const hint = getHint(model, id);
+      const index = getHints(model).findIndex((h) => h.id === id);
+      model.authoring.parts[0].hints = getHints(model).filter((h) => h.id !== id);
+      post({
+        description: 'Removed a hint',
+        operations: [
+          {
+            path: '$.authoring.parts[0].hints',
```
darrensiegel, line 29: The fact that this is part 0 should not be hardcoded here. In fact, almost none of this path makes sense to hardcode. This will make this component unusable in multi-part activities, and in activities that have non-standard structuress.  An easy way to address this is to allow this path to be passed in to this component as a prop.

darrensiegel (changes requested):

assets/src/components/activities/common/choices/authoring/choiceActions.ts
LEDThereBeLight: This is the problem with using immer without redux backing it, and is one of the reasons I had spent time moving things over to use redux. Immer only lets you mutate a model `draft` or return an entirely new model immutably, but you can't return anything if you mutate the `draft`. So I couldn't find a good way to compose actions when further updates need to know what changes were made.

The trouble with moving to redux was that with the way the models are set up (relational data without being normalized), all of the redux actions to update models were super complicated and depended on the model structure for each activity. The actions were composable, but it was hard to follow the logic when an update made changes to multiple parts of the model.

I played around with a normalization/denormalization step as part of loading/saving, which allowed the actions to be composable and pretty simple, but the normalization had to be run with each keystroke edit, and it forced writing custom `fromPersistence`/`toPersistence` logic for each part of the model.

Bottom line, I'm not sure how to make these actions re-usable. I pushed a change so that the calling actions (in MC and CATA) get the choice and index before they remove the choice, and then use it to post the update. But there's duplication.

assets/src/components/activities/common/hints/authoring/hintActions.ts
LEDThereBeLight: I'm not sure what the best way to fix this is. Immer makes mutable updates, so the hints actions have to update the property of an object, which means you can't just pass a string path for the `PostUndoable`, you have to have access to the parent object where the hints are stored. So the part ID could be passed as an argument, but these hint actions aren't useable for non-standard structures. 

I could back out these changes and have every activity write its own hint update logic, but I couldn't figure out a way to fix all the duplication that we had without moving to redux (see comment above, this created its own problems).

assets/src/components/activities/common/choices/authoring/choiceActions.ts
darrensiegel: Elevating the logic to trigger the post of the undo to where you did makes sense.  There is a bit of logic duplication here, I admit, but it keeps the actions more reusable.  The only problem I see is that the undo of removing a choice isn't complete: it also needs to undo removing the corresponding response.

assets/src/components/activities/common/hints/authoring/hintActions.ts
darrensiegel: I think the easiest way would be to expose a prop on `HintsAuthoring` called "hintPath" that the MCQ Authoring component would set to `"$.authoring.parts[0].hints"`.  That would have to get passed all the way into the code then that creates and invokes the `post` method.

darrensiegel (approved):

PR #1210, opened 2021-06-29
Feedback not displaying all content

schowkekarats360:
In this scenario we have both text an image in the feedback. The way feedback styling is handled, the width of the text is overwritten with 'auto' in other most of the cases. but in here as there is an image its width is coming from json and its going out of bound. Setting the display of div (.content) seems to be handling this issue.


assets/src/apps/delivery/layouts/deck/DeckLayoutFooter.tsx
```diff
@@ -431,7 +431,7 @@ const DeckLayoutFooter: React.FC = () => {
               </button>
             </div>
             <style type="text/css" aria-hidden="true" />
-            <div className="content">
+            <div className="content" style={{display:'inline-table'}}>
```
bsparks, line 434: im not sure this is always going to be good? I think in spr it applies the width and height perhaps

bsparks (approved):

PR #1215, opened 2021-07-01
[BUG] SIM issue - Progress bar 

dtiwarATS:
-- contains fix for progress bar not working in multiple lessons

### DIFF
--- a/assets/src/components/parts/janus-capi-iframe/ExternalActivity.tsx
+++ b/assets/src/components/parts/janus-capi-iframe/ExternalActivity.tsx
@@ -325,20 +325,23 @@ const ExternalActivity: React.FC<any> = (props) => {
         return false;
       }
       const internalValue = externalActivityMap.get(ms);
+      // mineValue is the value that was passed on to this part component
+      // internalVal is the value that is stored locally in key-value pair for value changes comparions
       let mineValue = vars[ms];
-      let internalVal: any = internalValue?.value;
+      let internalVal = internalValue?.value;
+      const typeOfMS = typeof vars[ms];
+      const typeofInternalVal = typeof internalVal;
 
-      if (ms.type === CapiVariableTypes.BOOLEAN && typeof internalVal === 'string') {
+      //handle case where internalVal = 'true' and mineValue =true
+      if (ms.type === CapiVariableTypes.BOOLEAN && typeofInternalVal === 'string') {
         mineValue = JSON.stringify(mineValue);
       }
-      if (ms.type === CapiVariableTypes.NUMBER && typeof internalVal === 'string') {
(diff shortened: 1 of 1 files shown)


assets/src/components/parts/janus-capi-iframe/ExternalActivity.tsx
```diff
@@ -318,6 +318,56 @@ const ExternalActivity: React.FC<any> = (props) => {
...
+        Array.isArray(vars[ms]) &&
+        typeof internalVal === 'object' &&
+        Array.isArray(internalVal)
+      ) {
+        return JSON.stringify(internalVal) !== JSON.stringify(mineValue);
+      }
+      if (mineValue == '' && internalVal == null) return false;
+      return !internalValue || internalVal != mineValue;
+    });
+    if (interested.length) {
```
bsparks, line 362: you don't need this branch here, the reduce of empty length array is going to return `{}` also

assets/src/components/parts/janus-capi-iframe/ExternalActivity.tsx
```diff
@@ -318,6 +318,56 @@ const ExternalActivity: React.FC<any> = (props) => {
...
+        return false;
+      }
+      const internalValue = externalActivityMap.get(ms);
+      let mineValue = vars[ms];
+      let internalVal: any = internalValue?.value;
+
+      if (ms.type === CapiVariableTypes.BOOLEAN && typeof internalVal === 'string') {
+        mineValue = JSON.stringify(mineValue);
+      }
+      if (ms.type === CapiVariableTypes.NUMBER && typeof internalVal === 'string') {
```
bsparks, line 334: `typeof internalVal === 'string'` is used several times, good to factor into a variable at the top

assets/src/components/parts/janus-capi-iframe/ExternalActivity.tsx
```diff
@@ -318,6 +318,56 @@ const ExternalActivity: React.FC<any> = (props) => {
     }
   };
 
+  const getInterestedVars = (vars: any) => {
+    const interested = Object.keys(vars).filter((ms: any) => {
+      const isMine = ms.indexOf(`stage.${id}.`) === 0;
+      if (!isMine) {
+        return false;
```
bsparks, line 325: this is a fairly big method with multiple return types (boolean here, object elsewhere) that makes it difficult to reason about. perhaps some of it can be factored into smaller functions. also if there can be one return type that would help

bsparks: nevermind the multiple return types, I see that is a filter, maybe that shows however that its big and easy to get lost :D

bsparks (approved):

PR #1218, opened 2021-07-02
Missing Purple Question Icon

schowkekarats360:


assets/src/apps/delivery/components/PartsLayoutRenderer.tsx
```diff
@@ -27,9 +27,13 @@ const PartsLayoutRenderer: React.FC<PartsLayoutRendererProps> = ({
   onPartSave = defaultHandler,
   onPartSubmit = defaultHandler,
 }) => {
+  const popups = parts.filter((part) => part.type === 'janus-popup');
+  const partsWithoutPopups = parts.filter((part) => part.type !== 'janus-popup');
+
+  const updatedParts = [...partsWithoutPopups, ...popups];
```
bsparks, line 33: I'm confused as to what this is doing, because it looks like you are splitting 2 groups from the array and then combining them back making it the same array.

bsparks (approved):

PR #1238, opened 2021-07-08
[BUG FIX] fix error in multiple choice submission

LEDThereBeLight:
Anders noticed a problem with multiple choice questions not evaluating correctly in preview mode. There were two issues:

1. The MC "add choice" logic was wrong. It created new feedback for each added choice, when they should just fall under the catch-all response.
2. The `activity_bridge` wasn't handling evaluation errors and threw instead of passing the evaluation error to the activity evaluation renderer which already handles these errors. 

To test:
1. Create a MCQ
2. Toggle the second choice to "Correct" under the answer key tab
3. Preview question, select the first choice (the incorrect choice which was correct originally when the question was created)
4. The submission should work (it didn't before).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 - Fix an issue preventing deletion of projects whose names contain special characters
 - Fix an issue related to persisting sessions across server restarts
 - Fix an issue where modals and rearrange were broken in curriculum view
+- Fix an issue where toggling multiple choice answer correctness could cause submission failures
 
 ## 0.11.0 (2021-6-15)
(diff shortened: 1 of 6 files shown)


LEDThereBeLight:
@andersweinstein This PR is a stop-gap to prevent the "processing error" from happening with new questions. The problem was that multiple choice activities don’t have a catch-all response, and changing the correct answer could put you in a state where there was no response to handle certain incorrect answers. 

I have a fix to add catch-all responses to existing MC questions, but it's in my main activity-redesign branch.

PR #756, opened 2021-01-08
Improve lti workflow

bitbldr:
This PR simplifies institution onboarding and LTI configuration by offering a simpler, more streamlined workflow with approval process for both institutions and torus admins. Also includes email support for when an institution has been approved and slack support for when a new request is made. You will need the SLACK_WEBHOOK_URL to test this functionality.

Instructions for configuring an LTI connection are now available on the publish page:
Screen Shot 2021-01-08 at 2 41 35 PM

Unregistered institutions are presented with a registration request form:
Screen Shot 2021-01-08 at 2 42 03 PM

Torus admins will recieve a slack notification with a link to the pending requests page in torus where they can approve or decline a request:
Screen Shot 2021-01-08 at 2 42 37 PM

Closes #636

PR #1241, opened 2021-07-09
[FEATURE] dockerize application

ChristianMurphy:
* add dockerfile to build torus application
* update docker compose such that both the database and the application can be spun up in a single command.
* add a new script that sets up env files for docker-compose environment.

It can be tested locally with

```sh
./configure-env-for-compose.sh
docker-compose up
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,20 @@
+FROM elixir:latest
+
+# use bash as shell
+SHELL ["/bin/bash", "-c"]
+
+# install node.js
+RUN curl -fsSL https://deb.nodesource.com/setup_16.x | bash -
+RUN apt-get update
+RUN apt-get install nodejs --yes
+
+# copy project files
+COPY . .
+
+# install elixir dependencies
+RUN mix local.hex --force
+RUN mix local.rebar --force
+RUN mix deps.get
+
+# install node modules
+RUN cd assets && npm install --legacy-peer-deps && cd ..
--- a/configure-env-for-compose.sh
+++ b/configure-env-for-compose.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+cp postgres.example.env postgres.env
+cp oli.example.env oli.env
+sed -i'' 's/DB_HOST=/# DB_HOST=/g' oli.env
+sed -i'' 's/FRESHDESK_API_URL=/# FRESHDESK_API_URL=/g' oli.env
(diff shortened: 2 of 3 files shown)


malav2110 (approved):
Looks Good. Approving.

darrensiegel:
Thanks for the PR.  

I would like to understand more what is motivating the need for this and what the intended usage is, as this change would be a reversion of a change we made a month ago where we removed full Docker support. 

Is this Docker support intended for use in setting up and running a development environment? If it is, we would also need to update the relevant WIKI page (https://github.com/Simon-Initiative/oli-torus/wiki/Setup-Dev-Environment).

ChristianMurphy:
> I would like to understand more what is motivating the need for this and what the intended usage is

The motivation is to smooth over cross platform differences, https://github.com/Simon-Initiative/oli-torus/wiki/Setup-Dev-Environment appears to work only on MacOS, partially working on Linux, and not at all on Windows.
It also resolves ambiguity as to what platform versions are supported (it appears elixir 1.12 and node 16 are)

More broadly it's a step toward being able to potentially have a one click dev environment setup cross platform with https://code.visualstudio.com/updates/v1_58#_open-in-vs-code-badge and  https://code.visualstudio.com/docs/remote/containers

---

It's also worth noting that adding the application container does not remove support for using docker compose to run just the database.
This can still be done with `docker-compose up postgres`, for developers who prefer to work code locally and do the manual setup.

eliknebel (approved):

ChristianMurphy:
> > Is this Docker support intended for use in setting up and running a development environment? If it is, we would also need to update the relevant WIKI page
>
> Yes and yes.
> Getting the wiki updated would also be in order.

@darrensiegel It appears I don't have access to be able to edit the wiki page https://github.com/Simon-Initiative/oli-torus/wiki/Setup-Dev-Environment

Would it be possible to get access?
Or alternatively, would there be interest in

> Another discussion to be had outside the scope of this PR would be, would it make sense to pull the docs into a folder in the repository?
> So they can be versioned along side the code, instead of running side car on the wiki.

I could move the wiki back into the docs folder, so that docs can be updated in the same PR as code changes, and don't require additional access.
Thoughts? :thought_balloon: 

It appears there may have been some offline discussion around where docs should live as part of https://github.com/Simon-Initiative/oli-torus/pull/756, but I don't have the context on what that discussion was.

PR #1242, opened 2021-07-09
[FEATURE] Redesign Short Answer and Ordering Activities

LEDThereBeLight:
Main changes:

Short answer
- Add rule options for correct/targeted answers based on feedback that the response defaults to "regex" matching but is not clear. Numeric rules also better support the numeric answer input type.

Ordering
- Restyle the delivery component to use drag and drop + keyboard rearrangement, instead of clicking answer options to place into the correct ordering. Feedback was that it was hard to tell what order your answer options were in if the answer choices were more than a few words. Also prevents submission without selecting all answer choices.

Model migrations
- Very simple model versioning, mostly just to eliminate a lot of duplicated code with slightly different logic for each activity. This will make it easier to support partial credit and changing activity types.

Multiple Choice
- Add targeted feedback support

Rules engine
- Fix a bug with the `eq` operator
- Add support for "contains" string matching

Accessibility
- Fix buttons which were not keyboard accessible


assets/src/components/activities/AuthoringElement.tsx
```diff
@@ -51,6 +51,10 @@ export abstract class AuthoringElement<T extends ActivityModelSchema> extends HT
     };
   }
 
+  transformModel(model: any): T {
```
darrensiegel, line 54: I understand this to be the inline model versioning upgrade, but the name "transformModel" is confusing because activities already have a notion of "transformed models" (for delivery purposes).  Can we change this name to be "migrateModelVersion" ? or something similar?

darrensiegel (changes requested):
I pushed one change that fixes submission of activities in delivery mode (the "BadMapError"). This was due to an incorrect implementation by the adaptive activity work to detect when to use adaptive evaluation. 

All in all this implementation looks great, and works well.  There is one gap that I see that we have to address, however: the "transformModel" inline upgrading of model versions that is found in the AuthoringElement framework must exist in the DeliveryElement framework to allow *existing* activities to have their models upgraded.  As noted in my inline comment, we should change the name of "transformModel" to something else.

darrensiegel (approved):

PR #1249, opened 2021-07-10
test: add type coverage checking

ChristianMurphy:
add tooling to check how much typescript has typings provided.

This tool will check type of all identifiers, the type coverage rate = the count of identifiers whose type is not any / the total count of identifiers, the higher, the better.

sets an initial threshold of 84.5% to aim to stay above (code is currently at 84.68%. with 47215 of 55753 identifiers typed).


ChristianMurphy:
Merged as part of https://github.com/Simon-Initiative/oli-torus/pull/1250

PR #1250, opened 2021-07-10
chore: increase type coverage

ChristianMurphy:
builds on #1249 

replaces `any` with a specific type, and addresses warning/errors that arise.

increases coverage from 84.68%. with 47215 of 55753 identifiers covered, to 84.78%  with 47273 of 55755 identifiers covered.


darrensiegel (approved):
This is a great addition, thanks!

PR #1263, opened 2021-07-13
avoid node json serialization issues by using the original value

bsparks:


darrensiegel (commented):
How big does the "rules" list get?  If it never gets large then the N^2 here isn't an issue and we can land this as-is.

bsparks:
> How big does the "rules" list get? If it never gets large then the N^2 here isn't an issue and we can land this as-is.

well "rules" is all of the rules for the activity, there could be perhaps 20 or 30 for a larger than average activity. the results are often only 1 but sometimes 2-5, much smaller in comparison

darrensiegel (approved):

PR #1267, opened 2021-07-14
PMP-821 - saving screen settings

schowkekarats360:


assets/src/apps/authoring/components/RightMenu/RightMenu.tsx
```diff
@@ -1,19 +1,21 @@
...
-import ContextAwareToggle from '../Accordion/ContextAwareToggle';
 import PropertyEditor from '../PropertyEditor/PropertyEditor';
 import lessonSchema, {
   lessonUiSchema,
   transformModelToSchema as transformLessonModel,
   transformSchemaToModel as transformLessonSchema,
 } from '../PropertyEditor/schemas/lesson';
-import screenSchema, { getScreenData, screenUiSchema } from '../PropertyEditor/schemas/screen';
+import screenSchema, {
+  transformScreenModeltoSchema as transformScreenModel,
```
bsparks, line 16: dont really need to rename these, the only reason I did above is because in the lesson file I just called them `transformModelToSchema` without the word Lesson in it. (since that was in the context of the filename). but if you are going to include "Screen" in the function name then it wouldn't be ambiguous. alternatively you could rename the exported method to just be `transformModelToSchema` and then rename it here

assets/src/apps/authoring/components/RightMenu/RightMenu.tsx
```diff
@@ -36,9 +37,30 @@ const RightMenu: React.FC<any> = (props) => {
...
   const screenPropertyChangeHandler = (properties: any) => {
-    console.log('SCREEN PROP CHANGED', { properties });
+    const modelChanges = transformScreenSchema(properties);
+    console.log(modelChanges);
+    const screenChanges = {
+      ...currentActivity?.content?.custom,
+      ...modelChanges
+    };
+    if(currentActivity){
+      const clone = JSON.parse(JSON.stringify(currentActivity));
```
bsparks, line 47: I think just yesterday a common function was added for this
https://github.com/Simon-Initiative/oli-torus/blob/master/assets/src/utils/common.ts#L21
(maybe rename to `const clonedActivity` since the method is called `clone`)

assets/src/apps/authoring/components/PropertyEditor/schemas/screen.ts
```diff
@@ -116,16 +116,32 @@ export const screenUiSchema = {
...
+  return {
+    width: schema.Size.width,
+    height: schema.Size.height,
+    customCssClass: schema.customCssClass,
+    combineFeedback: schema.combineFeedback,
+    showCheckBtn:schema.checkButton.showCheckBtn,
+    checkButtonLabel:schema.checkButton.checkButtonLabel,
+    maxAttempt:schema.max.maxAttempt,
+    maxScore:schema.max.maxScore,
+    palette: schema.palette,
```
bsparks, line 140: I think in here is where we want to force the `useHtmlProps`
```suggestion
    palette: { useHtmlProps: true, ...schema.palette },
```

assets/src/apps/authoring/components/PropertyEditor/schemas/screen.ts
```diff
@@ -116,16 +116,32 @@ export const screenUiSchema = {
   },
 };
 
-export const getScreenData = (data: any) => {
+export const transformScreenModeltoSchema = (data: any) => {
```
bsparks, line 119: in this method we need to do the conversion from the old properties
so the schema palette should only have the backgroundColor etc type properties as it does, but need to test
```
if (!data.palette.useHtmlProps) {
schema.palette.backgroundColor = `rgba(${
                    data.palette?.fillColor ||
                    data.palette?.fillColor === 0
                        ? chroma(data.palette?.fillColor).rgb().join(',')
                        : '255, 255, 255';
}
```
etc... this is similar to the code found here:
https://github.com/Simon-Initiative/oli-torus/blob/master/assets/src/apps/delivery/layouts/deck/DeckLayoutView.tsx#L398

assets/src/apps/authoring/components/PropertyEditor/schemas/screen.ts
```diff
@@ -116,16 +117,48 @@ export const screenUiSchema = {
...
+        : '255, 255, 255'
+        },${data.palette.fillAlpha})`
+    }
     return {
       ...data,
-      useHtmlProps: data.palette.useHtmlProps,
       Size: { width: data.width, height: data.height },
       checkButton: { showCheckBtn: data.showCheckBtn, checkButtonLabel: data.checkButtonLabel },
       max: { maxAttempt: data.maxAttempt, maxScore: data.maxScore },
+      palette: schemaPalette
```
bsparks, line 142: good but we only want to override these if useHtmlProps is falsey. that would mean that it hasn't yet come into this editor before (because we're going to flip it to true on the way out). So, we only want to convert the values if that is the case, because otherwise it won't retain the settings they've chosen for backgroundColor directly.

bsparks (approved):

darrensiegel (approved):

PR #1278, opened 2021-07-19
ability to edit the Activity Title

schowkekarats360:
saving title for activity is done. have doubt about syncing it with the sequence name.
Do we have a method to update the sequence(I saw one for adding sequence item but not for updating the existing sesquence) or do I need to write a new one??

### DIFF
--- a/assets/src/apps/authoring/components/RightMenu/RightMenu.tsx
+++ b/assets/src/apps/authoring/components/RightMenu/RightMenu.tsx
@@ -1,3 +1,5 @@
+import { findInSequence } from '../../../delivery/store/features/groups/actions/sequence';
+import { selectCurrentSequenceId, selectSequence } from '../../../delivery/store/features/groups/selectors/deck';
 import { JSONSchema7 } from 'json-schema';
 import { debounce } from 'lodash';
 import React, { useCallback, useEffect, useState } from 'react';
@@ -39,16 +41,18 @@ const RightMenu: React.FC<any> = () => {
   const currentActivity = useSelector(selectCurrentActivity);
   const currentLesson = useSelector(selectPageState);
 
+  const currentSequenceId = useSelector(selectCurrentSequenceId);
+  const sequence = useSelector(selectSequence);
   // TODO: dynamically load schema from Part Component configuration
   const componentSchema: JSONSchema7 = { type: 'object' };
   const currentComponent = null;
 
   const [screenData, setScreenData] = useState(
-    transformScreenModeltoSchema(currentActivity?.content?.custom),
+    transformScreenModeltoSchema(currentActivity),
   );
   useEffect(() => {
     console.log('CURRENT', { currentActivity, currentLesson });
-    setScreenData(transformScreenModeltoSchema(currentActivity?.content?.custom));
+    setScreenData(transformScreenModeltoSchema(currentActivity));
   }, [currentActivity]);
 
   // should probably wrap this in state too, but it doesn't change really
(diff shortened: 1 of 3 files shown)


bsparks:
> saving title for activity is done. have doubt about syncing it with the sequence name.
> Do we have a method to update the sequence(I saw one for adding sequence item but not for updating the existing sesquence) or do I need to write a new one??

You might want to create a new one, so that it can do more of the sequence logic based on activity, however the sequence itself is saved using `savePage`. The model exists in the group and `savePage` will always pull the group info out of redux and write it. So you really just need to make changes to the item and then `upsertGroup` and then call `savePage` to write it. That being said though, it might make sense to create a `updateSequenceItem` async thunk method that you pass the activity resource Id,  sequenceName (and perhaps other values) to. It would then handle the part of finding the item in the sequence and upserting the group, and then call `savePage` at the end. Just so the editing menu can have a simpler implementation.

assets/src/apps/authoring/components/RightMenu/RightMenu.tsx
```diff
@@ -62,14 +66,23 @@ const RightMenu: React.FC<any> = () => {
...
+        delete modelChanges.title;
         const screenChanges = {
           ...currentActivity?.content?.custom,
           ...modelChanges,
         };
         const cloneActivity = clone(currentActivity);
         cloneActivity.content.custom = screenChanges;
+        if(title){
+          cloneActivity.title = title;
+          cloneActivity.activitySlug = title.toLowerCase().split(' ').join('_');
```
bsparks, line 83: the slug should be generated on the server, need to check if the new slug comes back from the save (edit) call or not. if it does, then we need to update the redux item in `saveActivity` if it doesn't then will need to look into adding that server side I think

assets/src/apps/authoring/components/RightMenu/RightMenu.tsx
```diff
@@ -62,14 +66,23 @@ const RightMenu: React.FC<any> = () => {
   const screenPropertyChangeHandler = useCallback(
     (properties: any) => {
       if (currentActivity) {
+        const currentSequence = findInSequence(sequence, currentSequenceId);
```
bsparks, line 69: this stuff is what I'm talking about moving into a separate thunk, so that below it can be like
```
if (titleChanged) {
  dispatch(updateSequenceItemFromActivity({ activity: cloneActivity}));
}
```
so this method only needs to know that the sequence needs to be updated, not the details of *how*

assets/src/apps/authoring/store/groups/layouts/deck/actions/updateSequenceItemFromActivity.ts
```diff
@@ -0,0 +1,24 @@
...
+  `${GroupsSlice}/updateSequenceItemFromActivity`,
+  async (
+    payload: {
+      activity?: any;
+      group?: any;
+    },
+    { dispatch },
+  ) => {
+    const { activity = {}, group = {} } = payload;
+    const colnedGroup = clone(group);
```
bsparks, line 15:
```suggestion
    const clonedGroup = clone(group);
```

assets/src/apps/authoring/store/groups/layouts/deck/actions/updateSequenceItemFromActivity.ts
```diff
@@ -0,0 +1,24 @@
...
+    const sequencyEntry = colnedGroup.children.find(
```
bsparks, line 16:
```suggestion
    const sequenceEntry = clonedGroup.children.find(
```

assets/src/apps/authoring/store/groups/layouts/deck/actions/updateSequenceItemFromActivity.ts
```diff
@@ -0,0 +1,24 @@
...
+      (entry: any) => entry.resourceId === activity.resourceId,
+    );
+    sequencyEntry.custom.sequenceName = activity.title;
```
bsparks, line 19:
```suggestion
    sequenceEntry.custom.sequenceName = activity.title;
```

assets/src/apps/authoring/store/groups/layouts/deck/actions/updateSequenceItemFromActivity.ts
```diff
@@ -0,0 +1,24 @@
...
+    dispatch(upsertGroup({ group: colnedGroup }));
```
bsparks, line 20:
```suggestion
    dispatch(upsertGroup({ group: clonedGroup }));
```

schowkekarats360:
New PR is created for this one as there are some test failures.

PR #1286, opened 2021-07-21
[Feature] Lesson History  Part - 2

dtiwarATS:


assets/src/apps/delivery/components/ActivityRenderer.tsx
```diff
@@ -291,10 +294,18 @@ const ActivityRenderer: React.FC<ActivityRendererProps> = ({
...
     ref.current.notify(NotificationType.CONTEXT_CHANGED, {
       currentActivityId,
-      mode: 'VIEWER', // TODO: based on isPreviewMOde
+      mode: historyModeNavigation ? contexts.REVIEW : contexts.VIEWER,
       snapshot,
     });
   };
+  const historyNavigationActivity = useSelector(selectHistoryNavigationActivity);
+
+  useEffect(() => {
```
bsparks, line 303: this doesn't do anything?

assets/src/apps/delivery/layouts/deck/DeckLayoutFooter.tsx
```diff
@@ -1,6 +1,8 @@
 /* eslint-disable react/prop-types */
-import React, { useEffect, useState } from 'react';
+import { cursor } from 'components/misc/resizer/utils';
```
bsparks, line 2: probably accidental import

assets/src/apps/delivery/store/features/activities/slice.ts
```diff
@@ -58,6 +58,23 @@ const slice: Slice<ActivitiesState> = createSlice({
       state.currentActivityId = action.payload.activityId;
     },
   },
+
+  extraReducers: (builder) => {
+    builder.addCase(loadPageState, (state, action) => {
```
bsparks, line 63: hmm why is this coming back? I think this will make a flicker on resume because it will pick the first item then the navigation will send to the previous item

assets/src/utils/common.ts
```diff
@@ -134,3 +134,10 @@ export const parseBool = (val: any) => {
   // have to ignore the false searchValue in 'replace'
   return !isNaN(num) ? !!num : !!String(val).toLowerCase().replace('false', '');
 };
+
+export enum contexts {
```
bsparks, line 138: this would probably be better in a different file, but not totally sure where, for now it's fine, but it's not really a *util* more of an application context

bsparks (changes requested):
I think just remove the yarn.lock changes they look like they are undoing a recent update

bsparks (approved):

PR #1293, opened 2021-07-22
[BUG] ITTAC Lesson scoring issue + fixes mutate state evaluation

dtiwarATS:


assets/src/apps/delivery/layouts/deck/DeckLayoutFooter.tsx
```diff
@@ -117,6 +117,30 @@ const DeckLayoutFooter: React.FC = () => {
   const [nextButtonText, setNextButtonText] = useState('Next');
   const [nextCheckButtonText, setNextCheckButtonText] = useState('Next');
 
+  const handleValueExpression = (operationValue: string) => {
```
bsparks, line 120: lets separate this function out and export it at the top (passing in the currentActivityTree) so that we can unit test it with various types of values. I worry about the variable parsing.

assets/src/adaptivity/rules-engine.ts
```diff
@@ -147,27 +151,38 @@ export const check = async (
   /* console.log('RE CHECK', { checkResult }); */
   let resultEvents: Event[] = [];
   const successEvents = checkResult.events.sort((a, b) => a.params?.order - b.params?.order);
+
   // if there are any correct in the success, get rid of the incorrect (defaultWrong most likely)
   const isCorrect = successEvents.some((evt) => evt.params?.correct === true);
-  if (isCorrect) {
-    resultEvents = successEvents.filter((evt) => evt.params?.correct === true);
+  //These are correct rules that do not have any condition so we need to make sure that the result does not have a wrong trap state otherwise it will
```
bsparks, line 157: hmm maybe we can split the rules stuff into a separate PR because I am not sure about this

dtiwarATS: A separate PR created for value expression - https://github.com/Simon-Initiative/oli-torus/pull/1297

PR #1295, opened 2021-07-22
WIP - [PMP-1594, PMP-1595] Creating unit tests for lesson and screen settings

schowkekarats360:
Create Unit Tests


assets/test/advanced_authoring/right_menu/lesson/lesson_mocks.ts
```diff
@@ -0,0 +1,102 @@
...
+  resourceId: 93,
+  advancedAuthoring: true,
+  advancedDelivery: true,
+  displayApplicationChrome: false,
+  additionalStylesheets: [
+    'default',
+    'https://etx-nec.s3-us-west-2.amazonaws.com/css/etx/styles/style_season.css',
+  ],
+  customCss:
+    '',
```
bsparks, line 65: formatting seems off

assets/test/advanced_authoring/right_menu/screen/screen_mocks.ts
```diff
@@ -0,0 +1,66 @@
...
+    custom: {
+      applyBtnFlag: false,
+      applyBtnLabel: '',
+      combineFeedback: false,
+      facts: [],
+      lockCanvasSize: false,
+      mainBtnLabel: '',
+      negativeScoreAllowed: false,
+      palette: {
+        useHtmlProps: true,
```
bsparks, line 46: I think we expect the palette to always have the backgroundColor, etc. when useHtmlProps is true

assets/test/advanced_authoring/right_menu/screen/screen_mocks.ts
```diff
@@ -0,0 +1,66 @@
+import { IActivity } from "apps/delivery/store/features/activities/slice";
+
+export const screenSchema = {
```
bsparks, line 3: I think there are maybe really only 2 objects we want to mock actually, one is the actual screen data, this is what we transform into something that matches up with the schema in order to render the property fields. The other would be a "changes" object. We'd be simulating the changes being made on the edit form, and those are what get transformed  into model compatible format. we should probably have multiple variations of changes, simulating different things being "edited" and testing that the model version comes out of the method as expected. also it would be good to have a couple of palette versions to make sure that the conversion from legacy properties is happening

assets/test/advanced_authoring/right_menu/screen/screen_mocks.ts
```diff
@@ -0,0 +1,64 @@
...
+        borderStyle: "solid",
+        borderWidth: "0.1px",
+        fillAlpha: 1,
+        fillColor: 16777215,
+        lineAlpha: 1,
+        lineColor: 16777215,
+        lineStyle: 0,
+        lineThickness: 0.1,
+        borderRadius: "10px",
+        useHtmlProps: false,
```
bsparks, line 44: think about the logic that is happening, and create scenarios to test it. so with the palette for example, what we want to happen is: if it encounters a falsey `useHtmlProps` then we want to convert it using the "old" properties like `fillColor` etc into the `backgroundColor`. If `useHtmlProps` is truthy then we want to simply use (and expect that they are there) the `backgroundColor` etc properties and ignore the others. We want to permanently convert them every time. So when it comes out of the editor after they make *any* change then it should have `useHtmlProps` set to true, and the `backgroundColor` and related properties should be filled out. We want to save that, so that next time they come in, it will not do the conversion anymore.

So we need to create multiple versions of screen data here, some that have `useHtmlProps` as falsey and some as truthy. The tests should make sure that the functions behave in line with the business rules above.

not only that, but the `transformScreenModeltoSchema` method transforms more than just the palette, so it should check that passing in a screen data custom object returns a properly formed schema object.

schowkekarats360: I have implemented the palette related explanation you have mentioned. and other properties are covered with the method i have used i.e. toMatchObject(), which checks for the properties and its values.

bsparks (approved):

PR #1298, opened 2021-07-23
[BUG] PMP-1579

dtiwarATS:


assets/src/adaptivity/scripting.ts
```diff
@@ -177,7 +177,9 @@ export const applyState = (
               } else {
                 // it's not supposed to be a string however, so we need to coerce it
                 const coerced = coerceCapiValue(newValue, targetType);
-                script += `= ${coerced};`;
+                // need to wrap the content in "" as it's a string else it will be evaluated as undefined as the script
+                //generates the expression as 'let {stage.something.value} = Finches'
+                script += `= "${coerced}";`;
```
bsparks, line 182: since in this branch it's not supposed to be a string, my guess is that maybe its an ENUM? I think we should check
if the type of `coerced` is a string *then* wrap it in quotes, we wouldn't want to wrap numbers at this point

assets/src/adaptivity/capi.ts
```diff
@@ -59,6 +59,9 @@ export const coerceCapiValue = (
           );
         }
       }
+      if (typeof value === 'string') {
```
bsparks, line 62: we need to do this only in the scripting, this method is supposed to return the actual value, here it would be returning a sort of stringified version. the enum will always be a string already.

bsparks:
will take care of this in another branch today

PR #1303, opened 2021-07-26
[BUG] PMP-1580

dtiwarATS:


assets/src/adaptivity/operators/contains.ts
```diff
@@ -5,6 +5,14 @@ export const parseNumString = (item: string): string | number => {
   return !Number.isNaN(parseFloat(item)) ? parseFloat(item) : item.toString().trim();
 };
 
+export const parseArrayString = (value: (string | number)[] | string): (string | number)[] => {
```
bsparks, line 8: in utils/common I think we already have a stringy array parser? we can probably move `parseNumString` into there as well. also lets start getting into the habit of writing unit tests for these as we add them

assets/src/apps/delivery/layouts/deck/DeckLayoutFooter.tsx
```diff
@@ -59,7 +59,10 @@ export const handleValueExpression = (
           const ownerActivity = currentActivityTree?.find(
             (activity) => !!activity.content.partsLayout.find((p: any) => p.id === lstVar[1]),
           );
-          value = value.replace(`${item}`, `{${ownerActivity.id}|${modifiedValue}}`);
+          //ownerActivity is undefined for app.spr.adaptivity.something i.e. Beagle app variables
+          if (ownerActivity) {
```
bsparks, line 63: this too is a prime example of why we need unit tests for this method. lets make sure we add all the different types of input and match with expected output

bsparks (approved):

PR #1304, opened 2021-07-26
[bug] PMP-1611 - Focus jumping in the Edit Screen name field

schowkekarats360:


assets/src/apps/authoring/components/PropertyEditor/PropertyEditor.tsx
```diff
@@ -14,6 +15,7 @@ interface PropertyEditorProps {
 
 const widgets = {
   CheckboxWidget: CustomCheckbox,
+  TextWidget: CustomTextBox,
```
bsparks, line 18: the built in TextWidget has a lot of features that I wonder if we're missing out on for this? https://github.com/rjsf-team/react-jsonschema-form/blob/master/packages/bootstrap-4/src/TextWidget/TextWidget.tsx

I wonder if it's really the text widget that is causing this or the change handler for the whole property editor. I also wonder if having this one field only save on blur while all the others are immediate might lead to some minor data loss issues (though I suppose it's not super likely that it'll keep focus until they close the browser, but I think that would be a scenario where it wouldn't save).

assets/src/apps/authoring/components/PropertyEditor/custom/CustomTextbox.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import { useEffect, useState } from 'react';
+
+interface CustomTextboxProps {
+  label: string;
+  id: any;
```
bsparks, line 6:
```suggestion
  id: string;
```

can this be a string? it seems it's used for the dom attr

assets/src/apps/authoring/components/PropertyEditor/custom/CustomTextbox.tsx
```diff
@@ -0,0 +1,29 @@
...
+  value: string;
+}
+const CustomTextbox: React.FC<CustomTextboxProps> = (props: any) => {
```
bsparks, line 9:
```suggestion
const CustomTextbox: React.FC<CustomTextboxProps> = (props) => {
```
the generic props definition already passed in so props will have that `CustomTextboxProps` type already

bsparks:
I tried this one out locally and indeed we lose too much by switching to this text box, numbers are just text and also it doesn't respect disabled. closing will need to search for another solution

PR #1315, opened 2021-07-28
[Bug] PMP-1496: Feedback not displaying all content

schowkekarats360:
I am not sure if this solution has any hidden impact, but as far as i have tested it with couple of lessons, adding display block to the part components tag, hasnt showed any UI related changes. 
Let me know if this approach has any possible issues.


bsparks (approved):

bsparks:
let's try it out :D

PR #1318, opened 2021-07-29
[BUG] Combined Feedback Issue

dtiwarATS:


assets/src/apps/delivery/store/features/groups/actions/deck.ts
```diff
@@ -143,7 +143,8 @@ export const initializeActivity = createAsyncThunk(
       // maybe we don't need to write the local versions ever?? instead just whenever anything
       // is asking for it we can just give the localized snapshot?
       const idsToBeRemoved: any[] = Object.keys(globalSnapshot).filter(
-        (key: string) => key.indexOf('stage.') === 0,
+        (key: string) =>
+          key.indexOf('stage.') === 0 || key.indexOf(`${currentSequenceId}|stage.`) === 0,
```
bsparks, line 147: hmm need to understand this piece better, I just took that off because I found that it was deleting variables that were needed for something (I can't remember specifically, but it would have been in KIP open & free)

bsparks (approved):

PR #1335, opened 2021-07-30
Feature/torus aa integration

dsuriano:
This will need a little more carrying across the finish line, but it's 97% there.

1. On the Torus side, `/projects` needs a blank breadcrumb, and there will probably be some more small spacing adjustments to make below the header on various non-project pages.
2. I still need to remove the `🚀 Launch` button and all the logic that goes with it, but that will come after...
3. Dark mode. Exists in Torus, but not the Advanced Authoring until we have time to make it a priority.


bsparks (approved):

eliknebel (approved):
Looks great

PR #1350, opened 2021-08-02
[BUG FIX] Fix some delivery mode rendering issues

LEDThereBeLight:
- Fix iframe rendering when elements contain captions (webpage, youtube)
- Fix iframe rendering in activities
- Escape activity content item attributes
- ~~**Add max-width for all content in delivery mode** (maybe controversial, but I find it to be too hard to read the pages with as wide as they are)~~
- Remove extra styles not being used now that all images/embeds are "block only" (and can't be floated)

### DIFF
--- a/assets/src/components/editing/models/blockcode/Blockcode.scss
+++ b/assets/src/components/editing/models/blockcode/Blockcode.scss
@@ -1,7 +1,6 @@
 @use 'delivery/variables' as delivery;
 
 .code-editor {
-  clear: both;
   margin-bottom: 10px;
 
   pre {
--- a/assets/src/data/content/writers/html.ts
+++ b/assets/src/data/content/writers/html.ts
@@ -24,6 +24,7 @@ import {
   CodeLine,
   Blockquote,
   Hyperlink,
+  Webpage,
 } from '../model';
 import { Text } from 'slate';
 import { WriterContext } from './context';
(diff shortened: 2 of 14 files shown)


assets/src/components/editing/models/blockcode/Blockcode.scss
```diff
@@ -1,7 +1,6 @@
 @use 'delivery/variables' as delivery;
 
 .code-editor {
-  clear: both;
   margin-bottom: 10px;
```
LEDThereBeLight, line 4: These clears used to be necessary when we were floating images left and right, but we don't support that any more

assets/src/data/content/writers/html.ts
```diff
@@ -61,31 +62,17 @@ export class HtmlParser implements WriterImpl {
       .reduce((acc, mark) => `<${mark}>${acc}</${mark}>`, text);
   }
```
LEDThereBeLight, line 64: `displayClass` was used to float images/etc left and right, and I left the code there in case we decided to add it back in, but it's probably time to just remove it.

assets/src/data/content/writers/html.ts
```diff
@@ -61,31 +62,17 @@ export class HtmlParser implements WriterImpl {
...
-          return 'd-block';
-      }
-    }
-    return '';
-  }
-
   private figure(attrs: any, content: string) {
     if (!attrs.caption) {
       return content;
     }
```
LEDThereBeLight, line 68: wrap figures in bootstrap responsive classes

assets/src/data/content/writers/writer.ts
```diff
@@ -20,6 +16,7 @@ export interface WriterImpl {
   h6: ElementWriter;
   img: ElementWriter;
   youtube: ElementWriter;
+  iframe: ElementWriter;
```
LEDThereBeLight, line 19: iframes were not supported in the html renderer used by activities

eliknebel:
I think we are going to have issues with the max-width. Even the first login page is messed up, but not only that, this is going to be a pervasive issue given our fundamental use of bootstrap and how it is based on the container and grid system.
Screen Shot 2021-08-02 at 8 52 39 PM

We have decided to base our layouts of delivery on bootstrap scaffolding because it is tried and true and highly customizable. We plan to base all our delivery themes on this down the road and having this max-width restriction in place is not going to work.
Screen Shot 2021-08-02 at 8 56 13 PM

I know this is something you feel strongly about but I very much disagree on this ux decision. I think the browser is made to adjust the viewport as necessary for preferences/accessibility and constraining the width in this way breaks the way those are supposed to work. If users wish to have a more constrained view of the content then they should be able to use the reader mode in their browser or resize the window.

assets/src/data/content/writers/writer.ts
eliknebel: good catch

assets/styles/delivery/layout/delivery.scss
```diff
@@ -93,6 +93,7 @@ main.container {
   padding-right: 0;
   padding-left: 0;
   display: block;
+  max-width: 768px;
```
eliknebel, line 96: this should be removed

eliknebel (changes requested):

eliknebel (approved):

PR #1353, opened 2021-08-03
[BUG] MCQs not functioning when "array" is used

dtiwarATS:
-- by mistake I replace 'value' with 'factvalue' in the parsing method which caused this issue


assets/src/adaptivity/operators/equality.ts
```diff
@@ -20,7 +20,7 @@ export const isEqual = (factValue: any, value: any): boolean => {
     if (Array.isArray(value)) {
       // ** We are doing this for the cases where factValue comes [2 , 5] but the values comes as ['2','5'] */
       // ** DT - making sure that value is of array type else value.map() will throw error. */
-      compareValue = parseArrayString(factValue).sort();
+      compareValue = parseArrayString(value).sort();
```
bsparks, line 23: I think there should be a unit test for this case

bsparks (approved):

PR #1354, opened 2021-08-03
[BUG] Fill in the Blanks component rendering issue

dtiwarATS:


assets/src/apps/delivery/layouts/deck/DeckLayoutView.tsx
```diff
@@ -199,9 +199,6 @@ const DeckLayoutView: React.FC<LayoutProps> = ({ pageTitle, pageContent, preview
       setActivityClasses(
         [...new Set([...defaultClasses, ...customClasses])].map((str) => str.trim()),
       );
-    } else if (currentActivity?.content?.partsLayout) {
```
dtiwarATS, line 202: We don't need this entire section. It was done to add vft class based on the Id but now we have getCustomClassAncestry() that does all the things.

I already removed below code but looks like entire if block needs to be removed

```
let hasVft = false;
      currentActivityTree.forEach((activity) => {
        if (!hasVft) {
          hasVft = activity?.content?.partsLayout.some((part: any) => part.id === 'vft');
        }
      });
```

assets/src/components/parts/janus-fill-blanks/FillBlanks.tsx
```diff
@@ -117,7 +121,7 @@ const FillBlanks: React.FC<JanusFillBlanksProperties> = (props) => {
     }
 
     const sCustomCss = currentStateSnapshot[`stage.${id}.customCss`];
-    if (sCustomCss) {
+    if (sCustomCss?.length) {
```
bsparks, line 124: why do we need the length check? even a `''` is falsey already

bsparks (approved):

PR #1355, opened 2021-08-03
[BUG] Lesson History:  Using Screen List doesn't show screen in Read Only Mode

dtiwarATS:


assets/src/apps/delivery/layouts/deck/components/HistoryPanel.tsx
```diff
@@ -26,7 +26,9 @@ const HistoryPanel: React.FC<HistoryPanelProps> = ({ items, onMinimize, onRestar
   const itemClickHandler = (item: HistoryEntry) => {
     dispatch(navigateToActivity(item.id));
 
-    const nextHistoryActivityIndex = items.findIndex((item: any) => item.id === item.id);
+    const nextHistoryActivityIndex = items.findIndex(
+      (historyItem: any) => historyItem.id === item.id,
```
dtiwarATS, line 30: The variable name 'item.id' was matching with the function parameter hence it was not working. changed the variable to historyItem

bsparks (approved):

PR #1363, opened 2021-08-03
[ENHANCEMENT] Redesign overview page

LEDThereBeLight:
Change overview page styling / wording based on feedback that some parts were confusing. Make styling more consistent.


lib/oli_web/live/projects/visibility.ex
```diff
@@ -28,110 +28,105 @@ defmodule OliWeb.Projects.VisibilityLive do
...
+          <div class="form-check form-switch">
+            <div class="form-group">
+              <%= label class: "form-check-label" do %>
+                <%= radio_button :visibility, :option, "authors", class: "form-check-input", checked: @project.visibility == :authors or is_nil(@project.visibility) %>
+                <div class="d-flex align-items-center">
+                  <p>Project authors</p>
+                  <span class="ml-2 badge badge-pill badge-light">default</span>
                 </div>
-              </form>
+                <small>Only authors added as collaborators to this project</small>
```
eliknebel, line 46: Consider rephrasing: "Only instructors who are collaborators and have linked their accounts". We want to try and minimize confusion by using the correct domain roles. From the user delivery domain, it's really instructors who have access

lib/oli_web/live/projects/visibility.ex
```diff
@@ -28,110 +28,105 @@ defmodule OliWeb.Projects.VisibilityLive do
...
+                <%= radio_button :visibility, :option, "global", class: "form-check-input", checked: @project.visibility == :global %>
+                <p>Open</p>
+                <small>Any user who links their account to Proton as an instructor</small>
+              <% end %>
+            </div>
+            <div class="form-group">
+              <%= label class: "form-check-label" do %>
+                <%= radio_button :visibility, :option, "selected", class: "form-check-input", checked:  @project.visibility == :selected %>
+                <p>Restricted</p>
+                <small>Only these users or institutions...</small>
```
eliknebel, line 60: "Only these instructors or institutions..."

eliknebel (changes requested):
This looks really great, just a few things:

1. See comments on terminology
2. As nice as it is, we really should reconsider the "Search for users by email" because it is a security vulnerability. We are essentially leaking emails from the system by giving authors the ability to search. The best we should do here is simply allow a user to enter emails and click invite and show a "Sent" message. We shouldn't tell the users if the users emails exist or not either unfortunately. Searching institutions should be okay, however. This doesn't need to change here though, since it already exists. But we should have a conversation about this.
3. We also should a captcha for sending invite emails in project visibility so that bots don't get ahold of it and spam users. Also does not require a change as part of this PR.

eliknebel (approved):

PR #1383, opened 2021-08-06
Feature/authoring add parts

bsparks:
there is a DB migration in here to update the part component registrations


priv/repo/migrations/20210804999992_refresh_part_components.exs
```diff
@@ -0,0 +1,15 @@
...
+  use Ecto.Migration
+
+  def change do
+    alter table(:part_component_registrations) do
+      add :author, :string
+    end
+
+    flush()
+
+    execute "DELETE FROM part_component_registrations;"
```
eliknebel, line 11: Just want to verify, the intent here is to delete all records from this table?

bsparks: yeah because they are populated by reading the manifest one time, and the updates are to the manifest values it seemed like the easiest way rather than coding something (or copying) the initial manifest code and turning it into a delta

eliknebel (approved):

PR #1384, opened 2021-08-09
[BUG] PMP-1674 - making regex changes

dtiwarATS:


bsparks (approved):

assets/src/adaptivity/scripting.ts
```diff
@@ -47,7 +47,7 @@ export const getExpressionStringForValue = (v: { type: CapiVariableTypes; value:
         try {
           //trying to check if it is a CSS string.This might not handle any advance CSS string.
           const matchingCssElements = val.match(
-            /([#.@]?[\w.:> ]+)[\s]{[\r\n]?([A-Za-z\- \r\n\t]+[:][\s]*[\w .\\/()\-!]+;[\r\n]*(?:[A-Za-z\- \r\n\t]+[:][\s]*[\w .\\/()\-!]+;[\r\n]*(2)*)*)}/gi,
+            /^(([a-z0-9\\[\]=:]+\s?)|((div|span|body.*|.box-sizing:*|.columns-container.*|background-color.*)?(#|\.){1}[a-z0-9\-_\s?:]+\s?)+)(\{[\s\S][^}]*})$/im,
```
bsparks, line 50: actually, we should add new cases to the unit tests for whatever failure this solves

bsparks (changes requested):
add unit test cases pls

dtiwarATS:
new unit test added

assets/test/adaptivity/scripting_test.ts
```diff
@@ -194,7 +194,24 @@ describe('Scripting Interface', () => {
...
... solid 1px #FFC627;border-radius: 100px;margin: 0;padding: 0;top: 50%;transform: translateY(-50%);right: 5px;}.columns-container .right-column .item {border-radius: 100px 4px 4px 100px;margin-right: 0;}.columns-container .right-column .item .item-label:after {content: "";width: 20px;height: 20px;box-sizing: border-box;position: absolute;background-color: transparent;border: solid 1px #FFC627;border-radius: 100px;margin: 0;padding: 0;top: 50%;transform: translateY(-50%);left: 5px;}.columns-container .items-column .item.link-hover {border: solid 1px #FFC627;background-color: #222;padding: 0;box-sizing: border-box;box-shadow: none;}.columns-container .items-column .item[linked] {border: solid 2px #FFC627;background-color: #291b30;box-shadow: none;}.columns-container .items-column .item.preview-source-hover, .columns-container .items-column .item.preview-target-hover {border: solid 3px #ffcc00;background-color: #291b30;box-shadow: none;}.columns-container .items-column .item .remove-links:before {content: "X";margin-left: auto;margin-right: auto;position: relative;}.line-renderer line {stroke: #FFC627;stroke-width: 2px;}.line-renderer line.pointer {stroke: #ffcc00;stroke-width: 2px;}';
```
bsparks, line 197: are these 2 css blocks the same? if so, can we factor out to a variable? also can we describe more precisely what the "it should" is doing? like is it "it should correctly detect CSS values when xyz characters exist in the string" or something like that so we know what exactly the regex supports and doesn't support?

dtiwarATS: These 2 are different CSS blocks. I have refactored the code to clean things and removed repeated things.

bsparks (approved):

PR #1385, opened 2021-08-09
Feature/edit component json

schowkekarats360:


assets/src/apps/authoring/components/PropertyEditor/custom/CustomButtonTemplate.tsx
```diff
@@ -0,0 +1,62 @@
+import React, { CSSProperties, useRef, useState } from 'react';
+interface CustomButtonProps {
+  label: string;
+  onChange: any;
+}
+const CustomButtonTemplate: React.FC<CustomButtonProps> = (props: any) => {
```
bsparks, line 6: can we name this something like "EditJsonButtonTemplate" ? because I feel like this is a bit more specific than a button

bsparks:
actually, nevermind the button comment, I don't think that we need to add it to the model. the enitre model *is* the JSON. I think we can just add a button at the top, not part of the PropertyEditor, but above it. it's not a property that we want to edit, it's the whole component

PR #1389, opened 2021-08-10
[BUG] PMP-1652

dtiwarATS:


assets/src/apps/delivery/layouts/deck/DeckLayoutView.tsx
```diff
@@ -140,9 +140,9 @@ const DeckLayoutView: React.FC<LayoutProps> = ({ pageTitle, pageContent, preview
     // Need to clear out snapshot for the current activity before we send the init trap state.
     // this is needed for use cases where, when we re-visit an activity screen, it needs to restart fresh otherwise
     // some screens go in loop
-    // Don't do anything id enableHistory is ON
+    // Don't do anything id enableHistory/historyModeNavigation is ON
 
-    if (!enableHistory && currentActivityTree) {
+    if (!historyModeNavigation || (!enableHistory && currentActivityTree)) {
```
bsparks, line 145: this is difficult to reason about with the NOT and OR is there another way we can write it? it seems like this will be satisfied a lot more often and currentActivityTree might not even be ready when you go in

PR #1391, opened 2021-08-10
PMP-1688

schowkekarats360:


bsparks:
the branch got merged into master, so I think you can open direct against master now

assets/src/apps/authoring/components/PropertyEditor/custom/CompJsonEditor.tsx
```diff
@@ -0,0 +1,61 @@
+import React, { CSSProperties, useState } from 'react';
+interface JsonEditoProps {
```
bsparks, line 2: typo missing "r"

assets/src/apps/authoring/components/PropertyEditor/custom/CompJsonEditor.tsx
```diff
@@ -0,0 +1,61 @@
...
+  jsonValue: any;
+  onChangeHandler: any;
```
bsparks, line 4: I think usually for the props it's meant to be an event, so just `onChange` here, the method that you pass in from the outside would be the "handler" of that event;

also I think the type is more known for the method
maybe
`onChange: (changedJson: any) => void;`

assets/src/apps/authoring/components/PropertyEditor/custom/CompJsonEditor.tsx
```diff
@@ -0,0 +1,61 @@
...
+}
+const CompJsonEditor: React.FC<JsonEditoProps> = (props: any) => {
+  const { jsonValue, onChangeHandler } = props;
+  const val = { id: jsonValue.id, custom: jsonValue.custom };
+  const [value, setValue] = useState<string>(JSON.stringify(val, null, 4));
+  const popup: CSSProperties = {
```
bsparks, line 10: perhaps we should use a bootstrap modal instead since it already has the backdrop support and whatnot

assets/src/apps/authoring/components/PropertyEditor/custom/CompJsonEditor.tsx
```diff
@@ -0,0 +1,61 @@
...
+          setValue(e.target.value);
+        }}
+      >
+        {value}
+      </textarea>
+      <div className="d-flex justify-content-end">
+        <button
+          style={buttonStyle}
+          className="form-control"
+          onClick={() => onChangeHandler(JSON.parse(value))}
```
bsparks, line 45: this is dangerous because `JSON.parse` can throw exceptions if it's not valid JSON, perhaps we should watch (useEffect) the `value` and do a try/catch on the parse, any time it throws an error display an error message and disable the Save button

assets/src/apps/authoring/components/PropertyEditor/custom/CompJsonEditor.tsx
```diff
@@ -0,0 +1,61 @@
...
+        >
+          Save
+        </button>
+        <button
+          style={buttonStyle}
+          className="form-control"
+          onClick={() => onChangeHandler({}, true)}
```
bsparks, line 52: I think it would be better to have a separate `onCancel` event

assets/src/apps/authoring/components/RightMenu/RightMenu.tsx
```diff
@@ -302,7 +304,28 @@ const RightMenu: React.FC<any> = () => {
...
+      return;
+    const cloneActivity = clone(currentActivity);
+    const ogPart = cloneActivity.content?.partsLayout.find((part: any) => part.id === currentPartSelection);
+    if (!ogPart) {
+      console.warn(
+        'couldnt find part in current activity, most like lives on a layer; you need to update they layer copy directly',
+      );
+      return;
+    }
+    if (newJson.id !== ogPart.id) {
```
bsparks, line 319: should check that `newJson.id` is also not blank. it would also be good (same with the other component editor) now that I think of it to check all the part id in the `currentActivityTree` to make sure it isn't duplicated. both of those are a bit tricky however since we don't have any other error messages anywhere... at least with this modal we can pass in an array of all the ids I think and then it can use that for validation, displaying an error message and disabling the Save button.

assets/src/apps/authoring/components/RightMenu/RightMenu.tsx
```diff
@@ -302,7 +304,28 @@ const RightMenu: React.FC<any> = () => {
...
+      ogPart.id = newJson.id;
+      // in case the id changes, update the selection
+      dispatch(setCurrentSelection({ selection: newJson.id }));
+    }
+    ogPart.custom = newJson.custom;
```
bsparks, line 324: we need to do some kind of validation here, since we don't have the form fields from the property editor restricting what they can input. perhaps like above we should also pass into the component the schema data (or just use it in the component) and when it does the parse test on update make sure that they don't delete any required fields. also need to make sure that the same thing with the palette happens, so need to make sure that `useHTMLProps` gets set to true, and it has the `backgroundColor` and friends properties.

bsparks (changes requested):
changes based on comments

schowkekarats360:
Getting build error issue while running test.

PR #1394, opened 2021-08-11
[BUG] PMP-1723

dtiwarATS:


assets/src/utils/common.ts
```diff
@@ -134,3 +135,8 @@ export const parseBool = (val: any) => {
   // have to ignore the false searchValue in 'replace'
   return !isNaN(num) ? !!num : !!String(val).toLowerCase().replace('false', '');
 };
+
+export const parseNumString = (item: string): string | number => {
```
bsparks, line 139: good to put this in common, I think we can have the contains operator use this version now too. also we can write some unit tests for this method too

bsparks (approved):

PR #1398, opened 2021-08-12
[BUG] PMP-1674-firefox

dtiwarATS:


assets/src/adaptivity/scripting.ts
```diff
@@ -43,7 +43,7 @@ export const getExpressionStringForValue = (v: { type: CapiVariableTypes; value:
       const evalResult = evalScript(`let foo = ${val};`);
       // when evalScript is executed successfully, evalResult.result is null.
       // evalScript does not trigger catch block even though there is error and add the error in stack property.
-      if (evalResult?.result?.stack?.indexOf('Error') !== -1) {
+      if (evalResult?.result !== null) {
```
dtiwarATS, line 46: when evalScript is executed successfully, evalResult.result is null. There was an issue with Firefox and the way it was giving the properties. It was not going inside the if loop where we do CSS checks so removed the existing and made it more generic.

bsparks (approved):

PR #1404, opened 2021-08-13
[BUG] PMP-1730

dtiwarATS:


assets/src/components/parts/janus-mcq/MultipleChoiceQuestion.tsx
```diff
@@ -373,6 +373,11 @@ const MultipleChoiceQuestion: React.FC<JanusMultipleChoiceQuestionItemProperties
                       type: CapiVariableTypes.NUMBER,
                       value: updatedValues?.length,
                     },
+                    {
```
bsparks, line 376: im not sure about this because the reason that we're here in the first place is that it's been set from the outside, and the *only* reason that we are needing to save the `numberOfSelectedChoices` is because that value is calculated from `selectedChoices` so unless by updating the `selectedChoices` it is actually changing the value should we save it.

dtiwarATS:
Closing this PR. I will create a new one with more changes

PR #1405, opened 2021-08-13
[BUG] PMP-1666

dtiwarATS:


assets/src/apps/delivery/layouts/deck/DeckLayoutFooter.tsx
```diff
@@ -381,6 +381,7 @@ const DeckLayoutFooter: React.FC = () => {
   useEffect(() => {
     if (hasOnlyMutation) {
       setIsLoading(false);
+      setHasOnlyMutation(false);
```
dtiwarATS, line 384: Once hasOnlyMutation was set to true, if there is another only mutation happens on the same screen, it was disabling the Next button. I forgot to set it to false in previous fix.

bsparks (approved):

PR #1407, opened 2021-08-13
Component json editor

schowkekarats360:


assets/src/apps/authoring/components/RightMenu/RightMenu.tsx
```diff
@@ -63,6 +64,9 @@ const RightMenu: React.FC<any> = () => {
     }
     console.log('CURRENT', { currentActivity, currentLesson });
     setScreenData(transformScreenModeltoSchema(currentActivity));
+    currentActivity?.content?.partsLayout?.forEach((part: any) => {
```
bsparks, line 67: actually for these IDs we don't want duplicates from any layers either, so need to look through the whole tree.
also probably better to reduce that to a single array and call `setExisitngIds` one time instead of in a loop
```
const currentIds = currentActivityTree.reduce((acc, activity) => acc.concat(activity.content.partsLayout.map(p => p.id)), []);
```

assets/src/apps/authoring/components/RightMenu/RightMenu.tsx
```diff
@@ -234,8 +239,13 @@ const RightMenu: React.FC<any> = () => {
         // schema
         if (instance.getSchema) {
           const customPartSchema = instance.getSchema();
+          let requiredFields = [];
+          if(instance.getRequiredFields){
```
bsparks, line 243: do we need to have a separate method for this? I'm pretty sure that `required` *is* part of the json schema format already, so we can just include it in the schema directly for any object. alternatively we could just consider any property that is in the provided schema to be "required" for the purposes of this json editor

bsparks: also the schema in the individual components is *just* for the custom section, externally we are requiring that `id` and `type` is present, and we can set that in the part schema. also, since we're sort of moving size and position information into that "common" schema (even though they still end up living under the `custom` property for now, we can handle them being required outside as well. speaking of, we should probably move palette into the common section and have it not be something that each component needs to declare, but we dont' have to do that during this pr

assets/src/apps/authoring/components/PropertyEditor/custom/CompJsonEditor.tsx
```diff
@@ -0,0 +1,77 @@
...
+  const [validationMsg, setValidationMsg] = useState<string>('');
+  const currentPartSelection = useSelector(selectCurrentSelection);
+  const [displayEditor, setDisplayEditor] = useState<boolean>(false);
+  const textAreaStyle: CSSProperties = {
+    width: '100%',
+  };
+  useEffect(() => {
+    try {
+      const jsonVal = JSON.parse(value);
+      if (existingPartIds.indexOf(jsonVal.id) !== -1 && currentPartSelection !== jsonVal.id) {
```
bsparks, line 25: if we have `currentPartSelection !== jsonVal.id` doesn't it mean that they will be forced to change the ID?

assets/src/apps/authoring/components/PropertyEditor/custom/CompJsonEditor.tsx
```diff
@@ -0,0 +1,77 @@
...
+        setValidationMsg('ID you have used is already exist in the current Activity.');
+      } else {
+        setValidationMsg(jsonVal.id === '' ? 'ID is required Field and can not be empty' : '');
```
bsparks, line 28: it's a little more verbose but I think it'd be easier to follow if you split these
```
} else if (!jsonVali.id) {
  setValidationMsg('ID is required and cannot be empty');
} else {
  setValidationMsg('');
}
```

assets/src/apps/authoring/components/PropertyEditor/custom/CompJsonEditor.tsx
schowkekarats360: existingIds will also contain the currentPartSelection.... so just to restrict the duplicate ID validation this extra condition is there.

assets/src/apps/authoring/components/RightMenu/RightMenu.tsx
schowkekarats360: In the schema, we are getting properties section of the actual schema, whereas the required array is at the level of properties in schema json.... so i had to put separate method for that.

bsparks: each one doesn't even do ID though, I'm saying just put required in the partSchema (part.ts) for the id, and then for the component itself (at least for the purposes of the json editor) just dynamically grab all the keys and call them required so that they are forced to match the component (do that in the editor not the component)

schowkekarats360: ok.. so part of the suggested changes are made.... but the whole required fields functionality is not addded in this PR.
I have created a different ticket for that... so I will create a diff PR for this requirement.

bsparks (approved):

PR #1408, opened 2021-08-13
[BUG] PMP-1643

dtiwarATS:


assets/src/components/parts/janus-capi-iframe/ExternalActivity.tsx
```diff
@@ -131,6 +131,9 @@ const ExternalActivity: React.FC<any> = (props) => {
     // result of init has a state snapshot with latest (init state applied)
     writeCapiLog('INIT RESULT CAPI', initResult);
     const currentStateSnapshot = initResult.snapshot;
+    if (initResult.context.currentActivity) {
+      setActivityId(initResult.context.currentActivity);
```
dtiwarATS, line 135: Only setting this here as we want the owner activity Id to be appended. So far test this with KIP open and free, small word preview etc

bsparks: is the `currentActivity` the owner? we don't want the current screen if it's on a layer right? or else it will set a new one each time?

dtiwarATS: It will be the owner activity id.

bsparks (approved):

PR #1409, opened 2021-08-13
[BUG] PMP-1739

dtiwarATS:
-- passed snapshot to pop-up component
-- contains fix for PMP-1651 as well

### DIFF
--- a/assets/src/components/parts/janus-popup/Popup.tsx
+++ b/assets/src/components/parts/janus-popup/Popup.tsx
@@ -16,7 +16,7 @@ interface ContextProps {
   mode: string;
 }
 interface InitResultProps extends ContextProps {
-  snapshot: string[];
+  snapshot: { [key: string]: string };
   context: ContextProps;
 }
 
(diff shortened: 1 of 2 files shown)


assets/src/components/parts/janus-popup/Popup.tsx
```diff
@@ -22,6 +22,7 @@ const Popup: React.FC<any> = (props) => {
   const [popupVisible, setPopupVisible] = useState(true);
   const [iconSrc, setIconSrc] = useState('');
 
+  const [initSnapshot, setInitSnapshot] = useState<any>();
```
ChristianMurphy, line 25: Is there a more specific interface than `any` that could be used here?

bsparks: maybe it could be `Record<string, any>`

ChristianMurphy: Or `Record<string, unknown>`

dtiwarATS: Done

ChristianMurphy (approved):

assets/src/components/parts/janus-popup/Popup.tsx
dtiwarATS: I was planning using this

 snapshot: { [key: string]: string };

but something is happening with my Git. Hang On..

bsparks: `Record<string, string>` would be the same as that, but can't it have other values than string? `Record<string, unknown>` sounds good

bsparks (approved):

PR #1416, opened 2021-08-16
Feature/question bank component

schowkekarats360:


assets/src/apps/authoring/components/SequenceEditor/SequenceEditor.tsx
```diff
@@ -52,12 +52,15 @@ const SequenceEditor: React.FC<any> = (props) => {
...
     isLayer = false,
+    isBank = false,
   ) => {
     let layerRef: string | undefined;
     if (parentItem) {
       layerRef = parentItem.custom.sequenceId;
     }
-    const newTitle = `New ${layerRef ? 'Child' : ''}${isLayer ? 'Layer' : 'Screen'}`;
+    const newTitle = `New ${layerRef && !isBank ? 'Child' : ''}${
+      isLayer ? 'Layer' : isBank ? 'Question Bank' : 'Screen'
```
bsparks, line 62: not usually a good idea to keep going with nested tertiary conditions. it looks like maybe one is missing from this? anyway if you have more than one, then you should break it out into a variable

assets/src/apps/authoring/components/SequenceEditor/SequenceEditor.tsx
```diff
@@ -391,7 +410,8 @@ const SequenceEditor: React.FC<any> = (props) => {
                 }}
               >
                 <i className="fas fa-clipboard align-text-top mr-2" /> Copy{' '}
-                {item.custom.isLayer ? 'Layer' : 'Screen'} ID
+                {item.custom.isLayer ? 'Layer ' : item.custom.isBank ? 'Qustion Bank ' : 'Screen '}{' '}
```
bsparks, line 413: same here with the nested tertiary. also typo in "Question" ... maybe a good idea to put the strings "Layer", "Question Bank" and "Screen" into constants since they are reused a lot in this file that way can prevent typo and also allow for quick renaming if we decide to call it something other than "Question Bank"

assets/src/apps/delivery/store/features/groups/actions/sequence.ts
```diff
@@ -139,9 +139,17 @@ export const getSequenceLineage = (
   return lineage;
 };
 
-export const getIsLayer = () => {
+export const getSequenceInstance = () => {
   const currentSequenceId = useSelector(selectCurrentSequenceId);
```
bsparks, line 143: I didn't think it was possible to use hooks inside a "normal" method... I realize this wasn't you. making a mental note to refactor these 3 methods to take the sequenceId and sequence as input instead. should useSelector inside the component and pass to these methods (also makes them testable). I assume this must be working only because they are being called inside a component in the end? I can't imagine they are properly reactive however...

bsparks (approved):

PR #1421, opened 2021-08-18
[BUG] PMP-1168

dtiwarATS:


assets/src/components/parts/janus-capi-iframe/ExternalActivity.tsx
```diff
@@ -843,6 +843,61 @@ const ExternalActivity: React.FC<any> = (props) => {
...
+  const simulationHacks = () => {
+    const AllowNextOnCacheCaseVars: Record<string, unknown> = Object.keys(initState).reduce(
+      (formatted: any, key) => {
+        const baseKey = key.replace(`stage.${id}.`, '');
+        const value = initState[key];
+        const cVar = new CapiVariable({
+          key: baseKey,
+          value,
+        });
+        if (cVar.value === 'false' || cVar.value === false) {
```
bsparks, line 857:
```suggestion
        cVar.value = !parseBoolean(cVar.value);
```

bsparks:
hmmm I don't see anything about this in the original code, so my only conclusion is that we must be doing something fundamentally wrong in the CAPI processing... I guess we can proceed with the hack for now, but I feel like I can have no confidence that it won't happen again for some other sim and other variable

bsparks:
as decided in scrum we'll not apply any one off hacks and instead focus on finding a true fix

PR #1423, opened 2021-08-18
limit the amount of author invite emails at a time to 20

eliknebel:
This PR introduces a limit to the amount of comma-separated collaborator emails that can be sent at a time. This is a security measure to limit the capability of blasting unlimited emails from the system without reverifying the captcha.

Also includes some richer logging of errors for collaborator invitations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 ### Bug Fixes
+- Fix an issue where unlimited collaborator emails could be sent at once
 ### Enhancements
 
 ## 0.12.8 (2021-08-11)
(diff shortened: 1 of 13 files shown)


LEDThereBeLight:
Should we remove the captcha?

eliknebel:
This is based on my usage of importing some courses, I've been adding anywhere for 5-15 collaborators. I chose 20 because it seemed like a good balance of, "enough that I can reasonably batch add collaborators" and "not allowing an infinite amount of emails".

The captcha is still the linchpin of security here, since it prevents a bot from using an authoring account to send unlimited emails. The issue that I exposed by allowing multiple emails was that it only takes 1 captcha then an author can send unlimited emails. This fix prevents that by limiting the amount of emails at once to a predefined max, per captcha (so a bot/malicious user would not be able to automate spam through this)

LEDThereBeLight (approved):

PR #1429, opened 2021-08-19
Feature/question bank component

schowkekarats360:


assets/src/apps/authoring/components/PropertyEditor/schemas/screen.ts
```diff
@@ -148,19 +155,30 @@ export const transformScreenModeltoSchema = (activity?: IActivity) => {
...
       title: activity?.title || '',
       Size: { width: data.width, height: data.height },
       checkButton: { showCheckBtn: data.showCheckBtn, checkButtonLabel: data.checkButtonLabel },
       max: { maxAttempt: data.maxAttempt, maxScore: data.maxScore },
       palette: data.palette.useHtmlProps ? data.palette : schemaPalette,
     };
+    if (currentSequence?.custom.isBank) {
+      schemaData = {
+        ...schemaData,
+        bankShowCount: currentSequence.custom.bankShowCount,
```
bsparks, line 169: should probably add some defaults here
```suggestion
        bankShowCount: currentSequence.custom.bankShowCount || 1,
```

assets/src/apps/authoring/components/PropertyEditor/schemas/screen.ts
```diff
@@ -148,19 +155,30 @@ export const transformScreenModeltoSchema = (activity?: IActivity) => {
...
+        bankEndTarget: currentSequence.custom.bankEndTarget,
```
bsparks, line 170:
```suggestion
        bankEndTarget: currentSequence.custom.bankEndTarget || 'next',
```

assets/src/apps/authoring/components/PropertyEditor/schemas/screen.ts
```diff
@@ -174,6 +192,24 @@ export const transformScreenSchematoModel = (schema: any) => {
...
 
-export default screenSchema;
+export const getScreenSchema = (seq: SequenceEntry<SequenceEntryChild> | null) => {
+  if (seq?.custom.isBank) {
+    return {
+      type: schema.type,
+      properties: {
+        Size: schema.properties.Size,
+        palette: schema.properties.palette,
+        bankShowCount: { type: 'number', title: 'Randomly selects question(s) from the bank' },
```
bsparks, line 209: perhaps we should do the collapsible section for "Question Bank Settings" OR a non collapsible section could make sense as well, but I think it would be good to separate them visually

bsparks:
ok, so the more I think about it, I think it would be better to keep the Question Bank details separate, because they don't actually exist on the Activity (screen) itself. We still want it to show up under the Screen tab on the right, but it should be a totally separate Property Editor. So hopefully we can conditionally render 2 Property Editor instances under that tab. That way it can have its own schema and things don't have to be "hacked" into interfaces where they don't belong. It can have its own change handler, etc.

assets/src/apps/authoring/components/PropertyEditor/schemas/screen.ts
```diff
@@ -123,7 +127,10 @@ export const screenUiSchema = {
   },
 };
 
-export const transformScreenModeltoSchema = (activity?: IActivity) => {
+export const transformScreenModeltoSchema = (
+  currentSequence: SequenceEntry<SequenceEntryChild> | null,
```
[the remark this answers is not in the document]
schowkekarats360: but inside the methods we want to use those 2 properties(bankshowcount, bankendtarget) which are in the sequence, thats why i had to send whole sequence.

assets/src/apps/authoring/components/PropertyEditor/custom/CustomDropdownTemplate.tsx
```diff
@@ -0,0 +1,59 @@
...
+interface CustomDropdownProps {
+  id: string;
+  label: string;
+  uiSchema: UiSchema;
+  description: string;
+  properties: object;
+  value: string;
+  onChange: (value: string) => void;
+}
+const CustomDropdownTemplate: React.FC<CustomDropdownProps> = (props) => {
```
bsparks, line 23: since this isn't just *any* custom dropdown, it's specifically the Sequence / Screen picker dropdown, let's name it that way please. `ScreenSelectDropDown` or something like that

assets/src/apps/authoring/components/PropertyEditor/custom/CustomDropdownTemplate.tsx
```diff
@@ -0,0 +1,59 @@
...
+  const { id, label, uiSchema, description, properties, value, onChange } = props;
+  const sequence = useSelector(selectSequence);
+  const hierarchy = getHierarchy(sequence);
+  const [buttonLabel, setButtonLabel] = useState<string>('');
+  const [val, setVal] = useState<string>(value);
+  useEffect(() => {
+    const seq = findInHierarchy(hierarchy, val);
+    setButtonLabel(seq?.custom.sequenceName || 'Next');
+  }, [val]);
```
bsparks, line 32: hierarchy is also a dependency since you are using it inside the effect

assets/src/apps/authoring/components/PropertyEditor/custom/CustomDropdownTemplate.tsx
```diff
@@ -0,0 +1,59 @@
...
+  properties: object;
+  value: string;
+  onChange: (value: string) => void;
+}
+const CustomDropdownTemplate: React.FC<CustomDropdownProps> = (props) => {
+  const { id, label, uiSchema, description, properties, value, onChange } = props;
+  const sequence = useSelector(selectSequence);
+  const hierarchy = getHierarchy(sequence);
+  const [buttonLabel, setButtonLabel] = useState<string>('');
+  const [val, setVal] = useState<string>(value);
```
bsparks, line 28: what is val? `selectedSequenceId` ? can we rename for clarity?

assets/src/apps/authoring/components/PropertyEditor/custom/CustomDropdownTemplate.tsx
```diff
@@ -0,0 +1,59 @@
...
+  useEffect(() => {
+    const seq = findInHierarchy(hierarchy, val);
+    setButtonLabel(seq?.custom.sequenceName || 'Next');
+  }, [val]);
+  const onChangeHandler = (item: SequenceEntry<SequenceEntryChild>) => {
+    setVal(item.custom.sequenceId);
+    onChange(item.custom.sequenceId);
```
bsparks, line 35: how do you pick `next` in this one? (the data I think needs to be lowercase `next` even if the label is `Next`)

assets/src/apps/authoring/components/PropertyEditor/schemas/bank.ts
```diff
@@ -0,0 +1,129 @@
...
+  properties: {
+    Size: {
+      type: 'object',
+      title: 'Dimensions',
+      properties: {
+        width: { type: 'number' },
+        height: { type: 'number' },
+      },
+    },
+    palette: {
```
bsparks, line 24: I dont think you need all this stuff, just the items that go in the sequence for banks, just the 2 properties; the rest of the screen stuff doesn't exist in the sequence

assets/src/apps/delivery/store/features/groups/actions/sequence.ts
```diff
@@ -7,6 +7,8 @@ export interface SequenceEntryChild {
   layerRef?: string;
   isBank?: boolean;
   isLayer?: boolean;
+  bankShowCount: number;
```
bsparks, line 10: there is already `SequenceBank` below

PR #1430, opened 2021-08-19
[FEATURE] Allow for finalization without answering all questions or parts

### LINKED ISSUE #1425: Allow for page finalization with partial activity completeness
The adaptive page implementation requires the ability to finalize pages even when some activities in the page have not been visited (i.e., they have no state that can be evaluated)

darrensiegel:
Closes #1425 

This PR allows graded pages to be submitted without the student having answered all questions.   

The change to allow this was fairly straightforward: There was an explicit check in the activity lifecycle evaluation logic that would fail submission if a single part of any activity had no state available (aka, was `nil`).  I removed this and made the eval's handling of `nil` states robust.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,11 +3,15 @@
 ## Unreleased
 
 ### Bug Fixes
+
 ### Enhancements
+
+- Allow for submission of graded pages without answering all questions
 
 ## 0.12.8 (2021-08-11)
 
 ### Bug Fixes
+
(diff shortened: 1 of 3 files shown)


bsparks (approved):

eliknebel (approved):
Works as expected. I think in the future we may want to add a dialog on assessment submission that prompts a user if not all questions have been answered, requiring them to confirm. This can be added later if necessary.

PR #1438, opened 2021-08-23
[FEATURE] Activity bank editor

### LINKED ISSUE #1434: Create, edit and filter / page banked activities
Add the ability for a user to be able to create banked activities, filter and search through the existing banked activities, and edit banked activities.  

This represents PR3 from the Activity Bank design approach: https://docs.google.com/document/d/1U-il9imFI4PIGhdU2UWrVTz93ZqJtegIZNFGUTHzMQk/edit

darrensiegel:
Closes #1434 

This PR introduces the "Activity Bank Editor" - a new, top-level view within a project where an author can create new banked activities, and filter and search through to edit existing ones. 

The main structure of this work is a React client-side app that encapsulates the filter/search UIs and orchestrates the editing of individual activities.  It is supported by a new server-side `actvitiy_bank_controller.ex` that allows access to the banked activities within a project.  

The client side editor implements an explicit locking system whereby a user must first click to "unlock" an activity to edit it.  Only one activity can be edited at a time, and the client side app enforces that.  

For filtering and searching there are three fixed "Expression" editors that a user can use to pare down the list of activities to find specific ones.  This UI ultimately translates to the Logic model introduces in previous work, which is directly queryable against the database.  Note, the intent is to reuse the `Expression` component in the next PR that introduces creation of bank selections within pages.  That work will use the `Clause` component which is included here but not used here (and probably not complete).


eliknebel (changes requested):
This looks really good. I think this type of functionality is going to be very powerful. Just 3 thing I wanted to bring up:

1. The "Unable to Edit Activity" Modal should have a different ok_label than "Delete". Ideally, it should probably just be "Ok" with no cancel button.
Screen Shot 2021-08-23 at 2 18 45 PM

It might be useful to display the user name/email who has the lock, like we do with pages. But I don't think it's required as part of this PR.

2. This is kind of nit-picky, but I think the Activity Bank item would be better after Objectives but before Curriculum. Couple reasons for this:
    - Curriculum is probably the most used item and it is easier to quickly select the last item in the list than somewhere in the middle.
    - Current users are probably already used to quickly selecting the Curriculum as the last item in the dropdown, changing this might be jarring
    - In the abstract flow of course development, you must create your activities before inserting them into the curriculum. So it would probably be a step before curriculum in the development process.
Screen Shot 2021-08-23 at 2 20 01 PM

3. Do we plan to later add the ability to create multiple banks? As opposed to a single top level bank per project. I could see this being a useful level of abstraction as opposed to managing a top level set of tags to accomplish this.

darrensiegel:
1. Fixed by setting that dialog to only display an "Ok" button.
2. Resolved by reordering. 
3. No, we do not plan to support more than one activity bank.  The idea here is that the concept is simple for the author to understand and work with: there is a global bank of activities that can be "drawn" from using various criteria.

eliknebel (approved):

PR #1439, opened 2021-08-24
[FEATURE] Bulk Edit API - Authoring

dtiwarATS:
Modify the Authoring to Call the Bulk Edit API


assets/src/apps/authoring/store/groups/layouts/deck/actions/updateActivityPartInheritance.ts
```diff
@@ -69,15 +69,25 @@ export const updateActivityPartInheritance = createAsyncThunk(
...
           content: { ...activity.content, authoring: activity.authoring },
+          resource_id: resourceId,
         };
-        return edit(projectSlug, resourceId, activity.resourceId, changeData, false);
+        const update = Object.assign({}, changeData, false);
+        update.content = Object.assign({}, update.content);
+
+        // Here we pull the "authoring" key out of "content" and elevate it
+        // as a top-level key
+        if (update.content.authoring !== undefined) {
```
bsparks, line 84: since we're doing this here, we really don't need to go through all this, we can construct the object directly into this format. the reason the other one didn't do that is because other Torus activities were already using the "edit" call that expected the format in that way.

PR #1443, opened 2021-08-24
[BUG FIX] Fix an issue where changing the title rendered the current slug invalid

### LINKED ISSUE #848: Trying to access an activity after page title change fails
**Describe the bug**
After creating an activity on a page and clicking "Edit ...", an error is thrown.

**To Reproduce**
1. Go to a page
2. Change the title
2.  Create an activity (Multiple Choice, for example)
2. Immediately click "Edit Multiple Choice"
3. An error will be thrown

**Expected behavior**
A user should be able to edit an activity after changing the page title

**Screenshots**
Screen Shot 2021-03-22 at 12 57 03 PM

 - OS: macOS 11.2.3
 - Browser Chrome
 - Version 89

**Additional context**
I've seen this a couple times, usually when creating the first activity in a project.

eliknebel:
This fixes an issue where changing the title of a page makes the current slug invalid by refreshing the page after a slug change. It also appears retain any subsequent edits between a title change and page reload.

Closes #848


darrensiegel:
This fix shouldn't be necessary.  There is server side slug resolution in place that allows editing to continue after a title change and continued use of an "invalid" slug.

I just did some testing and I don't see any problem.

darrensiegel (changes requested):
This does work, in that it forces a refresh, but it does so only after the DeferredPersistenceStrategy reaches its quiet period.  So after editing the title, I have around two seconds before the reload happens, which could be jarring for a user and cause the user not to realize that it was the title edit that caused the release. 

Take another look here and see if there is an easy way to eliminate this delay.  One could either add a "flush" method to the strategy... or altogether side-step the strategy and in the case of a title edit directly invoke the persistence endpoint and when it finishes reload the page.

eliknebel:
Added a saveImmediate function to the persistence strategy which is called when the title changes.

darrensiegel (approved):

darrensiegel (approved):

PR #1445, opened 2021-08-25
Feature/adaptive author by activity

bsparks:
first part of WYSIWYG overhaul switching from Canvas to HTML. also renders the authoring screens by actual activity component (while leaving it open for other activity types (sort of ))


dsuriano:
@bsparks  A few things I'm noticing while testing locally: 

1. I can no longer drag or scroll the stage around to show parts of the activity that are cut off.
2. When I drag and release an element, it bounces between the previous position and the new one. 

3. There doesn't appear to be a way to deselect a selected component. Also, clicking an empty section of canvas no longer selects the "Screen" tab

assets/src/apps/authoring/components/EditingCanvas/AuthoringActivityRenderer.tsx
```diff
@@ -0,0 +1,69 @@
...
+          continuation(result);
+        }
+      }
+    };
+    // for now just do this, todo we need to setup events and listen
+    document.addEventListener('customEvent', customEventHandler);
+    setIsReady(true);
+
+    return () => {
+      document.removeEventListener('customEvent', customEventHandler);
```
dsuriano, line 60: Does `setIsReady` need to reset to `false` here?

dsuriano (approved):

PR #1459, opened 2021-08-31
[JAN-215, PMP-751] text author issues

bsparks:


dsuriano (approved):
tested locally and confirmed focus bug was resolved

PR #1466, opened 2021-09-01
[BUG FIX] Provide specific impl to realize activities from adaptive page content

darrensiegel:
This PR fixes an issue introduced by the new `ActivityProvider` implementation, namely that deeply nested activity references within adaptive page content were no longer being realized. The core issue was that the new impl only looks at the top-level `model` list.  A singular implementation was going to be fairly complex (due to content transformation requirements from bank selections), so instead I pattern matched and just do a `flat_filter` to find all activity references in the adaptive page case.


bsparks (approved):
looks good to me :)

PR #1475, opened 2021-09-02
[FEATURE] Allow admins to delete user and author accounts

eliknebel:
This PR refactors the DB schema to allow for author and user records to be deleted gracefully without violating constraints. It also adds the ability to do so via the admin Accounts page. Users and authors who have been deleted will be able to recreate accounts if they choose, but if a email needs to be locked out for whatever reason, then the lock feature should suffice.

One more thing to note is that when user accounts are deleted, they will not longer show up in the datashop export. Please refer to the migrations file in this PR to see the policies taken on foreign key records when an account is deleted.

Also includes fixes for the Manage dropdown where duplicate actions were being shown and CSRF issues.
Screen Shot 2021-09-02 at 6 25 09 PM


darrensiegel (approved):
Tested locally and it works as expected.

PR #1482, opened 2021-09-03
[JAN-587 ] - [PMP-1769] send the User to the 'Overview' screen

schowkekarats360:


assets/src/apps/delivery/layouts/deck/LessonFinishedDialog.tsx
```diff
@@ -17,15 +18,20 @@ const LessonFinishedDialog: React.FC<LessonFinishedDialogProps> = ({
   const [isOpen, setIsOpen] = useState(true);
   const isPreviewMode = useSelector(selectPreviewMode);
   const resourceAttemptGuid = useSelector(selectResourceAttemptGuid);
+  const { graded } = useDeliveryElementContext();
```
bsparks, line 21: oh hmm are you sure `useDeliveryElementContext` is working? I think it says that it needs to have that provider wrapped, and also the DeliveryElement is an activity (screen) not the whole lesson. I have to check but now that I think about it, not sure if we have that graded value passed in or not

bsparks (approved):

PR #1487, opened 2021-09-03
expose graded flag so that authoring and delivery can use when necessary

bsparks:
this is mostly needed so that delivery can tell whether to send to the finalization page or overview page at the end of the lesson

### DIFF
--- a/assets/src/apps/app.tsx
+++ b/assets/src/apps/app.tsx
@@ -38,7 +38,7 @@ export function defineApplication<T extends State>(Component: React.FunctionComp
       partComponentTypes: parsedPartComponentTypes,
     };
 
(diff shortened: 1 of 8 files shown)


assets/src/apps/app.tsx
```diff
@@ -38,6 +38,8 @@ export function defineApplication<T extends State>(Component: React.FunctionComp
       partComponentTypes: parsedPartComponentTypes,
     };
 
+    console.log('MOUNT UP', { props, params, mountPoint });
```
darrensiegel, line 41: Remove console.log

darrensiegel (changes requested):

PR #1490, opened 2021-09-07
PMP-1785 | JAN-606

dtiwarATS:


bsparks:
since we need to move all positioning logic to the web component element itself, the inner contents cannot do it anymore. I am thinking that this stuff (and maybe also palette) need to be wired through PartComponent because that is what has access to the element.

PR #1492, opened 2021-09-07
[BUG FIX] Allow use of image coding in graded pages

darrensiegel:
This PR fixes a couple of problems around use of Image Coding in graded contexts. 

1. It enables the "Submit" button for the activity even when graded.  This is the only way that a student can actually answer the question, given that it is a client-side evaluated activity. 
2. The PR adjusts evaluation logic to ensure that even unanswered questions will be counted as "0 out of 1" in evaluation and subsequent roll-up.  
3. It makes some page rendering logic robust against any existing or future cases where invalid values exist as the "out of" for a resource attempt, specifically any value less than or equal to 0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Restore ability to realize deeply nested activity references within adaptive page content
 - Fix an issue in admin accounts interface where manage options sometimes appear twice
 - Allow graded adaptive pages to render the prologue page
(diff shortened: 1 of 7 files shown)


eliknebel (approved):
I Tested over LTI connection, works as expected. I also checked that the grade was properly reporting back to the LMS grade book and that worked as expected too.

PR #1503, opened 2021-09-09
[BUG] PMP-1781 | JAN-603

dtiwarATS:
--This also contains partial fix for PMP-1647 (Issue#1 - pop-up not opening automatically)


bsparks:
taken care of in another

PR #1512, opened 2021-09-13
PMP-1780 | JAN-602

dtiwarATS:


assets/src/components/parts/janus-mcq/MultipleChoiceQuestion.tsx
```diff
@@ -648,7 +672,7 @@ const MultipleChoiceQuestion: React.FC<PartComponentProps<McqModel>> = (props) =
   }
 
   return ready ? (
-    <div data-janus-type={tagName} style={styles} className={`mcq-input`}>
+    <div data-janus-type={tagName} style={styles} className={`mcq-input ${customCssClass}`}>
```
bsparks, line 675: the css class is now on the custom element parent, what do we need this one for? if at all possible we just want to have them fix the CSS override

dtiwarATS: Ok . I will remove this and update the PR. We can merge the rest

bsparks (approved):

PR #1515, opened 2021-09-14
PMP-1808 | JAN-622

dtiwarATS:


assets/src/apps/delivery/layouts/deck/DeckLayoutFooter.tsx
```diff
@@ -225,13 +225,16 @@ const DeckLayoutFooter: React.FC = () => {
         let scopedTarget = op.params.target;
 
         if (scopedTarget.indexOf('stage') === 0) {
-          const ownerActivity = currentActivityTree?.find(
-            (activity) =>
-              !!activity.content.partsLayout.find((p: any) => p.id === op.params.target),
```
dtiwarATS, line 230: The value of "op.params.target" was 'stage.partId.value' and we were directly checking the p.id===op.params.target which will never return owner Id.

So, We need to get the part Id from op.params.target and then check that with p.id.

bsparks (approved):

PR #1519, opened 2021-09-15
[BUG]PMP-1829 | JAN-645

dtiwarATS:


assets/src/adaptivity/scripting.ts
```diff
@@ -161,6 +166,10 @@ export const getAssignStatements = (state: Record<string, any>): string[] => {
   return letStatements;
 };
 
+export const getAssignScriptArray = (state: Record<string, any>): string | string[] => {
```
dtiwarATS, line 169: Added this new function else I had to make changes on multiple locations including the unit test cases.

bsparks: why not just call `getAssignStatements` directly?

bsparks (approved):

PR #1520, opened 2021-09-15
PMP-1785 | JAN-606

dtiwarATS:


assets/src/components/activities/adaptive/components/common/PartComponent.tsx
```diff
@@ -35,7 +35,42 @@ const PartComponent: React.FC<AuthorProps | DeliveryProps> = (props) => {
   const [componentStyle, setComponentStyle] = useState<CSSProperties>(initialStyles);
 
   const [customCssClass, setCustomCssClass] = useState<string>(props.model.customCssClass || '');
+  const handleiFrameStylingChanges = (currentStateSnapshot: Record<string, unknown>) => {
+    //janus-capi-iframe is the only component that allows a user to change it's position and some other style attributes
+    if (props.type === 'janus-capi-iframe') {
```
bsparks, line 40: I don't think we need to filter this out if some other component wants to use it, though I am regretting the `IFRAME` prefix now, but there is no reason that I can think of that changing *any* component this way wouldn't be OK

assets/src/components/activities/adaptive/components/common/PartComponent.tsx
```diff
@@ -35,7 +35,42 @@ const PartComponent: React.FC<AuthorProps | DeliveryProps> = (props) => {
   const [componentStyle, setComponentStyle] = useState<CSSProperties>(initialStyles);
 
   const [customCssClass, setCustomCssClass] = useState<string>(props.model.customCssClass || '');
+  const handleiFrameStylingChanges = (currentStateSnapshot: Record<string, unknown>) => {
```
bsparks, line 38: maybe we can also just call this `handleStylingChanges` later if we want to we can add other keys to check for

assets/src/components/activities/adaptive/components/common/PartComponent.tsx
```diff
@@ -35,7 +35,42 @@ const PartComponent: React.FC<AuthorProps | DeliveryProps> = (props) => {
...
+    //janus-capi-iframe is the only component that allows a user to change it's position and some other style attributes
+    if (props.type === 'janus-capi-iframe') {
+      const externalActivityStyles: CSSProperties = {};
```
bsparks, line 41: maybe change to `styleChanges`

bsparks (approved):

PR #1521, opened 2021-09-15
[FEATURE] adjusting delivery JSON

dtiwarATS:


assets/src/components/parts/janus-Carousel/Carousel.tsx
```diff
@@ -133,24 +133,11 @@ const Carousel: React.FC<PartComponentProps<CarouselModel>> = (props) => {
...
-        pState = JSON.parse(props.state);
-        setState(pState);
-      } catch (err) {
-        // bad json, what do?
-      }
-    }
+    const pModel = props.model;
+    setModel(pModel);
+
+    const pState = props.state;
```
bsparks, line 139: we don't actually send state anymore I think

assets/src/components/parts/janus-Carousel/Carousel.tsx
```diff
@@ -133,24 +133,11 @@ const Carousel: React.FC<PartComponentProps<CarouselModel>> = (props) => {
...
-    if (typeof props?.state === 'string') {
-      try {
-        pState = JSON.parse(props.state);
-        setState(pState);
-      } catch (err) {
-        // bad json, what do?
-      }
-    }
+    const pModel = props.model;
+    setModel(pModel);
```
bsparks, line 137: it's more work to refactor, but I think one big reason to do this switch at all would be to revert from using model as a state to just using it from props directly. cutting down on the number of times the render function needs to be called. though hmm now that I say that maybe it's because of the events that can change the model values? still I think the initial value can come directly from props

dtiwarATS:
Closing it for now.

PR #1522, opened 2021-09-15
[FEATURE] Multi input activity

LEDThereBeLight:
Add support for multi-input activities. Multi-input activities let users insert answer inputs inline into a question stem. The activity supports text inputs, numeric inputs, and dropdowns (html selects).

Originally I tried to avoid injecting the inputs into the slate editor model, but it added enormous complexity in delivery and styling was awkward. Eventually I decided to add slate editor support for `input_ref`s, but that came with a lot of challenges.

I had to make changes to the activity HTML content parser to render react components instead of strings because rendering input refs required a lot of extra context and state management.

Some slate editor bugs and activity editor race conditions are also fixed.

- [x] Add multi-input activity framework
- [x] Create dropdown-type question part based on multiple choice
- [x] Create dropdown authoring/delivery rendering
- [x] Refactor activity utils to support multi-part activities
- [x] Expand `Undo` infrastructure into JSON path queries to find/update activity parts by path
- [x] Add `inputRef` element support to slate editors
- [x] Add `inputRef` rendering support - authoring (new editor, commands)
- [x] Add `inputRef` rendering support - delivery
- [x] Refactor `writer.ts` and `html.ts` (client-side activity rendering parsers) to use React elements instead of strings (to support `inputRef` context and state management)
- [x] Support adding a new input ref into an editor
- [x] Support reconciling missing inputs/parts
- [x] Support switching between input types using the slate editor
- [x] Support selecting an active input ref to edit details
- [x] Support removing an input ref from the stem
- [x] Support adding parts
- [x] Update `InlineActivityEditor`, `PageEditor`, and `Objectives` to support adding/removing parts for an activity
- [x] Support removing a part and reconciling the stem
- [x] Support adding/removing choices for dropdowns
- [x] Support dropdown choice shuffling
- [x] Support targeted feedback for all multi input types
- [x] Prevent copy/pasting input refs
- [x] Restrict input refs to only multi-input slate editors using a restricted element whitelist 
- [ ] Add safety reconciliation logic for malformed multi-inputs on detection (outside of edits made)
- [ ] Add safety project warnings for malformed multi-inputs
- [ ] Clean up / consolidate activity utils, remove duplication
- [ ] Improve input ref editor active / selected look and feel


darrensiegel (changes requested):
This is really some great work.  Nice UX impl that seems very robust. 

I saw a few console.logs (see inline notes).  

I am able to circumvent the check that prohibits deleting the last input ref.  It is tricky to do, but it seems like if I have a multi line stem and one input ref, if I then CTRL-A select all text then press "delete" key that I can sometimes get it to delete all input refs.  Once I get into that state, the editor for the stem is really horked up.  So we do need to prevent "invalid" multi-input activities from being saved.  There is a pattern here that we can follow for this: take a look at `adjustContentForConstraints` in `PageEditor.tsx`.  That method is invoked from the `save` but its results are not pushed down as updated props to the component tree.  This allows an "on the fly" transform during a save.  The currentl impl only targets the contents of the page - we need to add an identical but separate mechanism to allow transformations to be made for activity content saving.  This approach would have to be plugged into both the `PageEditor` and the `ActivityBankEditor`.

darrensiegel (approved):
We will land as-is and follow up with a fix to implement the on the fly model adjustment

PR #1526, opened 2021-09-16
[PMP-1771/JAN-633]-cant-delete-last-content-in-custom-css

malav2110:


assets/src/apps/authoring/store/page/slice.ts
```diff
@@ -69,6 +69,9 @@ const slice: Slice<PageState> = createSlice({
       }
       if (action.payload.customCss !== undefined) {
         state.customCss = action.payload.customCss;
+      } else {
```
bsparks, line 72: took me a min to think about this, so the reason that all of these are checking undefined is that it is supposed to allow you to send in only a partial model (only the things that need to be updated). so in this case if someone were to call updatePage and *not* include `customCss` then it would write it to `''` even if it was simply meant to remain unchanged.

so I think there must be somewhere earlier in the code process that is causing this glitch

bsparks (changes requested):
pls find another way

assets/src/apps/authoring/store/page/slice.ts
malav2110: Taking a look.

assets/src/apps/authoring/store/page/slice.ts
```diff
@@ -69,7 +69,7 @@ const slice: Slice<PageState> = createSlice({
       }
       if (action.payload.customCss !== undefined) {
         state.customCss = action.payload.customCss;
-      }
+      }
```
bsparks, line 72: is this an extra space?

bsparks:
```suggestion
      }
```

bsparks (approved):

PR #1538, opened 2021-09-17
[BUGFIX] also don't deselect when doing configuration/edit of a part

bsparks:


dsuriano (approved):
tested locally and works as expected.

PR #1539, opened 2021-09-17
[PMP-1844/JAN-660]-number-input-change-unit-and-remove-deleteenable

malav2110:


assets/src/apps/authoring/components/RightMenu/RightMenu.tsx
```diff
@@ -309,6 +309,12 @@ const RightMenu: React.FC<any> = () => {
         // schema
         if (instance.getSchema) {
           const customPartSchema = instance.getSchema();
+
+          // For Number Input, do not show 'Delete Enable' to Authors
+          if (customPartSchema.deleteEnabled) {
```
bsparks, line 314: you can just remove deleteEnabled from the schema.ts file, that schema is only used for the side bar it won't affect the operation of delivery. the side bar also shouldn't delete any properties that aren't in the schema that come along, so it should be fine however that property actually ends up in the data

malav2110: Updating the code

bsparks (approved):

PR #1561, opened 2021-09-21
[BUGFIX] Carousel Issues PMP-1881 JAN-696

bsparks:
clean up schema; refactor to use json model; remove custom css; change custom css class to match;


dsuriano (changes requested):
Looks like we've lost the ability to add images from the right panel.

dsuriano (approved):

PR #1565, opened 2021-09-22

dsuriano:
**This PR includes** work for the following sub-tasks:
* PMP-1852: Init & Mutate State - Limit to the Current Screen
* PMP-1853: Dropdown for Init And Mutate: Filter out Add for non numbers
* PMP-1855: Fix the Screen Ids and Pipes - when picking IDs from other screens needs to be a pipe then variable

**Remaining work** (future PR) to close PMP-1559 / JAN-332:
* PMP-1854: Dropdown should limit selection to valid combinations

### DIFF
--- a/assets/src/apps/authoring/components/AdaptivityEditor/ActionMutateEditor.tsx
+++ b/assets/src/apps/authoring/components/AdaptivityEditor/ActionMutateEditor.tsx
@@ -4,22 +4,7 @@ import React, { useEffect, useRef, useState } from 'react';
 import { OverlayTrigger, Tooltip } from 'react-bootstrap';
 import guid from 'utils/guid';
 import { VariablePicker, OverlayPlacements } from './VariablePicker';
-
-const typeOptions = [
-  { key: 'string', text: 'String', value: CapiVariableTypes.STRING },
-  { key: 'number', text: 'Number', value: CapiVariableTypes.NUMBER },
-  { key: 'array', text: 'Array', value: CapiVariableTypes.ARRAY },
-  { key: 'boolean', text: 'Boolean', value: CapiVariableTypes.BOOLEAN },
-  { key: 'enum', text: 'Enum', value: CapiVariableTypes.ENUM },
-  { key: 'math', text: 'Math Expression', value: CapiVariableTypes.MATH_EXPR },
-  { key: 'parray', text: 'Point Array', value: CapiVariableTypes.ARRAY_POINT },
-];
-const opOptions = [
-  { key: 'equal', text: '=', value: '=' },
-  { key: 'add', text: 'Adding', value: 'adding' },
-  { key: 'bind', text: 'Bind To', value: 'bind to' },
-  { key: 'set', text: 'Setting To', value: 'setting to' },
-];
+import { OperatorOption, operatorOptions, TypeOption, typeOptions } from './AdaptiveItemOptions';
 
 interface ActionMutateEditorProps {
   action: MutateStateAction;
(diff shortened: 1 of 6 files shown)


assets/src/apps/authoring/components/AdaptivityEditor/ActionMutateEditor.tsx
```diff
@@ -5,19 +5,19 @@ import guid from 'utils/guid';
 import { VariablePicker, OverlayPlacements } from './VariablePicker';
 
 const typeOptions = [
-  { key: 'number', text: 'Number', value: 1 },
   { key: 'string', text: 'String', value: 2 },
+  { key: 'number', text: 'Number', value: 1 },
```
bsparks, line 9: it would be good instead of the numbers to use the CapiVariableTypes enum everywhere

dsuriano: good catch. fixed.

assets/src/apps/authoring/components/AdaptivityEditor/ActionMutateEditor.tsx
```diff
@@ -93,6 +93,18 @@ const ActionMutateEditor: React.FC<ActionMutateEditorProps> = (props) => {
     }
   }, [isDirty]);
 
+  // update adding operator if targetType changes from number
+  useEffect(() => {
+    if (targetType !== ('1' as unknown)) {
```
bsparks, line 99: `'1' as unknown` is weird, can we parseInt on targetType?

assets/src/apps/authoring/components/AdaptivityEditor/ActionMutateEditor.tsx
```diff
@@ -138,11 +152,19 @@ const ActionMutateEditor: React.FC<ActionMutateEditorProps> = (props) => {
...
+        value={operator}
         onChange={(e) => handleOperatorChange(e)}
       >
         {opOptions.map((option) => (
-          <option key={option.key} value={option.value}>
+          <option
+            key={option.key}
+            value={option.value}
+            style={
+              targetType !== ('1' as unknown) && option.key === 'add'
```
bsparks, line 164: maybe instead of hiding the option should do like `opOptions.filter().map()` ?

assets/src/apps/authoring/components/AdaptivityEditor/InitStateEditor.tsx
```diff
@@ -142,8 +157,8 @@ interface TypeOption {
   value: number;
 }
 const typeOptions: TypeOption[] = [
-  { key: 'number', text: 'Number', value: 1 },
   { key: 'string', text: 'String', value: 2 },
+  { key: 'number', text: 'Number', value: 1 },
```
bsparks: should probably factor these options out to a shared file

bsparks (approved):

assets/src/apps/authoring/components/AdaptivityEditor/ActionMutateEditor.tsx
dsuriano: fixed.

assets/src/apps/authoring/components/AdaptivityEditor/ActionMutateEditor.tsx
dsuriano: fixed.

assets/src/apps/authoring/components/AdaptivityEditor/InitStateEditor.tsx
dsuriano: fixed

PR #1569, opened 2021-09-23
[FEATURE] CAPI Authoring Contd. JAN-672 | JAN-674

dtiwarATS:
-- fixes "config" click and save glitches
-- code clean-up
-- variable picker support for CAPI

### DIFF
--- a/assets/src/apps/authoring/components/AdaptivityEditor/VariablePicker.tsx
+++ b/assets/src/apps/authoring/components/AdaptivityEditor/VariablePicker.tsx
@@ -190,7 +190,13 @@ export const VariablePicker: React.FC<VariablePickerProps> = ({
               if (Array.isArray(custom.configData)) {
                 adaptivitySchema = custom.configData.reduce(
                   (acc: any, typeToAdaptivitySchemaMap: any) => {
-                    acc[typeToAdaptivitySchemaMap.key] = typeToAdaptivitySchemaMap.type;
+                    if (context === 'mutate') {
+                      if (!typeToAdaptivitySchemaMap.readonly) {
+                        acc[typeToAdaptivitySchemaMap.key] = typeToAdaptivitySchemaMap.type;
+                      }
+                    } else {
+                      acc[typeToAdaptivitySchemaMap.key] = typeToAdaptivitySchemaMap.type;
+                    }
                     return acc;
                   },
                   {},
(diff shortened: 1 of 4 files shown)


assets/src/components/parts/janus-capi-iframe/schema.ts
```diff
@@ -19,7 +19,7 @@ export const schema: JSONSchema7Object = {
 
 export const getCapabilities = () => ({
   configure: true,
-  save: true,
+  isCapi: true,
```
bsparks, line 22: I think as a capability it should just be `capi`

dtiwarATS: Fixed

assets/src/apps/authoring/components/AdaptivityEditor/VariablePicker.tsx
```diff
@@ -183,6 +183,27 @@ export const VariablePicker: React.FC<VariablePickerProps> = ({
           if (instance.getAdaptivitySchema) {
             adaptivitySchema = await instance.getAdaptivitySchema();
           }
+          if (instance.getCapabilities) {
```
bsparks, line 186: I think that we should pass a "context" to `getAdaptivitySchema` so that the component can decide what properties to provide and how to provide them. also so that we can have only *1* way of doing it.
@dsuriano I think this would affect the other vars;
```
adaptivitySchema = await instance.getAdaptivitySchema({ currentModel: part, editorContext: context });
```
so this way the editor doesn't need to know to use configData or even if the variables are "readonly" as it could filter those out... hmm actually might need more thought on that because we still might want all that info to be attached to each item. like enum for example needs to have more than the type, it needs to have the allowedValues; anyway, the context part still holds, we can send in the currentModel data so that you can either use the configData for it OR use the `src` to spin up a hidden iframe and ask it.

as a side note, at some point we might want to change this from pulling async every time the variable picker is opened to going through and caching it all once per screen. each part type (and with capi part src url) should only need to be read once.

dtiwarATS: @bsparks, I have removed the variable picker changes so that at least we can merger this branch. I will work on variable picker in different branch once we decided on the approach

dtiwarATS: Hey @bsparks @dsuriano , I implemented it with the way you suggested above:

`adaptivitySchema = await instance.getAdaptivitySchema({ currentModel: part, editorContext: context });`

It is working as expected. Also, this does not any changes for existing schema definitions. Take a look at the code and let me know.

assets/src/components/parts/janus-capi-iframe/schema.ts
```diff
@@ -19,9 +19,33 @@ export const schema: JSONSchema7Object = {
 
 export const getCapabilities = () => ({
   configure: true,
-  save: true,
 });
 
+export const adaptivitySchema = ({
```
bsparks, line 24: it is supposed to be an async function, though I suppose it doesn't matter until you use async inside it

bsparks (approved):

PR #1570, opened 2021-09-23
[JAN-659 | PMP-1843] remove elements from Question Bank panel that are not applicable

schowkekarats360:


bsparks:
don't we already have a `bank.ts` file, and it was only separate because it was being appended to the regular screen, but now if we're going to have a completely separate schema for banks then it seems like we might as well include the question bank specific properties in it.

bsparks (approved):

PR #1576, opened 2021-09-24
[ENHANCEMENT] Export analytics / insights raw data

LEDThereBeLight:
Per request from ASU.

Button added to insights page.


darrensiegel (changes requested):
This downloads the derived analytics data.  What we need actually is a download of the "raw snaphot" data.

LEDThereBeLight:
> This downloads the derived analytics data. What we need actually is a download of the "raw snaphot" data.

I added the "raw" snapshot data minus the ids. The download lists all snapshots rather than aggregating by section/activity/objective.

Since the derived data download was already in place, I figured it was worth leaving in place.

lib/oli_web/live/insights/insights.ex
```diff
@@ -141,4 +144,128 @@ defmodule OliWeb.Insights do
...
+      "Correct?",
+      "Activity Score",
+      "Activity Out Of",
+      "Hints Requested",
+      "Part Score",
+      "Part Out Of",
+      "Student Response",
+      "Feedback",
+      "Section Title",
+      "Section Slug"
```
darrensiegel, line 182: Add in the creation date field

darrensiegel (changes requested):

darrensiegel (approved):

PR #1580, opened 2021-09-24

dsuriano:
Includes work for PMP-1854: Dropdown should limit selection to valid combinations


bsparks (approved):
haven't tried it, but don't see any issues from just looking at the code

PR #1591, opened 2021-09-29

dsuriano:
Includes:
* PMP-1885: Wire in Session and Lesson Variables to variable picker
* PMP-1915: Show a placeholder Value in Rule Fields
* BUGFIX: adding a new adaptive rule from the bottom panel did not insert in proper order
* BUGFIX: changing the lesson title and reloading the page would crash due to mismatched slug
* BUGFIX: using keyboard tab navigation would trigger DB saves on every RightPanel field blur.


bsparks (approved):
Have not tested but looks good

PR #1604, opened 2021-10-01
PMP-1895 | JAN-713

dtiwarATS:


assets/src/apps/delivery/store/features/groups/actions/deck.ts
```diff
@@ -150,7 +151,6 @@ export const initializeActivity = createAsyncThunk(
         // shouldn't happen, but ignore I guess
         return { ...s, value: modifiedValue };
       }
-      arrInitFacts.push(`${s.target}`);
```
dtiwarATS, line 153: when it was here, it will miss the popup component's init state config because in that case, the function returns before it reaches here i.e. it returns from line number 152 because it won't find the ownerActivity in case of popup so just a minor change

bsparks (approved):

PR #1605, opened 2021-10-01
[JAN-661 | PMP-1845] - slider css issues

schowkekarats360:


assets/src/components/parts/janus-slider/Slider.tsx
```diff
@@ -285,16 +291,21 @@ const Slider: React.FC<PartComponentProps<SliderModel>> = (props) => {
               className={` slider `}
               id={internalId}
               onChange={handleSliderChange}
+              list={showTicks ? `lst${internalId}` : ''}
             />
+            {showTicks && (
+              <datalist id={`lst${internalId}`}>
```
bsparks, line 297: I tried this locally and didn't see anything. is there supposed to be some CSS changes to go with it? also, not sure about what appears to be hard coded "ticks" shouldn't they be reflected dynamically based on the increment value and the total? (so if increment is 1 there would be 100 tick marks if the max is 100)

assets/src/components/parts/janus-slider/SliderAuthor.tsx
```diff
@@ -75,6 +75,11 @@ const SliderAuthor: React.FC<AuthorPartComponentProps<SliderModel>> = (props) =>
 
   return (
     <div data-janus-type={tagName} style={styles} className={`slider`}>
+      {showLabel && (
+        <label className="input-label" htmlFor={internalId}>
+          {label}
+        </label>
+      )}
```
bsparks, line 82: shouldn't we also show the tick marks for authoring?

assets/src/components/parts/janus-slider/Slider.tsx
schowkekarats360: yeah... this part got checked-in by mistake.... actually this is in progress.. other two issues mentioned in the ticket are resolved in this PR.

bsparks (approved):

PR #1607, opened 2021-10-01
[FEATURE] [MER-318] - Added support for underline formatting.

chrislawson:
This PR adds support for underlining text within the editor as well as displaying the underlined text.

- Added ability to underline selected code in the editor either using the toolbar or standard cmd+U key shortcut
- Added underline rendering to editor and preview

To Test:

1. Login as Author and create new project or use existing.
2. Add a new practice page. 
3. Within this page add some text then highlight it. When the toolbar appears, click the U button to underline the content.
4. Press the Preview button to confirm that text is underlined then exit preview mode.
5. Select the underlined text. Press the {cmd|ctrl}+U keyboard shortcut to remove underline.

### DIFF
--- a/assets/src/components/editing/editor/modelEditorDispatch.tsx
+++ b/assets/src/components/editing/editor/modelEditorDispatch.tsx
@@ -106,6 +106,8 @@ export function markFor(mark: ContentModel.Mark, children: any): JSX.Element {
       return <sub>{children}</sub>;
     case 'sup':
       return <sup>{children}</sup>;
+    case 'u':
+      return <u>{children}</u>;
     default:
       return <span>{children}</span>;
   }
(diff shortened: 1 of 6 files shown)


bsparks:
Along with other pure styling elements, the original HTML Underline (<u>) element was deprecated in HTML 4; however, <u> was restored in HTML 5 with a new, semantic, meaning: to mark text as having some form of non-textual annotation applied.

The current "correct" way is to use CSS with `text-decoration: underline`

darrensiegel:
@chrislawson  Just pinging you to see what you plan to do with this PR.  I think Ben's suggestion on implementing a more "correct" way to underline makes sense.  If you can make that change and get the branch up to date I'm sure we can land this soon.

LEDThereBeLight:
Closing from inactivity

PR #1608, opened 2021-10-01
[BUGFIX] Feedback Rendering

bsparks:
multiple feedbacks are supposed to render individually.
applying some styles to get closer to SS defaults of most themes.
applying other "screen" settings that were previously ignored.

### DIFF
--- a/assets/src/apps/delivery/layouts/deck/components/FeedbackRenderer.tsx
+++ b/assets/src/apps/delivery/layouts/deck/components/FeedbackRenderer.tsx
@@ -29,7 +29,9 @@ const FeedbackRenderer: React.FC<FeedbackRendererProps> = ({ feedbacks, snapshot
       <style>
         {`
           .feedback-item > * {
-            position: relative !important;
+            position: static !important;
+          }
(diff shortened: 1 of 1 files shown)


assets/src/apps/delivery/layouts/deck/components/FeedbackRenderer.tsx
```diff
@@ -30,19 +24,36 @@ const FeedbackRenderer: React.FC<FeedbackRendererProps> = ({ feedbacks, snapshot
     [snapshot],
   );
 
-  // TODO: other handlers for parts, "advanced" things like tracking part responses within feedback??
-
   return (
-    <div className="feedback-item">
+    <Fragment>
```
dsuriano, line 28: 🎨 React auto-includes (no import required) a shorthand version of `<Fragment>` as: `<>` and `</>`.

dsuriano: Forgot to mention, the only caveat is that the shorthand version doesn't support the `key` prop if you need it...

PR #1611, opened 2021-10-04
PMP-1933 | JAN-744

dtiwarATS:


assets/src/components/parts/janus-popup/Popup.tsx
```diff
@@ -55,10 +55,6 @@ const Popup: React.FC<PartComponentProps<PopupModel>> = (props) => {
     if (isOpen !== undefined) {
       setShowPopup(isOpen);
     }
-    const openByDefault: boolean | undefined = currentStateSnapshot[`stage.${id}.openByDefault`];
```
dtiwarATS, line 58: We don't need this here. We already handle the "openByDefault" on line number 101.

`setShowPopup(!!pModel.openByDefault);`

bsparks (approved):

PR #1612, opened 2021-10-04
PMP-1931 | JAN-743

dtiwarATS:


assets/src/adaptivity/operators/equality.ts
```diff
@@ -16,10 +16,11 @@ export const isEqual = (factValue: any, value: any): boolean => {
   if (Array.isArray(factValue)) {
     let compareValue = value;
     const updatedFactValue = parseArray(factValue);
-    if (Array.isArray(value)) {
```
dtiwarATS, line 19: if the value  = '[0,0]', it was never going inside the if condition as  

`if (Array.isArray(value)) {}` is never true for value '[0,0]'

So, we had to parse it before checking that. It should not impact anything.

bsparks (approved):

PR #1628, opened 2021-10-05
[FEATURE] allow for env var driven provider changes

darrensiegel:
This PR adjusts how the system allows configuration of the payment provider.  Previously, a recompile was necessary to change providers. This PR allows the `PAYMENT_PROVIDER` env variable to determine the payment provider.  The valid values are `"stripe"` or `"none"`.  The default is `"stripe"` and any other value besides stripe or none defaults to `"none"`.  

When the payment provider is set to `"none"`, no link to initiate a direct payment is shown, only a link to redeem a payment code.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -52,7 +52,7 @@ config :oli,
       ),
     favicons: System.get_env("BRANDING_FAVICONS_DIR", "/favicons")
   ],
-  payment_provider: OliWeb.PaymentProviders.StripeController
+  payment_provider: System.get_env("PAYMENT_PROVIDER", "stripe")
 
 config :oli, :stripe_provider,
   public_secret: System.get_env("STRIPE_PUBLIC_SECRET"),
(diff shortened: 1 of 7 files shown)


eliknebel (approved):

config/config.exs
```diff
@@ -52,7 +52,7 @@ config :oli,
       ),
     favicons: System.get_env("BRANDING_FAVICONS_DIR", "/favicons")
   ],
-  payment_provider: OliWeb.PaymentProviders.StripeController
+  payment_provider: System.get_env("PAYMENT_PROVIDER", "stripe")
```
eliknebel, line 55: A configuration pulled from config.exs will be compile time. If the intent is a release-time config, then this value should be read from the environment in `config/releases.exs`

eliknebel (changes requested):

eliknebel:
Should the default be 'none' since it is more likely someone using the system will forget to set up their stripe configs, and therefore should see the 'Direct payments not enabled' message?

eliknebel (approved):

PR #1633, opened 2021-10-06
[BUG] PMP-1647 | JAN-390

dtiwarATS:


assets/src/components/parts/janus-popup/Popup.tsx
```diff
@@ -51,8 +51,12 @@ const Popup: React.FC<PartComponentProps<PopupModel>> = (props) => {
     // result of init has a state snapshot with latest (init state applied)
     setInitSnapshot(initResult);
     const currentStateSnapshot = initResult.snapshot;
+    const isOpenByDefault: boolean | undefined = currentStateSnapshot[`stage.${id}.openByDefault`];
```
dtiwarATS, line 54: For fixing one of the issue in Seasons lesson, I removed this one but we need this. Also, I have changed the sequence of these 2 variables i.e. isopen and openbydefault.

assets/src/components/parts/janus-popup/Popup.tsx
```diff
@@ -51,8 +51,12 @@ const Popup: React.FC<PartComponentProps<PopupModel>> = (props) => {
...
+    if (isOpenByDefault !== undefined) {
+      setShowPopup(isOpenByDefault);
+    }
     const isOpen: boolean | undefined = currentStateSnapshot[`stage.${id}.isOpen`];
-    if (isOpen !== undefined) {
+    if (isOpen !== undefined && !isOpenByDefault) {
```
dtiwarATS, line 59: when the component loads for the first time, if "OpenByDefault" = true then we have to ignore the "isOpen". This only happens for the first time and after that user can change it from trap states.

I have tested this with both the lessons

PR #1636, opened 2021-10-06
MCQ Spacing Issue

dtiwarATS:


assets/src/components/parts/janus-mcq/MultipleChoiceQuestion.tsx
```diff
@@ -31,7 +31,7 @@ const getNodeText = (node: any): any => {
 
 const MCQItemContent: React.FC<any> = ({ nodes, state }) => {
   return (
-    <div style={{ marginLeft: 18 }}>
+    <div style={{ left: 18, position: 'relative' }}>
```
dtiwarATS, line 34: The attribute that I was supposed to set was 'Left' and instead of that I added "marginLeft". Also, compared it with SS, the position needs to be 'relative' only for the <div> element.

bsparks (approved):

PR #1644, opened 2021-10-06
[BUGFIX][PMP-1650 / JAN-396] Video not appearing in Firefox on MAC

dsuriano:
Firefox chokes on some `.mp4` videos, but those same videos referenced as `.webm` work fine 🤷‍♂️.


assets/src/components/parts/janus-video/Video.tsx
```diff
@@ -438,6 +441,7 @@ const Video: React.FC<PartComponentProps<VideoModel>> = (props) => {
       onPause={handleVideoPause}
     >
       <source src={src} />
+      {isFirefox && <source src={updatedFFSrc} />}
```
bsparks, line 444: would it be bad to leave the browser test out and just put this as an alt video for all?

PR #1647, opened 2021-10-07
PMP-1951 | JAN-759

dtiwarATS:


dsuriano:
Tested this locally, and I can now see the new classes coming thru in the `customcssclass` prop (so I think that means you fixed the main issue), but they're not actually being applied to the component. Maybe that's a separate issue with `janus-input-text`?

image

PR #1655, opened 2021-10-07
[BUGFIX] cannibal parts

bsparks:
force property editor to refresh per part id


dsuriano (approved):
tested locally and fixes issue.

PR #1669, opened 2021-10-11
[BUG] Popup Issues

dtiwarATS:


assets/src/components/parts/janus-popup/Popup.tsx
```diff
@@ -251,6 +251,15 @@ const Popup: React.FC<PartComponentProps<PopupModel>> = (props) => {
       ],
     });
   };
+  useEffect(() => {
+    const popupModalZ = config?.z || 1000;
+    const zIndexIcon = z;
+    const finalZIndex = Math.max(zIndexIcon + popupModalZ || 0, popupModalZ);
```
bsparks, line 257: I don't think you need the `|| 0` on there, probably should have it on `const zIndexIcon = z || 0;` instead;

dtiwarATS: Fixed

PR #1677, opened 2021-10-12
[BUG] PMP-1951 | JAN-759

dtiwarATS:


assets/src/components/activities/adaptive/components/common/PartComponent.tsx
```diff
@@ -222,6 +222,7 @@ const PartComponent: React.FC<AuthorProps | DeliveryProps> = (props) => {
     model: JSON.stringify(props.model),
     state: JSON.stringify(props.state),
     customCssClass,
+    class: customCssClass,
```
dtiwarATS, line 225: Need to change the class as well. We were only updating the customCSSClass attribute of the web component.

bsparks (approved):

PR #1699, opened 2021-10-14
PMP-1962 - Configure a ThemeId for new Torus lessons

schowkekarats360:
code for saving the changes is present below on line 68


assets/src/apps/authoring/store/page/actions/initializeFromContext.ts
```diff
@@ -31,6 +31,9 @@ export const initializeFromContext = createAsyncThunk(
       customCss: params.context.content.customCss,
       custom: params.context.content.custom,
```
bsparks, line 32:
```suggestion
      custom: params.context.content.custom || {},
```
without this, it gives me an error and fails

PR #1710, opened 2021-10-15
[FEATURE] New admin landing page

darrensiegel:
This PR introduces a new landing page for Admin functionality, and removes the pop-up menu from the sidebar.  This is done largely for future proofing the scalability of the Admin UX.

This PR also addresses review comments from the last PR, specifically:
```
1. We should probably redirect back to the users/authors view when an account is deleted.
2. If you try accessing a deleted user by reloading the page, it shows an unauthorized message (instead of 404). Probably not a huge deal but could be confusing
3. It would be nice include some sort of link to go back to the users/authors view from the details page
```

For 3) above, this PR applies Breadcrumbs to all Admin views.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### Enhancements
 
+- New Admin landing page
+
 ## 0.14.0 (2021-10-13)
 
 ### Bug Fixes
@@ -32,7 +34,6 @@
 - Add ability to remix materials from multiple projects
 - Fix insert content popup in page editor
 - Add blackboard LTI 1.3 configuration instructions
-
 
 ### Release Notes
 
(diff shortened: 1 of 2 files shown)


lib/oli_web/live/admin/admin_view.ex
```diff
@@ -0,0 +1,79 @@
...
+    {:ok,
+     assign(socket,
+       author: author,
+       breadcrumbs: breadcrumb()
+     )}
+  end
+
+  def render(assigns) do
+    ~F"""
+    <Groups>
```
bsparks, line 25: seeing this makes me feel like it belongs in a template file, I presume there is a reason that its not

darrensiegel: Yes, indeed.  Although right this "Admin View" landing page contains no "dynamic" content and could be instead represented with a statically rendered template, I expect that there will be future functionality on this page.  So instead of having to reformulate it again in the future, I simply went with a LiveView implementation from the beginning.

PR #1731, opened 2021-10-21
[ENHANCEMENT] Popup Content

LEDThereBeLight:
_**Popup Content**_

**Changes**
- Popup: button in formatting toolbar. Can be triggered by "click" (or focus), or "hover" (or focus).
- Standardized most Z-indexes - certain elements weren't at the right hierarchy (e.g. clipping with page header when scrolling)
- Insert toolbar - positioning/tooltips/icons not closing after insertion
- Formatting toolbar - positioning/tooltips
- Upgrade popover library - positioning issues
- Stop persisting selection along with RichText - fixes several selection problems, e.g. selection would be reset after making selections in between save/persistence calls

Outstanding issues:
- Image resizer collapses on mousemove
- Now that tooltips are showing for the formatting toolbar, they sometimes don't close when the toolbar itself closes. The same change that was made to the insert toolbar needs to be applied, but I ran out of time.

**Pictures**

_Formatting Toolbar_
image
image

_Popup Content Editor_
image


darrensiegel (changes requested):
This looks and works well.  I see only one regression that we need to fix before we can land this:  table formatting within the editor is messed up.  If you  insert a table in a page or an activity, the borders of the cells no longer render, which makes it very difficult to actually click into a cell to edit.  This seems to be working fine on master.

darrensiegel (approved):

PR #1791, opened 2021-11-01
[BUGFIX] popup window

bsparks:
enable popup to render its window within the context of the activity;
having both the window and icon live within the bounds of the same web component proved to be problematic.
these changes send a reference to the container element that the activity renders its contents into all the way down to the part level. this allows a part to render secondary content (popup's window) within the confines of the activity along side of other parts. this way the window and the icon can have different z indexes and positions.

### DIFF
--- a/assets/src/components/parts/janus-popup/PopupWindow.tsx
+++ b/assets/src/components/parts/janus-popup/PopupWindow.tsx
@@ -50,7 +50,7 @@ const PopupWindow: React.FC<PopupWindowProps> = ({
   // position is an offset from the parent element now
   popupModalStyles.left = config.x || 0;
   popupModalStyles.top = config.y || 0;
-  popupModalStyles.zIndex = config.z || 0;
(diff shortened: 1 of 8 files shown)


assets/src/components/parts/janus-popup/PopupWindow.tsx
```diff
@@ -74,18 +81,34 @@ const PopupWindow: React.FC<PopupWindowProps> = ({ config, parts, context, onClo
     height: '100%',
   };
 
+  const closeButtonSpanStyles: CSSProperties = {
+    marginLeft: 12,
```
dsuriano, line 85: do these unit-less values actually work? i always thought any non-zero integer required a unit value.

dsuriano (approved):

assets/src/components/parts/janus-popup/PopupWindow.tsx
bsparks: React writes them as px by default

PR #1794, opened 2021-11-02
[BUG FIX] wrap create section in db transaction

### LINKED ISSUE #1790: Section creation from LMS may result in 2 active sections at once
**Describe the bug**
If multiple users try to create a section from LMS at the same time, it may result in 2 active sections which will cause an Internal Server Error to throw on subsequent launches. To mitigate this, a check should be performed when creating the section and database operations should be wrapped in a transaction in the create section call in delivery controller.

eliknebel:
This PR fixes an issue that allows multiple active sections to be simultaneously created with the same context id. It also adds a database unique constraint on the `context_id and status = 'active'` so that future sections cannot get caught in this state. Adding this constraint required updating some unit tests.

Closes #1790

### DIFF
--- a/priv/repo/migrations/20211102141706_ensure_unique_active_section.exs
+++ b/priv/repo/migrations/20211102141706_ensure_unique_active_section.exs
@@ -1,8 +1,0 @@
-defmodule Oli.Repo.Migrations.EnsureUniqueActiveSection do
-  use Ecto.Migration
-
-  def change do
-    # guarantee there is only ever one active section with a given context id
-    create unique_index(:sections, [:context_id], where: "status = 'active'", name: :sections_active_context_id_unique_index)
-  end
-end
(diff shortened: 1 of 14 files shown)


priv/repo/migrations/20211102141706_ensure_unique_active_section.exs
```diff
@@ -0,0 +1,8 @@
+defmodule Oli.Repo.Migrations.EnsureUniqueActiveSection do
+  use Ecto.Migration
+
+  def change do
+    # guarantee there is only ever one active section with a given context id
+    create unique_index(:sections, [:context_id], where: "status = 'active'", name: :sections_active_context_id_unique_index)
```
darrensiegel, line 6: How well will this work in an existing database like Proton?  Do we need to run some queries to ensure that Tokamak or Proton are not in a state already where it would be impossible to add this index?

darrensiegel (changes requested):

darrensiegel (approved):

PR #1798, opened 2021-11-03
[FEATURE] PR1 for gating and scheduling

eliknebel:
This PR contains the first pieces of work toward gating and scheduling:
1. add migration for gating conditions
2. add schema and context
3. implement query by resource_ids
4. implement gating condition evaluator interface

### DIFF
--- a/lib/oli/delivery/gating.ex
+++ b/lib/oli/delivery/gating.ex
@@ -1,0 +1,120 @@
+defmodule Oli.Delivery.Gating do
+  @moduledoc """
+  The Delivery.Gating context.
+  """
+
+  import Ecto.Query, warn: false
+  alias Oli.Repo
+
+  alias Oli.Delivery.Gating.GatingCondition
+
+  @doc """
+  Returns the list of gating_conditions.
+
+  ## Examples
+
+      iex> list_gating_conditions()
+      [%GatingCondition{}, ...]
+
+  """
+  def list_gating_conditions do
+    Repo.all(GatingCondition)
+  end
+
+  @doc """
+  Returns the list of gating_conditions for the list of given resource ids
+
+  ## Examples
+
+      iex> list_gating_conditions([1,2,3])
+      [%GatingCondition{}, ...]
+
+  """
+  def list_gating_conditions(resource_ids) do
+    from(gc in GatingCondition,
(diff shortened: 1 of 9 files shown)


lib/oli/delivery/gating.ex
```diff
@@ -0,0 +1,120 @@
...
+  @doc """
+  Returns the list of gating_conditions for the list of given resource ids
+
+  ## Examples
+
+      iex> list_gating_conditions([1,2,3])
+      [%GatingCondition{}, ...]
+
+  """
+  def list_gating_conditions(resource_ids) do
```
darrensiegel, line 33: This must also include a section or section_slug as an argument, otherwise it will return gatingconditions across all course sections that utilize a particular resource.

lib/oli/delivery/gating.ex
```diff
@@ -0,0 +1,120 @@
...
+  @doc """
+  Returns the list of gating_conditions.
+
+  ## Examples
+
+      iex> list_gating_conditions()
+      [%GatingCondition{}, ...]
+
+  """
+  def list_gating_conditions do
```
darrensiegel, line 20: I can't imagine this is useful, perhaps only in a testing context.

lib/oli/delivery/gating/gating_condition.ex
```diff
@@ -0,0 +1,35 @@
...
+  schema "gating_conditions" do
+    field :type, Ecto.Enum,
+      values: [
+        :start_datetime,
+        :end_datetime
+      ]
+
+    # data used by the condition evaluator, e.g. start or end datetime, a list of
+    # resource_ids, etc.
+    field :data, :map
```
darrensiegel, line 18: We probably want to model this more explicitly in the database, and model this as a "subtype" in a single table approach (similiar to how Revision has different types).  So instead of an opaque "data" blob, create specific fields for each type of GatingCondition.  For now, with only supporting scheduling, we would simply add a "date_time" field.  As we add in additional GatingCondition types, such as "required completed attempt of resouruce", we will add any additional fields.  The :type attribute informs the code which "subtype" this condition is, and thus which subset of fields to use.

darrensiegel (changes requested):

darrensiegel (approved):

PR #1808, opened 2021-11-04
[BugFix]-Set-wider-iFrame-for-Bio-Blue-Planet-Report

malav2110:


bsparks (changes requested):
need to find a generic way to do this, based off the classes that the theme support most likely

PR #1828, opened 2021-11-08
PMP-2090 | JAN-879

dtiwarATS:


assets/src/adaptivity/operators/equality.ts
```diff
@@ -13,11 +13,17 @@ export const isEqual = (factValue: any, value: any): boolean => {
...
 
+  const typeOfValue = typeof value;
+  const typeOfFactValue = typeof factValue;
   if (Array.isArray(factValue)) {
     let compareValue = value;
     const updatedFactValue = parseArray(factValue);
+    let updatedValue = value;
     //Need to parse before we check "if" condition else it will fail for cases where value = '[0,0]'.
-    const updatedValue = parseArray(value);
+    if (!Array.isArray(value) && typeOfValue === 'number') {
```
dtiwarATS, line 23: Use case : fact value = [3], value = 3. parseArray was throwing invalid array item error so if fact is a array value and value is  of type number, wrapping it inside '[]'.

bsparks (approved):

PR #1833, opened 2021-11-09
[FEATURE] PR2 for gating and scheduling

eliknebel:
This PR contains the following PR2 work for gating and scheduling:

 - New field and migration for Section schema
 - Implement the function that creates the resource gating index
 - Implement a “check” function that given a resource id returns if it is gated or not, and if gated, which resources are currently gating this resource

### DIFF
--- a/lib/oli/delivery/hierarchy.ex
+++ b/lib/oli/delivery/hierarchy.ex
@@ -266,6 +266,36 @@ defmodule Oli.Delivery.Hierarchy do
   end
 
   @doc """
+  Build a map that contains a list of ancestor resource_ids for each node
+  """
+  def gated_ancestry_map(%HierarchyNode{} = root, %{} = gated_resource_id_map) do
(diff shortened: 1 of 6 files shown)


lib/oli/delivery/hierarchy.ex
```diff
@@ -265,6 +265,36 @@ defmodule Oli.Delivery.Hierarchy do
     end)
   end
 
+  @doc """
+  Build a map that contains a list of ancestor resource_ids for each node
```
darrensiegel, line 269: It would be helpful to include more documentation here, especially indicating that a gated resource list entry will contain itself.  maybe icluding a sample hierarchy and sample set of gates, with a resultant gating index is a good idea.  We want to be able to look at this code two years from now and immediately undertand what it is building and why

darrensiegel (changes requested):

darrensiegel (approved):

PR #1840, opened 2021-11-09
Pmp 1421 text flow

candert1:
This is a big PR. A few highlights are:

- Updated the light.scss file by adding imports of new style files
- Added a image.scss file in case I need it late. I may not though.
- Addressed margins in text-flow.scss for when you have multiple <p> elements within one text-flow activity
- Worked on popup.scss and figured out how to display popup buttons and the close button for the popup window.
- Styled the image-carousel.scss. This one _should_ be good to go. 
- Styled the FITB activity. Still working on the correct/incorrect states of the dropdowns. The fill-blanks.scss file is just a reference file since the style needs to be minified and placed into the activity's JSON.
- I haven't touched slider.scss much

### DIFF
--- a/assets/styles/themes/delivery/adaptive_themes/default/activities/fill-blanks.scss
+++ b/assets/styles/themes/delivery/adaptive_themes/default/activities/fill-blanks.scss
@@ -142,8 +142,8 @@ p .dropdown-blot .select2-container .select2-selection--single .select2-selectio
   width: 16px;
   height: 16px;
   border-radius: 10px;
-  background-color: #4ad3ad;
-  background-image: url(data:image/svg+xml;utf8,<svg xmlns=http://www.w3.org/2000/svg width=8 height=6 viewBox=0 0 8 6> <g fill=none fill-rule=evenodd transform=translate(-3 -4)> <rect width=14 height=14 fill=%23D8D8D8 opacity=0/> <polygon fill=%23FFF points=5.857 10 3 7.121 3.806 6.309 5.857 8.37 10.194 4 11 4.818/> </g> </svg>);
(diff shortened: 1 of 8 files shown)


assets/styles/themes/delivery/adaptive_themes/default/activities/fill-blanks.scss
```diff
@@ -0,0 +1,158 @@
+.ql-container.ql-snow {
```
dsuriano, line 1: What are you thoughts on nesting / making more specific all these classes with something like `.fib-container` or `janus-fill-blanks`? I'm slightly concerned that they could spill over to unintended elements elsewhere in a lesson...

assets/styles/themes/delivery/adaptive_themes/default/activities/image-carousel.scss
```diff
@@ -0,0 +1,88 @@
+/*-----------------------------------------
+    Carousel
+-----------------------------------------*/
+/*-- Container --*/
+#stage-stage janus-image-carousel div .swiper-container {
```
dsuriano, line 5: Curious if the `#stage-stage` is necessary... have you tested without?

dsuriano (commented):
Overall things are looking good, but I'm curious if the "leaky styles" will come back to bite later.

I would think as a rule-of-thumb that each component should encapsulate all of it's own styles... and some of them seem to be doing it, some seem to be partially doing it, and others not so much... 

Is there a particular reason all component styles aren't fully encapsulated?

dsuriano (approved):
Discussed nesting and other comments on Zoom.

PR #1843, opened 2021-11-10
PMP-2125 | JAN-915

dtiwarATS:


assets/src/adaptivity/rules-engine.ts
```diff
@@ -1,3 +1,4 @@
+import { JanusConditionProperties } from 'apps/authoring/components/AdaptivityEditor/ConditionsBlockEditor';
```
bsparks, line 1: I think we need to split this type out into a different file. the rules engine is part of the server build for the engine and should not depend on a file inside the app. we can probably reverse it and have the app depend on a type in the rules-engine

assets/src/adaptivity/rules-engine.ts
```diff
@@ -113,6 +115,15 @@ const processRules = (rules: JanusRuleProperties[], env: Environment) => {
           }
         }
       }
+      //if it type ===3 then it is a array. We need to wrap it in [] if it is not already wrapped.
+      if (
+        typeof ogValue === 'string' &&
+        condition?.type === 3 &&
```
bsparks, line 121: would be good to use the enum so it's clear what 3 is

bsparks (changes requested):

assets/src/adaptivity/rules-engine.ts
dtiwarATS: Done

bsparks (approved):

PR #1846, opened 2021-11-11
PMP-2067 | JAN-859

dtiwarATS:


assets/src/apps/delivery/components/ActivityRenderer.tsx
```diff
@@ -420,7 +420,33 @@ const ActivityRenderer: React.FC<ActivityRendererProps> = ({
     }
     notifyStateMutation();
   }, [mutationTriggered]);
+  const debounceStateChanges = async (changes: any) => {
+    if (!ref.current) {
+      return;
+    }
+    const currentStateSnapshot: any = {};
+    if (changes.changed.length > 1 && Array.isArray(changes.changed)) {
```
bsparks, line 428: seems like if we need to check if it is an array that should happen before the length check, but maybe we don't need to check

assets/src/apps/delivery/components/ActivityRenderer.tsx
```diff
@@ -420,7 +420,33 @@ const ActivityRenderer: React.FC<ActivityRendererProps> = ({
     }
     notifyStateMutation();
   }, [mutationTriggered]);
+  const debounceStateChanges = async (changes: any) => {
```
bsparks, line 423: not sure why it is called `debounceStateChanges` as there is no debounce, maybe just `handleStateChangeEvents`

bsparks (changes requested):

assets/src/apps/delivery/components/ActivityRenderer.tsx
dtiwarATS: Done

bsparks (approved):

PR #1848, opened 2021-11-11
[FEATURE] Implement more efficient determination of Previous and Next links 

### LINKED ISSUE #1837: Previous Next calculation should be precalculated and cached: CMU-220
The "Previous" and "Next" page in course section delivery is calculated at page render time. This calculate requires fetching all course section resources, constructing the Hierarchy, then walking it to determine the previous and next. 

This is not efficient at all, and should be replaced by an approach that precalculates and caches the "Previous" and "Next" pages for all pages in the course, in a new, top-level "previous_next" JSON blob at the Course Section level.

darrensiegel:
Closes #1837

During the instructor preview work, it was remembered/discovered that the implementation of the determination of the "Previous" and "Next" pages in delivery was a potentially expensive operation.  This calculation was relying on the full delivery hierarchy being queried (returning all revisions of all resources in the section) and then reconstructed, then traversed to determine the preview and next pages. This was happening on every single page access in delivery. 

This PR significantly streamlines this by precalculating the "previous" and "next" pages link information in a new `previous_next_index` field in the `sections` schema.  With the section already being in the `conn.assigns` all that has to be done is at most three (likely constant time, maybe log n) lookups into this map.  One for the current page, and one each for the previous and next. 

Notice that the migration to add this field does nothing to set its content for existing sections. This is handled, just in time, by the `Oli.Delivery.PreviousNextIndex.retrieve` method that will detect when this field is `nil` and automatically build the index.  

The index is updated after every "remix" operation and every "apply course updates" operation.

### DIFF
--- a/lib/oli/authoring/editing/page_editor.ex
+++ b/lib/oli/authoring/editing/page_editor.ex
@@ -19,7 +19,6 @@ defmodule Oli.Authoring.Editing.PageEditor do
   alias Oli.Delivery.Page.ActivityContext
   alias Oli.Rendering.Activity.ActivitySummary
   alias Oli.Activities
-  alias Oli.Delivery.Page.PageContext
   alias Oli.Authoring.Editing.ActivityEditor
 
   import Ecto.Query, warn: false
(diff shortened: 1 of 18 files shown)


eliknebel (changes requested):
When I try to access an existing section, I'm running into the following error:

Screen Shot 2021-11-16 at 10 20 47 AM

eliknebel (approved):

PR #1849, opened 2021-11-11
[FEATURE] everapp structure closer match JAN-919 PMP-2129

bsparks:


assets/src/apps/delivery/layouts/deck/EverApps.ts
```diff
@@ -28,14 +28,14 @@ export const everAppActivityState: ActivityState = {
 
 export const EverAppActivity = {
   id: 'aa_3864718503',
```
dsuriano, line 30: Should this be a hardcoded value?

dsuriano (approved):
Not sure how to test this locally, but otherwise code looks good... just one little question below.

assets/src/apps/delivery/layouts/deck/EverApps.ts
bsparks: it gets overwritten on render

PR #1855, opened 2021-11-12
PMP-1783 formatted, hover/focus styles added, streamlined.

candert1:
formatted to SASS, added some hover styles, and removed unnecessary styles and !important tags
https://jira.unicon.net/jira/browse/PMP-1783


dsuriano (approved):
tested locally and looks good.

PR #1857, opened 2021-11-12
PMP-2035 fill in the blanks

candert1:
improved correct and incorrect dropdown states
https://jira.unicon.net/jira/browse/PMP-2035


dsuriano (approved):
tested locally by manually adding correct/incorrect classes and looks good.

PR #1858, opened 2021-11-12
PMP-1960 Formatted to Sass, updated feedback

candert1:
Formatted to Sass, updated feedback so that it behaves as expected again.

https://jira.unicon.net/jira/browse/PMP-1960


dsuriano (approved):
tested locally and looks good.

PR #1860, opened 2021-11-15
PMP-2125 | JAN-915

dtiwarATS:


assets/src/adaptivity/operators/contains.ts
```diff
@@ -21,13 +25,7 @@ const handleContainsOperator = (
       return factValue.toLocaleLowerCase().includes(value.toLocaleLowerCase());
     }
     value = parseArray(value);
-    return (
```
dtiwarATS, line 24: No Need to parse it again because we are already calling parseArray on value.

bsparks (approved):

PR #1862, opened 2021-11-15
[FEATURE] Preview activity bank selections

darrensiegel:
This PR augments the instructor preview feature by allowing activity bank selections to be previewed.  "Previewing" an activity bank selection is a link that takes the user to a new view that allows paged access to all possible activities - in preview mode - that could be selected for fulfillment for the bank.


eliknebel (approved):
My only comment here would be there should be a way for a previewer to return to the page they came from, maybe a back link or some sort of navigation breadcrumb.

This can probably be added in a followup pr

PR #1865, opened 2021-11-16
[JAN-947 | PMP-2144]   [JAN-679 | PMP-1964] - Wrong trap states firing

schowkekarats360:
PMP-2144 - sessons.visit variable for the page is getting set to 1 as soon as we come on the page which is causing the trapstate to be wrong as the criteria is sessions.visit should be 0.

PMP-1964 - The score was getting calculated for wrong attempts also when the negativescoring is not allowed, so that condition is added.

### DIFF
--- a/assets/src/adaptivity/rules-engine.ts
+++ b/assets/src/adaptivity/rules-engine.ts
@@ -306,6 +306,8 @@ export const check = async (
   }
 
   let score = 0;
+  //below condition make sure the score calculation will happen only if the answer is correct and
+  //in case of incorrect answer if negative scoring is allowed then calculation will proceed.
   if (isCorrect || scoringContext.negativeScoreAllowed) {
     if (scoringContext.trapStateScoreScheme) {
       // apply all the actions from the resultEvents that mutate the state
--- a/assets/src/adaptivity/rules-engine.ts
+++ b/assets/src/adaptivity/rules-engine.ts
@@ -306,31 +306,34 @@ export const check = async (
   }
 
   let score = 0;
-  if (scoringContext.trapStateScoreScheme) {
-    // apply all the actions from the resultEvents that mutate the state
-    // then check the session.currentQuestionScore and clamp it against the maxScore
-    // setting that value to score
-    const mutations = resultEvents.reduce((acc, evt) => {
-      const { actions } = evt.params as Record<string, any>;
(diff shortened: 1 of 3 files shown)


assets/src/apps/delivery/store/features/groups/actions/deck.ts
```diff
@@ -65,7 +65,7 @@ export const initializeActivity = createAsyncThunk(
     const visitOperation: ApplyStateOperation = {
       target: `session.visits.${currentSequenceId}`,
       operator: '+',
-      value: 1,
+      value: 0,
```
bsparks, line 68: hmm if we really aren't supposed to do this here, then I think we should just remove it all together. no point in *adding* zero

assets/src/adaptivity/rules-engine.ts
```diff
@@ -306,31 +306,34 @@ export const check = async (
...
-        (action: any) =>
-          action.type === 'mutateState' && action.params.target === 'session.currentQuestionScore',
-      );
-      return acc.concat(...acc, mActions);
-    }, []);
-    if (mutations.length) {
-      const mutApplies = mutations.map(({ params }) => params);
-      bulkApplyState(mutApplies, env);
-      score = getValue('session.currentQuestionScore', env) || 0;
+  if (isCorrect || scoringContext.negativeScoreAllowed) {
```
bsparks, line 309: im not sure that isCorrect should be involved here? or if negativeScoreAllowed? I'm non following the business rules logic, maybe we can explain it in a comment or something

schowkekarats360: I didnt acually got your point.. however i have added comments. let me know if that helps.

bsparks: well I think one thing that I'm nervous about is that I think this will prevent mutations from updating the current score how they see fit unless it is correct. so it could be an incorrect with a mutation to change current score and that wouldn't work. also, the other scoring method that is based on attempts won't be able to save the in progress score... maybe that's what the issue is? hmm yes, looking at the ticket they just don't want to *see* the score until it is correct... hmmm OK well I guess then I can just assume that they won't be using mutate unless negativeScore is allowed...

bsparks (approved):

PR #1866, opened 2021-11-16
[FEATURE] [MER-312] Community and admin CRUD

mcass19:
## Context 
Communities introduces a mechanism to group users at a level higher than an organization - primarily to provide instructors with a broader collection of shared projects.

## Description
This PR implements the CRUD operations for Communities which includes:
- Communities index view for system and community admins
  - Include a checkbox to view the deleted ones
- Community show view for system and community admins
  -  Ability to edit the community
  -  Ability to delete the community
  - Ability to add and remove community admins
- Community create view for system admins
- Added a communities section in the sidebar when user is an author and admin for at least one community


assets/styles/authoring/community.scss
```diff
@@ -0,0 +1,20 @@
+#communities-table .pagination .page-item {
+  &.active {
```
darrensiegel, line 2: These style definitions/overrides for the paging component should be removed.  I'm not sure what the intent was here, but these are causing the "Paging" component to render (in light mode) with dark gray backgrounds that make it difficult to read the text.  Removing lines 1-15 seems to restore the correct rendering.

lib/oli_web/live/community_live/index.ex
```diff
@@ -0,0 +1,120 @@
...
+      Enum.reduce(filter, %{}, fn {field, value}, acc ->
+        Map.put(acc, field, String.split(value, ","))
+      end)
+
+    query_str = String.downcase(query)
+    status_list = filter["status"]
+
+    Enum.filter(socket.assigns.communities, fn c ->
+      String.contains?(String.downcase(c.name), query_str) and
+        Enum.member?(status_list, Atom.to_string(c.status))
```
darrensiegel, line 41: Given that "status" is effectively user input (anyone can manually manipulate the query string), this implementation should probably filter the status list against a whitelist of what it knows to be valid statuses.  The potential for bad data (especially malicious) becomes more problematic when we consider that `Enum.member?` within `Enum.filter` is n^2.

lib/oli_web/plugs/set_user.ex
```diff
@@ -21,13 +22,23 @@ defmodule Oli.Plugs.SetCurrentUser do
     if author = Pow.Plug.current_user(conn, pow_config) do
       cond do
         current_author = Repo.get(Author, author.id) ->
+          is_community_admin =
+            current_author.id
+            |> Accounts.list_admin_communities()
```
darrensiegel, line 27: Given that this plug executes on virtually every end user request, we need to be careful here about keeping this lean.  This change introduces a second database read after the retrieval of the author in order to fetch all admin communities that the user belongs to.  It would be optimal, instead, to replace the `Repo.get(Author, author.id)` with a call to a new function that retrieves both the Author and pulls in a count of the number of communities that they are an admin to (say, perhaps, as a virtual field in the Author schema).  In this manner, we avoid the latency of multiple, sequential database reads.

darrensiegel (changes requested):
Overall, this is an **extremely** impressive collection of work, especially given the fact that the author of the PR has had zero assistance from the core team with regards to both this feature and just the system implementation in general. 

Beyond the couple of things that we want to change that I noted inline, this impl seems to favor a naming convention regarding LiveView views and components that differs from the one that has naturally evolved in Torus.  This impl explicitly suffixes the module names of components with `Component` and leaves the root live view module names without any special designation.  To bring this impl in line with Torus, I would recommend removing the `Component` suffix from these new components and adding a `View` suffix to the new LiveViews.

mcass19:
Hey @darrensiegel! Thanks for your response.

Yes, we know it is a lot, but we wanted to make the foundations of communities all at once. The development of next related features will surely be smaller and with a more continuous flow.

Regarding the naming convention, we are glad to hear your thoughts and change or discuss anything you want. As you said, we didn't have assistance from the core team, so we will make those changes and review in detail as much as you want :)

lib/oli_web/plugs/set_user.ex
mcass19: I like it! Thanks

assets/styles/authoring/community.scss
mcass19: We changed the color of the pagination component from green to some kind of gray due to some suggestion we received. In order not to contaminate other parts of the app, we only did it for communities. Maybe we can try other colors to make it more visible in light mode or if we prefer consistency, we can remove it without problem

jrissler: @mcass19 IMO probably just revert color change, and we can address later as part of upcoming design work.

lib/oli/accounts.ex
```diff
@@ -371,6 +371,33 @@ defmodule Oli.Accounts do
   """
   def get_author!(id), do: Repo.get!(Author, id)
 
+  @doc """
+  Gets a single author with the count of the communities for wich the author is an admin.
```
jrissler, line 375: Quick typo

lib/oli/accounts.ex
```diff
@@ -371,6 +371,33 @@ defmodule Oli.Accounts do
...
+
+  ## Examples
+      iex> get_author_with_community_admin_count(1)
+      %Author{community_admin_count: 1}
+
+      iex> get_author_with_community_admin_count(456)
+      nil
+
+  """
+  def get_author_with_community_admin_count(id) do
```
jrissler, line 385: Can you clean this function / query up a bit? I think it does capture the heart of what he wanted.

darrensiegel (approved):

PR #1872, opened 2021-11-17
PMP-2198 | JAN-1009

dtiwarATS:


assets/src/utils/common.ts
```diff
@@ -141,9 +141,16 @@ export const parseBool = (val: any) => {
...
+  const itemType = typeof item;
   if (!item?.length) return item;
-  // check if items are strings or numbers and converts if number
-  return !Number.isNaN(Number(item)) ? parseFloat(item) : item.trim();
+  if (!Number.isNaN(Number(item))) {
+    // check if items are strings or numbers and converts if number
+    return parseFloat(item);
+  } else if (itemType === 'string') {
+    //trim() only works on strings
+    return item.trim();
```
dtiwarATS, line 151: item.trim(); caused the because it was throwing error when calling .trim() on an array object.

bsparks (approved):

PR #1883, opened 2021-11-18
Small World

dtiwarATS:


assets/src/apps/delivery/components/TextParser.ts
```diff
@@ -0,0 +1,115 @@
+import { defaultGlobalEnv, evalScript, getAssignScript, getValue } from 'adaptivity/scripting';
```
dtiwarATS, line 1: I wanted to move this new file in Utils/ directory but for some reason, it was giving webpack compile error.

bsparks (approved):

PR #1884, opened 2021-11-18
PMP-2213 | JAN-1029

dtiwarATS:


assets/src/apps/delivery/components/preview-tools/Inspector.tsx
```diff
@@ -69,13 +69,22 @@ const Inspector: React.FC<InspectorProps> = ({ currentActivity }) => {
             (activity) => !!activity.content.partsLayout.find((p: any) => p.id === part),
           );
           if (ownerActivity.id === activity.id) {
-            const next = { ...col, ...statePuff[`${activity.id}|stage`] };
```
dtiwarATS, line 72: I saw an issue with the inspector removing MCQ  variables from list after performing trigger check. So fixed this and also tested Plate tectonics- inspector getting out of sync to make sure both works.

assets/src/components/parts/janus-mcq/MultipleChoiceQuestion.tsx
```diff
@@ -625,8 +625,8 @@ const MultipleChoiceQuestion: React.FC<PartComponentProps<McqModel>> = (props) =
     if (shouldSave) {
       saveState({
         numberOfSelectedChoices: newCount,
-        selectedChoice: newSelectedChoices.slice(-1)[0] || 0,
-        selectedChoiceText: updatedChoicesText.slice(-1)[0] || '',
+        selectedChoice: newSelectedChoices.length ? newSelectedChoices.sort()[0] : 0,
```
dtiwarATS, line 628: As per SS, It does not matter in which order you select the MCQ option (Multi select), the lowest one in the array gets set to 'selectedChoice' property.

bsparks (approved):

PR #1889, opened 2021-11-18
[Bugfix] [PMP-2173 / JAN-977] Popup Missing CustomCSS

dsuriano:
popup `customCssClass` - use 'em if you got em.


bsparks:
hmm well the customCSSClass should be being applied to the main component already, in PartComponent.tsx through the web component wrapper. so my concern is that there may be cases where "double" styles are applied. I think that we might need to fix via CSS instead as the class is probably targetting the wrong thing

PR #1893, opened 2021-11-22
PMP-2041 | JAN-837

dtiwarATS:


assets/src/apps/delivery/layouts/deck/components/FeedbackContainer.tsx
```diff
@@ -70,7 +70,14 @@ const FeedbackContainer: React.FC<FeedbackContainerProps> = ({
             </button>
           </div>
           <style type="text/css" aria-hidden="true" />
-          <div className="content" style={{ overflow: 'hidden auto !important' }}>
```
dtiwarATS, line 73: For some reason, the inline style` style={{ overflow: 'hidden auto !important' }}` was not getting applied so change it to selector base. Starting the selector with '#stage-feedback ' so that this only gets applied to feedback section because there is another '.content' in the main content view.

bsparks: inline styles in React don't work with important, the only way is through text in a style block;

assets/src/apps/delivery/layouts/deck/components/FeedbackContainer.tsx
```diff
@@ -70,7 +70,14 @@ const FeedbackContainer: React.FC<FeedbackContainerProps> = ({
             </button>
           </div>
           <style type="text/css" aria-hidden="true" />
```
bsparks, line 72: probably could have used this one, not sure why it was here... not sure if aria-hidden is needed for style blocks and screen readers, anyway, we can delete it now

bsparks (approved):

PR #1894, opened 2021-11-22
PMP-2169 | JAN-972

dtiwarATS:


assets/src/apps/delivery/store/features/groups/actions/deck.ts
```diff
@@ -145,8 +147,22 @@ export const initializeActivity = createAsyncThunk(
...
         return { ...s, value: modifiedValue };
       }
+      if (s.operator === 'bind to') {
+        arrBindToVars.push(`${modifiedValue}`);
+      }
       return { ...s, target: `${ownerActivity.id}|${s.target}`, value: modifiedValue };
     });
+    //if a variable is 'bindTo' another variable then we need to make sure that we reset the value of the original variable to ''
+    //otherwise if that original variable is already evaluated on previous screen then on the current screen, those values will be automatically populated
+    // for the current bindto target on page load by default which don't want
```
bsparks, line 157: hmm, why don't we want that? it seems like odd behavior. also seems like we probably shouldn't set everything to an empty string, what if it's a number type or something else? also, maybe if we really need to do this we can do it at a lower level either in the scripting applyState or maybe even in the scripting itself... but it only needs to be done here in init state and not mutate?

bsparks:
closing I believe we are looking for an alternate solution

PR #1900, opened 2021-11-23
default wrong trap state not moving to next screen

schowkekarats360:


assets/src/apps/authoring/components/AdaptivityEditor/AdaptivityEditor.tsx
```diff
@@ -152,7 +152,7 @@ export const AdaptivityEditor: React.FC<AdaptivityEditorProps> = () => {
             onChange={(changes: any) => {
               handleActionChange(action, changes);
             }}
-            allowDelete={!currentRule.correct}
+            allowDelete={!currentRule.default}
```
bsparks, line 155: seems like maybe this should be only if it is the default correct rule that we can't delete, not the defaultWrong rule

schowkekarats360: ohkay.... but in smartsparrow I saw that the defaultwrong for this was not allowed to be deleted either... so I make this change... Should I remove this or do you want to take a look?

bsparks: hmm when I'm looking in a lesson I don't see that, maybe it is something specific with forceProgress, i'll check another lesson

bsparks: yeah, when I check the lesson in the ticket, I see that you can't delete that one, I can't see anything in the editor that lets you set that though, maybe it's a legacy sparrow thing?? not sure, but IF I click the + to add an action and pick go to another screen then it changes it from next to something else and then I can delete it. so I think we are good to have this, and actually I just checked, it should be any correct that have to have a navigation on there, so I think we can just revert this part

schowkekarats360: ohk... I reverted back this change.

bsparks (approved):

PR #1924, opened 2021-12-02
PMP-1064 | JAN-163

dtiwarATS:


assets/src/components/parts/janus-video/Video.tsx
```diff
@@ -454,7 +454,7 @@ const Video: React.FC<PartComponentProps<VideoModel>> = (props) => {
       height="100%"
       /* className={cssClass} */
       autoPlay={autoPlay}
-      loop={autoPlay}
+      loop={enableReplay}
```
dtiwarATS, line 457: This is not related to this ticket. I saw this bug with  video component so fixed it here.

assets/src/adaptivity/scripting.ts
```diff
@@ -75,6 +75,11 @@ export const getExpressionStringForValue = (
         } catch (e) {
           actuallyAString = true;
         }
+      } else {
+        const testVal = getValue('foo', testEnv);
```
dtiwarATS, line 79: the variable was getting set as "let stage.iframe.css = video {width: 100%;}*{padding:0;margin:0;}"  which evaluates to undefined. So, I am just returning the actual value if this is the case.

dtiwarATS:
Closing this PR. I will create new one.

PR #1929, opened 2021-12-03
PMP-1064 | JAN-163

dtiwarATS:


assets/src/adaptivity/scripting.ts
```diff
@@ -57,25 +57,15 @@ export const getExpressionStringForValue = (
...
     let actuallyAString = false;
+    const expressions = extractAllExpressionsFromText(val);
+    // A expression will not have a ';' inside it. So if there is a ';' inside it, it is CSS.
+    const isCSSString = expressions.filter((e) => e.includes(';'));
+    if (isCSSString?.length) {
+      actuallyAString = true;
+    }
     try {
       const testEnv = new Environment(env);
-      const evalResult = evalScript(`let foo = ${val};`, testEnv);
```
dtiwarATS, line 62: Removing all the previous CSS related fixes. We don't need them anymore. I have test few CSS related tickets and they we working properly.

assets/src/adaptivity/scripting.ts
```diff
@@ -57,25 +57,15 @@ export const getExpressionStringForValue = (
...
-          );
-          //matchingCssElements !== null then it means it's a CSS string so set actuallyAString=true so that it can be wrapped in ""
-          if (matchingCssElements) {
-            actuallyAString = true;
-          }
-        } catch (e) {
-          actuallyAString = true;
-        }
-      }
+      evalScript(`let foo = ${val};`, testEnv);
```
bsparks, line 68: still we need to test `evalResult != null` for things besides CSS. also I think if we have determined via the CSS test above that `if (actuallyAString === true)` then we don't need to continue testing

dtiwarATS: Added changes for the second point. I am still not sure why we need the evalResult != null section because if it fails we already setting  "actuallyAString = true;" in catch section.

assets/src/adaptivity/scripting.ts
```diff
@@ -57,30 +57,21 @@ export const getExpressionStringForValue = (
...
+    if (isCSSString?.length) {
       actuallyAString = true;
     }
-
+    if (!actuallyAString) {
+      try {
+        const testEnv = new Environment(env);
+        evalScript(`let foo = ${val};`, testEnv);
+      } catch (e) {
+        // if we have parsing error then we're guessing it's CSS
```
bsparks, line 71: I think if we're here we're no longer guessing it's CSS because of above, but have determined that whatever it is can't be evaluated by the scripting engine and so we're assuming its a string

assets/src/adaptivity/scripting.ts
```diff
@@ -57,30 +57,21 @@ export const getExpressionStringForValue = (
...
+    // A expression will not have a ';' inside it. So if there is a ';' inside it, it is CSS.
+    const isCSSString = expressions.filter((e) => e.includes(';'));
+    if (isCSSString?.length) {
       actuallyAString = true;
     }
-
+    if (!actuallyAString) {
+      try {
+        const testEnv = new Environment(env);
+        evalScript(`let foo = ${val};`, testEnv);
```
bsparks, line 69: I think we still want to do
```
const testResult = evalScript(`let foo = ${val};`, testEnv);
if (testResult?.result !== null) {
  actuallyAString = true;
}
```
as well, because the catch is only for exceptions, or maybe we should be checking `testResult.result.message` exists?

dtiwarATS: Done

assets/src/adaptivity/scripting.ts
```diff
@@ -57,30 +57,25 @@ export const getExpressionStringForValue = (
...
+    if (isCSSString?.length) {
+      actuallyAString = true;
+    }
+    if (!actuallyAString) {
+      try {
+        const testEnv = new Environment(env);
+        const testResult = evalScript(`let foo = ${val};`, testEnv);
+        if (testResult?.result !== null) {
+          //expression {stage.foo} + {stage.bar} was failling if we set actuallyAString= true
+          actuallyAString = expressions?.length ? false : true;
```
bsparks, line 72: hmm what about if it's `round(37.234)` ?

dtiwarATS: It works properly and gets evaluated to 37 so the testResult?.result is null for round(37.234)

bsparks (approved):

PR #1936, opened 2021-12-06
[Feature] Advanced Author Lesson Diagnostics

bsparks:
this adds a "Lesson Diagnostics" window which right now just does a part ID validation for duplicates and problematic ids with an option to fix them from the window


dsuriano (approved):
Just the 1 small style suggestion 🎨

PR #1942, opened 2021-12-08
PMP-1263 | JAN-119

dtiwarATS:


assets/src/components/parts/janus-capi-iframe/ExternalActivity.tsx
```diff
@@ -857,9 +857,11 @@ const ExternalActivity: React.FC<PartComponentProps<CapiIframeModel>> = (props)
         const formatted: Record<string, unknown> = {};
         const baseKey = key.replace(`stage.${id}.`, '').replace(`app.${id}.`, '');
         const value = initState[key];
+        const typeOfValue = typeof value;
         const cVar = new CapiVariable({
           key: baseKey,
           value,
+          type: typeOfValue === 'string' ? CapiVariableTypes.STRING : CapiVariableTypes.UNKNOWN,
```
bsparks, line 864: this doesn't seem right, surely sometimes there are other types than string, plus UNKNOWN is made up in our system, doesn't exist in original capi

dtiwarATS: There is logic in CapiVariable(), if we pass UNKNOWN type, then it has a method getCapiType that finds the type. It does not send the CapiVariableTypes.UNKNOWN

dtiwarATS:
```
if (this.type === CapiVariableTypes.UNKNOWN) {
      this.type = getCapiType(options.value);
    }
```

This is the code

bsparks: ok, but then why are you overriding that logic by pre-setting it to STRING based on purely if it's a string, I think the logic inside is such that if you don't pass type in then it will start as UNKNOWN and then determine based on the value

bsparks (approved):

PR #1944, opened 2021-12-08
[BUG] PMP-2199 | JAN-1010 | Question bank issue

dtiwarATS:
@bsparks, I have done the changes only for Dropdown and MCQ component for now. I wanted to get your thoughts on it before rolling out to all the required components.

### DIFF
--- a/assets/src/apps/delivery/components/ActivityRenderer.tsx
+++ b/assets/src/apps/delivery/components/ActivityRenderer.tsx
@@ -21,7 +21,10 @@ import React, { useEffect, useRef, useState } from 'react';
 import { useSelector } from 'react-redux';
 import { clone } from 'utils/common';
 import { contexts } from '../../../types/applicationContext';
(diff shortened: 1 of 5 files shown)


dtiwarATS:
@bsparks , here is the other approach that we can look at. [https://github.com/Simon-Initiative/oli-torus/compare/master...dtiwarATS:Question_Bank_New_Approach?expand=1](url)

bsparks:
> @bsparks , here is the other approach that we can look at. https://github.com/Simon-Initiative/oli-torus/compare/master...dtiwarATS:Question_Bank_New_Approach?expand=1

hmm well we can't have all the custom component knowledge in there, we'll have to do something generic for all part components. hmm maybe expose a method on it like `getInitDefaults` or something and use it like we do in authoring.

dtiwarATS:
Closing this PR. I will create another one with new approach

PR #1955, opened 2021-12-09
Fixes the width of a rendered FillBlanks component

rmathis:
This fixes an issue where the user entered value for the width on Fill in the Blanks components was not being respected in the preview mode.

### DIFF
--- a/assets/src/components/parts/janus-fill-blanks/FillBlanks.tsx
+++ b/assets/src/components/parts/janus-fill-blanks/FillBlanks.tsx
@@ -91,11 +91,7 @@ const FillBlanks: React.FC<PartComponentProps<FIBModel>> = (props) => {
   );
   const [ready, setReady] = useState<boolean>(false);
   const wrapperStyles: CSSProperties = {
-    /* position: 'absolute',
-    top: y,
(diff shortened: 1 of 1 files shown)


assets/src/components/parts/janus-fill-blanks/FillBlanks.tsx
```diff
@@ -91,11 +91,7 @@ const FillBlanks: React.FC<PartComponentProps<FIBModel>> = (props) => {
   );
   const [ready, setReady] = useState<boolean>(false);
   const wrapperStyles: CSSProperties = {
-    /* position: 'absolute',
-    top: y,
-    left: x,
     width,
-    zIndex: z, */
```
dtiwarATS, line 98: the width and height of each part component is set on the custom web component which is in partComponent.tsx. We don't apply them directly on the component. Each component calls props.onResize and pass it width and height.

It seems that these issues are there in other components as well so I went ahead and created a PR that will fix all the component issues related to width and height. https://github.com/Simon-Initiative/oli-torus/pull/1957

bsparks:
alternate solution #1957 used instead

PR #1958, opened 2021-12-10
[BUG] PMP-2199 | JAN-1010 | Question Bank Reset Issue

dtiwarATS:


bsparks (changes requested):
change to shared static method as discussed

bsparks:
please check on my question about layers above the question bank; also test in open & free just in case

dtiwarATS:
@bsparks, if there are layers above the question bank, we need to ignore it. we should not reset it's parts. We only reset parts of the question bank layer part and it's activities.

I have made changes for the same and have tested it with adding a part on non-question bank layer. It is behaving similar now.

assets/src/apps/delivery/store/features/groups/actions/deck.ts
```diff
@@ -126,11 +152,39 @@ export const initializeActivity = createAsyncThunk(
...
+    const isQuestionBankActivity = currentSequenceLineage.some((s) => s.custom.isBank);
+
+    const quetionBankResetScript: ApplyStateOperation[] = [];
+    if (isQuestionBankActivity && currentActivityTree) {
+      // question bank children cannot be nested, so the parent must be the bank
+      const parentBank = currentSequenceLineage[currentSequenceLineage.length - 1];
+      // in delivery activity id is sequence id
+      const bankActivity = selectActivityById(rootState, parentBank.custom.sequenceId);
+      if (bankActivity) {
+        currentActivityTree.forEach((activity) => {
```
bsparks, line 170: I dont understand, doing the entire activity tree seems the opposite of not resetting layers, all of the layer ancestors will be in the tree. before I was *only* doing the question bank, not even the current activity (though the current activity should be doing it on its own during init anyway).

dtiwarATS: But we do have condition there which only reset variables of question bank activity and question bank layer.

`if (activity.id === parentBank.custom.layerRef || activity.id === bankActivity.id) {}`

Also, the reason for that is, layer don't get rendered as screen so the components that are on layer, they never get called if we do current activity reset

bsparks:
after further investigation I believe this functionality does *not* exist in SPR and we should not apply it.

PR #1966, opened 2021-12-13
PMP-2304 - cover screen numbered list color is black

schowkekarats360:


assets/src/components/parts/janus-text-flow/Markup.tsx
```diff
@@ -303,8 +303,17 @@ const Markup: React.FC<any> = ({
       );
     case 'li':
       // eslint-disable-next-line
-      const spanChildren = children.filter((child: any) => {
-        return child.props.tag === 'span';
+      let spanChildren: any[] = [];
+      children.map((child: any) => {
```
bsparks, line 307: you are using map as foreach, use either foreach or a reduce

PR #1968, opened 2021-12-13
[BUG FIX] Send actual total points available during line item creation

### LINKED ISSUE #1962: Send actual total points for line item instead of hardcoded 1: CMU-413

darrensiegel:
This PR changes LMS grade sync to send the actual maximum points available in a graded page during line item creation.  Previous to this PR, the impl sent a hard coded "1.0" value.  Which then effectively normalized the score and out_of from actual scores to a percentage (since the grade book considered the max point value as 1.0). 

This PR changes line item creation to first derive the max points possible. For basic pages this is simply a count of the activities that are present.  Note, this impl cannot be "attempt" based since line item creation can be invoked before any grade is generated.  Therefore it makes sense to only have a singular impl that is page content and activity content driven. 

Another interesting aspect of this is that because activity bank selections exist there can be situations where one student has a multi-input activity and another doesn't.  Prior to this PR, that would result in a different "out_of" value for each student.  To avoid this, this PR also changes "part attempt" score rollup to normalize to a value 1 for the activity attempt. This has the effect of making every activity present in a page worth the same 1 point, regardless of the number of parts.  With that change in place, we can simply just count the number of activities to determine max points available.  

For adaptive pages, we cannot do this, so we fetch all scoreable activity content and count the number of parts. 

I've tested this PR locally from canvas by both
1. Manually creating line items from the Admin/Instructor "manage LMS grades" view
2. Just in time line item creation by a student submitting a graded page. 

Closes #1962


eliknebel (approved):

dkantergithub:
I'm not sure if this is the right ticket for this issue.  I just tested using Torus 0.18.1 ported to ArgosProd with integration to Unicon's Canvas and unfortunately I didn't get the expected result.

As instructor, initially check Grades and confirmed there were no assessments.
Then tried to "update LMS line items"
**Got this error: Error getting LMS access token**
When I checked Grades, there were no assessments.

I went ahead and as a student I took the assessment (first lesson of first unit in BioBeyond product = a course like no other) and at the end I saw this:
Attempt Submitted
You've completed this graded assessment attempt.
**There was a problem updating your grade in your LMS**

Went back in as instructor and still saw no assignments under grades.

PR #1970, opened 2021-12-13
[BUG FIX] Allow cost to be determined for product-less sections MER-535

### LINKED ISSUE #1961: Student cannot access paid product.: CMU-412
https://eliterate.atlassian.net/browse/MER-535

darrensiegel:
Closes #1961 

The PR fixes an issue where the cost for a paid section that was not created from a product could not be calculated.  The reason the too strict pattern matching in `Oli.Delivery.Paywall.calculate_product_cost` where that impl asserted that a section had to be a blueprint.  When a course section is created from a product we do indeed pass the product to this function to determine the cost of the enrollable section.  But when a section is paid and has not been created from a product we pass the enrollable section itself to this method.


lib/oli/delivery/paywall.ex
```diff
@@ -148,26 +148,26 @@ defmodule Oli.Delivery.Paywall do
   end
 
   @doc """
-  Given a section blueprint (aka a product), calculate the cost to use this product for
-  a specific insituttion, taking into account any product-wide and product-specific discounts
+  Given a section (blueprint or enrollable), calculate the cost to use it for
+  a specific insitution, taking into account any product-wide and product-specific discounts
```
eliknebel, line 152: spelling *institution :)

eliknebel (approved):

eliknebel (approved):

PR #1973, opened 2021-12-13
[CHORE] silence datashop test error logs

eliknebel:
There might be a better solution to this issue but this PR attempts to silence the errors being logged during execution of some datashop tests


darrensiegel:
Closing as the root cause was addressed on the LMS-Lite branch.  `mix test` on `master` now runs clean

PR #1975, opened 2021-12-14
[BUG] PMP-2034 | JAN-831

dtiwarATS:


assets/src/components/parts/janus-mcq/MultipleChoiceQuestion.tsx
```diff
@@ -828,7 +828,11 @@ const MultipleChoiceQuestion: React.FC<PartComponentProps<McqModel>> = (props) =
   }
 
   return ready ? (
-    <div data-janus-type={tagName} style={styles} className={`mcq-input`}>
+    <div
+      data-janus-type={tagName}
+      style={styles}
+      className={`mcq-input ${customCssClass === 'no-wrap' ? customCssClass : ''}`}
```
dtiwarATS, line 834: Need to add custom class if it is 'no-wrap'. Rest of the classes already gets applied on custom component. I am not sure if we can add 'no-wrap' class from override file. @bsparks  thoughts?

dtiwarATS:
closing this PR. Something has happened and these changes are not resolving the issues now. They were working 2 days back. I am not sure if something changed somewhere that is having impact on this.

dtiwarATS:
@bsparks , I saw a strange behavior. The changes in this PR work when I point the SIM URL to

https://vft.inspark.education/VFTDresser/panos/DresserTTG4/Dresser.html

but it does not work with

https://vft.asu.edu/migrated/VFTDresser/panos/DresserTTG4/Dresser.html

It seems to be some issue with this VFT URL.

assets/src/components/parts/janus-mcq/MultipleChoiceQuestion.tsx
bsparks: is it not possible to adjust via css using something like `janus-mcq.no-wrap .mcq-input {`

dtiwarATS: Done. I will update the CSS override file

bsparks (approved):

PR #1976, opened 2021-12-14
[JAN-1144 | PMP-2348] - CSS seems to be blocking the background image

schowkekarats360:


bsparks:
from what I could tell, the CSS that is messing this up by using !important is an override file, so I think we should just alter the override file instead of doing this in the code which would make it harder to override using an override file

schowkekarats360:
closing the PR as the changes are done in the css file.

PR #1983, opened 2021-12-14
[FEATURE] move registrations view to a live view

eliknebel:
This PR is a followup to the deployment model fix. It moves the registrations listing into a live view with table sorting, searching and pagination so that this page can properly scale when more registrations are added to the system.

Screen Shot 2021-12-14 at 3 19 20 PM


darrensiegel (approved):
Tested locally, works great.

PR #1987, opened 2021-12-15
PMP-2328 | JAN-1124

dtiwarATS:


assets/src/components/parts/janus-capi-iframe/ExternalActivity.tsx
```diff
@@ -869,6 +869,13 @@ const ExternalActivity: React.FC<PartComponentProps<CapiIframeModel>> = (props)
         key: baseKey,
         value,
       });
+      const typeOfValue = typeof value;
```
dtiwarATS, line 872: This fixes the absorbance issue.

Note:- we can make this changes in ParseArray() but i did it here because in SIMs, they expect it as string instead of array object for multi-dimensional array.

bsparks (approved):

PR #1996, opened 2021-12-15
[FEATURE] [MER-350] Instructors permissions based on communities logic

mcass19:
Story: https://eliterate.atlassian.net/browse/MER-350

This feature improves the instructor permissions to select projects and products based on communities logic.

Places where things changed:
- LTI instructor in getting started create course section
- LMS Lite instructor in create new open and free sections
- Remix when editing a section (just listing publications)

Instructor will see projects and products when:
- They are part of a community assigned to their institution
- They are part of a community as a user
- If they have a linked author account
  - They are an author of content
  - An author has made content visible to their institution
  - Another author has shared content with them
- Global -> will see them when they are not associated to any community, or one of the associated communities allows it

### DIFF
--- a/lib/oli/accounts.ex
+++ b/lib/oli/accounts.ex
@@ -2,9 +2,6 @@ defmodule Oli.Accounts do
   import Ecto.Query, warn: false
   import Oli.Utils, only: [value_or: 2]
 
-  alias Oli.Repo
-  alias Oli.Repo.{Paging, Sorting}
-
   alias Oli.Accounts.{
     User,
     Author,
@@ -14,7 +11,14 @@ defmodule Oli.Accounts do
     AuthorPreferences
   }
 
+  alias Oli.Authoring.Course.{Project, ProjectVisibility}
+  alias Oli.Delivery.Sections.Section
   alias Oli.Groups.CommunityAccount
+  alias Oli.Institutions.Institution
+  alias Oli.Publishing
+  alias Oli.Publishing.Publication
+  alias Oli.Repo
+  alias Oli.Repo.{Paging, Sorting}
 
   def browse_users(
         %Paging{limit: limit, offset: offset},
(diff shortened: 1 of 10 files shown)


lib/oli/accounts.ex
```diff
@@ -597,4 +601,70 @@ defmodule Oli.Accounts do
...
+
+  ## Examples
+
+      iex> list_available_publications_and_products(nil, nil, false)
+      []
+
+      iex> list_available_publications_and_products(123, 1, true)
+      [{%Publication{project: %Project{}}, %Section{}}, ...]
+  """
+  def list_available_publications_and_products(nil, nil, false), do: []
```
darrensiegel, line 619: Adding this functionality to the `Oli.Accounts` context introduces a dependency cycle between `Oli.Accounts` and `Oli.Publishing` and`Oli.Delivery.Sections`.  I think these new functions should simply go into `Oli.Publishing'

lib/oli/delivery/sections.ex
```diff
@@ -771,31 +771,94 @@ defmodule Oli.Delivery.Sections do
...
+
+  ## Examples
+
+      iex> retrieve_visible_publications(%User{}, %Institution{})
+      [%Publication{project: %Project{}}, ...]
+
+      iex> retrieve_visible_publications(%User{}, %Institution{})
+      []
+  """
+  def retrieve_visible_publications(user, institution) do
```
darrensiegel, line 792: I'd make the same comment here (about dependency cycles), but it looks like there was already a cycle in the code between `Oli.Delivery.Sections` and `Oli.Publishing`.  The intent is that Publishing depends on Sections context.

darrensiegel (changes requested):
Again, fantastic work. Just one comment regarding placement of new functionality within the `Oli.Accounts` context.

mcass19:
Thanks for the comments @darrensiegel! 

I agree with you, moved the functions to the Publishing context in the last commit. 

Let me know your thoughts, thanks!

darrensiegel (approved):
Thanks!

PR #2000, opened 2021-12-16
[BUG] PMP-2143 | JAN-946

dtiwarATS:


assets/src/apps/delivery/components/ActivityRenderer.tsx
```diff
@@ -404,13 +402,7 @@ const ActivityRenderer: React.FC<ActivityRendererProps> = ({
           target = ownerActivity ? `${ownerActivity.id}|${target}` : `${target}`;
         }
       }
-      const originalValue = snapshot[target];
```
dtiwarATS, line 407: We don't need to do this check as we are fetching values from snapshot

dsuriano: Curious, do you have any idea why it would behave differently between a `localhost` open+free vs a `Tokamak` open+free?

dtiwarATS: I am not exactly sure about that but may be it due to the some timing issue in the sequence of things when they are executed.  Still I think it should have any impact in open and free.

The only thing is that we have to be very patience when testing the Open & free for Small world like sims as they are very heavy and some times after clicking, we have to wait till it completes the check.

dsuriano (approved):

assets/src/apps/delivery/components/ActivityRenderer.tsx
```diff
@@ -404,13 +402,7 @@ const ActivityRenderer: React.FC<ActivityRendererProps> = ({
...
-      const typeOfOriginalValue = typeof originalValue;
-      const evaluatedValue =
-        typeOfOriginalValue === 'string' && initStateFacts[key] !== CapiVariableTypes.MATH_EXPR
```
bsparks, line 410: pretty sure that the only thing that was added was about the math expr, we needed to parse these to get the values to show up in mission control

bsparks (changes requested):
not sure we should remove all processing

assets/src/apps/delivery/components/ActivityRenderer.tsx
dtiwarATS: Hey @bsparks,

Actually we don't need that because we are already doing that in deck.ts when we were loading the fact. We were doing the same thing here so it shouldn't matter..

Also, I have tested the mission control thing, it displays the selected planet text in mission control logs

bsparks (approved):

PR #2004, opened 2021-12-16
[MER-339] - Added Cognito SSO support

chrislawson:


lib/oli_web/controllers/cognito_controller.ex
```diff
@@ -0,0 +1,151 @@
...
+  require Logger
+
+  alias Oli.Repo
+  alias Oli.Accounts
+  alias Oli.Groups.CommunityAccount
+  alias Oli.Institutions.SsoJwks
+
+  def launch(conn, params) do
+    try do
+      with {:ok, valid_params} <- validate_query_params(params) do
```
gastonabella, line 16: You could pattern match all the fields used below:
```elixir
{:ok, %{id_token: id_token, community_id: community_id, course_section_id: course_section_id} = valid_params} <- validate_query_params(params)
```

and then replace the places where they are used accordingly.

Also, I think a `case` makes more sense than a `with` clause. Wdyt?

lib/oli_web/controllers/cognito_controller.ex
```diff
@@ -0,0 +1,151 @@
...
+        jwk = get_jwk(Map.get(params, "id_token"))
+        jwk_from_pem = JOSE.JWK.from_pem(jwk.pem)
+        case JOSE.JWT.verify_strict(jwk_from_pem, ["RS256"], valid_params.id_token) do
+          {true, verified_jwt, _} ->
```
gastonabella, line 20: Same here. We could pattern match `fields` and avoid calling it multiple times

```elixir
{true, %{fields: jwt_fields}, _} ->
```

lib/oli_web/controllers/cognito_controller.ex
```diff
@@ -0,0 +1,151 @@
...
+            case Accounts.insert_or_update_lms_user(%{
```
gastonabella, line 21: We're handling the successful case but we're not covering the case when this function returns `{:error, changeset}`

lib/oli_web/controllers/cognito_controller.ex
```diff
@@ -0,0 +1,151 @@
...
+                  preferred_username: Map.get(verified_jwt.fields, "cognito:username"),
+                  email: Map.get(verified_jwt.fields, "email"),
+                  can_create_sections: true
+                }) do
+              {:ok, user} ->
+                  attrs = %{user_id: user.id, community_id: valid_params.community_id}
+
+                  %CommunityAccount{}
+                  |> CommunityAccount.changeset(attrs)
+                  |> Repo.insert()
```
gastonabella, line 32: We could use `Oli.Groups.create_community_account(attrs)` method that does basically the same

lib/oli_web/controllers/cognito_controller.ex
```diff
@@ -0,0 +1,151 @@
...
+            end
+
+          {false, _, _} ->
+            error_url = get_error_url(conn, valid_params)
+            conn
+            |> redirect(external: "#{error_url}?error=Unable to verify credentials")
+            |> halt()
+        end
+      else
+        {:error, changeset} ->
```
gastonabella, line 47: `{:error, %Ecto.Changeset{errors: errors, changes: changes}` would also work :)

lib/oli_web/controllers/cognito_controller.ex
```diff
@@ -0,0 +1,151 @@
...
+  def get_jwk(id_token) do
+    jwt_header = id_token
+    |> Joken.peek_header()
+
+    jwt_payload = id_token
+    |> JOSE.JWT.peek_payload()
+
+    issuer = Map.get(jwt_payload.fields, "iss")
+
+    case Joken.peek_header(id_token) do
```
gastonabella, line 77: We're not handling the `:error` case here either

lib/oli_web/controllers/cognito_controller.ex
```diff
@@ -0,0 +1,151 @@
...
+      {:ok, result} ->
```
gastonabella, line 78: `{:ok, %{kid: kid}}` and avoid the calls to `Map.get` below

lib/oli_web/controllers/cognito_controller.ex
```diff
@@ -0,0 +1,151 @@
...
+        case Repo.get_by(SsoJwks, kid: Map.get(result, "kid")) do
+          nil ->
+            jwks = get_cognito_jwks(issuer)
+
+            sso_jwks = Enum.map(elem(jwks.keys, 1), fn(s) -> get_sso_jwks(s) end)
```
gastonabella, line 83: I think this could work:

```elixir
sso_jwks =
  issuer
  |> get_cognito_jwks()
  |> Map.get(:keys)
  |> elem(1)
  |> Enum.map(& get_sso_jwks(&1))
```

or as we suggested via email

```elixir
%JOSE.JWK{keys: {_, keys_list}} = get_cognito_jwks(issuer)
sso_jwks = Enum.map(keys_list, & get_sso_jwks(&1))
```

lib/oli_web/controllers/cognito_controller.ex
```diff
@@ -0,0 +1,151 @@
...
+
+            Enum.each(sso_jwks, fn(attrs) ->
+              %SsoJwks{}
+              |> SsoJwks.changeset(attrs)
+              |> Repo.insert()
```
gastonabella, line 88: What do you think about defining this block in a separate function in `Oli.Institutions.SsoJwks` ?

lib/oli_web/controllers/cognito_controller.ex
```diff
@@ -0,0 +1,151 @@
...
+            end)
+
+            case Repo.get_by(SsoJwks, kid: Map.get(result, "kid")) do
```
gastonabella, line 91: We are not handling the error case here

lib/oli_web/controllers/cognito_controller.ex
```diff
@@ -0,0 +1,151 @@
...
+
+  def get_sso_jwks(item) do
+    pem = JOSE.JWK.to_pem(item)
+
+    sso_jwks = Map.put(%{typ: "JWT"}, :alg, Map.get(elem(item, 3), "alg"))
+    |> Map.put(:kid, Map.get(elem(item, 3), "kid"))
+    |> Map.put(:pem, elem(pem, 1))
+
+    sso_jwks
+  end
```
gastonabella, line 109: You could leverage pattern matching here and do something like the following:
```elixir
  def get_sso_jwks({_, _, _, %{"alg" => alg_value, "kid" => kid_value}} = item) do
    {_, pem} = JOSE.JWK.to_pem(item)
    %{
      pem: pem,
      alg: alg_value,
      kid: kid_value
    }
  end
```

lib/oli_web/controllers/cognito_controller.ex
```diff
@@ -0,0 +1,151 @@
...
+
+  def get_cognito_jwks(issuer) do
+    url = "#{issuer}/.well-known/jwks.json"
+    case http().get(url) do
```
gastonabella, line 113: Error case not handled

gastonabella (commented):
Great job and welcome to the Elixir world :):)

I left some comments and most of them are pretty similar (leverage pattern matching, handle error cases). Let me know what you think 🙂

PR #2025, opened 2021-12-22
PMP-2318 | JAN-1118

dtiwarATS:


assets/src/components/activities/DeliveryElement.tsx
```diff
@@ -63,6 +63,11 @@ export interface DeliveryElementProps<T extends ActivityModelSchema> {
     partAttemptGuid: string,
     response: StudentResponse,
   ) => Promise<Success>;
+  onInitPart: (
```
bsparks, line 66: we need to make this optional, or else the other Torus activities will have errors.

assets/src/components/parts/janus-text-flow/Markup.tsx
```diff
@@ -64,7 +65,11 @@ const Markup: React.FC<any> = ({
   let processedText = text;
   // allow (authoring usually) skipping the template processing
   if (!displayRawText) {
-    processedText = templatizeText(text, state, env);
+    let updatedText = text;
+    expressions.forEach((exp: Record<string, string>) => {
+      updatedText = text.replaceAll(exp.expression, exp.newExpression);
+    });
+    processedText = templatizeText(updatedText, state, env);
```
bsparks, line 72: so I guess that I was thinking that `state` would just contain only the required values. i'm honestly having a hard time visualizing what is coming through for the expressions here that we are replacing in text. state usually comes from the snapshot being sent back from init, which has ALL of the state, are we still sending that? I was thinking that we'd just be able to send a list of required values and filter the snapshot being sent back. maybe you can describe what is happening here

dtiwarATS: expressions is an array object that contains the original expression used in the text i.e. exp.expression    and exp.newExpression which is the expression with actual ownerId of the part component that is used in the variable.
 we are trying to replace the expression with the actual ownerId (i.e. the question Id) of the part component that is used in the variable.

Also, I think I can filter the state and send only the required variables.

assets/src/apps/delivery/layouts/deck/DeckLayoutView.tsx
```diff
@@ -42,13 +43,40 @@ const InjectedStyles: React.FC<{ css?: string }> = (props) => {
...
+      updatedExpression = activityLayerId + '|' + part;
+    } else if (activityLayerId) {
+      updatedExpression = activityLayerId + '|' + part;
+      updatedExpression = getActualOwnerDetails(updatedExpression, sequence);
+    }
+  }
+  return updatedExpression;
+};
+
+export const getPartOwnerVariableId = (expression: string, sequence: any) => {
```
bsparks, line 64: it would be better to export things from library type files rather than react components

assets/src/apps/delivery/layouts/deck/DeckLayoutView.tsx
```diff
@@ -42,13 +43,40 @@ const InjectedStyles: React.FC<{ css?: string }> = (props) => {
...
+  if (expression.indexOf('stage.') === -1) {
+    return { expression, newExpression: expression };
+  }
+  const updatedExpression = getActualOwnerDetails(expression, sequence);
+  return { expression, newExpression: updatedExpression };
```
bsparks, line 69: the name of this method confuses me a bit, `getPartOwnerVariableId` returns `{ expression: string, newExpression: string }` that doesn't seem like an id or variable id to me.

assets/src/apps/delivery/layouts/deck/DeckLayoutView.tsx
```diff
@@ -42,13 +43,40 @@ const InjectedStyles: React.FC<{ css?: string }> = (props) => {
...
+    } else if (activityLayerId) {
+      updatedExpression = activityLayerId + '|' + part;
+      updatedExpression = getActualOwnerDetails(updatedExpression, sequence);
+    }
+  }
+  return updatedExpression;
+};
+
+export const getPartOwnerVariableId = (expression: string, sequence: any) => {
+  if (expression.indexOf('stage.') === -1) {
```
bsparks, line 65: shouldn't we be checking for a `|` instead? or maybe `|stage.` ?

assets/src/apps/delivery/layouts/deck/DeckLayoutView.tsx
```diff
@@ -42,13 +43,40 @@ const InjectedStyles: React.FC<{ css?: string }> = (props) => {
   return injected ? <style>{injected}</style> : null;
 };
 
+const getActualOwnerDetails = (expression: string, sequence: any) => {
+  const [activityId, part] = expression.split('|');
```
bsparks, line 47: seems like it's possible there is no `|` for whatever reason, should check for that probably

assets/src/apps/delivery/layouts/deck/DeckLayoutView.tsx
```diff
@@ -42,13 +43,40 @@ const InjectedStyles: React.FC<{ css?: string }> = (props) => {
   return injected ? <style>{injected}</style> : null;
 };
 
+const getActualOwnerDetails = (expression: string, sequence: any) => {
```
bsparks, line 46: I think it would help (me at least) to name it `identifier: string` OR `identifierExpression: string` because expression by itself is a bit to broad

assets/src/apps/delivery/layouts/deck/DeckLayoutView.tsx
```diff
@@ -42,13 +43,40 @@ const InjectedStyles: React.FC<{ css?: string }> = (props) => {
...
+  const [activityId, part] = expression.split('|');
+  const activitySequence = sequence.find((entry: any) => entry?.custom?.sequenceId === activityId);
+  const activityLayerId = activitySequence?.custom?.layerRef;
+  let expressionValue = getValue(expression, defaultGlobalEnv);
```
bsparks, line 50: should pass in `env = defaultGlobalEnv` as a parameter instead of requiring it here, or possibly without a default, makes it more flexible & testable

assets/src/apps/delivery/layouts/deck/DeckLayoutView.tsx
```diff
@@ -42,13 +43,40 @@ const InjectedStyles: React.FC<{ css?: string }> = (props) => {
   return injected ? <style>{injected}</style> : null;
 };
 
+const getActualOwnerDetails = (expression: string, sequence: any) => {
+  const [activityId, part] = expression.split('|');
+  const activitySequence = sequence.find((entry: any) => entry?.custom?.sequenceId === activityId);
+  const activityLayerId = activitySequence?.custom?.layerRef;
```
bsparks, line 49: this could end up as undefined if it doesn't find a match in the sequence, and later when it's used you aren't checking that

bsparks: oh hmm, perhaps you are, but I think if it's undefined then you probably don't need to bother checking the value or anything at that point and either throw an error or just return the original

assets/src/apps/delivery/layouts/deck/DeckLayoutView.tsx
```diff
@@ -42,13 +43,40 @@ const InjectedStyles: React.FC<{ css?: string }> = (props) => {
...
+  let expressionValue = getValue(expression, defaultGlobalEnv);
+  let updatedExpression = expression;
+  if (expressionValue == undefined) {
+    expressionValue = getValue(activityLayerId + '|' + part, defaultGlobalEnv);
```
bsparks, line 53: DRY : ```const parentIdentifier = `${activityLayerId}|${part}`;```

assets/src/apps/delivery/layouts/deck/DeckLayoutView.tsx
bsparks, line 46: now that I've read through the method, should probably call it `getActualOwnerIdentifier`

assets/src/apps/delivery/layouts/deck/DeckLayoutView.tsx
```diff
@@ -42,13 +43,40 @@ const InjectedStyles: React.FC<{ css?: string }> = (props) => {
   return injected ? <style>{injected}</style> : null;
 };
 
+const getActualOwnerDetails = (expression: string, sequence: any) => {
+  const [activityId, part] = expression.split('|');
+  const activitySequence = sequence.find((entry: any) => entry?.custom?.sequenceId === activityId);
+  const activityLayerId = activitySequence?.custom?.layerRef;
+  let expressionValue = getValue(expression, defaultGlobalEnv);
+  let updatedExpression = expression;
+  if (expressionValue == undefined) {
```
bsparks, line 52: should be triple =

assets/src/components/activities/DeliveryElement.tsx
```diff
@@ -151,6 +160,8 @@ export abstract class DeliveryElement<T extends ActivityModelSchema> extends HTM
 
     this.onReady = (attemptGuid: string) => this.dispatch('activityReady', attemptGuid, undefined);
     this.onResize = (attemptGuid: string) => this.dispatch('resizePart', attemptGuid, undefined);
+    this.onInitPart = (attemptGuid: string, partAttemptGuid: string, expression: string[]) =>
+      this.dispatch('partInit', attemptGuid, partAttemptGuid, expression);
```
bsparks, line 164: why are we flip flopping the name? initPart / partInit

assets/src/components/activities/adaptive/AdaptiveDelivery.tsx
```diff
@@ -192,16 +192,29 @@ const Adaptive = (props: DeliveryElementProps<AdaptiveModelSchema>) => {
...
+    responses: any[];
+    expressions?: string[];
+  }) => {
     // a part should send initial state values
     if (payload.responses.length) {
       const saveResults = await handlePartSave(payload);
     }
-
+    let expressionResult: any = [];
+    if (payload.expressions && payload.expressions.length && props.onInitPart) {
```
bsparks, line 205: seems like if we're going to have a part init event then we might as well fire it all the time, and the result of it should be a payload object (that could possibly contain "expressions" in the result). this would make it more generic for all activities to know what to expect.

assets/src/components/parts/janus-text-flow/TextFlow.tsx
```diff
@@ -111,14 +117,37 @@ const TextFlow: React.FC<PartComponentProps<TextFlowModel>> = (props: any) => {
...
+    if (pNodes && typeof pNodes === 'string') {
+      tree = JSON.parse(pNodes as string) as MarkupTree[];
+    } else if (Array.isArray(pNodes)) {
+      tree = pNodes;
+    }
+    const allExpression = tree?.map((subtree: MarkupTree) => {
+      return getTextExpressions(subtree || []);
+    });
+
+    const finalExppression = allExpression.filter((expression) => expression);
```
bsparks, line 144: typo, extra p

bsparks:
ok, so I think the part that I'm still struggling with is that we do still send the entire env as a reference, so at least that part of the memory still exists. what I *think* you're doing having looked through this is going through and replacing everything on the value side with the actual owner; so if there is some adaptivity that says use `{stage.foo.bar}` and `foo` lives on a parent layer, then instead it would replace that with `{parent|stage.foo.bar}` as if it had been originally authored that way. is that right? and then what we *were* doing before with the sync is making a screen specific copy of for each child, so `{grandparent|stage.foo.bar}` `{parent|stage.foo.bar}` `{current|stage.foo.bar}` etc and in that case keeping all those values in sync with grandparent. is that also right? if that is true, then I wonder if it wouldn't just be simpler to automatically create bindings to the grandparent so that `{parent|stage.foo.bar}` and `{current|stage.foo.bar}` were both "bind to" `{grandparent|stage.foo.bar}` . and then there would be the issue I think of updating a child and expecting grandparent to update. one thing I can think of for that is possibly testing for an error, it should throw an error if you attempt to assign a value to a bound id, at that point we could then find the bind target and update that instead. perhaps though it would be unnecessary if we can just handle that when we save, but wait, maybe that isn't an issue since we're already doing that in savePartToTree I think...

bsparks:
closing in favor of alternate approach

PR #2027, opened 2021-12-22
[MER-624] Make help provider a configurable env variable

gastonabella:
[MER-624](https://eliterate.atlassian.net/browse/MER-624)

Make Help Provider a configurable and dynamic env variable.

### Release Notes (important)
Ensure `HELP_PROVIDER` env variable is added to each Torus instance with the corresponding value:
- `HELP_PROVIDER=FreshdeskHelp` or `HELP_PROVIDER=EmailHelp` depending on the case.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -115,7 +115,14 @@ config :oli, :recaptcha,
   secret: System.get_env("RECAPTCHA_PRIVATE_KEY")
 
 # Configure help
-config :oli, :help, dispatcher: Oli.Help.Providers.EmailHelp
+# HELP_PROVIDER env var must be a string representing an existing provider module, such as "EmailHelp"
+help_provider =
+  case System.get_env("HELP_PROVIDER") do
+    nil -> Oli.Help.Providers.EmailHelp
+    provider -> Module.concat([Oli, Help, Providers, provider])
+  end
+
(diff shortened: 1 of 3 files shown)


config/config.exs
```diff
@@ -115,7 +115,14 @@ config :oli, :recaptcha,
   secret: System.get_env("RECAPTCHA_PRIVATE_KEY")
 
 # Configure help
-config :oli, :help, dispatcher: Oli.Help.Providers.EmailHelp
+# HELP_PROVIDER env var must be a string representing an existing provider module, such as "EmailHelp"
+help_provider =
+  case System.get_env("HELP_PROVIDER") do
+    nil -> Oli.Help.Providers.EmailHelp
+    provider -> Module.concat([Oli, Help, Providers, provider])
```
darrensiegel, line 122: I did not know about `Module.concat`, very interesting.

darrensiegel (approved):

config/config.exs
gastonabella: Actually, it is safer to use [Module.safe_concat/1](https://hexdocs.pm/elixir/1.12/Module.html#safe_concat/1) but in this case we can't since this file is loaded before any dependency and `safe_concat` raises an error if the module wasn't referenced yet.

PR #2031, opened 2021-12-22
[Bugfix] everapp state

bsparks:


dsuriano (approved):
tested locally and works as expected (for init state)

PR #2033, opened 2021-12-23
[MER-555] Make privacy policies url an env var

gastonabella:
[MER-555](https://eliterate.atlassian.net/browse/MER-555)

Make "Privacy Notice" link a configurable and dynamic env variable.

### Release Notes (important)
Ensure `PRIVACY_POLICIES_URL` env variable is added to each Torus instance with the corresponding value:
- `PRIVACY_POLICIES_URL={INSERT A COMPLETE URL HERE}`.

It is important that the url must include the protocol (`https://www.example.com` instead of `www.example.com`) in order to work.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -157,6 +157,10 @@ config :surface, :components, [
   }
 ]
 
+# Configure Privacy Policies link
+config :oli, :privacy_policies,
+  url: System.get_env("PRIVACY_POLICIES_URL", "https://www.cmu.edu/legal/privacy-notice.html")
+
 # Import environment specific config. This must remain at the bottom
 # of this file so it overrides the configuration defined above.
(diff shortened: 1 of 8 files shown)


config/config.exs
```diff
@@ -157,6 +157,10 @@ config :surface, :components, [
   }
 ]
 
+# Configure Privacy Policies link
+config :oli, :privacy_policies,
```
darrensiegel, line 161: Does a similar change to this need to be made in `config/releases.exs` ?

darrensiegel (changes requested):

config/config.exs
gastonabella: Done 👍

darrensiegel (approved):

PR #2038, opened 2021-12-28
PMP-2393 | JAN-1212

dtiwarATS:


dtiwarATS:
I will created another PR with more changes

dsuriano (approved):
Tested locally and confirmed fluid in `xM CuCl2` beaker is now visible upon lesson load.

assets/src/apps/delivery/components/TextParser.ts
```diff
@@ -8,7 +13,7 @@ export const templatizeText = (
   env?: Environment,
   isFromTrapStates = false,
 ): string => {
-  let innerEnv = env;
+  let innerEnv = env || defaultGlobalEnv;
```
bsparks, line 16: I don't think we can import the defaultGlobalEnv into here for use in externalActivity, I think this is why I'm sending env reference in init or context change (or both). when you import them, they become part of a bundle, and the components are their *own* web component bundle, it won't be the same reference.

bsparks: if you need to have a default here, which I don't think you do, then you might as well just say `|| new Environment()`

dtiwarATS: Oops, I didn't knew that env reference was passed in part components. Changes made and PR Updated. followed the same patterns as of text flow part component.

assets/src/components/parts/janus-capi-iframe/ExternalActivity.tsx
```diff
@@ -520,7 +529,23 @@ const ExternalActivity: React.FC<PartComponentProps<CapiIframeModel>> = (props)
...
+      writeCapiLog('SENDING SIM CONFIG DATA !!!!', 3, {
+        simLife,
+        configData: filterVars,
+      });
+      Object.keys(filterVars).forEach((variable: any) => {
+        const formatted: Record<string, unknown> = {};
+        const simVariable = filterVars[variable];
+        const typeOfValue = typeof simVariable.value;
+        //Now we know that configData variable can contain a expression in value field so we need to parse that before sending it to SIM
+        if (typeOfValue === 'string' && simVariable.value.indexOf('{') !== -1) {
```
bsparks, line 541: I don't think that you *need* to check for the curly brace here, the templatizeText will handle that check

assets/src/components/parts/janus-capi-iframe/ExternalActivity.tsx
```diff
@@ -138,6 +141,10 @@ const ExternalActivity: React.FC<PartComponentProps<CapiIframeModel>> = (props)
     if (initResult.context.mode) {
       context = initResult.context.mode;
     }
+    if (initResult.env) {
+      const Env = new Environment(initResult.env);
```
bsparks, line 145: shouldn't have a capital E

assets/src/components/parts/janus-capi-iframe/ExternalActivity.tsx
```diff
@@ -520,7 +529,23 @@ const ExternalActivity: React.FC<PartComponentProps<CapiIframeModel>> = (props)
...
+        const simVariable = filterVars[variable];
+        const typeOfValue = typeof simVariable.value;
+        //Now we know that configData variable can contain a expression in value field so we need to parse that before sending it to SIM
+        if (typeOfValue === 'string' && simVariable.value.indexOf('{') !== -1) {
+          const gotid = templatizeText(simVariable.value, simLife.snapshot, scriptEnv);
+          simVariable.value = gotid;
+        }
+
+        formatted[variable] = simVariable;
+        sendFormedResponse(simLife.handshake, {}, JanusCAPIRequestTypes.VALUE_CHANGE, formatted);
```
bsparks, line 547: so now we not only need to process the possible expressions in string values but also send them one at a time?

assets/src/components/parts/janus-capi-iframe/ExternalActivity.tsx
```diff
@@ -520,7 +529,23 @@ const ExternalActivity: React.FC<PartComponentProps<CapiIframeModel>> = (props)
...
     if (filterVars && Object.keys(filterVars)?.length !== 0) {
       handleIFrameSpecificProperties(simLife.currentState, simLife.domain);
-      sendFormedResponse(simLife.handshake, {}, JanusCAPIRequestTypes.VALUE_CHANGE, filterVars);
+      writeCapiLog('SENDING SIM CONFIG DATA !!!!', 3, {
+        simLife,
+        configData: filterVars,
+      });
+      Object.keys(filterVars).forEach((variable: any) => {
+        const formatted: Record<string, unknown> = {};
+        const simVariable = filterVars[variable];
```
bsparks, line 538: I found this a bit confusing since you are mutating items via reference.  I think that you might as well just use it directly; so like
```
formatted[variable] = filterVars[variable];
if (typeof formatted[variable].value === 'string') {
   formatted[variable].value = templatizeText(formatted[variable].value, simLife.snapshot, scriptEnv);
}
sendFormedResponse(simLife.handshake, {}, JanusCAPIRequestTypes.VALUE_CHANGE, formatted);
```

assets/src/components/parts/janus-capi-iframe/ExternalActivity.tsx
dtiwarATS: I was only checking to avoid processing each text if it does not have expression. I will remove that.

assets/src/components/parts/janus-capi-iframe/ExternalActivity.tsx
dtiwarATS: Yes. During KIP investigation, we found that we need to send it one variable at a time and we were already sending the 'init state' variables like this. This one is configData. We never changed it to send one at  a time so refactored that code.

PR #2039, opened 2021-12-29
[FEATURE] [MER-656] Add custom footer text and links

mcass19:
https://eliterate.atlassian.net/browse/MER-656

Make footer configurable and dynamic by env variables.

### Release Notes (important)
Ensure the following environment variables are added to each Torus instance with the corresponding values:
- `FOOTER_TEXT`
- `FOOTER_LINK_1_LOCATION`
- `FOOTER_LINK_1_TEXT`
- `FOOTER_LINK_2_LOCATION`
- `FOOTER_LINK_2_TEXT`

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -168,6 +168,14 @@ config :surface, :components, [
 config :oli, :privacy_policies,
   url: System.get_env("PRIVACY_POLICIES_URL", "https://www.cmu.edu/legal/privacy-notice.html")
 
+# Configure footer text and links
+config :oli, :footer,
+  text: System.get_env("FOOTER_TEXT", "By accessing this site, you agree to the "),
+  link_1_location: System.get_env("FOOTER_LINK_1_LOCATION", "https://www.argos.education/terms"),
+  link_1_text: System.get_env("FOOTER_LINK_1_TEXT", "terms of use. "),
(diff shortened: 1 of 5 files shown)


config/config.exs
```diff
@@ -168,6 +168,14 @@ config :surface, :components, [
 config :oli, :privacy_policies,
   url: System.get_env("PRIVACY_POLICIES_URL", "https://www.cmu.edu/legal/privacy-notice.html")
 
+# Configure footer text and links
+config :oli, :footer,
+  text: System.get_env("FOOTER_TEXT", "By accessing this site, you agree to the "),
+  link_1_location: System.get_env("FOOTER_LINK_1_LOCATION", "https://www.argos.education/terms"),
```
darrensiegel, line 174: The default values for link locations should not be Argos related, but rather should be CMU OLI links.

darrensiegel (changes requested):

config/config.exs
mcass19: Okay makes sense @darrensiegel ! What values do you propose? Or should we leave them empty?

ehilfer: We should make them default to empty, as Proton currently does not have them populated

darrensiegel (approved):

PR #2043, opened 2022-01-03
MER-613 If a users email exists, pass to stripe when creating intent.

jrissler:
If Torus has an email for the user, it should be passed along to stripe since they will not get a receipt otherwise. As a user with an unverified email this feels bad since I am paying with money.

_Note: Rather than checking for verified emails here - it would be better to not let them end up here without a verified email if we feel that is important. Could be worth discussing and opening alternate ticket if needed._


ehilfer (approved):
This is the best pragmatic approach for launch, and I agree we should enforce email verification prior to showing the paywall in the future.

darrensiegel (approved):

PR #2047, opened 2022-01-03
[MER-548] Help modal fix

ehilfer:


lib/oli_web/templates/shared/_help.html.eex
```diff
@@ -113,7 +113,7 @@
                 successElement.innerHTML = json.info;
                 successElement.classList.remove('d-none');
                 successElement.classList.add('d-block');
-                helpForm.classList.add('d-none');
+                //helpForm.classList.add('d-none');
```
bsparks, line 116: what's the reason to keep a line and comment it out rather than delete it?

ehilfer:
Too many commits got into this PR - closing

PR #2049, opened 2022-01-04
Screen scrolls to far.

schowkekarats360:
I am not sure if the whole style tag is needed, but pretty much sure that the width is not. beagleContainer getting widhth applied from the css files. this one only adding width to all oli-adaptive-delivery wrappers and its taking up the space on page, causing scrolling.

### DIFF
--- a/assets/src/apps/delivery/layouts/deck/components/EverappContainer.tsx
+++ b/assets/src/apps/delivery/layouts/deck/components/EverappContainer.tsx
@@ -18,7 +18,7 @@ const EverappContainer: React.FC<EverappContainerProps> = ({ apps }) => {
 
   return (
     <div className={`beagleContainer ${secondaryClass}`}>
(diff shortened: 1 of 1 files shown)


assets/src/apps/delivery/layouts/deck/components/EverappContainer.tsx
```diff
@@ -18,7 +18,7 @@ const EverappContainer: React.FC<EverappContainerProps> = ({ apps }) => {
 
   return (
     <div className={`beagleContainer ${secondaryClass}`}>
-      <style>{`oli-adaptive-delivery { width: 256px; height: 100%; display: block; }`}</style>
+      <style>{`oli-adaptive-delivery { height: 100%; display: block; }`}</style>
```
bsparks, line 21: maybe we can just be specific instead, like 
```
<style>{`.beagleContainer oli-adaptive-delivery { height: 100%; display: block; }`}</style>
```

bsparks (approved):

PR #2056, opened 2022-01-06
PMP-2420 | JAN-1241

dtiwarATS:


assets/src/components/parts/janus-capi-iframe/ExternalActivity.tsx
```diff
@@ -529,6 +529,7 @@ const ExternalActivity: React.FC<PartComponentProps<CapiIframeModel>> = (props)
     const filterVars = createCapiObjectFromStateVars(simLife.currentState, simLife.domain);
     if (filterVars && Object.keys(filterVars)?.length !== 0) {
       handleIFrameSpecificProperties(simLife.currentState, simLife.domain);
+      sendFormedResponse(simLife.handshake, {}, JanusCAPIRequestTypes.VALUE_CHANGE, filterVars);
```
bsparks, line 532: wait so now we need to send them all at once unaltered, then resend them each one at a time possibly parsing the string values?

dtiwarATS: Oops. it was a mistake. I think I was investigating some other issue and checked-in this by mistake. reverted the changes

assets/src/apps/delivery/components/TextParser.ts
```diff
@@ -9,21 +9,13 @@ export const templatizeText = (
   isFromTrapStates = false,
 ): string => {
   let innerEnv = env;
-  const vars = extractAllExpressionsFromText(text);
+  let vars = extractAllExpressionsFromText(text);
+  // A expression will not have a ';' inside it. So if there is a ';' inside it, it is CSS and we should filter it.
+  vars = vars.filter((e) => !e.includes(';'));
   /* console.log('templatizeText call: ', { text, vars, state, env }); */
   if (!vars) {
```
bsparks, line 16: this line leads me to believe that it's possible for vars to be falsey from above, where we are definitely expecting it to be an array or else the filter call will be an exception

bsparks:
I think this function is a good candidate for unit tests where we can throw all different kinds of data at it and make sure that it has the desired results

bsparks (approved):

PR #2058, opened 2022-01-06
[BUG FIX] Use bulk insertion queries for attempt hierarchy creation MER-684

darrensiegel:
Performance testing (and perhaps real world usage) shows dramatically long page loads.  Investigation yields the culprit to be the logic that creates the activity and part attempt hierarchy (done one time, upon first page visit).  These inserts were being done in serial, so a large adaptive page with say 100 activities (screens) with 3 parts each would require 300 insertion queries, executed serially and incurring the DB latency requests in aggregate.  

The solution in this PR is to rework this logic to use a constant number of bulk insertion queries.

This should fix MER-683 as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+## 0.18.4 (2022-01-06)
+
+### Bug Fixes
+
+- Improve performance of initial page visits by introducing bulk insertions of attempts
+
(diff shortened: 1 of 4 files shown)


eliknebel (approved):

eliknebel:
Confirmed `lib/oli/delivery/attempts/page_lifecycle.ex:42` wraps all the logic in a transaction

PR #2069, opened 2022-01-07
[BUG FIX] [MER-672] Add default layout class to fix footer overlapping

mcass19:


darrensiegel (approved):

eliknebel:
This PR is breaking the overall layout of authoring, therefore it is set to be reverted in https://github.com/Simon-Initiative/oli-torus/pull/2114

PR #2084, opened 2022-01-12
PMP-2143 | Small World Related Issues | New Approach

dtiwarATS:


assets/src/components/parts/janus-Carousel/Carousel.tsx
```diff
@@ -115,7 +116,8 @@ const Carousel: React.FC<PartComponentProps<CarouselModel>> = (props) => {
             break;
           case NotificationType.CONTEXT_CHANGED:
             {
-              const { initStateFacts: changes } = payload;
+              const { initStateFacts: initStateFacts } = payload;
+              const changes = getFormattedVariables(initStateFacts);
```
dtiwarATS, line 120: Now, we are sending the init Variables as 'q:14541d:5421:1|stage.foo.val:20', we need to format that to 'stage.foo.val:20'

assets/src/utils/common.ts
```diff
@@ -156,3 +156,16 @@ export const parseNumString = (item: string): string | number => {
 // Zips two arrays. E.g. zip([1,2,3], [4,5,6,7]) == [[1, 4], [2, 5], [3, 6]]
 export const zip = <T, U>(xs1: T[], xs2: U[]): [T, U][] =>
   xs1.reduce((acc, x, i) => (i > xs2.length - 1 ? acc : acc.concat([[x, xs2[i]]])), [] as [T, U][]);
+
+export const getFormattedVariables = (initStateFacts: Record<string, any>) => {
```
bsparks, line 160: this method should go in a different common type file, something related to adaptive lessons since *this* common file is generic for all things

assets/src/components/parts/janus-capi-iframe/ExternalActivity.tsx
```diff
@@ -426,9 +426,22 @@ const ExternalActivity: React.FC<PartComponentProps<CapiIframeModel>> = (props)
                 context: payload.mode,
                 questionId: payload.currentActivityId,
               };
+              const initStateFacts = payload.initStateFacts;
+              let processedInitStateFacts = {};
+              if (scriptEnv) {
+                processedInitStateFacts = Object.keys(initStateFacts).reduce(
+                  (acc: any, key: string) => {
+                    const target = key.split('|')[1];
+                    acc[target] = scriptEnv.Get(key).toJS();
```
bsparks, line 435: hmm I was wondering why we were changing it, and now I remember. one problem I have is that I'm not sure that I like how this forces each component to strip out the screen id. I do see *why* that might be necessary, however. it might be possible to do it without doing this though, I'm pretty sure that the reference env that is getting sent to each activity is a child of the defaultGlobalEnv meaning that whatever changes happen to it remain within that context, but it still has read access to the parent context. that being the case we should be able to instead just populate that environment with the localized copies of the variables. also I think the preferred way to get the value would be to use `getValue` from scripting

PR #2087, opened 2022-01-12
[MER-503] Invited authors can confirm email and resend confirmation link

gastonabella:
[MER-503](https://eliterate.atlassian.net/browse/MER-503)

### Description
There are different states where an author could be depending on if it was invited by an admin or if it created an account by himself. These states are translated into different combinations of the following author fields added by the Pow library: `email_confirmation_token`, `email_confirmed_at`, `unconfirmed_email`, `invitation_token`, `invitation_accepted_at`.

There are some cases where an author is not waiting for email confirmation, and the author details view still allows a sys admin to Resend a confirmation link or to Confirm an author email manually, causing an Internal Server Error when attempting to perform these actions.

**Note**:
The original description in the Jira ticket is not a bug. It is a decision of Pow devs not to send a confirmation link via email when an author was invited by an admin and he accepted the invitation using the same email as the one sent by the admin. A different case is when an author receives an admin invitation and he accepts it but using a different email. In that case he receives another email with a confirmation link.

### DIFF
--- a/lib/oli/accounts.ex
+++ b/lib/oli/accounts.ex
@@ -14,6 +14,7 @@ defmodule Oli.Accounts do
   alias Oli.Groups.CommunityAccount
   alias Oli.Repo
   alias Oli.Repo.{Paging, Sorting}
+  alias PowEmailConfirmation.Ecto.Context, as: EmailConfirmationContext
 
   def browse_users(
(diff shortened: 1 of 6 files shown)


gastonabella:
Opening this PR in the fork repo first

PR #2098, opened 2022-01-13
Fixed issue with rule condition type toggling between rules

rmathis:
This change fixes an issue where the radio buttons were not being updated when the rule type changed.


dsuriano (approved):
Tested locally and confirmed issue is fixed. Will hold off on merging until it's confirmed safe to do so.

PR #2104, opened 2022-01-17
Inorganic Carbon Cycle: 127870 Incorrect trap state

schowkekarats360:
while applying mutate state changes the multiselect mcq is not getting the previously selected values in selectedChoices property.

### DIFF
--- a/assets/src/components/parts/janus-mcq/MultipleChoiceQuestion.tsx
+++ b/assets/src/components/parts/janus-mcq/MultipleChoiceQuestion.tsx
@@ -713,68 +713,69 @@ const MultipleChoiceQuestion: React.FC<PartComponentProps<McqModel>> = (props) =
     }
   };
 
-  const handleItemSelection = (
-    { value, textValue, checked }: ItemSelectionInput,
-    snapshot: any,
-    shouldSave = true,
-  ) => {
-    const originalValue = parseInt(value.toString(), 10);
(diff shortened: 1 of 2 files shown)


assets/src/components/parts/janus-mcq/MultipleChoiceQuestion.tsx
```diff
@@ -711,6 +715,7 @@ const MultipleChoiceQuestion: React.FC<PartComponentProps<McqModel>> = (props) =
 
   const handleItemSelection = (
     { value, textValue, checked }: ItemSelectionInput,
+    snapshot: any,
```
bsparks, line 718: I think that instead all that might be necessary is to make this into a callback and make sure it updates when `selectedChoices` does, because it should already have the updated value I think

bsparks (changes requested):
pls try callback approach

assets/src/components/parts/janus-mcq/MultipleChoiceQuestion.tsx
schowkekarats360: I have just made this function as useCallback. but it was not retaining the selectedChoices value. and there is another variable `multipleSelection ` which was being undefined after useCallback changes so i have added `model ` in the dependencies. 
Let me know what other changes need to be made.

PR #2111, opened 2022-01-17
PMP-2430: fixes history navigation arrows missing, feedback displaced

dsuriano:
Problem was due to the way `legacyLessons` did not nest the feedback, history, and nav containers within `checkContainer`. Tested in a few lessons where `isLegacyLesson` is `true`. 

image


assets/src/apps/delivery/layouts/deck/DeckLayoutFooter.tsx
```diff
@@ -582,6 +583,7 @@ const DeckLayoutFooter: React.FC = () => {
             <HistoryNavigation />
           </>
         )}
+        {isLegacyTheme && <HistoryNavigation />}
```
bsparks, line 586: does this still need a condition? it seems like it's here in both cases

dsuriano: good catch... fixed.

bsparks (approved):

PR #2112, opened 2022-01-18
A Telltale Trace 127896: No trap state is firing

schowkekarats360:
This function is to overwrite the actual owner id of the variable in the rule's condition.


bsparks:
well I don't really want to start editing lessons for them. the things that we already auto "fix" are more for platform wide things or sort of like data version updates. even if we wanted to do this, then we'd have to also do it everywhere they try to create a condition with this in it, technically there is nothing stopping them and they could just put it right back. then the update only happens the when it gets opened in edit mode, so it could get tricky. need to think of another way

bsparks:
we'll fix this one in authoring and/or come up with a different approach

PR #2153, opened 2022-01-27
PMP-2199 | JAN-1010

dtiwarATS:


assets/src/components/parts/janus-dropdown/Dropdown.tsx
```diff
@@ -276,7 +276,7 @@ const Dropdown: React.FC<PartComponentProps<DropdownModel>> = (props) => {
               const sSelectedIndex = changes[`stage.${id}.selectedIndex`];
               if (sSelectedIndex !== undefined) {
                 const stateSelection = Number(sSelectedIndex);
-                if (selectedIndex !== stateSelection) {
+                if (selectedIndex !== stateSelection || stateSelection === -1) {
```
dtiwarATS, line 279: Need to add this condition else if we try to clear a already selected dropdown in child screens that exist on a layer, it does not work and value is not cleared. I have also tested this original ticket for which these changes were made i.e. https://jira.unicon.net/jira/browse/PMP-2074 and the PR was https://github.com/Simon-Initiative/oli-torus/pull/1908.

Both are working as expected.

bsparks (approved):

PR #2157, opened 2022-01-27
[BUGFIX] [PMP-2206 / JAN-1022]: Feedback Alignment

dsuriano:
Fixes misaligned feedback window in certain legacy lessons. Tested with multiple lessons, both legacy and non-legacy.

### DIFF
--- a/assets/src/apps/delivery/layouts/deck/components/FeedbackContainer.tsx
+++ b/assets/src/apps/delivery/layouts/deck/components/FeedbackContainer.tsx
@@ -22,7 +22,7 @@ const FeedbackContainer: React.FC<FeedbackContainerProps> = ({
   feedbacks,
   onMinimize,
   onMaximize,
-  style,
+  style = {},
 }) => {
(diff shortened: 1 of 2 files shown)


assets/src/apps/delivery/layouts/deck/components/FeedbackContainer.tsx
```diff
@@ -42,7 +44,7 @@ const FeedbackContainer: React.FC<FeedbackContainerProps> = ({
   return (
     <div
       className={`feedbackContainer rowRestriction ${isLegacyTheme ? 'columnRestriction' : ''}`}
-      style={{ top: '525px' }}
+      style={{ top: '525px', ...style }}
```
bsparks, line 47: looks like it's optional, maybe above needs `style = {}` just in case so it doesn't error on the spread here

dsuriano: 👍

bsparks (approved):

PR #2159, opened 2022-01-31
[FEATURE] Content Element Enhancements

LEDThereBeLight:
Part 1 of core editor changes. Missing some features like text and content element settings, plus styling fixes that will have to be a part of the legacy content elements PR.

Changes:
- Update editor toolbar
- Fix a dozen selection issues
- Update most content element editors
- Add some additional font settings

Other:
- Upgrade slate js

Known issues:
- Code editor needs 2 clicks to focus


eliknebel:
There seems to be an issue with adding a header to an empty content block where a bunch of white space is added. This issue continues to expand the content block when more content is added making the content element unruly. 

See video:
https://user-images.githubusercontent.com/6248894/152410750-56d53ccf-4237-424f-82a3-1e2ddb11a348.mov

LEDThereBeLight:
> 

good catch. resolved.

eliknebel:
Screen Shot 2022-02-07 at 2 52 17 PM

Adding an image to a multiple choice and clicking "Upload" results in an empty page

eliknebel:
These can be resolved in a followup PR, but there are some general UI inconsistencies that should be addressed, for example:

1. In general, border styles and overall feel is inconsistent with the rest of the UI. We should add some styling to the toolbar and its menus to make it feel more consistent, like its part of the overall system.
1. There is no action to button for links to "Apply" to hide to popup and give a user the indication that their url will be saved
Screen Shot 2022-02-07 at 2 50 05 PM
1. Modals should use bootstrap styling so that they are consistent with the rest of the system (see Add Webpage or Media Manager for example)
Screen Shot 2022-02-07 at 2 49 51 PM
1. Insert from URL shows a browser popup, which is inconsistent with the other modals
Screen Shot 2022-02-07 at 2 59 44 PM

eliknebel (approved):

PR #2160, opened 2022-01-31
[BUG] Small World Issues

dtiwarATS:


assets/src/apps/delivery/components/ActivityRenderer.tsx
```diff
@@ -371,17 +367,17 @@ const ActivityRenderer: React.FC<ActivityRendererProps> = ({
   // TODO: check if it needs to come from somewhere higher
   const currentActivityId = useSelector(selectCurrentActivityId);
   const initPhaseComplete = useSelector(selectInitPhaseComplete);
-  const initStateFacts = useSelector(selectInitStateFacts);
   const initStateBindToFacts: any = {};
   const currentActivityTree = useSelector(selectCurrentActivityTree);
   const updateGlobalState = async (snapshot: any, stateFacts: any) => {
     const payloadData = {} as any;
-    Object.keys(stateFacts).map((fact: string) => {
+    stateFacts.map((fact: any) => {
```
bsparks, line 374: maybe you mean reduce

bsparks: well, map returns an array, and you're using it like foreach, but reduce would cover that
```
const payloadData = stateFacts.reduce((data, facts) => {... return data; }, {});
```

dtiwarATS: I can change that to following but I don't know how to confirm that Ever app is working after this changes

```
const payloadData = stateFacts.reduce((data: any, fact: any) => {
      const target = fact.target;
      // EverApp Information
      if (target.startsWith('app.')) {
        const data = target.split('.');
        const objId = data.splice(2).join('.');
        const value = snapshot[target];
        data[data[1]] = { ...data[data[1]], [objId]: value };
      }
      return data;
    }, {});
```

dtiwarATS: I tested the ever app notepad and it was working as well as SW so I will go ahead and check-in these changes

bsparks (approved):

PR #2167, opened 2022-02-02
[FEATURE] Centralize API error handling

darrensiegel:
This PR creates a centralized place where API endpoint errors (typically anything that we trap and then return as a `500` error) can be handled and logged.  

The logging of error details is needed in order to troubleshoot and identify the root causes. 

The introduction of the centralized `log_error` method is to provide a future place for extensibility.  In the future we may want to, for example, post an AppSignal error in addition to just logging the error.

### DIFF
--- a/lib/oli/utils/error_logger.ex
+++ b/lib/oli/utils/error_logger.ex
@@ -1,0 +1,7 @@
+defmodule Oli.Utils.ErrorLogger do
+  require Logger
+
+  def log_error(e, label \\ "Unexpected error") do
+    Logger.error(label <> ": " <> Kernel.inspect(e))
+  end
+end
--- a/lib/oli_web/controllers/api/activity_controller.ex
+++ b/lib/oli_web/controllers/api/activity_controller.ex
@@ -161,7 +161,8 @@ defmodule OliWeb.Api.ActivityController do
       {:error, {:not_authorized}} ->
         error(conn, 403, "unauthorized")
 
-      _ ->
+      e ->
+        Oli.Utils.ErrorLogger.log_error(e, "Could not create secondary activity document")
         error(conn, 500, "server error")
     end
   end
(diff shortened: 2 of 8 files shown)


bsparks (approved):
looks good to me, though I wonder if we are able to have as detailed as info as we had the other day where we got ` errors: [part_id: {"can't be blank", [validation: :required]}],`

eliknebel (approved):

lib/oli_web/controllers/api/activity_controller.ex
```diff
@@ -161,7 +161,8 @@ defmodule OliWeb.Api.ActivityController do
       {:error, {:not_authorized}} ->
         error(conn, 403, "unauthorized")
 
-      _ ->
+      e ->
+        Oli.Utils.ErrorLogger.log_error(e, "Could not create secondary activity document")
```
eliknebel, line 165: This function returns a tuple with an error id and message. The message contains the error given as a parameter as well as a message that says "Contact support with issue #1234" so that the serverside error can easily be tracked down.

If we wanted to improve support here, we could capture the error message returned: `{_id, error_msg} = Oli.Utils.ErrorLogger.log_error(...` then return that in `error©nn, error_msg)`

lib/oli/utils/error_logger.ex
```diff
@@ -0,0 +1,7 @@
+defmodule Oli.Utils.ErrorLogger do
+  require Logger
+
+  def log_error(e, label \\ "Unexpected error") do
+    Logger.error(label <> ": " <> Kernel.inspect(e))
```
eliknebel, line 5: There is already a `log_error` function in Oli.Utils. Should we move that function into this new module?

eliknebel (changes requested):

lib/oli_web/controllers/api/activity_controller.ex
eliknebel: Sorry, I was confused as well. I thought this`Oli.Utils.ErrorLogger.log_error` function was the existing `Oli.Utils.log_error` function. Should we combine these two into a single function for consistency?

darrensiegel: Yes, good idea.  I just submitted a commit that does that.

eliknebel (approved):

PR #2170, opened 2022-02-03
PMP-2404 | JAN-1228

dtiwarATS:


assets/src/apps/delivery/components/ActivityRenderer.tsx
```diff
@@ -412,7 +412,7 @@ const ActivityRenderer: React.FC<ActivityRendererProps> = ({
       }
       const operator = initObject.operator;
       if (operator === 'bind to') {
-        initStateBindToFacts[initObject.target] = snapshot[key];
+        initStateBindToFacts[initObject.target] = snapshot[key] || '';
       } else {
```
dtiwarATS, line 416: If the value is undefined then we should send it as ''. Here we are just sending it to SIM so it doesn't matter, if the type of value is string | number etc. Once we send the value_change to SIM, it sends us the right variables so this should not cause any issues.

bsparks (approved):

PR #2171, opened 2022-02-03
[FEATURE] [MER-511] Horizontal scaling 

mcass19:
https://eliterate.atlassian.net/browse/MER-511

⚠️ ⚠️ DO NOT MERGE (we will do more testing) ⚠️ ⚠️ 

### Notes
- For a single node/instance this will continue to work as it is right now
- We use [libcluster](https://github.com/bitwalker/libcluster) for automatic cluster formation
- We are using the Gossip strategy for default config
- For prod env (releases), we will use the [EC2 strategy](https://github.com/kyleaa/libcluster_ec2) that uses `ex_aws` to query the EC2 DescribeInstances API endpoint. Access to this API should be granted to the EC2 instance profile.
- Take into account that different strategies could use different config options
- We are using the mnesia [distribution capability](https://hexdocs.pm/pow/Pow.Store.Backend.MnesiaCache.html#module-distribution) so that all new nodes connected to the cluster have the same data in the cache, therefore the pow session as well -> [more info here](https://github.com/pow-auth/pow_site/issues/10)


mcass19:
Closing this, will open a new one with some updates after doing some tests/experiments.

PR #2183, opened 2022-02-04
PMP-2470 | JAN-1302

dtiwarATS:


assets/src/adaptivity/capi.ts
```diff
@@ -98,12 +99,70 @@ export interface ICapiVariableOptions {
...
   allowedValues?: string[];
   bindTo?: string;
+  shouldConvertNumbers?: boolean;
 }
 
 export interface JanusConditionProperties extends ConditionProperties {
   id: string;
   type?: CapiVariableTypes;
 }
+export const parseArrayWithStringConversion = (val: unknown): unknown[] => {
```
dtiwarATS, line 109: Created a new function to handle this just for shake of minimal impact on any other areas.

bsparks, line 109: isn't this *without* string conversion? this version is *not* converting the numbers to numbers that are strings

assets/src/adaptivity/capi.ts
```diff
@@ -1,5 +1,5 @@
 import { ConditionProperties } from 'json-rules-engine';
-import { isStringArray, parseArray, parseBoolean } from 'utils/common';
+import { isStringArray, parseArray, parseBoolean, parseNumString } from 'utils/common';
```
bsparks, line 2: why do we need to add that?

assets/src/adaptivity/capi.ts
bsparks, line 109: also this one maybe should go into the utils/common and have a unit test to show that it isn't altering the arrays

bsparks (approved):

PR #2186, opened 2022-02-05
MER-700 Infiniscope Cognito

jrissler:
Competes initial Infiniscope Cognito requirements.

Next features for this start almost immediately where a suggested enhancement will be to move most of the JWK auth into a plug or shared module to streamline this controller.

### DIFF
--- a/lib/oli/groups.ex
+++ b/lib/oli/groups.ex
@@ -123,6 +123,35 @@ defmodule Oli.Groups do
   # Communities accounts
 
   @doc """
+  Finds or creates a community account for a user. More efficient than using
+  many-to-many assoc.
+
+  ## Examples
+
+      iex> find_or_create_community_user_account(user_id, community_id)
+      {:ok, %CommunityAccount{}}
+
+      iex> find_or_create_community_user_account(bad_user_id, community_id)
+      {:error, %Ecto.Changeset{}}
+
+  """
+  def find_or_create_community_user_account(user_id, community_id) do
+    case Repo.one(
+           from(account in CommunityAccount,
+             where:
+               account.user_id == ^user_id and
+                 account.community_id == ^community_id
+           )
+         ) do
+      nil ->
+        create_community_account(%{user_id: user_id, community_id: community_id})
+
+      account ->
+        {:ok, account}
+    end
+  end
+
+  @doc """
   Creates a community account.
 
   ## Examples
--- a/lib/oli/groups.ex
+++ b/lib/oli/groups.ex
@@ -136,19 +136,24 @@ defmodule Oli.Groups do
 
   """
   def find_or_create_community_user_account(user_id, community_id) do
-    case Repo.one(
-           from(account in CommunityAccount,
-             where:
-               account.user_id == ^user_id and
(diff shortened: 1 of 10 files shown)


lib/oli/groups.ex
```diff
@@ -122,6 +122,35 @@ defmodule Oli.Groups do
...
+
+      iex> find_or_create_community_user_account(user_id, community_id)
+      {:ok, %CommunityAccount{}}
+
+      iex> find_or_create_community_user_account(bad_user_id, community_id)
+      {:error, %Ecto.Changeset{}}
+
+  """
+  def find_or_create_community_user_account(user_id, community_id) do
+    case Repo.one(
```
darrensiegel, line 139: This "find" and "create" as two separate steps needs to be wrapped within a transaction.

lib/oli_web/controllers/cognito_controller.ex
```diff
@@ -0,0 +1,133 @@
...
+        %{
+          "product_id" => product_id,
+          "cognito_id_token" => jwt,
+          "error_url" => _error_url,
+          "community_id" => community_id
+        } = params
+      ) do
+    with {:ok, jwk} <- get_jwk(jwt),
+         {true, %{fields: jwt_fields}, _} <- JOSE.JWT.verify_strict(jwk, ["RS256"], jwt),
+         %Section{} = section <- Sections.get_section_by(slug: product_id),
```
darrensiegel, line 26: We probably want "create lms user" and "create community account" to be atomic operations, therefore this could should be wrapped inside of a transaction

darrensiegel (changes requested):

lib/oli_web/controllers/cognito_controller.ex
```diff
@@ -0,0 +1,143 @@
+defmodule OliWeb.CognitoController do
+  use OliWeb, :controller
+
+  import Oli.{HTTP, Utils}
+
+  require Logger
```
gastonabella, line 6: We are requiring but not using this.

lib/oli_web/controllers/cognito_controller.ex
```diff
@@ -0,0 +1,143 @@
...
+         {:ok, %{"iss" => issuer}} <- Joken.peek_claims(jwt),
+         {:ok, sso_jwk} <- find_or_fetch_jwt(kid, issuer) do
+      {:ok, JOSE.JWK.from_pem(sso_jwk.pem)}
+    else
+      {:error, error} ->
+        {:error, error}
+    end
+  end
+
+  def find_or_fetch_jwt(kid, issuer) do
```
gastonabella, line 62: Nit: I realized this and some other functions below are not private. Is there any specific reason? If so I would move all the private ones to the bottom.

gastonabella (commented):
LGTM! Just left a couple of small comments

lib/oli_web/controllers/cognito_controller.ex
jrissler: Awesome will apply to next ticket! Great catch!

PR #2203, opened 2022-02-10
Bind To Issues | PMP-2404

dtiwarATS:


assets/src/apps/delivery/components/ActivityRenderer.tsx
```diff
@@ -353,6 +353,7 @@ const ActivityRenderer: React.FC<ActivityRendererProps> = ({
       }
       if (initObject.operator === 'bind to') {
         initStateBindToFacts[initObject.target] = snapshot[key] || '';
+        return acc;
```
dtiwarATS, line 356: We don't send the BindTo trap states. I missed it with the new approach to fix the SW

assets/src/components/parts/janus-capi-iframe/ExternalActivity.tsx
```diff
@@ -33,7 +33,7 @@ const ExternalActivity: React.FC<PartComponentProps<CapiIframeModel>> = (props)
   const [ready, setReady] = useState<boolean>(false);
   const [initState, setInitState] = useState<any>(null);
   const [initStateReceived, setInitStateReceived] = useState(false);
-  const [initStateBindToFacts, setInitStateBindToFacts] = useState<any>(null);
+  const [initStateBindToFacts, setInitStateBindToFacts] = useState<any>({});
```
dtiwarATS, line 36: This needs to be {} by default instead of null

bsparks (approved):

PR #2225, opened 2022-02-17
[FEATURE] Limit paywall editing to admins MER-840

darrensiegel:
This PR limits paywall settings editing to author admin accounts only. 

This is a relatively low risk change. I've tested it interactively to ensure that only an admin can edit these settings now.


jrissler (approved):
Looks great!

PR #2227, opened 2022-02-17
MER-837 Add an index action to Cognito controller

jrissler:
Two notes:

- Created MER-843 and added to our current sprint, while I think this is "ok" - we can make this a little cleaner w small refactoring.
- Also asking Infiniscope to change JWT token parameter to be consistent - either `id_token` or `cognito_id_token`. So will add a future PR for that when they decide what to change it too.

### DIFF
--- a/lib/oli_web/controllers/cognito_controller.ex
+++ b/lib/oli_web/controllers/cognito_controller.ex
@@ -14,6 +14,33 @@ defmodule OliWeb.CognitoController do
 
   @jwks_relative_path "/.well-known/jwks.json"
 
+  def index(
+        conn,
+        %{
+          "id_token" => jwt,
+          "error_url" => _error_url,
+          "community_id" => community_id
+        } = params
+      ) do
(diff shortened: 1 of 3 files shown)


lib/oli_web/controllers/cognito_controller.ex
```diff
@@ -14,6 +14,33 @@ defmodule OliWeb.CognitoController do
 
   @jwks_relative_path "/.well-known/jwks.json"
 
+  def index(
```
darrensiegel, line 17: Should this function have a parameter-less version similar to:

```
  def launch(conn, params) do
    redirect_with_error(conn, params, "Missing parameters")
  end
```

darrensiegel (changes requested):

lib/oli_web/controllers/cognito_controller.ex
jrissler: Great catch - thanks!

darrensiegel (approved):

PR #2229, opened 2022-02-17
[FEATURE] Authoring/Clone Portal flow. WORK-IN-PROGRESS-DO-NOT-MERGE

darrensiegel:
This PR implements 95% of what I believe that we need to support a Cognito user launching into Torus to obtain an authoring account, join a community, and get their own cloned copy of a specific *Project*

The missing piece is how to implement support for creating the Cognito based authoring account.  This would take place in the cognito controller: 

```
  defp find_or_create_author(fields, _communit_id) do
    case Accounts.get_author_by_email(fields["email"]) do
      nil -> {:ok, nil}
      author -> {:ok, author}
    end
  end
```

You can see above that when the account does not exist, the implementation of the creation is stubbed out (returning `{:ok, nil}`. 

The catch here to properly support adding Cognito based author accounts is that we need to configure and add a new `PowAssent` provider (similar to what exists right now for Google and GitHub) to support Cognito.  I suspect an Oauth2 or perhaps an OIDC based provider would work - but I don't have all the details on Cognito to attempt this, or to test it out. 

[https://github.com/pow-auth/pow_assent/blob/master/README.md](https://github.com/pow-auth/pow_assent/blob/master/README.md) contains an example OIDC configuration.

### DIFF
--- a/lib/oli/authoring/clone.ex
+++ b/lib/oli/authoring/clone.ex
@@ -78,8 +78,7 @@ defmodule Oli.Authoring.Clone do
   end
 
   @doc """
-  Returns true if the given author already is a collaborator on a project that was cloned from the
-  project specified by the given project slug.
+  Returns the existing clones from a parent project that an author might have.
   """
   def existing_clones(project_slug, author) do
     from(
@@ -96,6 +95,10 @@ defmodule Oli.Authoring.Clone do
     |> Repo.all()
   end
 
+  @doc """
+  Returns true if the given author already is a collaborator on a project that was cloned from the
+  project specified by the given project slug.
+  """
   def already_has_clone?(project_slug, author) do
     existing_clones(project_slug, author)
     |> Enum.count() > 0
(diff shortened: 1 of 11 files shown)


eliknebel (commented):
Looks good. I would suggest protecting the launch and launch_clone routes using a plug that requires the SSO login, then once logged in, automatically create an account on their behalf if one doesn’t exist. Any POW assent OIDC/OAuth 2.0 or custom strategy should handle all of this.

This would result in the following:
- If for some reason someone uses the link without being logged in already, it will prompt them to log in/create an account
- If someone is already logged in with their SSO, then they will be redirected to “Authorize” torus which will result in an account being created
- If someone is logged in through SSO and already has a torus account, they will simply be prompted if they wish to clone the course

jrissler:
Fwiw - followed up with Infiniscope again about their staging env. Where I'll be able to setup links to unblock @darrensiegel

jrissler:
Closing - superseded by other PR's.

PR #2238, opened 2022-02-21
PMP-2509 | JAN-1350

dtiwarATS:


assets/src/components/parts/janus-text-flow/TextFlow.tsx
```diff
@@ -229,9 +229,12 @@ const TextFlow: React.FC<PartComponentProps<TextFlowModel>> = (props: any) => {
 
       let bgColor = 'transparent';
       if (palette.fillColor >= 0) {
-        bgColor = chroma(palette.fillColor || 0)
-          .alpha(palette.fillAlpha || 0)
-          .css();
+        const color = chroma(palette.fillColor || 0).alpha(palette.fillAlpha || 0);
+        // for few use cases, chroma(palette.fillColor || 0).alpha(palette.fillAlpha || 0) was sometimes getting evaluated to a numeric number(i.e. 1)
+        //hence 1.css() was throwing error as it is not a supported fuction for number.
+        if (typeof color === 'object') {
```
bsparks, line 235: so in those cases where it's 1 or a number, we want it to be transparent? or is that number somehow representing a different color?

bsparks (approved):

PR #2246, opened 2022-02-24
PMP-2479 | JAN-1310

dtiwarATS:


assets/src/adaptivity/operators/equality.ts
```diff
@@ -88,7 +88,7 @@ export const equalWithToleranceOperator = (
     return false;
   }
   if (Number.isNaN(modifiedFactValue)) {
-    return false;
+    return calledFromNotEqualToToleranceOperator ? true : false;
```
bsparks, line 91: no need to do a tertiary for a boolean, just return it

bsparks (approved):

PR #2250, opened 2022-02-24
[BUGFIX] [MER-842] [MER-853] System Message Banner issues

gastonabella:
[MER-853](https://eliterate.atlassian.net/browse/MER-853) & [MER-842](https://eliterate.atlassian.net/browse/MER-842)

Fix the following bugs:
- Internal Server Error being raised when an admin tries to manage system messages and the local timezone is not set (this is a separate issue: [MER-838](https://eliterate.atlassian.net/browse/MER-838)).
- Duplicate flash messages showing up.

### DIFF
--- a/lib/oli_web/live/system_message_live/index_view.ex
+++ b/lib/oli_web/live/system_message_live/index_view.ex
@@ -11,6 +11,8 @@ defmodule OliWeb.SystemMessageLive.IndexView do
   alias Surface.Components.Form
   alias Surface.Components.Form.{ErrorTag, Field, TextArea}
 
+  @utc_timezone "Etc/UTC"
+
   data title, :string, default: "Type a System Message"
   data breadcrumbs, :list
   data show_confirm, :boolean, default: false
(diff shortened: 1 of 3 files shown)


lib/oli_web/live/system_message_live/index_view.ex
```diff
@@ -100,8 +116,8 @@ defmodule OliWeb.SystemMessageLive.IndexView do
     new_message_attrs =
       attrs
       |> Oli.Utils.atomize_keys()
-      |> Map.update(:start, "", &datestring_to_utc_datetime(&1, socket.assigns.local_tz))
-      |> Map.update(:end, "", &datestring_to_utc_datetime(&1, socket.assigns.local_tz))
+      |> Map.update(:start, "", &datestring_to_utc_datetime(&1, socket.assigns.local_timezone))
```
eliknebel, line 119: There are some facilities that exist to format datetimes in the current user's timezone, then fallback to UTC if not is not available. See `OliWeb.Common.FormatDateTime.date`.

This helper function is already imported into all live views in `OliWeb` so yo should be able ot directly call `date(some_date, @context)` where @context is an assign set by calling `context = OliWeb.Common.SessionContext.init(session)` in the mount. See `lib/oli_web/live/qa/qa_live.ex` for an example

eliknebel (changes requested):

lib/oli_web/live/system_message_live/index_view.ex
eliknebel: Ah, nevermind. I now realize this is already utilizing this

eliknebel (approved):

PR #2259, opened 2022-02-28
[FEATURE] Student-specific exceptions for gating

darrensiegel:
This PR adds support for student specific exceptions for gating and scheduling. 

## Core Impl

The core of this design is the introduction of the `:parent_id` attribute of the `%GatingCondition` schema.  This is used to model a "parent-child" relationship where the parent is the original (student-wide) gate, and the children are student-specific gates.  So, for any gate that needs a student exception, a new GatingCondition record is created that 1) points to the parent gate via the `parent_id` attribute and that 2) specifies which specific user this exception applies to via the `user_id` attribute. 

The `Gating.resource_open` has been renamed to `Gating.blocked_by`.  It's return value is a list of the gates that are currently blocking access to a particular resource.  This change was made to allow the case where gates exist to be more efficient (avoiding a second lookup of the gate details). 

`Gating.blocked_by` implements a new algorithm for determining if a resource is blocked or open: Specifically it must account for student specific exceptions. It does this by "throwing out" any base (parent) gate if at least one student specific exception exists for the student in question.  With the introduction of a new, second gate type `:always_open` the following conditions can now be modeled:

1. A schedule gate exists for all students, but for one student the schedule is extended two days: This would be accomplished by a root schedule gate, and a student specific exception of type `:schedule` with the different scheduling dates.
2. A schedule gate exists for all students, but simply does not apply to one specific student.  This would be accomplished by a root schedule date and a student specific exception of type `:always_open`. 
3. A "inverted" gate can be created where an `:always_open` gate is applied to a resource for all students.  Then, for one specific student an exception is created as type `:schedule` to restrict just that one student.

## UI/UX

The UI/UX support for exceptions heavily leverages the existing gating and scheduling UI.  The changes here make the listing, editing and creation surface components reusable in the context of exceptions.  This is doable since student exceptions are merely gating conditions.


eliknebel (changes requested):
As a UI convenience, click on the "Select resource..." text box also opens the resource picker. However, while the "Select" button itself behaves as expected, it seems the "Select student..." text box still opens the resource picker, instead of the student picker

Screen Shot 2022-02-28 at 12 54 56 PM

darrensiegel:
Great catch, I updated the UI to fix that

eliknebel (approved):

PR #2269, opened 2022-03-03
[FEATURE] Render containers as pages with a table of contents

### LINKED ISSUE #2261: Render "Container" pages in delivery
The delivery system needs to be changed to render pages for all containers present in a course.  In Legacy, "container" pages are rendered with an auto-generated table of conents that include titles and links to the children of that container. 

Container pages should appear in the "in order" traversal of a course - so as a student clicks the "Next" page button they will encounter and view the container pages as they navigate into new units, modules, sections, etc.

LEDThereBeLight:
Closes #2261 

Tested:
- Page Preview
- Section Preview as instructor
- Section as student


darrensiegel:
I pushed some changes that provide specific routes for rendering containers.  This is necessary because allowing containers to render through the page routes results in `ResourceAccess` and `ResourceAttempt` records being created for visited containers, which we really do not want to have happen.

Also, I removed the usage of `Resolver.full_hierarchy` in rendering a container page.  This is a fairly expensive call to make (the fetching of the entire hierarchy).  Instead, this impl now requires only a single Resolver call (for the revision of the container itself).

LEDThereBeLight:
The container-specific routes make sense. I wasn't sure if we wanted to show container children deeply or not, so I had left the full hierarchy call. 

The changes work well.

darrensiegel (approved):

PR #2282, opened 2022-03-03
[MER-743] Add ability to delete a section

gastonabella:
[MER-743](https://eliterate.atlassian.net/browse/MER-743)

Add button in the Sections Overview page to delete (if there's no student associated data) a section or archive it.

### DIFF
--- a/lib/oli/delivery/sections.ex
+++ b/lib/oli/delivery/sections.ex
@@ -20,6 +20,7 @@ defmodule Oli.Delivery.Sections do
   alias Oli.Authoring.Course.Project
   alias Oli.Delivery.Hierarchy
   alias Oli.Delivery.Hierarchy.HierarchyNode
+  alias Oli.Delivery.Snapshots.Snapshot
   alias Oli.Resources.ResourceType
   alias Oli.Publishing.DeliveryResolver
   alias Oli.Resources.Revision
@@ -282,25 +283,18 @@ defmodule Oli.Delivery.Sections do
   end
 
   @doc """
-  Returns a listing of all student enrollments for a given section.
-
+  Returns true if there is student data associated to the given section.
   """
-  def list_student_enrollments(section_slug) do
-    learner_role_id = ContextRoles.get_role(:context_learner).id
-
+  def has_student_data?(section_slug) do
     query =
       from(
-        e in Enrollment,
-        join: s in assoc(e, :section),
-        join: ecr in "enrollments_context_roles",
-        on: ecr.enrollment_id == e.id,
-        where:
-          s.slug == ^section_slug and s.status == :active and
-            ecr.context_role_id == ^learner_role_id,
-        select: e
+        snapshot in Snapshot,
+        join: s in assoc(snapshot, :section),
+        where: s.slug == ^section_slug and s.status == :active,
+        select: snapshot
       )
 
-    Repo.all(query)
+    Repo.aggregate(query, :count, :id) > 0
   end
 
   def get_enrollment(section_slug, user_id) do
--- a/lib/oli/delivery/sections.ex
+++ b/lib/oli/delivery/sections.ex
@@ -291,8 +291,7 @@ defmodule Oli.Delivery.Sections do
     query =
       from(
         e in Enrollment,
-        join: s in Section,
-        on: e.section_id == s.id,
+        join: s in assoc(e, :section),
         join: ecr in "enrollments_context_roles",
         on: ecr.enrollment_id == e.id,
         where:
@@ -642,18 +641,6 @@ defmodule Oli.Delivery.Sections do
   """
   def soft_delete_section(%Section{} = section) do
     update_section(section, %{status: :deleted})
-  end
-
-  @doc """
-  Archives a section by marking the record as archived.
-  ## Examples
-      iex> archive_section(section)
-      {:ok, %Section{}}
-      iex> archive_section(section)
-      {:error, %Ecto.Changeset{}}
-  """
-  def archive_section(%Section{} = section) do
-    update_section(section, %{status: :archived})
   end
 
   @doc """
--- a/lib/oli/delivery/sections.ex
+++ b/lib/oli/delivery/sections.ex
@@ -290,7 +290,7 @@ defmodule Oli.Delivery.Sections do
       from(
         snapshot in Snapshot,
         join: s in assoc(snapshot, :section),
-        where: s.slug == ^section_slug and s.status == :active,
+        where: s.slug == ^section_slug,
         select: snapshot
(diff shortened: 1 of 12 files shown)


lib/oli_web/live/sections/overview_view.ex
```diff
@@ -144,4 +150,65 @@ defmodule OliWeb.Sections.OverviewView do
...
+        entity_type: "section",
+        entity_id: socket.assigns.section.id,
+        delete_enabled: true,
+        delete: "delete_section",
+        modal_action: action
+      }
+    }
+
+    {:noreply,
+     assign(socket, modal: modal, section_has_student_enrollments: section_has_student_enrollments)}
```
darrensiegel, line 184: `section_has_student_enrollments` should not be cached as an assign here, and then later used in the `delete_section` event.   This introduces a timing bug. Imagine the following:

1. Admin user clicks "Delete section"
2. The `show_delete_modal` event is handled and the check to see if section has student enrollments executes.  It returns `false`, indicating that a hard delete is possible.  The modal displays and the `section_has_student_enrollments` value is cached in the assign.
3. While the modal is displayed and before the user clicks "Ok" to proceed, a student launches from their LMS and enrolls in the section and visits the first page, thus generating student data.
4. The admin user clicks "Ok" to hard delete the section, which results in student data loss.

darrensiegel (changes requested):
I do not believe that the logic used for determining whether or not student data exists in a course section is correct.  With this implementation in place, consider the following scenario:

1. A student enrolls in a course section.
2. Student visits every practice page in the course and exercises all activities on these pages.
3. The student drops the class, and is then "Unenrolled" from the course section.

With the above scenario, this implementation would determine that "No student data exists" (since there are no enrollments), and thus allow the section to be deleted, which with the cascading deletes that this PR introduces, would delete attempt records and snapshots.  

I believe that a better approach would be to add a new `:status` enum field to the `Enrollment` schema, which can be either `:active` or `:disabled`.  Then any enrolled student that needs to be unenrolled can simply have their status set to `:disabled`.  They stop showing up in all Torus student/roster UIs, but their data still remains (this is important for analytics purposes).  If they happen to re-enroll, their status is simply set back to `:active`, which has the nice side effect of restoring all of their previous state. 

In addition to this new "soft delete" for Enrollments, the check to determine if student data exists should be changed to simply query the count of Snapshot records. If one or more Snapshot records exist that point to that section_id the section cannot be hard deleted.  If none exist, it is indeed safe to hard delete this section.  One edge case that this also allows (which this PR allows as well) is that if the only enrolled user in a section is an instructor it is safe to delete this section.  Snapshot records do not get created when Instructors exercise course material.

lib/oli_web/live/sections/overview_view.ex
gastonabella: This is an edge case but it might happen, good catch! If we don't cache `section_has_student_enrollments` in the assigns, then we'll need to call the function again in the `delete_section` event, but in the scenario you described we would have told the user that the section was going to be "deleted", and instead it will be "archived" which is not ideal.
One option would be to still cache it in the assigns, and in the `delete_section` event compare both boolean values. If they are different then we could re-render the page asking the user to try again or so. Thoughts?

gastonabella:
> I do not believe that the logic used for determining whether or not student data exists in a course section is correct. With this implementation in place, consider the following scenario:
> 
> 1. A student enrolls in a course section.
> 2. Student visits every practice page in the course and exercises all activities on these pages.
> 3. The student drops the class, and is then "Unenrolled" from the course section.
> 
> With the above scenario, this implementation would determine that "No student data exists" (since there are no enrollments), and thus allow the section to be deleted, which with the cascading deletes that this PR introduces, would delete attempt records and snapshots.
> 
> I believe that a better approach would be to add a new `:status` enum field to the `Enrollment` schema, which can be either `:active` or `:disabled`. Then any enrolled student that needs to be unenrolled can simply have their status set to `:disabled`. They stop showing up in all Torus student/roster UIs, but their data still remains (this is important for analytics purposes). If they happen to re-enroll, their status is simply set back to `:active`, which has the nice side effect of restoring all of their previous state.
> 
> In addition to this new "soft delete" for Enrollments, the check to determine if student data exists should be changed to simply query the count of Snapshot records. If one or more Snapshot records exist that point to that section_id the section cannot be hard deleted. If none exist, it is indeed safe to hard delete this section. One edge case that this also allows (which this PR allows as well) is that if the only enrolled user in a section is an instructor it is safe to delete this section. Snapshot records do not get created when Instructors exercise course material.

@darrensiegel thank you for taking the time to review this and explaining how the enrollments logic works. I don't have the full context so this is really helpful.
Some comments about your suggestion:
1. I realized that there might be multiple context roles for the same enrollment. For the case an enrollment has two or more context roles associated, which role does that `status` belongs to?
2. If 1) is not a problem and we want to proceed with the solution, does it make sense to do these logic changes in a separate story? I think it's cleaner to keep this story as simple as possible.
3. If 2) (we make the enrollments logic change separately) could we still determine if student data exists by counting the Snapshot records?

lib/oli_web/live/sections/overview_view.ex
darrensiegel: I think re-rendering with the update that it can now only be archived makes sense.

darrensiegel (changes requested):

darrensiegel:
Yes, the enrollments status change and soft delete should be a different story.  Just counting snapshot records is sufficient.

lib/oli/delivery/sections.ex
```diff
@@ -273,21 +274,36 @@ defmodule Oli.Delivery.Sections do
...
 
+  @doc """
+  Returns true if there is student data associated to the given section.
+  """
+  def has_student_data?(section_slug) do
+    query =
+      from(
+        snapshot in Snapshot,
+        join: s in assoc(snapshot, :section),
+        where: s.slug == ^section_slug and s.status == :active,
```
darrensiegel, line 293: It seems wrong to include "status == `:active`" within this query.   If `has_student_data?` was ever called on a section that was archived and that actually had student data, it would return `false`, and perhaps lead data being deleted.

darrensiegel (changes requested):
Seem my inline comment regarding one more change in `has_student_data?`

lib/oli/delivery/sections.ex
gastonabella: Actually there's no way for a user to "delete" an archived section in Torus now, but I see your point and this logic doesn't seem semantically right. I pushed the changes in the last commit, thanks!

darrensiegel (approved):

PR #2291, opened 2022-03-08
[ENHANCEMENT] Revision history monaco json validation

eliknebel:
This PR enhances the capability of the admin revision history editor but allowing direct JSON content editing and validation. It also adds a proper breadcrumb for easier navigation.

Screen Shot 2022-03-07 at 8 05 27 PM

By using JSON schema validation, the system should be capable of preventing any invalid json from being imported or edited into the system, minimizing risk of delivery-time crashes.

With this PR there is now a front-end react monaco editor and as well as a wrapper surface component for liveviews. This required some webpack changes for monaco.

This PR also add a high-level activity schema and refactors the existing schemas to be more straightforward and complete. Because the editor is not completely consistent in the JSON it produces across activity types, it is difficult to create a common schema. We will either have to change the editor to be more consistent (and add migration capabilities for existing content) or add a specific schema for each activity type. It might make sense in the future to add a schema for each activity as part of its manifest in order to be more flexible with any 3rd party activity extensibility torus may provide in the future.


darrensiegel (changes requested):
I noticed that graph nodes were not re-rendering selection changes.  This was due to the application of `phx-update="ignore"` at the top level `svg` element.  I removed that and added the necessary unique DOM ids to allow this to work.  Let me know, though, if that `phx-update="ignore"` was necessary to fix other issues.

After I get a failed validation during editing, I see a button that links me out to the relevant JSON schema.  That link does not resolved:

`http://torus.oli.cmu.edu/schemas/v0-1-0/page-content.schema.json`

eliknebel:
> I noticed that graph nodes were not re-rendering selection changes. This was due to the application of `phx-update="ignore"` at the top level `svg` element. I removed that and added the necessary unique DOM ids to allow this to work. Let me know, though, if that `phx-update="ignore"` was necessary to fix other issues.
> 
> After I get a failed validation during editing, I see a button that links me out to the relevant JSON schema. That link does not resolve:
> 
> `http://torus.oli.cmu.edu/schemas/v0-1-0/page-content.schema.json`

Good catch, thanks. I originally added that because liveview re-enders were breaking the graph but it appears to work now, maybe with the javascript changes.

The schemas wont actually be available at the urls until we deploy to production, as they are hosted by torus itself and torus.oli.cmu.edu is just a CNAME to proton.

I also discovered an issue where editing content seemed to render a bunch of duplicate rows, this was fixed by converting the table component into a surface component.

darrensiegel (approved):

PR #2295, opened 2022-03-08
[MER-807] Rename LMS-Lite to Direct Delivery

gastonabella:
[MER-807](https://eliterate.atlassian.net/browse/MER-807)

Replace every part of the code where it says "LMS-Lite" with "Direct Delivery".


bsparks (approved):
looks technically fine to me

darrensiegel (approved):

PR #2296, opened 2022-03-08
[BUGFIX] [MER-793] Open lesson in new tab

mcass19:
[MER-793](https://eliterate.atlassian.net/browse/MER-793)

Open lessons in new tab. I consider all the possible links a course could have. Is that fine for you?


darrensiegel (changes requested):
We do not want to take this impl, as the solution to this navigation challenge (difficulty in returning to the Overview page) is not to open all lessons in a new tab.

mcass19:
Thanks @darrensiegel ! Will close and continue necessary conversations.

PR #2302, opened 2022-03-08
[BUGFIX] add favicon code from advanced delivery

bsparks:


bsparks:
I just copied this from advanced delivery, honestly I wasn't sure how to test locally given that I don't have S3 setup locally. pls let me know if there is more to it than this

darrensiegel (approved):

PR #2318, opened 2022-03-12
[BUG FIX] Handle malformed objectives attrs AOPS-543

darrensiegel:
This fix addresses an issue where objectives rollup was breaking when encountering pages whose `objectives` attribute was an empty map, instead of a map with a single key of `attached` which points to the list of attached objective resource ids.

In making this fix, I have removed the feature here that defers to objectives attached to activity revisions when there are none attached to the page.  This brings this feature inline with how legacy OLI works - and will better support upcoming course migration.


jrissler (approved):
Looks great and love the new tests!

bsparks (approved):

PR #2330, opened 2022-03-15
[MER-402] Show auth providers login buttons if config vars are enabled

gastonabella:
[MER-402](https://eliterate.atlassian.net/browse/MER-402)

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -196,6 +196,12 @@ config :ex_json_schema,
 # Configure if age verification checkbox appears on learner account creation
 config :oli, :age_verification, is_enabled: System.get_env("IS_AGE_VERIFICATION_ENABLED", "")
 
+config :oli, :auth_providers,
+  google_client_id: System.get_env("GOOGLE_CLIENT_ID", ""),
+  google_client_secret: System.get_env("GOOGLE_CLIENT_SECRET", ""),
+  github_client_id: System.get_env("GITHUB_CLIENT_ID", ""),
+  github_client_secret: System.get_env("GITHUB_CLIENT_SECRET", "")
+
 # Import environment specific config. This must remain at the bottom
 # of this file so it overrides the configuration defined above.
 import_config "#{Mix.env()}.exs"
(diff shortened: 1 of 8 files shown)


eliknebel (changes requested):
This implementation combines the provider configs for both users and authors. This means whatever provider config is set for authoring will be the same for delivery. Because Github OAuth only supports a single callback uri, the github login will not work for both (authoring uses callback uri path `/authoring/auth/github/callback` while delivery is just `/auth/github/callback`.

To support both, one of the following changes is most likely required:
1. Either make the provider client_id and secret configs separately configurable for users and authors, e.g. `USER_GITHUB_CLIENT_ID, AUTHOR_GITHUB_CLIENT_ID ... )
2. Or make the path used for redirect_uri have a common root so that a single Github OAuth app can be configured for both. See https://docs.github.com/en/developers/apps/building-oauth-apps/authorizing-oauth-apps#redirect-urls for more info.

The first approach is preferred, since it allows torus to separately configure which OAuth methods to support for which domain (delivery/authoring)

gastonabella:
> This implementation combines the provider configs for both users and authors. This means whatever provider config is set for authoring will be the same for delivery. Because Github OAuth only supports a single callback uri, the github login will not work for both (authoring uses callback uri path `/authoring/auth/github/callback` while delivery is just `/auth/github/callback`.
> 
> To support both, one of the following changes is most likely required:
> 
> 1. Either make the provider client_id and secret configs separately configurable for users and authors, e.g. `USER_GITHUB_CLIENT_ID, AUTHOR_GITHUB_CLIENT_ID ... )
> 2. Or make the path used for redirect_uri have a common root so that a single Github OAuth app can be configured for both. See https://docs.github.com/en/developers/apps/building-oauth-apps/authorizing-oauth-apps#redirect-urls for more info.
> 
> The first approach is preferred, since it allows torus to separately configure which OAuth methods to support for which domain (delivery/authoring)

Thanks for the suggestions @eliknebel, I decided to go with the first approach. Now only github vars are configured separately based on user type. Do you think we should make it more generic?

eliknebel:
I think this should be fine for now. We will probably want to make it more generic at some point so that a deployment config can specify exactly which strategies to enable separately for users and authors, and then those strategies can pickup the necessary credentials accordingly.

I would imagine something like:
```
USER_AUTH_STRATEGIES=google,github,etc...
AUTHOR_AUTH_STRATEGIES=google,github,etc...

GOOGLE_CLIENT_ID=...
```

PR #2351, opened 2022-03-21
WIP: Don't merge [MER-636] Fixed XSS vulnerability on activity bank page.

marc-hughes:
WIP: Don't merge

There was a potential XSS vulnerability when entering a learning objective and then viewing the activity-bank page.

Steps to reproduce: 
1.	Login as an authoring user
2.	Create a course or select existing
3.	Navigate to “Create -> Objectives”
4.	Type a payload and click ‘Create’ to save it:
`</script><script>alert(document.location)</script><!--`
5.	Navigate to “Create -> Activity Bank”
6.	Observe executed JavaScript as an alert pop-up.

This was caused by the way we were passing params to the embedded react app on that page since @context contained some user-supplied data.
```
const params = <%= raw(Jason.encode!(@context)) %>;
```

I tried a few different ways of encoding the data:
- Manually encoding each value - this turned out pretty messy due to the deeper hierarchy of data passed.
- Using Phoenix.HTML.html_escape - this escapes characters you really want to preserve, like quotes. Tried to be clever with un-escaping those, but didn't feel right.

In the end, went with a URI encoded plain string that the client page would then parse out. It has the benefit of always working, being straight-forward, and easily readable. Downside of having to do a JSON.parse in JS land.

### DIFF
--- a/lib/oli_web/templates/activity_bank/index.html.eex
+++ b/lib/oli_web/templates/activity_bank/index.html.eex
@@ -7,7 +7,7 @@
 <div id="editor" class="container"/>
 
 <script>
-
-  const params = <%= raw(Jason.encode!(@context)) %>;
+  const encodedParams = "<%= json_escape(@context) %>";
+  const params = JSON.parse(decodeURIComponent(encodedParams));
   window.oliMountApplication(document.getElementById('editor'), params);
 </script>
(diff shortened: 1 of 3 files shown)


marc-hughes:
Going a different direction with this.

PR #2366, opened 2022-03-28
[ENHANCEMENT] Allow payment during grace period CMU-417

darrensiegel:
This PR implements the ability for a student that is accessing a paywalled course during the grace period to:
1. See and understand how much time they have left remaining in the grace period
2. Proceed forward with making a payment or applying an access code

The changes here are:
1. Provide a `summarize_access` method in the `Paywall` context that returns a struct that details not only whether or not a user can access a course section, but also the reason why (`:not_paid`, `:within_grace_period`, `:instructor`, etc). 
2. Adjust the `EnforcePaywall` plug to insert the `AccessSummary` struct into the connection assigns for later use in a banner display
3. Add a banner display when the user is accessing the course in the grace period.  

To test this out:
1. Create a paywalled course section, that doesn't require enrollment, with an active grace period.
2. Access the course as a student. You should now see a warning banner that details the time remaining in the grace period and that provides a link to initiate payment

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 - Instructor review of completed graded attempts
 - Allow gates to be defined in products
 - Hide subsequent purpose types for activities when the same purpose type is used in a series
+- Allow students to pay and apply codes during a grace period
 
 ## 0.18.4 (2022-02-24)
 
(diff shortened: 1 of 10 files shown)


eliknebel:
I've tested this and it works as expected, however I wonder if this should apply to Guest users (sections which do not require registration) given that a guest can simply leave the course and come back as a new guest at any time. This might be something we just want to consider in the future.

eliknebel (approved):

PR #2370, opened 2022-03-29
PMP-2580 | JAN-1417

dtiwarATS:
The correct trap state has codition 'session.visits.q:1484171837224:1023 = true'. Currently the value of 'session.visits.q:1484171837224:1023' gets set to numeric value i.e. 1; so made changes to fix this. We could compare that with '1==true' but since we are checking with '===', need the code to handle this.

### DIFF
--- a/assets/src/adaptivity/operators/equality.ts
+++ b/assets/src/adaptivity/operators/equality.ts
@@ -51,6 +51,9 @@ export const isEqual = (factValue: any, value: any): boolean => {
     return value.toString().toLowerCase() === 'true' ? true === factValue : false === factValue;
   }
   if (typeOfValue === 'boolean') {
+    if (typeOfFactValue === 'number') {
+      return factValue === 1 ? true : false;
+    }
     return factValue.toString().toLowerCase() === 'true' ? true === value : false === value;
   }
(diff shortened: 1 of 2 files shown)


assets/src/adaptivity/operators/equality.ts
```diff
@@ -51,6 +51,9 @@ export const isEqual = (factValue: any, value: any): boolean => {
     return value.toString().toLowerCase() === 'true' ? true === factValue : false === factValue;
   }
   if (typeOfValue === 'boolean') {
+    if (typeOfFactValue === 'number') {
+      return factValue === 1 ? true : false;
```
bsparks, line 55: I think with visits it can be more than 1, so probably you want to check `>0` for truthiness of numbers. also you don't need to do a tertiary for boolean, just `return factValue > 0;`

bsparks: oh and also we should add unit tests for these new cases

dtiwarATS: Unit test added and refactored the code to match all the use cases

bsparks (approved):

PR #2389, opened 2022-04-02
[ENHANCEMENT] improve sass styles compile performance

eliknebel:
This PR removes node-sass-glob imports in favor of directly importing styles in sass files to eliminate the performance burden of crawling the entire src and style directories.

On an intel macOS machine, this improves initial compilation time from about 1 minute to 30 seconds, and HMR updates from 16 seconds to 7 seconds, effectively cutting developer wait time in half.

Screen Shot 2022-04-01 at 7 49 27 PM
Screen Shot 2022-04-01 at 8 24 44 PM

As part of this refactor, any component styles that live under `assets/src` will now use style-loader and themes to use mini-css-extract-plugin. These style-loader styles are compiled even faster since they are separate from the bootstrap theming.

Screen Shot 2022-04-03 at 8 34 39 PM

This means any ts react component utilizing a scss file to style it should import the file directly. As a convention, this file should live right next to the component it supports and to ensure proper scoping [css modules](https://github.com/css-modules/css-modules) can now be used. For example:
```
MyComponent.tsx
MyComponent.module.scss
```

And the component should include in the top imports:
```ts
...
import styles from './MyComponent.modules.scss';
...

<div className={styles.myComponent}>
...
</div>
```

Theme variables like colors can be accessed in the component stylesheet and dark/delivery specific styles can be set accordingly, for example using css modules:
```sass
// optionally use the current authoring and/or delivery theme
@use 'authoring/theme' as authoring;
@use 'delivery/theme' as delivery;

// some basic styling using colors from the active theme
.myComponent {
 color: authoring.$primary;
}

:global(html.dark) {
  .myComponent {
   color: authoring.$white;
  }
}

// optionally scope styles to delivery using colors from the current delivery theme

:global(html.delivery) {
  .myComponent {
    background-color: delivery.$primary;
  }
}

:global(html.delivery.dark) {
  .myComponent {
    color: delivery.$white;
  }
}

```

**NOTICE:** Using [CSS Modules](https://github.com/css-modules/css-modules) for styles is highly recommended over straight CSS/SASS classes. Without modules, styles could be overridden by the base bootstrap styles because the themes import bootstrap into this scope, making them more specific by default. If necessary, however, regular SASS and CSS can still be used with a regular import statement `import './SomeStyles.scss'`.

Users can also set which dark-mode setting they wish to use directly in the account page. The preference is stored in browser local storage. Auto follows the operating system preference, as it did before (default).

Screen Shot 2022-04-03 at 5 06 02 PM
Screen Shot 2022-04-03 at 10 52 56 PM

This PR modifies core authoring and delivery styles for both basic and adaptive pages and therefore should be tested thoroughly before being deployed to a production environment.


lib/oli_web/templates/layout/authoring.html.leex
```diff
@@ -11,20 +11,9 @@
...
-    <link rel="stylesheet" href="/css/janus_slider_delivery.css"/>
-    <link rel="stylesheet" href="/css/janus_image_carousel_delivery.css"/>
-
     <!-- Theme CSS (contains all bootstrap styles as well as custom styles for a specific theme) -->
-    <link id="authoring-theme-light" rel="stylesheet" href="/css/authoring_torus_light.css" media="screen and (prefers-color-scheme: light)"/>
-    <link id="authoring-theme-dark" rel="stylesheet" href="/css/authoring_torus_dark.css" media="screen and (prefers-color-scheme: dark)"/>
-
-    <link id="delivery-theme-light" rel="stylesheet" href="/css/delivery_torus_light.css" media="(prefers-color-scheme: light)"/>
-    <link id="delivery-theme-dark" rel="stylesheet" href="/css/delivery_torus_dark.css" media="(prefers-color-scheme: dark)"/>
+    <link id="authoring-theme" rel="stylesheet" href="/css/authoring_default.css" />
```
bsparks, line 15: there is a bit of hackery that no longer works because the adapative author doesn't have a dark mode yet
https://github.com/Simon-Initiative/oli-torus/blob/master/assets/src/apps/authoring/Authoring.tsx#L128

eliknebel: Thanks Ben, I believe I have addressed this with https://github.com/Simon-Initiative/oli-torus/pull/2389/commits/48b0b1694081e1d7d730eec868d8bd9ef1fc9cf3 but please let me know if you still see an issue

LEDThereBeLight (approved):

LEDThereBeLight:
Approved/merged but left branch open for comments in case an issue is found related to the conversation above.

I did a fair amount of testing and it all looked good to me.

Great improvement.

PR #2390, opened 2022-04-04
PMP-2658 | JAN-1511

dtiwarATS:


assets/src/adaptivity/rules-engine.ts
```diff
@@ -328,6 +328,68 @@ export const getReferencedKeysInConditions = (conditions: any) => {
   return Array.from(references);
 };
 
+export const findReferencedActivitiesInMutateConditions = (conditions: any) => {
```
dtiwarATS, line 331: We need to look for the conditions in mutate state as well for variables Required For Evaluation. Just copied the conditions code and refactored it to use for Mutate State. The reason we need is, if there are mutation that changes the question  score, they get evaluated in the rules engine itself hence it needs all the variable to perform the calculation.

assets/src/adaptivity/rules-engine.ts
```diff
@@ -328,6 +328,68 @@ export const getReferencedKeysInConditions = (conditions: any) => {
...
+          }
+        });
+      }
+    }
+  });
+
+  return Array.from(referencedActivities);
+};
+
+export const getReferencedKeysInMutateConditions = (conditions: any) => {
```
dtiwarATS, line 356: Above comments applies here as well.

assets/src/adaptivity/rules-engine.ts
```diff
@@ -403,36 +465,35 @@ export const check = async (
...
+    if (mutations.length) {
+      const mutApplies = mutations.map(({ params }) => params);
+
+      bulkApplyState(mutApplies, env);
+      score = getValue('session.currentQuestionScore', env) || 0;
+    }
+  }
   //below condition make sure the score calculation will happen only if the answer is correct and
   //in case of incorrect answer if negative scoring is allowed then calculation will proceed.
-  if (isCorrect || scoringContext.negativeScoreAllowed) {
```
dtiwarATS, line 408: This needs to be applied only for screen level scoring calculations. If there a mutation at trap state level that changes the scoring then the scoring should be calculated.

assets/src/apps/authoring/store/groups/layouts/deck/actions/updateActivityRules.ts
```diff
@@ -110,6 +113,12 @@ export const updateActivityRules = createAsyncThunk(
             await updateNestedConditions(conditionsToUpdate, activityTree);
             referencedSequenceIds.push(...findReferencedActivitiesInConditions(conditionsToUpdate));
             referencedVariableKeys.push(...getReferencedKeysInConditions(conditionsToUpdate));
+            referencedSequenceIds.push(
```
bsparks, line 116: it's not only here that this gets called, also when the rule is edited in the author.

assets/src/adaptivity/rules-engine.ts
bsparks: do we need to have a separate method? or can we not just look at the actions while we're going through the original loop? also if we *do* need to have a separate method, it would be good to change the variable names to reflect what we are working with (actions) instead of conditions so that it is less confusing

dtiwarATS: PR Updated

assets/src/adaptivity/rules-engine.ts
```diff
@@ -261,7 +261,7 @@ export const defaultWrongRule = {
   },
 };
 
-export const findReferencedActivitiesInConditions = (conditions: any) => {
+export const findReferencedActivitiesInConditions = (conditions: any, actions: any) => {
```
bsparks, line 264: very sorry :D I was thinking that the actions were already going to be a part of the condition item; I do think that it would be good to have separate distinct methods for doing this, esp. since you had to edit the locations anyway

bsparks, line 264: very sorry :D I was thinking that the actions were already going to be a part of the condition item; I do think that it would be good to have separate distinct methods for doing this, esp. since you had to edit the locations anyway

bsparks (approved):

PR #2401, opened 2022-04-06
PMP-2400 | JAN-1224

dtiwarATS:


assets/src/adaptivity/scripting.ts
```diff
@@ -131,8 +131,17 @@ export const getExpressionStringForValue = (
           }
         }
       } catch (e) {
-        // if we have parsing error then we're guessing it's CSS
-        actuallyAString = true;
+        //lets evaluat everything if first and last char are {}
+        if (val[0] === '{' && val[val.length - 1] === '}') {
```
dtiwarATS, line 135: If the expression is a function then we need to make sure that we send (120/12) instead of {(120/10)}. We have to remove the '{}'. Scripting does evaluate with '{}'

bsparks (approved):

PR #2402, opened 2022-04-06
PMP-2665 | JAN-1518

dtiwarATS:


assets/src/components/parts/janus-capi-iframe/ExternalActivity.tsx
```diff
@@ -585,6 +585,15 @@ const ExternalActivity: React.FC<PartComponentProps<CapiIframeModel>> = (props)
             typeof changedVar.value === 'number'
           ) {
             changedVar.type = CapiVariableTypes.NUMBER;
+          } else if (
+            changedVar.type === CapiVariableTypes.STRING &&
```
dtiwarATS, line 589: so, the EverApp was sending value of 'app.spr-app-adaptivity.SW.Where.iNumPlanetsFound'  = "0" i.e. as a string value and there is a condition in trap state 

`let app.spr-app-adaptivity.SW.Where.iNumPlanetsFound = {app.spr-app-adaptivity.SW.Where.iNumPlanetsFound}+1`

and it was failing during evaluation in scripting.

assets/src/components/parts/janus-capi-iframe/ExternalActivity.tsx
```diff
@@ -585,6 +585,15 @@ const ExternalActivity: React.FC<PartComponentProps<CapiIframeModel>> = (props)
...
+            !Number.isNaN(changedVar.value)
+          ) {
+            const parsedValue = parseNumString(changedVar.value);
+            if (typeof parsedValue === 'number') {
```
dtiwarATS, line 593: value of 'parsedValue' can be anything so if it's a type number then only we modifying the value and type.

bsparks (approved):

PR #2411, opened 2022-04-07
[FEATURE] add content outline to page editor

eliknebel:
This PR refactors the page editor to show an outline and allow rearranging of content from the outline. This PR sets the groundwork to support groups and surveys. The outline can be expanded and collapsed using the toggle on the left.

Screen Shot 2022-04-07 at 12 00 25 PM


LEDThereBeLight:
This is awesome.

The only things I noticed are that in dark mode, the collapse button blends into the background: 
image

and the last item in the DnD list doesn't seem to be able to be moved. The others work fine for me. I have an activity bank selection last, in case that matters.

Nit: When you drag an item, placeholder space shows up even for the currently active item's position. So it ends up having twice the space. For example, try dragging the first item to the first slot (the one it's already in). You'll see a placeholder indicating it's moving, but it's really just reslotting itself to the same place.

assets/src/components/resource/editors/ContentOutline.tsx
```diff
@@ -1,12 +1,256 @@
...
+};
+
+const getViewportHeight = () =>
+  Math.max(document.documentElement.clientHeight || 0, window.innerHeight || 0);
+
+const calculateOutlineHeight = (scrollOffset: number) => {
+  const topMargin = 420;
+  const scrolledMargin = 200;
+  const scrollCompensation = Math.max(topMargin - scrollOffset, scrolledMargin);
+  return Math.max(getViewportHeight() - scrollCompensation, 220);
```
bsparks, line 45: 220 is a magic number, consider describing it in a variable like the others

LEDThereBeLight (approved):

PR #2414, opened 2022-04-08
Diagnostics Expression Check

dtiwarATS:


assets/src/apps/authoring/store/groups/layouts/deck/actions/validate.ts
```diff
@@ -273,23 +287,8 @@ export const validators = [
 
         return [...broken, ...brokenConditionValues];
       }, []);
-    },
-  },
-  {
-    type: DiagnosticTypes.INVALID_EXPRESSION_VALUE,
```
dtiwarATS, line 279: I created a separate method with same type 'INVALID_EXPRESSION_VALUE' for fact value expression check which caused some issues and it wasn't checking them all

bsparks (approved):

PR #2415, opened 2022-04-08
[FEATURE] [MER-733] UX restyle for select content

mcass19:
[MER-733](https://eliterate.atlassian.net/browse/MER-733)

Changes include:
- LMS instructor select source reuse existing live view (used for independent instructors and admins)
- Move creation logic from delivery controller to mentioned liveview
- Add a cards view and make possible to switch between card/list view (only for instructors, admin will remain only with the list)

### Before
before

### After
after


jrissler (approved):

dsuriano (approved):
UI looks correct based on discussions and screenshots. Have not tested the code locally.

eliknebel (approved):

PR #2416, opened 2022-04-08
[MER-741] Notify LTI user to login from LMS when logging directly into Torus

gastonabella:
[MER-741](https://eliterate.atlassian.net/browse/MER-741)

### DIFF
--- a/lib/oli_web/pow/messages.ex
+++ b/lib/oli_web/pow/messages.ex
@@ -1,0 +1,41 @@
+defmodule OliWeb.Pow.Messages do
+  @moduledoc """
+  Custom module that handles returned messages by pow actions.
+  """
+  use Pow.Phoenix.Messages
+
+  use Pow.Extension.Phoenix.Messages,
+    extensions: [
+      PowAssent,
+      PowResetPassword,
+      PowEmailConfirmation,
+      PowPersistentSession,
+      PowInvitation
+    ]
+
+  alias Oli.Accounts
+  alias OliWeb.Router.Helpers, as: Routes
+  alias Phoenix.HTML.Link
+  alias Pow.Phoenix.Messages
+
+  @before_signin_message """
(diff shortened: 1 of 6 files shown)


jrissler (approved):

lib/oli_web/pow/messages.ex
```diff
@@ -0,0 +1,41 @@
...
+
+  alias Oli.Accounts
+  alias OliWeb.Router.Helpers, as: Routes
+  alias Phoenix.HTML.Link
+  alias Pow.Phoenix.Messages
+
+  @before_signin_message """
+    We have detected an account using that email was previously created when you accessed the system from your LMS.
+    For your scores to count toward your course, you must return to your LMS and continue access from there.
+    If you wish to create a new account independent of that course
```
bsparks, line 24: seems like it should say "click here to `Create an Account`" or maybe "then you may `Create an Account`" ... anyway, reads weird to me, but I don't know if product already came up with the text

gastonabella: It should, and it's being added a few lines below as a link :)

bsparks: oh I saw that it will read `If you wish to create a new account independent of that course, Create an Account.` I just thought that read a little weird to me, like maybe `If you wish to create a new account independent of that course, you may Create an Account here.` or something like that. but I wasn't sure if this copy text was coming from Product or not, and I'm also not up in arms about it :)

gastonabella: Good point, but that exact text is coming from Product (Anita wrote it in the Jira ticket) so I think we could stick to it now and update in case they want something different, wdyt?

bsparks: sounds good to me :)

bsparks (approved):

PR #2419, opened 2022-04-11
PMP-2514 | JAN-1359

dtiwarATS:


assets/src/adaptivity/scripting.ts
```diff
@@ -133,14 +133,27 @@ export const getExpressionStringForValue = (
...
+          const evaluatedValues = evalScript(expressions[0], env).result;
+          if (evaluatedValues !== undefined) {
+            val = evaluatedValues;
             actuallyAString = false;
           }
         } else {
-          // if we have parsing error then we're guessing it's CSS
-          actuallyAString = true;
+          const containsExpression = val.match(/{([^{^}]+)}/g) || [];
+          if (containsExpression?.length) {
```
dtiwarATS, line 143: So, the scripting is not able to evaluate 

`{{q:1468628289415:435|stage.BMR.value}+({q:1468628289415:435|stage.BMR.value}*{q:1468628289656:443|stage.activityFactor.value})}*0.10,1` because it starts with '{'. We already had a code that checks if first and last brackets are '{}' and then it does a substring and sends the expression to scripting engine. 

but the above won't work that way so my I idea is to use the 'templateText' to replace the variables with  their values and then send them to scripting engine and it works. We can talk about it let today @bsparks

bsparks (approved):

PR #2421, opened 2022-04-11
[MER-936] Add publishers table and provide CRUD operations

gastonabella:
[MER-936](https://eliterate.atlassian.net/browse/MER-936)

### Description
This PR adds a new `publishers` table and define the CRUD operations for this new entity.
It is only accessible by system admins, under the "Content Management" Section in the main admin page.

**NOTE: This PR will be marked as "Ready for Review" and can be merged once it has 2 approvals.**

### DIFF
--- a/test/oli/publishing/publisher_test.exs
+++ b/test/oli/publishing/publisher_test.exs
@@ -8,25 +8,19 @@ defmodule Oli.Publishing.PublisherTest do
 
   describe "changeset/2" do
     test "changeset should be valid with correct attributes" do
-      changeset =
-        build(:publisher)
-        |> Publisher.changeset()
+      changeset = Publisher.changeset(build(:publisher))
 
       assert changeset.valid?
     end
 
     test "changeset should be invalid if name is empty" do
-      changeset =
-        build(:publisher, %{name: ""})
-        |> Publisher.changeset()
+      changeset = Publisher.changeset(build(:publisher, %{name: ""}))
 
       refute changeset.valid?
     end
 
     test "changeset should be invalid if email is empty" do
-      changeset =
-        build(:publisher, %{email: ""})
-        |> Publisher.changeset()
+      changeset = Publisher.changeset(build(:publisher, %{email: ""}))
 
       refute changeset.valid?
     end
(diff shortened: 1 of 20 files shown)


test/oli/publishing/publisher_test.exs
```diff
@@ -0,0 +1,46 @@
...
+
+  import Oli.Factory
+
+  alias Oli.Publishing.Publisher
+  alias Oli.Repo
+
+  describe "changeset/2" do
+    test "changeset should be valid with correct attributes" do
+      changeset =
+        build(:publisher)
```
jrissler, line 12: nit: instead of piping one item in, it would stylistically be better not to use pipe here re: general elixir style guide. (applies to few other lines)

jrissler (approved):
Minus small nit - great job 🚀 🎸

ehilfer:
In its initial state, the publisher entity would be the source of truth of ownership of payments associated with enrollments and be linked to products for that purpose. Later, when we have more sophisticated digital rights management and tracking of ownership fractions in composite projects the publisher would be used to route appropriate fractional payments to the right business entities.   We wanted to get this first increment in place asap so that we can start managing inspark versus openskill payments and be ready for CMU OLI payment distinctions. I imagine that in the future the communities construct may play a role in organizing around publishers but would still need this table to hold the list of product publishers.
The most immediate need is to have a source of truth for the list of publishers and some of their profile information for appearing in invoices or other payment and ordering-related communication.  The Publishers table CRUD operations will be accessible via API for the Magento storefront (other other possible store integrations) to use.
It is important for the Publishers table to reside in torus, so that CMU or other torus implementors have a credible way to provide bookstore fulfillment for multiple or independent content producers and have a torus-based mechanism for attributing ownership and income to the correct parties.  We didn’t want to embed this fully into Magento, since the source of truth should stay within torus for portability of implementation.

mcass19 (approved):
Awesome work ! 🚀 

I agree that maybe is better to move from the `Publishing` context to a new one (and leave that one more for the project publications etc), but not sure how to name it 😅

eliknebel (approved):

PR #2422, opened 2022-04-11
[ENHANCEMENT] [MER-636] New JS app embedding / XSS fix

marc-hughes:
This replaces our usages of our custom oliMountApplication function to bootstrap the JS apps on pages with [react-phoenix] (https://github.com/geolessel/react-phoenix) for a more phoenix-ey way of doing things.

The original motivation behind this change was to fix a XSS bug we had on the ActivityBank page caused by unsafe passing of parameters to the JS app on that page.

By replacing all JS embeds, we get a few other benefits:

- No more special-casing of specific params with varying ways of encoding them
- Ability to embed smaller pieces of react within pages in the future

In regards to the XSS fix, the core mechanism of the passing of params ends up looking like a data-attribute within a tag in the body of the document. The main thing to note there is this happens outside of a script tag, so normal HTML encoding rules apply instead of the (slightly weird and inconsistent) script-block body rules.


assets/src/apps/app.tsx
```diff
@@ -1,76 +1,51 @@
...
+ * that as window.Component.*name*
+ *
+ * @param name Name of the app to register. Is used in both the key on window.Components and the debugging label in react-dev-tools
+ * @param Component Any react component that represents a top-level applicacation
+ */
+export function registerApplication(name: string, Component: React.FunctionComponent<any>) {
+  // TODO - does it really make sense that every one of these gets the same redux store? Might we make that configurable?
+  //        or maybe it belongs up a level in the app component being passed in. One argument for here, is if we ever want
+  //        to share a single redux store across multiple embedded apps. One argument for up a level, is not every embedded
+  //        app might need it, and not every one might need the same reducers.
```
marc-hughes, line 19: @darrensiegel An unresolved question came up during our internal review on what to do with this and wanted to know if you had input. No rush, can wait until you're back.

mcass19 (approved):

jrissler:
IMO we should wait for @darrensiegel  to look at this. I was about to approve it however potentially has implications he may have insight into, that we do not.

assets/src/apps/app.tsx
darrensiegel: At the moment (and when this was all initially built) every "app" needed the same common Redux store.  This was an easy way to allow some common reusable components such as the Media Library to work, out of the box, with any app.  Until some new app has a need for its own custom store, I would not change any of this.

gastonabella (approved):

marc-hughes:
Updated that comment, fixed conflicts, and rebased onto master, I think this one's ready to merge now.

lib/oli_web/router.ex
```diff
@@ -379,7 +379,6 @@ defmodule OliWeb.Router do
 
     # Editors
     get("/:project_id/resource/:revision_slug", ResourceController, :edit)
-    get("/:project_id/resource/:revision_slug/activity/:activity_slug", ActivityController, :edit)
```
bsparks, line 382: why is this removed?

marc-hughes: The /js/activityeditor.js app it used was removed 9 months ago from the build, effectively breaking this page. I'm making the (hopefully reasonable) assumption that none of this is needed anymore and cleaning the template, view, and controller up.

See commit 52e906613add34ab859c1e35dff40fe9a8d350aa

PR #2424, opened 2022-04-12
[FEATURE] [MER-818] Move between accounts

mcass19:
[MER-818](https://eliterate.atlassian.net/browse/MER-818)

### Description
This PR adds a new dropdown in the authoring side for consistency. Having a signout and a way to jump into the linked user account.

Also take the opportunity to refactor the user dropdown to make it less repetitive code and clearer. Based on the logic, I ended with:
- Signout is always shown
- A user can edit its account if is independent learner 
  - Independent instructors are **not** marked as independent learners by default
- A user can see the My Courses if is independent learner or independent instructor
- A user can see the link account if it's not a student or is independent instructor
  - That last part because a user could be potentially both?

^^  Appreciate you let me know your thoughts and if I missed something or you think something is not taken into account.

**NOTE: This PR will be marked as "Ready for Review" and can be merged once it has 2 approvals.**

### DIFF
--- a/assets/styles/authoring/layout/workspace.scss
+++ b/assets/styles/authoring/layout/workspace.scss
@@ -23,7 +23,7 @@ $workspace-sidebar-width: 65px;
     box-shadow: 0 2px 4px 0 rgba($black, 0.15);
 
     .header-bar {
-      height: 40px;
+      height: 2.5em;
       background-color: $workspace-logo-bg;
     }
 
--- a/assets/styles/authoring/layout/workspace.scss
+++ b/assets/styles/authoring/layout/workspace.scss
@@ -23,7 +23,7 @@ $workspace-sidebar-width: 65px;
     box-shadow: 0 2px 4px 0 rgba($black, 0.15);
 
     .header-bar {
-      height: 38px;
+      height: 40px;
       background-color: $workspace-logo-bg;
     }
 
@@ -128,7 +128,6 @@ $workspace-sidebar-width: 65px;
     padding-top: 2px;
     font-size: 0.9em;
     font-weight: 600;
-    margin-right: 10px;
   }
 
   .custom-breadcrumb {
--- a/lib/oli_web/templates/layout/_author_account_dropdown.html.eex
+++ b/lib/oli_web/templates/layout/_author_account_dropdown.html.eex
@@ -3,7 +3,7 @@
     <a class="user nav-link dropdown-toggle" href="#" id="navbarDropdown" role="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
       <div class="block lg:inline-block lg:mt-0 text-grey-darkest no-underline hover:underline mr-4">
         <div class="username"><%= @current_author.name %></div>
-        <div class="role text-right" style="color: <%= author_role_color @current_author %>;"><%= author_role_text @current_author %></div>
+        <div class="role text-right <%= author_role_color @current_author %>"><%= author_role_text @current_author %></div>
       </div>
       <div class="user-icon"><%= author_icon(assigns) %></div>
     </a>
--- a/lib/oli_web/templates/layout/_author_account_dropdown.html.eex
+++ b/lib/oli_web/templates/layout/_author_account_dropdown.html.eex
@@ -1,0 +1,28 @@
+<nav class="navbar p-0 small">
+  <div class="nav-item dropdown form-inline my-2 my-lg-0">
+    <a class="user nav-link dropdown-toggle" href="#" id="navbarDropdown" role="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+      <div class="block lg:inline-block lg:mt-0 text-grey-darkest no-underline hover:underline mr-4">
+        <div class="username"><%= @current_author.name %></div>
+        <div class="role text-right" style="color: <%= author_role_color @current_author %>;"><%= author_role_text @current_author %></div>
+      </div>
+      <div class="user-icon"><%= author_icon(assigns) %></div>
+    </a>
+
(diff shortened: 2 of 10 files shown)


assets/styles/authoring/layout/workspace.scss
```diff
@@ -23,7 +23,7 @@ $workspace-sidebar-width: 65px;
     box-shadow: 0 2px 4px 0 rgba($black, 0.15);
 
     .header-bar {
-      height: 38px;
+      height: 40px;
```
jrissler, line 26: If it's not too much trouble rem would be better for responsiveness 🙉

lib/oli_web/templates/layout/_author_account_dropdown.html.eex
```diff
@@ -0,0 +1,28 @@
+<nav class="navbar p-0 small">
+  <div class="nav-item dropdown form-inline my-2 my-lg-0">
+    <a class="user nav-link dropdown-toggle" href="#" id="navbarDropdown" role="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+      <div class="block lg:inline-block lg:mt-0 text-grey-darkest no-underline hover:underline mr-4">
+        <div class="username"><%= @current_author.name %></div>
+        <div class="role text-right" style="color: <%= author_role_color @current_author %>;"><%= author_role_text @current_author %></div>
```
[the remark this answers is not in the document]
jrissler: +1 ☝️

lib/oli_web/views/authoring_view.ex
```diff
@@ -0,0 +1,51 @@
+defmodule OliWeb.AuthoringView do
+  use OliWeb, :view
+  use Phoenix.Component
+
+  alias Oli.Accounts
+  alias Oli.Accounts.{Author, User}
+
+  def author_role_text(author) do
```
jrissler, line 8: nit: Might be shorter and easier to read/scan - to just use pattern matching here

lib/oli_web/views/authoring_view.ex
```diff
@@ -0,0 +1,51 @@
...
+  def author_account_linked?(%Author{} = author),
+    do: Accounts.get_user_by(author_id: author.id) != nil
+
+  def author_user_associated_email(author) do
+    %User{email: email} = Accounts.get_user_by(author_id: author.id)
+    email
+  end
+
+  def author_icon(%{current_autor: current_autor} = assigns) do
+    case current_autor.picture do
```
jrissler, line 29: misspelling - should be author

jrissler (changes requested):
Great work - love the improvements as well :)

lib/oli_web/views/authoring_view.ex
mcass19: @jrissler Agree but I was thinking that if pattern matching here, I should do it in the author user role which will make this module to depend on the system roles, and if in the future our "admin logic" change, we will need to change it here apart from `Accounts.is_admin?`. Wdyt?

jrissler: I'm thinking even simpler the if statement could even be pattern matched to be cleaner / easier to read.

lib/oli_web/views/layout_view.ex
```diff
@@ -111,41 +120,7 @@ defmodule OliWeb.LayoutView do
   end
 
   def account_link(%{:assigns => assigns} = conn) do
```
gastonabella, line 122: I have three different comments about this function:
1. It used to return a link but now it's a dropdown, so maybe we could change the name accordingly, wdyt?
2. Super nit but we could use the shorthand syntax here (`%{assigns: assigns}`) which I like more 💁🏻‍♂️
3. This could also be a one-line function

gastonabella (commented):
This looks really good 🚀 Only left a minor comment.

Also, what do you think about adding tests for the `authoring_view` as you did for the `layout_view`?

mcass19:
> Also, what do you think about adding tests for the `authoring_view` as you did for the `layout_view`?

@gastonabella The tests added were for the `authoring_view` module (which is what I added in this feature, but agree that all the user side doesn't have tests either). I put the tests in the `layout_view_test` file because is in that view where we use the functions defined in `authoring_view`. 

Let me know if I understood you well 🙏🏼

gastonabella (approved):
> > Also, what do you think about adding tests for the `authoring_view` as you did for the `layout_view`?
> 
> @gastonabella The tests added were for the `authoring_view` module (which is what I added in this feature, but agree that all the user side doesn't have tests either). I put the tests in the `layout_view_test` file because is in that view where we use the functions defined in `authoring_view`.
> 
> Let me know if I understood you well 🙏🏼

I got confused with the names, my bad. About where to put the tests, I think is fine to write them in `layout_view_tests` but generally I try to put them in the module where they are defined instead of where they are used, and I think that's consistent with other tests in Torus.

Anyway, this looks good to me ✅ 🚀

jrissler (approved):

eliknebel:
I'm not sure "Go to My Courses" will always work here, consider the case where a user's linked delivery account is from an LTI launch, there is no way to get to their courses from this link which may cause confusion. I can think of a few solutions here:

1. Hide this link for non-independent accounts
2. Lookup a users institution if they are a non-independent account and show a message that they should login via their LMS

I think #1 is probably they best approach with least amount of friction here.

Also, I'm not sure we can really infer a 1:1 relationship between author accounts and users account. For example, a user may link an single authoring account to many user accounts

mcass19:
Hey @eliknebel ! Thanks for your reply.

You are right, we should hide the `Go to my courses` if the linked user is non-independent. Good catch!

Regarding having multiple users linked to one author, I think that's a bigger decision we need to discuss a little if we want to allow it.

PR #2431, opened 2022-04-14
Feature/MER-1016 hound test setup

marc-hughes:
Adds the ability to write & run hound based integration tests against torus and modifies the github action to run them automatically.

Developer info on how to write & run tests can be found in guides/starting/developer.md

Approach taken:

I didn't want to impose another requirement on developers by default, so one goal was that all current workflows, namely `mix test`, remain unchanged. Therefore, these new tests do not run during a `mix test`.

There is a new task, `mix test.hound` that executes the tests. As mentioned in developer.md, there is a requirement for running a chromedriver process (or another webdriver) while those tests execute. The configuration for this is in a new config environment called hound which is automatically used when you `mix test.hound`

The "Build & Test PR" github workflow has been modified to run these tests.  🚨

Upon a failed integration test step, screenshots of the failed tests will be uploaded to the github action artifacts to help debug why a test failed.

I took the approach of adding it to the end of  the `elixir-build-test` job. A different approach would be to make a separate job for these new tests. That would have the potential benefit of running the unit tests & integration tests simultaneously, but all the setup before those tests would be repeated in each. I'm happy to change it if there are opinions one way or another.

While in the GA workflow config, I noticed we don't cache dependencies. Adding something like this is a potential future improvement to speed up builds by a couple minutes.

```
      - name: 💾 Restore the deps cache
        id: mix-deps-cache
        uses: actions/cache@v3
        with:
          path: |
            deps
            _build
          key: ${{ runner.os }}-mix-${{ hashFiles('**/mix.lock') }}
          restore-keys: |
            ${{ runner.os }}-mix-




      - name: 💾 Restore the node cache
        id: yarn-cache
        uses: actions/cache@v3
        with:
          path: assets/node_modules
          key: ${{ runner.os }}-yarn-${{ hashFiles('assets/yarn.lock') }}
          restore-keys: |
            ${{ runner.os }}-yarn-

```

I believe the tests are currently in a stable state with 10+ consecutive successful builds on my fork, but we'll want to monitor them in case they become flaky over time.

# Additional changes:

I noticed our `yarn install` did not specify `--frozen-lockfile`.  That option is a requirement if you want to be able to reliably reproduce builds without dependency changes that a developer has not approved. The builds now specify this option.

Previously, it was impossible to run a dev server and unit tests on the same machine because the port that the prometheus telemetry listens on was hard coded.  That is now configurable, and the :test and :hound environment are now set to non-default ports.  This makes it possible to run `mix phx.server` `mix test` and `mix test.hound` concurrently on the same machine.


bsparks:
just want to say this is awesome, great work!

eliknebel (approved):
Very cool stuff, looks good!

I think adding the deps caching is a great idea too

jrissler (approved):
Agreed - great start to framework for adding functional tests!

gastonabella (approved):
Great job! 🚀 I only left a minor comment but it looks great

test/support/pages/login_page.ex
```diff
@@ -0,0 +1,36 @@
...
+    end
+  end
+
+  def go_to_author_login do
+    click({:link_text, "Authoring Sign In"})
+  end
+
+  def go_to_educator_login do
+    # TODO - It would be better for these to have explicit id or aria-label attributes instead of relying on the implemention-detail of the url they go to
+    click({:xpath, "//*[@href='/session/new']"})
```
mcass19, line 23: You can also use phoenix paths. eg: `Routes.pow_session_path(OliWeb.Endpoint, :new)`

mcass19 (approved):
Awesome 👏🏼 👏🏼 👏🏼

mcass19 (commented):
Agree 100% that caching dependencies is good and would speed up the build.

In regards of the GA, I'm ok with having it as other step of `elixir-build-test` for now but I was thinking that maybe we can also explore the [matrix](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix) strategy and split by `unit/hound`? (and maybe we can also split the setup and test jobs, and make one dependent of the other?, etc but future stories for that🙂 )

test/test_helper.exs
```diff
@@ -1,3 +1,24 @@
...
+    File.mkdir_p!("test-results/screenshots")
+    Application.ensure_all_started(:ex_machina)
+    Application.ensure_all_started(:hound)
+    ExUnit.start(exclude: [:test], include: [:hound])
+
+    # We can't use manual SQL sandbox mode here, because torus attempts to run
+    # several queries at startup that fail in that mode. However, the tests do
+    # requre the SQL sandbox, so we set up a shared strategy here at the outset.
+    # This means we should refrain from making these tests async for now.
+    # TODO: What's a better way to handle this?
```
mcass19, line 12: Reading a little more the [sandbox docs](https://hexdocs.pm/ecto_sql/Ecto.Adapters.SQL.Sandbox.html#module-collaborating-processes), if we want to make possible to run tests async, we should leave the `manual` mode and `allow` the processes trying to connecto to the db. It might be a bit tricky to do this, so if you need any help let me know.

marc-hughes: Quick note for follow-up: https://github.com/phoenixframework/phoenix_ecto#concurrent-browser-tests

lib/oli/utils/db_seeder.ex
```diff
@@ -1250,4 +1250,28 @@ defmodule Oli.Seeder do
 
     gating_condition
   end
+
+  def create_author_account(), do: create_user_account(SystemRole.role_id().author)
+  def create_admin_account(), do: create_user_account(SystemRole.role_id().admin)
```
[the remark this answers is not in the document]
marc-hughes: Updated to use the factory.

test/support/pages/login_page.ex
marc-hughes: Updated to use Routes, thanks!

PR #2432, opened 2022-04-14
BUGFIX: Mutate state race condition

rmathis:
This changes fixes an issue with two separate calls made to saveActivity causing a race condition where one would overwrite the other. The fix merges the code for the two calls into one.


assets/src/apps/authoring/components/AdaptivityEditor/AdaptivityEditor.tsx
```diff
@@ -100,11 +97,11 @@ export const AdaptivityEditor: React.FC<AdaptivityEditorProps> = () => {
   const handleRuleChange = (rule: any) => {
     const existing = currentActivity?.authoring.rules.find((r: any) => r.id === rule.id);
     const diff = JSON.stringify(rule) !== JSON.stringify(existing);
-    /* console.log('RULE CHANGE: ', {
+    console.log('RULE CHANGE: ', {
```
bsparks, line 100: do we need to leave this one?

bsparks (approved):

PR #2452, opened 2022-04-20
enable env variable to turn off DB logging on dev

bsparks:


jrissler:
What's the use case for this? I feel like you would almost always want to see this locally - but maybe it's noisy for your front-end work?

Minor tweak - using the pipe operator once is generally discouraged - see: https://github.com/christopheradams/elixir_style_guide#avoid-single-pipelines

bsparks:
> What's the use case for this? I feel like you would almost always want to see this locally - but maybe it's noisy for your front-end work?
> 
> Minor tweak - using the pipe operator once is generally discouraged - see: https://github.com/christopheradams/elixir_style_guide#avoid-single-pipelines

yeah I have never cared to see DB queries in the log, very noisy. :)

PR #2462, opened 2022-04-22
[MER-923] Add publisher to products and projects

gastonabella:
[MER-923](https://eliterate.atlassian.net/browse/MER-923)
[MER-1033](https://eliterate.atlassian.net/browse/MER-1033)

This PR makes the following changes:
- Add publisher association to projects and products.
- Make publisher required for projects and products.
- Define a default publisher
- Add default publisher to all existing projects and products
- Add email validation to publishers

**DON'T MERGE THIS PR UNTIL IT HAS AT LEAST TO APPROVALS**

### DIFF
--- a/lib/oli/authoring/course/project.ex
+++ b/lib/oli/authoring/course/project.ex
@@ -31,6 +31,8 @@ defmodule Oli.Authoring.Course.Project do
 
     has_many :publications, Oli.Publishing.Publication
 
+    belongs_to :publisher, Oli.Inventories.Publisher, foreign_key: :publisher_id
+
     field :owner_id, :integer, virtual: true
     field :owner_name, :string, virtual: true
 
@@ -49,9 +51,10 @@ defmodule Oli.Authoring.Course.Project do
       :project_id,
       :visibility,
       :status,
-      :allow_duplication
+      :allow_duplication,
+      :publisher_id
     ])
-    |> validate_required([:title, :version, :family_id])
+    |> validate_required([:title, :version, :family_id, :publisher_id])
     |> Slug.update_never("projects")
   end
 end
--- a/lib/oli/authoring/course/project.ex
+++ b/lib/oli/authoring/course/project.ex
@@ -31,7 +31,7 @@ defmodule Oli.Authoring.Course.Project do
 
     has_many :publications, Oli.Publishing.Publication
 
-    belongs_to :publisher, Oli.Inventories.Publisher, foreign_key: :publisher_id
+    belongs_to :publisher, Oli.Inventories.Publisher
 
     field :owner_id, :integer, virtual: true
     field :owner_name, :string, virtual: true
--- a/lib/oli_web/live/publisher_live/show_view.ex
+++ b/lib/oli_web/live/publisher_live/show_view.ex
@@ -17,6 +17,7 @@ defmodule OliWeb.PublisherLive.ShowView do
   data changeset, :changeset
   data breadcrumbs, :list
   data modal, :any, default: nil
+  data is_default_publisher, :boolean, default: false
 
   def breadcrumb(publisher_id) do
     IndexView.breadcrumb() ++
@@ -39,13 +40,13 @@ defmodule OliWeb.PublisherLive.ShowView do
         publisher ->
           changeset = Inventories.change_publisher(publisher)
 
-          default_publisher? = publisher.name == Inventories.default_publisher_name()
+          is_default_publisher = publisher.name == Inventories.default_publisher_name()
 
           assign(socket,
             publisher: publisher,
             changeset: changeset,
             breadcrumbs: breadcrumb(publisher_id),
-            default_publisher?: default_publisher?
+            is_default_publisher: is_default_publisher
           )
       end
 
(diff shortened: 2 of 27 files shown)


lib/oli/authoring/course/project.ex
```diff
@@ -31,6 +31,8 @@ defmodule Oli.Authoring.Course.Project do
 
     has_many :publications, Oli.Publishing.Publication
 
+    belongs_to :publisher, Oli.Inventories.Publisher, foreign_key: :publisher_id
```
mcass19, line 34: NIT: I think the foreign_key is the default

lib/oli_web/live/publisher_live/show_view.ex
```diff
@@ -39,10 +39,13 @@ defmodule OliWeb.PublisherLive.ShowView do
         publisher ->
           changeset = Inventories.change_publisher(publisher)
 
+          default_publisher? = publisher.name == Inventories.default_publisher_name()
```
mcass19, line 42: NIT: `?` in var name

mcass19 (approved):
🚀 🚀

lib/oli/authoring/course/project.ex
jrissler: 💯

priv/repo/migrations/20220418150726_add_publisher_to_projects_and_products.exs
```diff
@@ -0,0 +1,46 @@
...
+          add :publisher_id, references(:publishers), null: false, default: publisher_id
+        end
+
+        flush()
+
+        #Set default publisher for products
+        from(s in Section, where: s.type == :blueprint)
+        |> Repo.update_all(set: [publisher_id: publisher_id])
+
+      {:error, error} ->
```
jrissler, line 32: Safe to assume this will stop the migration - 99% sure it will lol :)

priv/repo/migrations/20220418150726_add_publisher_to_projects_and_products.exs
```diff
@@ -0,0 +1,46 @@
...
+
+  require Logger
+
+  def up do
+    default_publisher_attrs = %{name: Inventories.default_publisher_name(), email: "publisher@cmu.edu"}
+
+    case Inventories.find_or_create_publisher(default_publisher_attrs) do
+      {:ok, %Publisher{id: publisher_id}} ->
+        alter table(:sections) do
+          add :publisher_id, references(:publishers)
```
jrissler, line 19: should this be null false + defaulted to this publisher as well?

test/oli/inventories_test.exs
```diff
@@ -24,14 +24,45 @@ defmodule Oli.InventoriesTest do
...
+    end
+
+    test "find_or_create_publisher/1 with invalid_data returns error changeset" do
+      params = Map.delete(params_for(:publisher), :email)
+
+      assert {:error, %Ecto.Changeset{}} = Inventories.find_or_create_publisher(params)
+    end
+
+    test "default_publisher_name/0 returns default publisher name" do
+      assert "Torus Publisher" == Inventories.default_publisher_name()
```
jrissler, line 58: Nit: I think the assertion should be revered here 🛩️

test/oli/inventories_test.exs
```diff
@@ -24,14 +24,45 @@ defmodule Oli.InventoriesTest do
...
     end
 
     test "list_publishers/0 returns all the publishers" do
       insert_list(3, :publisher)
 
-      assert 3 = length(Inventories.list_publishers())
+      # There is an existing default publisher
+      assert 4 = length(Inventories.list_publishers())
```
jrissler, line 65: nit: reverse assertion?

jrissler (changes requested):
Great job @gastonabella 🎸 🚀

priv/repo/migrations/20220418150726_add_publisher_to_projects_and_products.exs
gastonabella: We can't add a `null: false` constraint here since we don't want publisher to be required in all sections. We only want that for products (`type == :blueprint`)

priv/repo/migrations/20220418150726_add_publisher_to_projects_and_products.exs
gastonabella: Decided to follow a different approach for the migration code so that it's independent from external modules. Also we're letting it fail when it has to now, so the migration isn't marked as successful in that case and it can be rerun

jrissler (approved):
Looks good after updates- ty

mcass19 (approved):

priv/repo/migrations/20220418150726_add_publisher_to_projects_and_products.exs
```diff
@@ -0,0 +1,52 @@
...
+    end
+
+    alter table(:projects) do
+      add :publisher_id, references(:publishers), null: false, default: publisher_id
+    end
+
+    flush()
+
+    # Set default publisher for products
+    from(s in Section, where: s.type == :blueprint)
```
darrensiegel, line 39: I would think that we need a similar UPDATE query to set the publisher_id for all existing projects.

darrensiegel (changes requested):

priv/repo/migrations/20220418150726_add_publisher_to_projects_and_products.exs
gastonabella: Actually we're doing that above by setting the default value when adding the `publisher_id` to projects.

gastonabella: It affects only existing records. We have the `null: false` constraint in the db and also a schema validation that doesn't allow new projects without a publisher.

darrensiegel: Ok, great.  I was unaware that `add :publisher_id, references(:publishers), null: false, default: publisher_id` in a migration would actually set the values in the existing rows.

darrensiegel (approved):

PR #2472, opened 2022-04-26
[MER-947] Order curriculum items equal to add materials

mcass19:
https://eliterate.atlassian.net/browse/MER-947

### DIFF
--- a/lib/oli_web/live/delivery/remix_section.html.heex
+++ b/lib/oli_web/live/delivery/remix_section.html.heex
@@ -28,7 +28,7 @@
             <p>There's nothing here.</p>
           </div>
         <% end %>
-        <%= for {node, index} <- Enum.with_index(@active.children) |> Enum.filter(fn {c, _i} -> c.uuid != @dragging end) do %>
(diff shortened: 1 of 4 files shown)


lib/oli_web/live/delivery/remix_section.html.heex
```diff
@@ -28,7 +28,7 @@
             <p>There's nothing here.</p>
           </div>
         <% end %>
-        <%= for {node, index} <- Enum.with_index(@active.children) |> Enum.filter(fn {c, _i} -> c.uuid != @dragging end) do %>
+        <%= for {node, index} <- HierarchyPicker.sort_items(@active.children, assigns) |> Enum.with_index() |> Enum.filter(fn {c, _i} -> c.uuid != @dragging end) do %>
```
jrissler, line 31: if possible, I think this would be a great view helper vs doing it inline with the comprehension.

jrissler (approved):
Looks great - nice tests!

lib/oli_web/live/delivery/remix_section.html.heex
mcass19: Makes sense, will do! Thank you 👍🏼

PR #2481, opened 2022-04-29
[ENHANCEMENT] [MER-1016] hound test setup

marc-hughes:
*note* - This is the same PR as https://github.com/Simon-Initiative/oli-torus/pull/2431 - I inadvertently closed it while doing some branch maintenance in my fork.

Adds the ability to write & run [hound](https://github.com/HashNuke/hound) based integration tests against torus and modifies the github action to run them automatically.

Developer info on how to write & run tests can be found in [guides/starting/developer.md](https://github.com/Simon-Initiative/oli-torus/pull/2431/files#diff-acd7e397c17395fd4e36b553f397c9b2b08d09851427bb1995556ea45da548a8)

Approach taken:

I didn't want to impose another requirement on developers by default, so one goal was that all current workflows, namely `mix test`, remain unchanged. Therefore, these new tests do not run during a `mix test`.

There is a new task, `mix test.hound` that executes the tests. As mentioned in developer.md, there is a requirement for running a chromedriver process (or another webdriver) while those tests execute. The configuration for this is in a new config environment called hound which is automatically used when you `mix test.hound`

The "Build & Test PR" github workflow has been modified to run these tests.  🚨

Upon a failed integration test step, screenshots of the failed tests will be uploaded to the github action artifacts to help debug why a test failed.

I took the approach of adding it to the end of  the `elixir-build-test` job. A different approach would be to make a separate job for these new tests. That would have the potential benefit of running the unit tests & integration tests simultaneously, but all the setup before those tests would be repeated in each. I'm happy to change it if there are opinions one way or another.

While in the GA workflow config, I noticed we don't cache dependencies. Adding something like this is a potential future improvement to speed up builds by a couple minutes.

```
      - name: 💾 Restore the deps cache
        id: mix-deps-cache
        uses: actions/cache@v3
        with:
          path: |
            deps
            _build
          key: ${{ runner.os }}-mix-${{ hashFiles('**/mix.lock') }}
          restore-keys: |
            ${{ runner.os }}-mix-




      - name: 💾 Restore the node cache
        id: yarn-cache
        uses: actions/cache@v3
        with:
          path: assets/node_modules
          key: ${{ runner.os }}-yarn-${{ hashFiles('assets/yarn.lock') }}
          restore-keys: |
            ${{ runner.os }}-yarn-

```

I believe the tests are currently in a stable state with 10+ consecutive successful builds on my fork, but we'll want to monitor them in case they become flaky over time.

# Additional changes:

I noticed our `yarn install` did not specify `--frozen-lockfile`.  That option is a requirement if you want to be able to reliably reproduce builds without dependency changes that a developer has not approved. The builds now specify this option.

Previously, it was impossible to run a dev server and unit tests on the same machine because the port that the prometheus telemetry listens on was hard coded.  That is now configurable, and the :test and :hound environment are now set to non-default ports.  This makes it possible to run `mix phx.server` `mix test` and `mix test.hound` concurrently on the same machine.


marc-hughes:
@mcass19 @jrissler @eliknebel @gastonabella - Mind hitting that approve button? It's the same changes as the earlier PR you approved. Thanks!

mcass19 (approved):

gastonabella (approved):

jrissler (approved):

jrissler:
@eliknebel curious what you think?

marc-hughes:
@eliknebel Does your chromedriver & chrome version match? Looks like a mismatch there is the most common source of that type of error.

eliknebel:
Yes, that was it! thanks! Works great

eliknebel (approved):

PR #2482, opened 2022-04-29
[ENHANCEMENT] [MER-636] New JS app embedding / XSS fix

marc-hughes:
*note* - This is the same PR as https://github.com/Simon-Initiative/oli-torus/pull/2422 - I inadvertently closed it while doing some branch maintenance in my fork.

This replaces our usages of our custom oliMountApplication function to bootstrap the JS apps on pages with [react-phoenix] (https://github.com/geolessel/react-phoenix) for a more phoenix-ey way of doing things.

The original motivation behind this change was to fix a XSS bug we had on the ActivityBank page caused by unsafe passing of parameters to the JS app on that page.

By replacing all JS embeds, we get a few other benefits:

- No more special-casing of specific params with varying ways of encoding them
- Ability to embed smaller pieces of react within pages in the future

In regards to the XSS fix, the core mechanism of the passing of params ends up looking like a data-attribute within a tag in the body of the document. The main thing to note there is this happens outside of a script tag, so normal HTML encoding rules apply instead of the (slightly weird and inconsistent) script-block body rules.


marc-hughes:
@bsparks @gastonabella @mcass19 - Mind hitting that approve button? It's the same changes as the earlier approved PR you approved. Thanks!

jrissler (approved):

mcass19 (approved):

eliknebel (approved):

PR #2483, opened 2022-04-29
[MER-1102] Fixing user registration on failed captcha

marc-hughes:
Bug:

1. Go to the new learner/educator registration form /authoring/registration/new
2. Fill in all the fields, but do not do the captcha
3. Click the register button

Expected: You are brought back to the same form with an error about the captcha which then lets you complete the process and create a learner/educator account.

Actual: You are brought to the create/link account form with an error about the captcha.  That form does not contain the "13 years old" checkbox. Completing that form does not create a learner/educator account like you would want.

### DIFF
--- a/lib/oli_web/plugs/registration_captcha.ex
+++ b/lib/oli_web/plugs/registration_captcha.ex
@@ -59,7 +59,7 @@ defmodule Oli.Plugs.RegistrationCaptcha do
     end
   end
 
-  defp render_captcha_error(conn, purpose) do
+  defp render_captcha_error(conn, :register_author) do
     conn =
       conn
       |> OliWeb.Pow.PowHelpers.use_pow_config(:author)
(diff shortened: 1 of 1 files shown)


lib/oli_web/plugs/registration_captcha.ex
```diff
@@ -59,39 +59,55 @@ defmodule Oli.Plugs.RegistrationCaptcha do
     end
   end
 
-  defp render_captcha_error(conn, purpose) do
+  defp render_captcha_error(conn, :register_author) do
```
marc-hughes, line 62: I refactored render_captcha_error to match on the reason instead of having a cond in the body because the `use_pow_config(:author)` part is not actually common to all the cases.

lib/oli_web/plugs/registration_captcha.ex
```diff
@@ -59,39 +59,55 @@ defmodule Oli.Plugs.RegistrationCaptcha do
...
+      |> Ecto.Changeset.apply_action(:insert)
+
+    conn
+    |> assign(:changeset, changeset)
+    |> assign(:action, Routes.pow_registration_path(conn, :create))
+    |> assign(:sign_in_path, Routes.pow_session_path(conn, :new))
+    |> assign(:cancel_path, Routes.delivery_path(conn, :index))
+    |> Phoenix.Controller.put_view(OliWeb.Pow.RegistrationView)
+    |> Phoenix.Controller.render("new.html")
+    |> halt()
```
marc-hughes, line 96: I don't like this bit.

The other 2 cases had a function inside DeliveryController to call, but I didn't see an equivalent place for a non-authoring account. I could go create a new controller with just this function in it if people prefer, but that seemed weird too.

Open to suggestions.

mcass19: I think that DeliveryController is a good place where to define a `render_user_create_form` function since define some other "user functions". With that, we should also change the above's name to include `author` or something like that

mcass19 (approved):
💪🏼 

If you want to move that function it's good but anyway LGTM. 
Thanks for fixing this!!

jrissler:
I'd like to see some automated tests for this - so it doesn't break without us knowing in the future.

marc-hughes:
> I'd like to see some automated tests for this - so it doesn't break without us knowing in the future (again).

Will do. Have to think a bit on how to automate something that's meant to prevent automation.

jrissler:
> > I'd like to see some automated tests for this - so it doesn't break without us knowing in the future (again).
> 
> Will do. Have to think a bit on how to automate something that's meant to prevent automation.

Maybe couple unit tests or something that just reasonably assumes it would work given the element is filled in type of thing, to give us any additional confidence?

marc-hughes:
> I'd like to see some automated tests for this - so it doesn't break without us knowing in the future (again).

Turns out, when you don't over-think it, those are a lot easier to write.

jrissler:
> 

Perfect - easy peasy 😄

jrissler (approved):
Great job

lib/oli_web/plugs/registration_captcha.ex
marc-hughes: That's done now, thanks!

gastonabella (approved):

PR #2489, opened 2022-05-03
PMP-2788 | JAN-1636

dtiwarATS:
The issues was occurring because of the changes made in [PR](https://github.com/Simon-Initiative/oli-torus/pull/2465/files) . Added a separate function to format the numbers with different scenarios

### DIFF
--- a/assets/src/adaptivity/scripting.ts
+++ b/assets/src/adaptivity/scripting.ts
@@ -1,5 +1,5 @@
 import { Environment, Evaluator, Lexer, Parser } from 'janus-script';
-import { parseArray, parseBoolean } from 'utils/common';
+import { formatNumber, parseArray, parseBoolean } from 'utils/common';
 import { CapiVariableTypes, getCapiType } from './capi';
 import { janus_std } from './janus-scripts/builtin_functions';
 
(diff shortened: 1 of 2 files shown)


assets/src/adaptivity/scripting.ts
```diff
@@ -468,6 +468,29 @@ export const extractAllExpressionsFromText = (text: string): string[] => {
   return expressions;
 };
 
+export const formatNumber = (number: number) => {
```
dtiwarATS, line 471: The idea here is to separate the decimal value from the number and handle it.

assets/src/adaptivity/scripting.ts
```diff
@@ -468,6 +468,29 @@ export const extractAllExpressionsFromText = (text: string): string[] => {
...
+    return number;
+  }
+  const decimalNumber = arrNumber[1];
+  const LeadingZerosInNumber = decimalNumber.toString().match(/\b0+/g);
+  let totalLeadingZerosInNumber = 0;
+  if (LeadingZerosInNumber?.length) {
+    totalLeadingZerosInNumber = LeadingZerosInNumber[0].length;
+  }
+
+  const modifiedNumber =
```
dtiwarATS, line 484: The result should be as follows:

12.4567 - 12.46
11.88  - 11.88
0.0000126 - 0.000017 (previously it was rounding of to 0.00)
0.000120 - 0.00012

assets/src/adaptivity/scripting.ts
bsparks, line 471: this should probably be a function in utils/common instead of in here

assets/src/adaptivity/scripting.ts
```diff
@@ -468,6 +468,29 @@ export const extractAllExpressionsFromText = (text: string): string[] => {
...
 };
 
+export const formatNumber = (number: number) => {
+  const arrNumber = number.toString().split('.');
+  const containsDecimal = arrNumber.length > 1;
+  if (!containsDecimal) {
+    return number;
+  }
+  const decimalNumber = arrNumber[1];
+  const LeadingZerosInNumber = decimalNumber.toString().match(/\b0+/g);
```
bsparks, line 478: this variable name should start with lowercase

bsparks (approved):

PR #2501, opened 2022-05-05
PMP-2755 | JAN-1603

dtiwarATS:
We send the environment object as part of  props.onInit result to all the parts but this was not sent when we render the pop-ups. So added the code to send the environment object.


assets/src/components/parts/janus-popup/PopupWindow.tsx
```diff
@@ -1,5 +1,7 @@
+import { defaultGlobalEnv } from 'adaptivity/scripting';
```
bsparks, line 1: we don't want to include the defaultGlobalEnv within one of the part components, it won't be the same anyway, this is why they get passed in the init. We need to either pass in the one from the popup icon, or just create a new one (but I think this one should be preferred to be the same as the popup has)

bsparks (approved):

PR #2508, opened 2022-05-06
[FEATURE] [MER-1092] Suppress Paywall

mcass19:
[MER-1092](https://eliterate.atlassian.net/browse/MER-1092) and [MER-1083](https://eliterate.atlassian.net/browse/MER-1083) 

## What's changed?
- Added the `pay_by_institution` field to sections table
- Updated the logic to take this setting into account when accessing the section being a student
  - With that, did some refactor in the enforce paywall plug
- Add a csv to download enrollments
  - Include the section cost and the applied discounts (if appropiate)
- Add a UI for manage institution and product/institution specific discounts
  - Added the constraint to only have one discount per product/institution

## Questions
- In the enforce plug, I assumed we should be enrolled at that point (when accessing those routes), even via lms or open and free invitation. So if you are not enrolled, instead of going to pay (as it was before, and fail consequently because there wasn't an enrollment) you are going to be redirected to not authorized. Is this ok or I'm missing some case/flow?
  - **UPDATE:** logic seems to be fine
- Does it make sense to have `amount` and `percentage` required in the discount, or should be depending on it's type?
  - **UPDATE:** fields are required based on discount type

### Discount ui example
image

### Enrollments csv example
image


lib/oli_web/controllers/page_delivery_controller.ex
```diff
@@ -681,6 +683,65 @@ defmodule OliWeb.PageDeliveryController do
...
+
+  defp build_enrollments_text(enrollments) do
+    ([["Student name", "Student email", "Enrolled on"]] ++
+      Enum.map(enrollments, fn record -> [record.user.name, record.user.email, record.inserted_at] end))
+    |> CSV.encode()
+    |> Enum.to_list()
+    |> to_string()
+  end
+
+  defp get_discount_string(discount) do
```
gastonabella, line 736: Nit: you could pattern match `discount` instead of using a `case`

lib/oli_web/controllers/page_delivery_controller.ex
```diff
@@ -681,6 +683,65 @@ defmodule OliWeb.PageDeliveryController do
     end
   end
 
+  def export_enrollments(conn, %{"section_slug" => section_slug}) do
+    user = conn.assigns.current_user
+
+    if is_admin?(conn) or ContextRoles.has_role?(user, section_slug, ContextRoles.get_role(:context_instructor)) do
```
gastonabella, line 689: Wdyt about moving this permissions check to a function plug or so?

lib/oli_web/live/products/details/edit.ex
```diff
@@ -74,6 +76,13 @@ defmodule OliWeb.Products.Details.Edit do
               <div><%= error_tag f, :amount %></div>
             </div>
 
+            <%= unless get_field(@changeset, :open_and_free) do %>
+              <div class="custom-control custom-switch" style="width: 200px;">
```
gastonabella, line 80: We could move this style to a css class instead, wdyt?

lib/oli_web/live/products/payments/discounts_form.ex
```diff
@@ -0,0 +1,45 @@
...
+          <Label/>
+          <TextInput class="form-control"/>
+          <ErrorTag/>
+        </Field>
+
+        <Field name={:percentage} class="form-group">
+          <Label/>
+          <NumberInput class="form-control"/>
+          <ErrorTag/>
+        </Field>
```
gastonabella, line 37: Does it make sense to disable one or the other depending on the discount type? Or maybe have only one input field that stands for the corresponding discount based on type.

gastonabella (commented):
Awesome job! I left some comments but looks great so far 🥇 
I think next time I would make it in more than one PR since this includes a lot of different changes

gastonabella (approved):
Thanks for doing the changes 🚀

jrissler:
@darrensiegel I think you've thought through this more than we have, any input on his questions above?

eliknebel (approved):
Changes look good.

- In the enforce plug, I assumed we should be enrolled at that point (when accessing those routes), even via lms or open and free invitation. So if you are not enrolled, instead of going to pay (as it was before, and fail consequently because there wasn't an enrollment) you are going to be redirected to not authorized. Is this ok or I'm missing some case/flow?
    - These changes seem reasonable to me. I can't think of any case where someone must have access to the pay unless they are enrolled in the course

- Does it make sense to have amount and percentage required in the discount, or should be depending on it's type?
    - Preferably, this would only show the input corresponding to the type to simplify the form. I think showing both might be confusing to a user and they may think they *have* to supply both in order to continue

mcass19:
Thanks @eliknebel for the answer !

I rebased the branch with master since this was opened many days ago, and update the discount fields to be required based on its type.

mcass19:
@darrensiegel any input from your side on this? Thank you!

eliknebel:
> @darrensiegel any input from your side on this? Thank you!

Darren had asked me to review this on his behalf, so looks good from our end

jrissler (approved):

PR #2511, opened 2022-05-10
Add missing default value to publishers default

gastonabella:


jrissler (approved):

mcass19 (approved):

darrensiegel (changes requested):
This updated migration won't run on any platform that has already had it applied.

darrensiegel (approved):

PR #2514, opened 2022-05-10
[BUG FIX] MER-1109 recovery

darrensiegel:
This PR adds a level of robustness in some authoring and delivery code to allow adaptive lessons that might have been affected by the "Edits made in a duplicate project affect original" bug. 

To reach this impl, I first placed an adaptive page in this affected state.  To do that, I had to revert locally some code in the `page_editor.ex` module that properly deletes activities.  Then I duplicated a project, and in the duplicate I made changes to an adaptive page that indeed "bled" back to the original.   When I was in that state, I could not access that page from both delivery or authoring.

I then worked through a series errors, and added some filtering by not `nil` steps to avoid handling revisions that either were deleted or were never part of that project.  This allowed the lesson to render to the student and, most importantly, to be launchable in the authoring view. 

I would recommend, with this in place, that affected lessons are reviewed in authoring, and any necessary fixes are made and then deployed out to course sections.


lib/oli/authoring/editing/activity_editor.ex
```diff
@@ -717,6 +717,7 @@ defmodule Oli.Authoring.Editing.ActivityEditor do
       |> Enum.reduce(%{}, fn t, m -> Map.put(m, t.id, t) end)
 
     AuthoringResolver.from_resource_id(project_slug, activity_ids)
+    |> Enum.filter(fn r -> !is_nil(r) end)
```
jrissler, line 720: This would be easier to read as: `Enum.filter(&(not is_nil(&1)))` (may have to tweak syntax)

lib/oli/authoring/editing/page_editor.ex
```diff
@@ -444,6 +444,7 @@ defmodule Oli.Authoring.Editing.PageEditor do
 
       activity_ids ->
         AuthoringResolver.from_resource_id(project_slug, activity_ids)
+        |> Enum.filter(fn r -> !is_nil(r) end)
```
jrissler, line 447: Same as above re: easier to read

lib/oli/delivery/attempts/page_lifecycle/hierarchy.ex
```diff
@@ -83,7 +83,8 @@ defmodule Oli.Delivery.Attempts.PageLifecycle.Hierarchy do
       |> DateTime.truncate(:second)
 
     # Create the activity attempts, in bulk
-    Enum.map(activity_revisions, fn r ->
+    Enum.filter(activity_revisions, fn r -> !is_nil(r) end)
```
jrissler, line 86: I think piping activity_revisions into the filter would make this easier to grok.

lib/oli_web/live/curriculum/entries/learning_summary.ex
```diff
@@ -8,6 +8,7 @@ defmodule OliWeb.Curriculum.LearningSummaryLive do
 
   defp determine_activities(activity_ids, activity_map) do
     Enum.map(activity_ids, fn id -> Map.get(activity_map, id) end)
```
jrissler, line 10: Be great to pipe activity_ids into map.

lib/oli_web/live/curriculum/rollup.ex
```diff
@@ -44,6 +44,7 @@ defmodule OliWeb.Curriculum.Rollup do
 
           partial_activity_map =
             resolved_activities
+            |> Enum.filter(fn r -> !is_nil(r) end)
```
jrissler, line 47: May apply to above - a small private function could save repetition here and make it easier to change in the future.

jrissler (approved):
Looks good - few small readability / scanability improvement ideas

bsparks (approved):

PR #2528, opened 2022-05-12
[MER-1097] - Automation test data setup

marc-hughes:
Here at Argos, we're developing a suite of automated tests using cypress to test lesson content as part of a post-deployment verification. This PR implements a new API that allows those tests to set up and tear down the data needed to run those tests. Since these are deployment-verification tests, we want the ability to run the tests against remote torus instances, including our production instance.

The easiest way to test this new API:
1. Create a new API key via the admin interface [Manage Third-Party API Keys](https://localhost:8043/admin/api_keys)
2. Make sure that key has `Automation Data Setup Enabled` enabled.
3. Take the generated key and base-64 encode it. Make note of this value, you won't see it again.
4. Go to the swagger page and pick the automated test data setup endpoint `/api/v1/docs#/Automated%20Test%20Data%20Setup%20Service/OliWeb.Api.AutomationSetupController.setup`
5. Click the lock icon and paste in the value from step 3
6. Click the Try It option
7. Select the api options and click execute. (You probably want all true values and pick a zip file of an exported project)
image

You should then get back information that an automated test can use to test the uploaded content.

```
{
  "author": {
    "email": "author-e2e-test-1652379065703091800@argos.test",
    "id": 78,
    "password": "k7vcskew@ree$s1pfnzb"
  },
  "educator": {
    "email": "educator-e2e-test-1652379066542668000@argos.test",
    "id": 88,
    "password": "aum6jc$yosbmjl3bpjyc"
  },
  "learner": {
    "email": "learner-e2e-test-1652379067188244800@argos.test",
    "id": 89,
    "password": "r14bwc0drzill#gfq!_6"
  },
  "project": {
    "id": 44,
    "slug": "cc_balancing_chemical_reaction_5odio",
    "title": "CC: Balancing Chemical Reactions 95858"
  },
  "section": {
    "id": 25,
    "slug": "automation_test_section_sasp2"
  },
  "success": true
}
```

Once your tests have completed running, you can use the teardown api endpoint to delete the data that had been created. Since that endpoint deletes data, we tried to be as conservative as possible to avoid deleting any real non-test data, therefore you must supply email & password credentials of each account to delete, and we run a variety of checks against deleted resources to make sure they were initialized by the setup function.


lib/oli/publishing.ex
```diff
@@ -512,6 +512,8 @@ defmodule Oli.Publishing do
       {:error, %Ecto.Changeset{}}
   """
   def delete_publication(%Publication{} = publication) do
```
marc-hughes, line 514: While this function existed prior to this PR, I found no reference to actually using it within the codebase.
It would not work if there were any PublishedResources associated with it, so I extended this function to delete any of those resources as part of deleting a publication

lib/oli_web/controllers/api/automation_setup_controller.ex
```diff
@@ -0,0 +1,332 @@
...
+              educator: format_user(educator, educator_password),
+              learner: format_user(learner, learner_password),
+              project: format_project(project),
+              section: format_section(section),
+              success: true
+            }
+          )
+      end
+    else
+      error(conn, 403, "invalid api key")
```
marc-hughes, line 239: Some of our other API calls send a 400 response here, 403 seemed more semantically correct for an invalid api key.

lib/oli/automation_setup.ex
```diff
@@ -0,0 +1,354 @@
...
+
+  defp validate_user(email, password, user_type, expected_name) do
+    config = OliWeb.Pow.PowHelpers.get_pow_config(user_type)
+
+    case Pow.Operations.get_by([{:name, expected_name}, {:email, email}], config) do
+      nil ->
+        {:error, "User not found"}
+
+      user ->
+        case user.__struct__.verify_password(user, password) do
```
marc-hughes, line 340: I'm not entirely sure this is the best way to validate user credentials. Searching through POW api docs, I found a lot of ways of authenticating a user based on their session or through a plug validating form values, but I came up short on finding a generic email/pass check.

lib/oli/automation_setup.ex
```diff
@@ -0,0 +1,354 @@
...
+      #
+      # Delete all the publications for this project, this will also get the publication_resources
+      Enum.each(project.publications, &Oli.Publishing.delete_publication/1)
+
+      # Delete all the associations between this project and resources
+      Repo.delete_all(from pr in ProjectResource, where: pr.project_id == ^project.id, select: pr)
+
+      # Delete the resources
+      for resource <- project.resources do
+        Repo.delete_all(from r in Oli.Resources.Revision, where: r.resource_id == ^resource.id)
```
marc-hughes, line 265: This seems to be the most potentially dangerous part of this PR. As far as I can tell, the system wasn't really designed to ever delete revisions. With the checks in place to make sure we're only deleting data from objects created through the automation setup api, I think this is safe, but lack some understanding of the overall design to know for sure.

gastonabella: While working on improving performance in torus queries I ran into some issues when querying the revisions table. It might be potentially big, and without an index on `resource_id` it could do a seq scan through the whole table.

lib/oli/automation_setup.ex
```diff
@@ -0,0 +1,354 @@
...
+        Repo.delete(resource)
+      end
+
+      {:ok, _} = Repo.delete(project)
```
gastonabella, line 269: If we want to raise when this action fails, I think it would be cleaner to use `delete!` instead so that we raise an accurate error.

lib/oli/publishing.ex
gastonabella: We could also do this in the db by adding a migration that updates the `publication` association in the `published_resources` table to have the `on_delete: :delete_all` option.

lib/oli/automation_setup.ex
```diff
@@ -0,0 +1,354 @@
...
+
+  defp can_delete_section(%{:title => "Automation test section"} = section) do
+    {:ok, section}
+  end
+
+  defp can_delete_section(_) do
+    {:error, "Can only delete sections with title 'Automation test section'"}
+  end
+
+  defp no_duplicates(project) do
```
mcass19, line 209: NIT: Wdyt about moving all private functions to the bottom?

priv/repo/migrations/20220505114827_automation_setup_scope.exs
```diff
@@ -0,0 +1,13 @@
...
+  use Ecto.Migration
+
+  def change do
+    alter table(:api_keys) do
+      add :automation_setup_enabled, :boolean, default: false, null: false
+    end
+
+    flush()
+
+    execute "UPDATE api_keys SET automation_setup_enabled = false;"
```
gastonabella, line 11: Isn't the `default: false` above doing the same as this line?

lib/oli/automation_setup.ex
```diff
@@ -0,0 +1,354 @@
...
+  alias Lti_1p3.Tool.ContextRoles
+
+  alias Oli.Repo
+  alias Oli.Authoring.Course.Project
+  alias Oli.Authoring.Course.ProjectResource
+
+  import Ecto.Query, warn: false
+  require Logger
+
+  def setup_data(
```
gastonabella, line 15: I wonder if we should enclose this function in a transaction, since any of the steps could fail and the user wouldn't know which records were created and which ones weren't

gastonabella (commented):
Awesome job! 🚀

lib/oli/publishing.ex
mcass19: What if we try to set the [on_delete option](https://hexdocs.pm/ecto/Ecto.Schema.html#has_many/3-options) in the publication table for the published resources ?

lib/oli/automation_setup.ex
```diff
@@ -0,0 +1,354 @@
...
+        project_archive,
+        create_learner?,
+        create_educator?,
+        create_author?,
+        create_section?
+      ) do
+    with {:ok, author, author_password} <-
```
mcass19, line 22: Maybe we can try to make this a transaction? If some of the steps fail, we can rollback changes. Ecto.Multi sounds like a good friend here

mcass19 (commented):
This is great 🚀 🚀 !!!

Just some little things and questions if you want to consider them

lib/oli/automation_setup.ex
marc-hughes: TODO: I think I could get a list of revisions to delete from the PublishedResources before deleting those.

priv/repo/migrations/20220505114827_automation_setup_scope.exs
marc-hughes: I was copying what Oli.Repo.Migrations.RegistrationScope did and assumed it had some purpose. I'll try without to see.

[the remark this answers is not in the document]
marc-hughes: Gaston caught this one too. Didn't seem necessary, but it's what Oli.Repo.Migrations.RegistrationScope had done so I copied it. Will try without.

lib/oli/publishing.ex
marc-hughes: I started down this path but got a `** (Postgrex.Error) ERROR 0A000 (feature_not_supported) cannot alter type of a column used by a view or rule` while running the migration.

Sure enough, there's a materialzed view called "part__mapping" that references these tables.

In the migration, I think I need to drop that view, update the foreign key definition, and then recreate the view to make this work, but I've never dealt with views in postgres and am not entirely sure if that's safe to do. Thoughts on that?

lib/oli/automation_setup.ex
marc-hughes: Without an index on resource_id, this turned out to be trickier than I hoped. I refactored the code to do everything in two delete_all calls instead of a loop like previous. This brings the number of seq-scans of the table from the number of resources to just once, but that's as good as I could get it without that index.

I populated a database with 600 imports of a project. After doing that, the tear-down takes 5-10 seconds which seems reasonable for an automated test api call.

marc-hughes: Looking at priv/repo/migrations/20220317164007_fix_previous_migration.exs, there's also a database trigger on the view which probably needs to be dropped/created as part of this. I think this migration would end up looking something like

```
Oli.Repo.Migrations.FixPreviousMigration.drop_trigger()
Oli.Repo.Migrations.FixPreviousMigration.drop_materialized_view()

drop the old constraint
add the new constraint with the on_delete option

Oli.Repo.Migrations.FixPreviousMigration.user = get_current_db_user()
Oli.Repo.Migrations.FixPreviousMigration.create_materialized_view(user)
Oli.Repo.Migrations.FixPreviousMigration.create_trigger(user)
Oli.Repo.Migrations.FixPreviousMigration.refresh_materialized_view()
```

Does this sound like the right path?

lib/oli/publishing.ex
marc-hughes: Looking at priv/repo/migrations/20220317164007_fix_previous_migration.exs, there's also a database trigger on the view which probably needs to be dropped/created as part of this. I think this migration would end up looking something like

```
Oli.Repo.Migrations.FixPreviousMigration.drop_trigger()
Oli.Repo.Migrations.FixPreviousMigration.drop_materialized_view()

drop the old constraint
add the new constraint with the on_delete option

Oli.Repo.Migrations.FixPreviousMigration.user = get_current_db_user()
Oli.Repo.Migrations.FixPreviousMigration.create_materialized_view(user)
Oli.Repo.Migrations.FixPreviousMigration.create_trigger(user)
Oli.Repo.Migrations.FixPreviousMigration.refresh_materialized_view()
```

Does this sound like the right path?

lib/oli_web/controllers/api/automation_setup_controller.ex
```diff
@@ -0,0 +1,332 @@
...
+            OliWeb.Api.AutomationSetupController.AutomationSetupResponse}
+       }
+  def setup(conn, %{
+        "create_author" => create_author,
+        "create_learner" => create_learner,
+        "create_section" => create_section,
+        "create_educator" => create_educator,
+        "project_archive" => project_archive
+      }) do
+    if is_valid_api_key?(conn, &Oli.Interop.validate_for_automation_setup/1) do
```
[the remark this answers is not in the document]
marc-hughes: Done. I left the other unrelated api's alone, but we may want to move them over too someday.

lib/oli/publishing.ex
gastonabella: I see... I've never dealt with views in postgres either but seems like dropping and recreating the trigger/view is the way to go. I think you could also avoid the drop constraint line if you use the `from` option in [`modify/3`](https://hexdocs.pm/ecto_sql/Ecto.Migration.html#modify/3) (I believe you could do `from: references(:publications)`)

Anyway, I assumed that modifying this field was way easier 😅. I'm ok with leaving it as is if it gets too complicated.

lib/oli/automation_setup.ex
marc-hughes: Got that, and also did project teardown in a transaction.

lib/oli/automation_setup.ex
marc-hughes: Got that, and also did project teardown in a transaction.

lib/oli/publishing.ex
marc-hughes: I put that in and it seems to work well, migration only took a few seconds locally.

https://github.com/Simon-Initiative/oli-torus/pull/2528/files#diff-76f59a86fa8b417205c670a829a79d79fd590797fe67a6b7203df23d96bd04a5

marc-hughes:
@gastonabella @mcass19 Thanks for the reviews, I think I resolved all the outstanding issues.

gastonabella (approved):
LGTM 👍
Thanks for doing the changes!

test/oli_web/controllers/api/automation/automation_setup_controller_test.exs
```diff
@@ -0,0 +1,232 @@
...
+          true -> {:ok, user}
+          false -> {:error, "Credentials didn't match"}
+        end
+    end
+  end
+
+  def create_project(_) do
+    {:ok,
+     %{
+       project: Repo.insert!(Oli.Factory.project_factory())
```
mcass19, line 214: You can import the factory here and call `insert(:project)`, avoiding the call to `Repo.insert`

mcass19 (approved):
Thanks for the changes, great work !!!

test/oli_web/controllers/api/automation/automation_setup_controller_test.exs
marc-hughes: Thanks! I guess it's time for me to actually go read those ExMachina docs 😄

eliknebel (approved):

PR #2532, opened 2022-05-12
[BUG FIX] MER-1125 improved guard to force new revision creation

darrensiegel:
This PR improves the logic to guard against edits made in duplicate projects affecting the upstream original. 

The root cause of these ongoing problems is reliance on the lock acquisition result (acquired or updated) to determine if a new revision is needed.  I believe there are cases with the bulk activity edit endpoint in the adaptive page where this strategy falls down. 

Rather than rework this lock based logic, I am replacing this with a much more straightforward approach.  This PR adds a check to see if any other PublishedResource records have the same `revision_id` as the id of the revision that is about to be edited.  For the project that we are editing within, if the only published resource record that contains this same revision id pertains to the "unpublished" publication, we are safe to proceed with editing the revision in place (assuming that the title hasn't changed or the lock wasn't just acquired).  More importantly, in all other cases, we force both the page and activity editor code to create a new revision.  

Imagine this scenario:

A course project containing one page is duplicated five times.  We would then have several `PublishedResource` records that all have the same `revision_id` for the single page in the course.  If we would attempt to edit this page in any of these projects, the `needs_new_revision_for_edit?` would return `true`, thus forcing a new revision to be created.  The new unit test in this PR exercises this exact scenario, albeit with just one clone. 

I've tested this interactively with the basic and adaptive editor, in addition to the unit test.

### DIFF
--- a/lib/oli/authoring/editing/activity_editor.ex
+++ b/lib/oli/authoring/editing/activity_editor.ex
@@ -302,7 +302,7 @@ defmodule Oli.Authoring.Editing.ActivityEditor do
 
           activity ->
             get_latest_revision(publication.id, activity.id)
-            |> maybe_create_new_revision(publication, activity, author.id, update)
+            |> maybe_create_new_revision(publication, project, activity, author.id, update)
             |> update_revision(update, project.slug)
         end
       end)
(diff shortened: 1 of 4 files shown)


bsparks (approved):

jrissler (approved):
looks good to me 🎸

eliknebel (approved):

PR #2535, opened 2022-05-13
[BUGFIX] hack the part component types onto the window pending a better fix

bsparks:
the screen editor for Feedback and Popups was using a hack on the window object because of the nested web components. this went away when we changed how apps were launched/registered. putting it back for now


ehilfer (approved):
looks good for a patch

PR #2537, opened 2022-05-13
PMP-2791 | JAN-1639

dtiwarATS:


assets/src/adaptivity/scripting.ts
```diff
@@ -576,9 +576,14 @@ export const templatizeText = (
...
     }
     let strValue = stateValue;
-    /* console.log({ strValue, typeOD: typeof stateValue }); */
-
-    if (Array.isArray(stateValue)) {
+    if (
+      vars.length === 1 &&
+      text === `{${vars[0]}}` &&
+      Array.isArray(stateValue) &&
+      typeof stateValue === 'object'
```
bsparks, line 583: I think if Array.isArray is true, then it's going to be object because arrays are objects

bsparks:
please add unit tests because this is difficult to understand

bsparks:
can you please also write a comment somewhere that explains this scenario? I am not following what we're doing here.

bsparks (approved):

PR #2538, opened 2022-05-13
PMP-2799 | JAN-1648

dtiwarATS:


assets/src/components/parts/janus-capi-iframe/ExternalActivity.tsx
```diff
@@ -214,6 +214,28 @@ const ExternalActivity: React.FC<PartComponentProps<CapiIframeModel>> = (props)
           } else {
             collect[key] = value;
           }
+          // if the array contains values like "00.12" then they need special handling as it's not same as "0.12" so
+          //the array needs to be modified from ['00.23'] to ['"00.23"'] otherwise SIM throws an errors.
```
bsparks, line 218: this makes me wonder if there isn't some way to tell that this should be an array of strings rather than an array of numbers wrapped as strings... very difficult to understand the rules and why this isn't a special case. should we be doing some different parsing somewhere else? maybe these values aren't used in trap states, but if we only have this logic in here then it seems like we have different logic in different places... any way we can make this easier to reason about would be good.

bsparks: also we may need to break this out into a function we can unit test

dtiwarATS: Hey @bsparks. I spent some more time looking at the responses from SIM and found another approach to fix this issue. A different PR created - https://github.com/Simon-Initiative/oli-torus/pull/2544

bsparks:
closed in favor of new PR #2544

PR #2544, opened 2022-05-16
PMP-2799 | JAN-1648 | New Approach

dtiwarATS:


assets/src/utils/common.ts
```diff
@@ -274,3 +274,25 @@ export const formatNumber = (number: number) => {
 
   return modifiedNumber;
 };
+
+export const isSpecialArrayString = (value: string) => {
```
bsparks, line 278: hmm since this "special" behavior is specifically related to capi, I think this method should belong in capi.ts ;

bsparks (approved):

PR #2547, opened 2022-05-16
[MER-1072] Wire lesson finish urls from backend

mcass19:
[MER-1072](https://eliterate.atlassian.net/browse/MER-1072)

## What's changed?

The urls that the lessons were redirecting to after finishing them when closing the congrats modal were wrong as they were hardcoded and calculated given the current url of window.

In the use case of reviewing again the lesson and "re-submitting" it, the  `/attempt/${resourceAttemptGuid}` path was being appended twice. 

With this change, the urls are sent from the phoenix app, so no matter where it comes from it will redirect just fine. Also, if tomorrow the path of those route changes, it's not necessary to update anything in the JS code.

### DIFF
--- a/assets/src/apps/delivery/Delivery.tsx
+++ b/assets/src/apps/delivery/Delivery.tsx
@@ -28,6 +28,8 @@ export interface DeliveryProps {
   enableHistory?: boolean;
   activityTypes?: any[];
   graded: boolean;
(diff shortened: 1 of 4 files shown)


bsparks (approved):

jrissler:
@darrensiegel can you take quick peek at this?

PR #2548, opened 2022-05-16
[FEATURE] Basic Likert activity

andersweinstein:
Basic working implementation of Likert series activity as an evaluated activity. Functions as a multi-part multiple choice question presented in tabular form with the common set of choices as the header row. 

Model can hold extra features from likert XML (non-default value for choices, required flag and group id for items) but no use is currently made of these anywhere. 

Still todo -- authoring correct answer/feedback/hints for multiple parts (currently only applies to first part); tweak css styling as desired; delivering hints.


assets/src/components/activities/likert/LikertAuthoring.tsx
```diff
@@ -0,0 +1,109 @@
...
+      <Provider store={store}>
+        <AuthoringElementProvider {...props}>
+          <Likert {...props} />
+        </AuthoringElementProvider>
+      </Provider>,
+      mountPoint,
+    );
+  }
+}
+// eslint-disable-next-line
```
eliknebel, line 107: We should be sparing `// eslint-disable-next-line` and imports should be at the top of the file.

I'm guessing this was copied from an existing reference activity so I'm guessing there was a good reason for it and is fine for now, we can take a task later to clean these up if necessary

eliknebel (approved):
Looks great and functions as expected. The only comment I have is we should display feedback in a block layout (one below another) possibly with their corresponding part.
Screen Shot 2022-05-19 at 11 15 56 AM


Also, adding "Incorrect" feedback does not seem to render in delivery.

Since grading and feedback are not the main intent of this activity and there is still more followup work to be done for authoring targeted feedback, I don't think either of these are show stoppers and these can be fixed later along with those other enhancements.

PR #2549, opened 2022-05-16
[MER-1087-MER-1086] Add section setting for hiding curriculum items numbering

julmontesdeoca:
[MER-1087](https://eliterate.atlassian.net/browse/MER-1087), [MER-1086](https://eliterate.atlassian.net/browse/MER-1086) and [MER-1137](https://eliterate.atlassian.net/browse/MER-1137)

### What does it change?

- Adds a section setting to hide/display curriculum item numbering to students, by default set to true.
   This is both available at the section and product level, having new sections that part from a product, inherit their value.
- Adds new functions to setup a base project with a minimal curriculum, using factories instead of the `db_seeder`

#### New setting to enable/disable numbering visibility 

**At product level: Product overview page**

Screen Shot 2022-05-20 at 10 20 02

**At section level: Section Overview > Edit Section Details > Settings**

Screen Shot 2022-05-20 at 10 20 02

#### Example of section not displaying unit numbers
image 

#### Example of section displaying unit numbers

image

----- 

#### Why a new setup for a base project with resources?

The `db_seeder` provides: `base_project_with_resource`, `base_project_with_resource2`, `base_project_with_resource3` and `base_project_with_resource4` for creating projects with resources.
However, it relies heavily on the `Oli.Delivery` modules, while performing some unnecessary creations and updates of revisions and published resources. 

Creating the test data through factories in this manner, it should be easier to override the default values for the project and sections, while performing half of the queries to the database, and being less coupled to the actual logic of the app :)

----

Previous implementation: setting in the remix section view

~~#### New setting to enable/disable numbering visibility (Customize curriculum view)~~ Previous

~~image~~

### DIFF
--- a/lib/oli_web/controllers/page_delivery_controller.ex
+++ b/lib/oli_web/controllers/page_delivery_controller.ex
@@ -45,6 +45,7 @@ defmodule OliWeb.PageDeliveryController do
               description: section.description,
               section_slug: section_slug,
               hierarchy: hierarchy,
+              display_curriculum_item_numbering: section.display_curriculum_item_numbering,
               preview_mode: true,
               page_link_url: &Routes.page_delivery_path(conn, :page_preview, section_slug, &1),
               container_link_url:
@@ -79,6 +80,7 @@ defmodule OliWeb.PageDeliveryController do
             description: section.description,
             section_slug: section_slug,
             hierarchy: hierarchy,
+            display_curriculum_item_numbering: section.display_curriculum_item_numbering,
             preview_mode: false,
             page_link_url: &Routes.page_delivery_path(conn, :page, section_slug, &1),
             container_link_url: &Routes.page_delivery_path(conn, :container, section_slug, &1)
@@ -154,7 +156,8 @@ defmodule OliWeb.PageDeliveryController do
             container_link_url: container_link_url,
             active_page: nil,
             revision: revision,
-            resource_slug: revision.slug
+            resource_slug: revision.slug,
+            display_curriculum_item_numbering: section.display_curriculum_item_numbering
           )
 
         # Any attempt to render a valid revision that is not container or page gets an error
@@ -298,7 +301,10 @@ defmodule OliWeb.PageDeliveryController do
         page_link_url: &Routes.page_delivery_path(conn, :page_preview, section_slug, &1),
         container_link_url:
           &Routes.page_delivery_path(conn, :container_preview, section_slug, &1),
-        resource_slug: revision.slug
+        resource_slug: revision.slug,
+        container_link_url:
+          &Routes.page_delivery_path(conn, :container_preview, section_slug, &1),
+        display_curriculum_item_numbering: section.display_curriculum_item_numbering
       }
     )
   end
--- a/lib/oli_web/controllers/page_delivery_controller.ex
+++ b/lib/oli_web/controllers/page_delivery_controller.ex
@@ -302,8 +302,6 @@ defmodule OliWeb.PageDeliveryController do
         container_link_url:
           &Routes.page_delivery_path(conn, :container_preview, section_slug, &1),
         resource_slug: revision.slug,
-        container_link_url:
-          &Routes.page_delivery_path(conn, :container_preview, section_slug, &1),
         display_curriculum_item_numbering: section.display_curriculum_item_numbering
       }
     )
--- a/lib/oli_web/live/delivery/remix_section.ex
+++ b/lib/oli_web/live/delivery/remix_section.ex
@@ -25,6 +25,7 @@ defmodule OliWeb.Delivery.RemixSection do
   alias Oli.Publishing
   alias Oli.Publishing.PublishedResource
   alias OliWeb.Sections.Mount
+  alias Oli.Delivery.Sections.Section
 
   def set_breadcrumbs(type, section) do
     type
@@ -179,6 +180,7 @@ defmodule OliWeb.Delivery.RemixSection do
      assign(socket,
        title: "Customize Content",
        section: section,
+       changeset: Section.changeset(section, %{}),
        pinned_project_publications: pinned_project_publications,
        previous_hierarchy: hierarchy,
        hierarchy: hierarchy,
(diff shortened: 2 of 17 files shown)


eliknebel:
I think the functionality here makes sense, but I wonder if the option control would be better placed under Section Overview > Edit Section Details > Settings. I think most users customizing their curriculum probably do not need to see this control and may add confusion

lib/oli_web/controllers/page_delivery_controller.ex
```diff
@@ -298,7 +301,10 @@ defmodule OliWeb.PageDeliveryController do
         page_link_url: &Routes.page_delivery_path(conn, :page_preview, section_slug, &1),
         container_link_url:
           &Routes.page_delivery_path(conn, :container_preview, section_slug, &1),
-        resource_slug: revision.slug
+        resource_slug: revision.slug,
+        container_link_url:
```
julmontesdeoca, line 305: Fixing this that ended up here twice

julmontesdeoca:
> I think the functionality here makes sense, but I wonder if the option control would be better placed under Section Overview > Edit Section Details > Settings. I think most users customizing their curriculum probably do not need to see this control and may add confusion

I get your idea here, wondered the same. 
Although I did went with was described in the Jira ticket [MER-1087](https://eliterate.atlassian.net/browse/MER-1087) 

> preferably within the Content section of the Product Overview page

And thinking about the use case, if an instructor/author wants to hide the numbering while editing, they can do so in one page, without having to go back and forth. What do you think @eliknebel? Would you instead have a content section inside Edit Section Details > Settings ?

eliknebel:
Good point about the product author use case. We will need to add this control in the Product Overview page as well so that it is accessible to product authors. However, we don't want to clutter the customize curriculum page which may be used frequently to manage content with a control that will most likely only be set once for the lifetime of the section/product.

Yes, I think having a Content section under Edit Section Details > Settings makes the most sense. According to the ticket acceptance criteria, it also will need to be added to the Product Overview > Content section as well so that this can be set on products

julmontesdeoca:
Got it, working on it.

eliknebel (approved):

eliknebel:
Nice work

mcass19 (commented):
Excellent job, looks great !!! 🚀 🚀 🚀

Just left really small things that you can address or ignore if you want, but this is great and that project with curriculum using factories is awesome!

lib/oli_web/live/products/details/content.ex
```diff
@@ -3,43 +3,73 @@ defmodule OliWeb.Products.Details.Content do
   alias Surface.Components.Link
   alias OliWeb.Router.Helpers, as: Routes
 
+  import Ecto.Changeset
```
gastonabella, line 6: Nit but we try to group all the `use`, `import` and `alias` together by type, and I think the convention says that's the order too.

gastonabella, line 6: Nit but we try to group all the `use`, `import` and `alias` together by type, and I think the convention says that's the order too.

Edit: as @mcass19 said 😂

lib/oli_web/live/products/details/content.ex
```diff
@@ -3,43 +3,73 @@ defmodule OliWeb.Products.Details.Content do
...
+  alias Surface.Components.Form
```
gastonabella, line 7: Nit: you can group this line and line 3:
```elixir
alias Surface.Components.{Form, Link}
```

lib/oli_web/live/sections/content_settings.ex
```diff
@@ -0,0 +1,28 @@
+defmodule OliWeb.Sections.ContentSettings do
+  use Surface.Component
+
+  alias Surface.Components.Form.{
+    Field,
+    Label,
+    Checkbox,
+  }
+
+  alias OliWeb.Common.Properties.{Group}
```
gastonabella, line 10: Nit: the brackets are not necessary if we only bring one module

test/oli_web/live/products_test.exs
```diff
@@ -0,0 +1,41 @@
+defmodule OliWeb.ProductsLiveTest do
+  use ExUnit.Case
+  use OliWeb.ConnCase
+
+  import Phoenix.LiveViewTest
+  import Oli.Factory
+  alias Oli.Delivery.Sections
```
gastonabella, line 7: Same here

test/oli_web/live/products_test.exs
```diff
@@ -0,0 +1,41 @@
...
+    setup [:admin_conn, :create_product]
+
+    test "save event updates curriculum numbering visibility", %{conn: conn, product: product} do
+      {:ok, view, _html} = live(conn, live_view_details_route(product.slug))
+
+      assert view
+        |> element("#section_display_curriculum_item_numbering")
+        |> render() =~ "checked"
+
+      render_hook(view, "save", %{
```
gastonabella, line 29: Couldn't you call a `render_submit` here instead?

Edit: same as @mcass19 , a `render_click` would also work

test/support/factory.ex
```diff
@@ -272,6 +272,13 @@ defmodule Oli.Factory do
     %Resource{}
   end
 
+  def project_resource_factory() do
+    %ProjectResource{
+      project_id: insert(:project).id,
```
gastonabella, line 277: Super nit but you can also do
```elixir 
project: insert(:project)
```

gastonabella (commented):
Great work! 🚀 👏 
Nits everywhere

lib/oli_web/live/delivery/remix_section.ex
```diff
@@ -179,6 +180,7 @@ defmodule OliWeb.Delivery.RemixSection do
      assign(socket,
        title: "Customize Content",
        section: section,
+       changeset: Section.changeset(section, %{}),
```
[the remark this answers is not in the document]
julmontesdeoca: Oops, leftover from the first implementation. Removing..

test/support/factory.ex
julmontesdeoca: Makes sense! Copied a bad habit from the schema 😅

jrissler (approved):
Great work 🎸 🚀 Fix nits and we're gtg!

julmontesdeoca:
@mcass19 @gastonabella pushed the changes cleaning up the imports and the tests improvements. 
Couldn't mange to do `render_click` on the tests, since the checkboxes have no `phx-click `or `phx-change` events (and it didn't seem worth it to change them, I rather keep the forms). I did use instead render_submit on their respective forms.

test/support/factory.ex
julmontesdeoca: I couldn't manage to do this, the ProjectResource schema is defined as such: 
```
  schema "projects_resources" do
    timestamps(type: :utc_datetime)
    field :project_id, :integer, primary_key: true
    field :resource_id, :integer, primary_key: true
  end
```
Therefore the `ex_machina` factory needs to be defined as:

```
 def project_resource_factory() do
    %ProjectResource{
      project_id: insert(:project).id,
      resource_id: insert(:resource).id
    }
  end
```

And it seems like `ex_machina` doesn't allow for receiving the whole struct and just taking the id.

gastonabella (approved):

mcass19 (approved):

PR #2553, opened 2022-05-17
[BUGFIX][MER-1117] Fix performance issues on remix page

gastonabella:
[MER-1117](https://eliterate.atlassian.net/browse/MER-1117)

### Context
Some users are experiencing long response times when saving an updated curriculum in the Remix Section Page, even sometimes causing a timeout and the materials not being added.

### Problem
The main issue was that when creating new section resources, in order to ensure that these have unique slugs, the logic was querying the `section_resources` table `n` times causing very slow responses.

### Solution
Instead of making `n` queries to ensure the slugs are unique, the logic now creates a set of slug candidates for the section resources that will be created, and makes only one query to reduce this set to only include the candidates that already exist in the db (and therefore aren't unique yet). It repeats this process until all slugs are unique.

This reduced the response time from **~20 seconds** to less than **2 seconds** in some cases.

### More improvements
Besides this performance issue, there are some other parts of the logic that could be improved:
1. [`create_section_resource`](https://github.com/Simon-Initiative/oli-torus/blob/master/lib/oli/delivery/sections.ex#L731) function iterates recursively and calls `Slug.generate/2`, which is making `n` queries to ensure slug uniqueness. Same happens with [`collapse_section_hierarchy`](https://github.com/Simon-Initiative/oli-torus/blob/master/lib/oli/delivery/sections.ex#L1296) function.
2. [`Publishing.get_published_resources_by_publication/2`](https://github.com/Simon-Initiative/oli-torus/blob/master/lib/oli/publishing.ex#L588) is taking `~2s` when running locally, and this is because of the preload clause that is making a separate query for each association. The `revision` association is the slowest and I was able to confirm that it's making a sequential scan through the whole table. Adding an index didn't solve it.

### DIFF
--- a/lib/oli/delivery/sections.ex
+++ b/lib/oli/delivery/sections.ex
@@ -30,6 +30,7 @@ defmodule Oli.Delivery.Sections do
   alias Lti_1p3.Tool.PlatformRoles
   alias Oli.Delivery.Updates.Broadcaster
   alias Oli.Delivery.Sections.EnrollmentBrowseOptions
+  alias Oli.Utils.Slug
 
   require Logger
 
@@ -772,7 +773,7 @@ defmodule Oli.Delivery.Sections do
         numbering_index: numbering_index,
         numbering_level: level,
         children: children,
-        slug: Oli.Utils.Slug.generate(:section_resources, revision.title),
+        slug: Slug.generate(:section_resources, revision.title),
         resource_id: revision.resource_id,
         project_id: publication.project_id,
         section_id: section.id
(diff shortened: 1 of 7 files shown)


lib/oli/delivery/sections.ex
```diff
@@ -1383,6 +1389,76 @@ defmodule Oli.Delivery.Sections do
...
+        title = pr.revision.title
+
+        # if a previous published resource has the same revision then generate a new initial slug different from the default
+        slug_attempt = if Map.has_key?(acc, Slug.slugify(title)), do: 1, else: 0
+        initial_slug = Slug.generate_nth(title, slug_attempt)
+
+        Map.put(acc, initial_slug, pr)
+      end)
+
+    # until all slugs are unique or up to 10 attempts
```
mcass19, line 1408: Should we raise or something if we achieve the 10th attempt and we still have non unique slugs? With current impl we are returning all of them even though we could have records in `prs_by_non_uniq_slug`, or I misunderstood something?

mcass19 (commented):
Great job 🚀 🚀 

Just left a minor question

eliknebel:
Build appears to be failing due to a non-deterministic unit test

gastonabella:
@mcass19 @eliknebel although it is unlikely, I agree we should be covered from that scenario so I just pushed the changes.
Thanks!

mcass19 (approved):

darrensiegel (approved):

PR #2564, opened 2022-05-20
PMP-2738 | JAN-1586

dtiwarATS:


bsparks:
I think we need to figure out a better looksLikeArray, because I think this will break:
image

dtiwarATS:
> I think we need to figure out a better looksLikeArray, because I think this will break: image

Fixed with another approach

assets/src/adaptivity/rules-engine.ts
```diff
@@ -182,6 +183,28 @@ const processRules = (rules: JanusRuleProperties[], env: Environment) => {
...
+          const modifiedTargetValue = `[${targetValue}]`;
+          const updateAttempt = [
+            {
+              target: `${condition.fact}`,
+              operator: '=',
+              value: modifiedTargetValue,
+            },
+          ];
+          // doing this to avoid updating the same fact multiple times if it exists in multiple rules
+          if (modifiedFacts[condition.fact] === undefined) {
```
bsparks, line 202: it shouldn't happen already right? because next time it hits it the targetValue will have the brackets and not enter the condition

bsparks (approved):

PR #2569, opened 2022-05-23
[BUGFIX] update janus scripting to take changes for JAN-1639

bsparks:
just aliasing `Math.max` to `max` and `Math.min` to `min` ... did it at the library level however because janus script does not support rest arguments


bsparks:
actually, nevermind, just thought of a better approach

PR #2575, opened 2022-05-24
PMP-2836 | JAN-1685 | Rounding Support in Text Flow

dtiwarATS:


assets/src/adaptivity/scripting.ts
```diff
@@ -582,23 +582,23 @@ export const templatizeText = (
...
-      } else if (typeof stateValue === 'number') {
-        const modifiedValue = formatNumber(strValue);
-        strValue = parseFloat(modifiedValue.toString());
       }
     } else {
       if (typeof stateValue === 'object' && !Array.isArray(stateValue)) {
         strValue = JSON.stringify(stateValue);
       }
     }
+    if (typeof stateValue === 'number') {
```
bsparks, line 591: this is supposed to be inside the `useFormattedText` we are only doing this for TextFlow

bsparks (approved):

PR #2576, opened 2022-05-24
[BUGFIX][MER-1144] Create Publisher field validation too restrictive

gastonabella:
[MER-1144](https://eliterate.atlassian.net/browse/MER-1144)

Update email regex to be less restrictive and consider emails such as `support@argos.education`.

### DIFF
--- a/lib/oli/utils.ex
+++ b/lib/oli/utils.ex
@@ -4,7 +4,6 @@ defmodule Oli.Utils do
   import Ecto.Changeset
 
   @url_regex ~r/(https?:\/\/(?:www\.|(?!www))[a-zA-Z0-9][a-zA-Z0-9-]+[a-zA-Z0-9]\.[^\s]{2,}|www\.[a-zA-Z0-9][a-zA-Z0-9-]+[a-zA-Z0-9]\.[^\s]{2,}|https?:\/\/(?:www\.|(?!www))[a-zA-Z0-9]+\.[^\s]{2,}|www\.[a-zA-Z0-9]+\.[^\s]{2,})/i
-  @email_regex ~r/^[\w-\.]+@([\w-]+\.)+[\w-]+$/
 
   @doc """
   Generates a random hex string of the given length
@@ -305,9 +304,4 @@ defmodule Oli.Utils do
       "<a href=\"#{absolute_url}\" target=\"_blank\">#{url}</a>"
     end)
   end
-
-  @doc """
-  Returns email regex
-  """
-  def email_regex, do: @email_regex
 end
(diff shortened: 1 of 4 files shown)


lib/oli/utils.ex
```diff
@@ -3,8 +3,8 @@ defmodule Oli.Utils do
 
   import Ecto.Changeset
 
-  @urlRegex ~r/(https?:\/\/(?:www\.|(?!www))[a-zA-Z0-9][a-zA-Z0-9-]+[a-zA-Z0-9]\.[^\s]{2,}|www\.[a-zA-Z0-9][a-zA-Z0-9-]+[a-zA-Z0-9]\.[^\s]{2,}|https?:\/\/(?:www\.|(?!www))[a-zA-Z0-9]+\.[^\s]{2,}|www\.[a-zA-Z0-9]+\.[^\s]{2,})/i
-  @emailRegex ~r/^[\w-\.]+@([\w-]+\.)+[\w-]{2,4}$/
```
marc-hughes, line 7: Not sure the exact requirements here, but a regex for valid RFC 5322 email address validations is a lot more complicated than this.

I think this will allow some addresses that it shouldn't like: `first..last@gmail.com`  (consecutive . are not allowed)
and it will reject some addresses that it should allow like `first+last@gmail.com` (+ should be allowed)

mcass19 (approved):

jrissler:
I'm generally a fan of validating emails by sending them an actual email and having them confirm since its the only sure way to know - there is no prefect regex after years of trying.

I think we should accept this change and re-visit with a much better way throughout the system. The request was pretty simple for this admin functionality change.

eliknebel:
I'd be curious to know what the underlying purpose of capturing an email is here. If it is ever to get in contact with that person, we should be validating with an actual email so that the feature can guarantee delivery in the future.

Debugging email delivery is difficult so storing potentially invalid emails only increases the difficulty. With that said, if we were going to take this as is, I would remove the highly specific regex altogether in favor of a simpler one so that users don't run into the issue of having a valid email that wont validate, for example myemail+modifier@gmail.com is a pretty common use case I think we would want to cover.

gastonabella:
@marc-hughes @jrissler @eliknebel I pushed the updates to have a simpler validation at the schema level so that we don't restrict valid emails, and added some validation in the view. Regarding the purpose of capturing an email here, I don't think there will be a need in the near future to deliver emails but might be worth looking into.

eliknebel (approved):

lib/oli/utils.ex
marc-hughes: Awesome, rejecting valid emails was my primary worry, so sounds good.

PR #2596, opened 2022-05-30
[FEATURE] [MER-729] Improve session timeout

mcass19:
[MER-729](https://eliterate.atlassian.net/browse/MER-729)

### Description

This PR improves the way we handle sessions, as persistent sessions were not working for the delivery side (users) and after 30 minutes, which is pow's default idle time limit, the session is expired. This is mainly because we can have two sessions (user + author) at the same time, and for example in some cases we were trying to get a current user with the author pow config.

### What's changed

- Make keep-alive work not only for authors (we also have delivery routes with live view)
  - **QUESTION:** I moved out the logic from the live template, if doesn't make sense, we can put it back there and have just one keep-alive endpoint (and tweak the plug/checks performed there)
- Make persistent session work properly
  - In some places we are creating the Pow.Plug session manually, so we need to make sure that we create the persistent session accordingly. Moved that logic to a helper (same for delete)
  - Use different persistent session cookie keys for user/author so they are independent.
- Refactor router plugs and pipelines
   - We were repeating unnecessary plugs, making the session not work properly. This change is intended to refactor that a bit, and make sure we always go through the `delivery` or `authoring` plugin on every authenticated route, which will set the correct plug + user configuration

### So how it looks with this?

We will have persistent sessions that work for both author and users independently. That is, that idle time limit is still 30 minutes, but that will reissue a new session token, and the user/author will still be logged in even if they close the tab and returns in some time ([docs](https://hexdocs.pm/pow/Pow.Plug.Session.html#module-session-expiration)). Last, if we are logged in on both sides at the same time and one side is logged out, the other will still be there (different cookies).


gastonabella (approved):
LGTM 🚀 
Great job refactoring the routes!

lib/oli_web/templates/pow/session/author_new.html.eex
```diff
@@ -14,10 +14,6 @@
       <%= error_tag f, :password %>
     </div>
     <div class="mb-4 d-flex flex-row">
-      <div class="custom-control custom-checkbox flex-grow-1">
-        <%= checkbox f, :persistent_session, class: "custom-control-input" %>
-        <%= label f, :persistent_session, "Remember me", class: "custom-control-label" %>
-      </div>
```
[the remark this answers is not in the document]
mcass19: Yes, to be consistent with users coming from lti or cognito which will have persistent session always. 

But is debatable, we could leave that in place for users that do the login through torus UI and make their own decision if they want to persist the session or not

lib/oli_web/templates/layout/authoring.html.leex
```diff
@@ -75,6 +75,42 @@
       <script type="text/javascript" src="<%= Routes.static_path(@conn, "/js/timezone.js") %>"></script>
     <% end %>
     <script type="text/javascript" src="<%= Routes.static_path(@conn, "/js/dark.js") %>"></script>
+
```
eliknebel, line 78: Since we are duplicating this code from another layout template, we should refactor it into it's own ts module, like timezone.js and dark.js above

lib/oli_web/templates/pow/session/author_new.html.eex
```diff
@@ -14,10 +14,6 @@
       <%= error_tag f, :password %>
     </div>
     <div class="mb-4 d-flex flex-row">
-      <div class="custom-control custom-checkbox flex-grow-1">
```
eliknebel, line 17: We should leave this checkbox here for users who may be accessing from a shared/non-secure browser and default to not setting a persistent session

lib/oli_web/templates/pow/session/user_new.html.eex
```diff
@@ -32,10 +32,6 @@
       <%= error_tag f, :password %>
     </div>
     <div class="mb-4 d-flex flex-row">
-      <div class="custom-control custom-checkbox flex-grow-1">
-        <%= checkbox f, :persistent_session, class: "custom-control-input" %>
```
eliknebel, line 36: We should leave this checkbox here for users who may be accessing from a shared/non-secure browser and default to not setting a persistent session

eliknebel (changes requested):
The original auth strategy intentionally required a user to "check-in" with the LMS periodically so that the system can receive updated LTI information from the LMS as well as ensure a user is still enrolled in the section. These changes effectively bypass torus ability to ensure up-to-date LTI details. There might be an alternate way to tackle the issue described, such as present a "reload your browser" page when the session expires in an LTI section inside an iframe, or store LMS details and request the user to re-authenticate/access through their LMS and possibly provide a link to do so.

With that said, the improved user experience of not asking a user to reload the page or re-authenticate might outweigh the downside of possibly out-of-date details, but we should understand and carefully weigh the trade offs here.

jrissler:
@eliknebel That's helpful insight. From my experience with this, it's very normal to have longer than 30 min sessions even for LTI launches (usually they span few hours at a minimum) - since 99.9% of the time, data isn't changing in an LMS and sections aren't continually being shuffled. Even if a student was removed from a section, and continued to do work for a bit in another tool - that's generally not a huge concern.

I don't see many people do it, usually they will just handle any updates on the next LTI launch. We do have things we could do to tackle this scenario in future tickets - such as polling Names and Roles service or other such mechanisms that could effectively keep the section more up-to-date, and have a mechanism for handling a student being removed from that section - even if they are logged in / doing Torus activities.

tl;dr - I think the trade-offs are ok here, and if we think that use case is important - I feel like we should handle that differently outside of this auth mechanism.

lib/oli_web/templates/layout/authoring.html.leex
eliknebel: I see these comments were marked as resolved, but I don't think the latest commits reflect the changes requested

mcass19: Working on them :) (need to rebase and fix some conflicts as well)

eliknebel: Great! Just wanted to make sure they didn't get lost in the dust :)

dsonaty:
Please do NOT merge until 06/10/22 as we are planning to do a code freeze and push to Stage. This needs to be tested and we do NOT have the time to execute it before the code freeze.

jrissler (approved):

PR #2599, opened 2022-05-31
[CHORE] [MER-1089] Dependency cache during builds.

marc-hughes:
This tweaks the GA builds to cache dependencies and not run mix.clean on non-master branches.

Non-scientific test sees build times go from ~14m to ~8m with the dependency caching to ~6m with the mix.clean removal.

image

Removing the mix.clean seemed safe since I don't see where we use any actual artifacts from these builds, but I kept it in place for master branch to be absolutely safe.

We had talked about doing this back in https://github.com/Simon-Initiative/oli-torus/pull/2481 but never got back around to doing it until now.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,6 +28,26 @@ jobs:
       - name: 🗄 Start test database
         run: docker-compose up -d postgres
 
+      - name: 💾 Restore the deps cache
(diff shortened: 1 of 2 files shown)


jrissler (approved):
Looks like it shaved 2.5m off or so, nice 🎸

PR #2602, opened 2022-06-01
[BUGFIX][MER-1203] Institution link must only appear for admins in the sections overview page

gastonabella:
[MER-1203](https://eliterate.atlassian.net/browse/MER-1203)

Before these changes, the institution link was appearing for both admins and instructors in the sections overview page, but only admins are allowed to access to an institution page.

With the new behavior:
- Show the section institution name and a link to the institution page when accessing as an **admin**.
- Show only the section institution name when accessing as an **instructor**.


jrissler (approved):
🎸

PR #2608, opened 2022-06-02
[MER-462] Add product image upload and make it accesible through API

julmontesdeoca:
[MER-462](https://eliterate.atlassian.net/browse/MER-462)

### What does it change?

- Adds an image section in the product overview page, for uploading/updating the product image. At
 this moment, this is only available at the product level. New sections inherit the image from the parting product.
- When a user-uploaded image is provided, it replaces the default `/images/course_default.jpg` image.

#### New product overview section

Default - for all products that have no current image

image

Once an image is uploaded

image

#### Upload example - success

ezgif-5-a21614fd52

#### Upload example - error handling

ezgif-5-a21614fd523

#### /sections as independent learner 

image

#### Enroll page

image

### DIFF
--- a/assets/styles/authoring/image-upload.scss
+++ b/assets/styles/authoring/image-upload.scss
@@ -1,0 +1,37 @@
+@import 'delivery/variables';
+
+.drag-and-drop-zone {
+  background-image: url("data:image/svg+xml,%3csvg width='100%25' height='100%25' xmlns='http://www.w3.org/2000/svg'%3e%3crect width='100%25' height='100%25' fill='none' rx='13' ry='13' stroke='%23C8C8C8FF' stroke-width='3' stroke-dasharray='7%2c 16' stroke-dashoffset='0' stroke-linecap='square'/%3e%3c/svg%3e");
+  border-radius: 13px;
+  background-color: $body-bg;
+}
+
+.on-drag {
+  background-color: darken($color: $body-bg, $amount: 10%);
+}
+
+.btn-tertiary {
+  margin: auto;
+  display: block;
+  width: 100%;
+  height: 100%;
+  cursor: pointer;
+}
+
+.img-input-file {
+	opacity: 0;
+	overflow: hidden;
+	z-index: 1;
+  cursor: pointer;
+  position: absolute;
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 100%;
+}
+
+.input-file-form-group {
+  width: 100%;
+  max-height: 75%;
+  background-color: $body-bg;
+}
--- a/lib/oli_web/common/source_image.ex
+++ b/lib/oli_web/common/source_image.ex
@@ -1,0 +1,14 @@
+defmodule OliWeb.Common.SourceImage do
+  alias OliWeb.Router.Helpers, as: Routes
+
+  @doc """
+    Given a section or source, returns the cover_image for the given element, or a default one.
+  """
+  def cover_image(%{cover_image: url}) when not is_nil(url) do
+      url
+  end
+
+  def cover_image(_) do
+    Routes.static_path(OliWeb.Endpoint, "/images/course_default.jpg")
+  end
+end
--- a/lib/oli_web/live/products/details/image_upload.ex
+++ b/lib/oli_web/live/products/details/image_upload.ex
@@ -31,7 +31,7 @@ defmodule OliWeb.Products.Details.ImageUpload do
                 <div class="col-6 d-flex justify-content-center">
 
                   <div class="form-group input-file-form-group">
-                    <LiveFileInput upload={@uploads.cover_image} class="img-input-file" opts={[disaled: @uploads.cover_image.entries != []]}/>
+                    <LiveFileInput upload={@uploads.cover_image} class="img-input-file"/>
                     <label class="btn btn-dark btn-tertiary js-labelFile">
                       <i class={"#{if @uploads.cover_image.entries != [], do: "fa-check", else: "fa-upload"} icon fa"}></i>
                       <span class="js-fileName">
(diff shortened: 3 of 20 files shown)


eliknebel (approved):
Looks great. We should eventually add this to the course project as well so that products that don't specify a cover image can inherit one specified by the base project.

lib/oli_web/live/products/details/image_upload.ex
```diff
@@ -0,0 +1,128 @@
...
+    Enum.any?(upload.entries, &entry_has_errors?(upload, &1))
+  end
+
+  defp entry_has_errors?(upload, entry) do
+    upload_errors(upload, entry) != []
+  end
+
+  defp upload_error(:too_large), do: "Image too large, try again with a image lower than 5 MB."
+  defp upload_error(:too_many_files), do: "Too many files, try again with a single file"
+  defp upload_error(:not_accepted), do: "Unacceptable file type, try again with a .jpg .jpeg or .png file"
```
gastonabella, line 127: Wdyt about adding one more to match everything in case the error is not in this list?

lib/oli_web/views/delivery_view.ex
```diff
@@ -2,11 +2,14 @@ defmodule OliWeb.DeliveryView do
   use OliWeb, :view
   use Phoenix.Component
 
+  import OliWeb.Common.SourceImage
```
gastonabella, line 5: Is this import necessary?

priv/repo/migrations/20220524233403_update_sections_add_cover_image.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Oli.Repo.Migrations.UpdateSectionsAddCoverImage do
+  use Ecto.Migration
+
+  def change do
+    alter table(:sections) do
+      add :cover_image, :string, null: true
```
gastonabella, line 6: Shouldn't we define a default here?

gastonabella (commented):
Great job! 👏 🚀

lib/oli_web/common/source_image.ex
```diff
@@ -0,0 +1,14 @@
+defmodule OliWeb.Common.SourceImage do
```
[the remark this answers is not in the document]
julmontesdeoca: Yeah, I mostly needed the functionality in a liveview and in a different controller, and wanted to avoid having it twice.

priv/repo/migrations/20220524233403_update_sections_add_cover_image.exs
julmontesdeoca: Yup, adding.

assets/styles/authoring/image-upload.scss
```diff
@@ -0,0 +1,37 @@
...
+.btn-tertiary {
+  margin: auto;
+  display: block;
+  width: 100%;
+  height: 100%;
+  cursor: pointer;
+}
+
+.img-input-file {
+	opacity: 0;
```
mcass19, line 22: NIT: indent

lib/oli_web/views/delivery_view.ex
mcass19: I assumed it is because we are using the functions defined there in some delivery templates, right? I wonder why we add the `Routes` alias below?

priv/repo/migrations/20220524233403_update_sections_add_cover_image.exs
mcass19: If we define a default, does it makes sense the `null: true`?

lib/oli_web/live/products/details_view.ex
```diff
@@ -145,4 +169,44 @@ defmodule OliWeb.Products.DetailsView do
         {:noreply, assign(socket, changeset: changeset)}
     end
   end
+
+  def handle_event("validate_image", _, socket) do
```
mcass19, line 173: I think we are using this event in the component but we are not passing it as an explicit prop, am I right?

lib/oli_web/views/delivery_view.ex
julmontesdeoca: Discussed offline - needed by the enroll template.

lib/oli_web/live/products/details/image_upload.ex
```diff
@@ -0,0 +1,128 @@
...
+          <Form for={@changeset} submit={@upload_event} change="validate_image" opts={[id: "img-upload-form"]}>
+
+          <section>
+            <div id="drag-and-drop-zone" class="drag-and-drop-zone mb-2 py-4 w-75" phx-drop-target={@uploads.cover_image.ref}>
+
+            <div class="row d-flex justify-content-center">
+                <div class="col-6 d-flex justify-content-center">
+
+                  <div class="form-group input-file-form-group">
+                    <LiveFileInput upload={@uploads.cover_image} class="img-input-file" opts={[disaled: @uploads.cover_image.entries != []]}/>
```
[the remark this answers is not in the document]
julmontesdeoca: Thanks for catching this, I'm actually removing it because it makes more sense that people can reselect after having uploaded something, without having to click on the cancel button to re-enable the input.

test/oli_web/live/products_test.exs
```diff
@@ -93,4 +94,92 @@ defmodule OliWeb.ProductsLiveTest do
       assert has_element?(view, "a[href=\"#{Routes.discount_path(OliWeb.Endpoint, :product, product.slug)}\"]")
     end
   end
+
+  describe "product overview image upload" do
+    setup [:admin_conn, :create_product]
+
+    test "displays the current image", %{conn: conn, product: product} do
+      Sections.update_section(product, %{cover_image: "https://example.com/some-image-url.png"})
```
mcass19, line 102: NIT: For this cases, I prefer to not use the `product` that comes in the context and instead insert a new one with the data we need (so we not call app logic here).

test/oli_web/live/products_test.exs
```diff
@@ -93,4 +94,92 @@ defmodule OliWeb.ProductsLiveTest do
...
+      {:ok, view, _html} = live(conn, live_view_details_route(product.slug))
+
+      assert view
+      |> element("#img-preview img")
+      |> render() =~ "src=\"https://example.com/some-image-url.png\""
+
```
mcass19, line 108: NIT: empty line

mcass19 (commented):
Really great job !!! 🚀 💪🏼 

Left some small comments, feel free to just take what you want

priv/repo/migrations/20220524233403_update_sections_add_cover_image.exs
julmontesdeoca: I was going for `null: true, default: null`. Thinking about the new sections parting from products that do not have a `cover_image` at the moment. To allow them but also set them by default

lib/oli_web/live/products/details_view.ex
julmontesdeoca: Yes, `live_file_input` (used in Surface `LiveFileInput`) requires to be in a form with `phx-submit` and `phx-change` events. Even if you do nothing on change.

assets/styles/authoring/image-upload.scss
julmontesdeoca: Could it be a github error? Seeing it fine locally

lib/oli_web/views/delivery_view.ex
julmontesdeoca: Routes alias not needed, removing

test/oli_web/controllers/page_delivery_controller_test.exs
```diff
@@ -678,6 +678,42 @@ defmodule OliWeb.PageDeliveryControllerTest do
       assert html_response(conn, 302) =~
         "You are being <a href=\"#{Routes.payment_path(conn, :guard, section.slug)}\">redirected"
     end
+
```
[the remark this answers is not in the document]
julmontesdeoca: Makes sense, I see now that delivery_controller redirects to the page_delivery_controller only on already rolled users.

test/oli_web/live/products_test.exs
julmontesdeoca: Agree, did think about it when doing tests for `page_delivery_controller` but they had a lot of logic in the creation of the section, besides just inserting into the table. In this case it makes sense because we can replicate the setup easily.

gastonabella (approved):
LGTM 👌

jrissler (approved):
Looks great once all feedback is addressed

PR #2614, opened 2022-06-03
[FEATURE] File Upload activity

darrensiegel:
This PR adds a new activity type, the "File Upload" Activity, which expects a student to upload one or more files as their submission. 

It basically works by allowing files to be uploaded into a folder structure of "section slug" / "activity attempt guid" / "part attempt guid", and recording uploaded file meta data in the part attempt `response` attribute. 

The large bulk of the PR is the client-side impl of the activity, with a new state service endpoint for uploading files. 

Given time constraints, there are a couple of rough edges here that we will eventually want to improve upon. These include:
1. Allowing author to specify a max file size per file (or perhaps as a total quota)
2. An atomic method for uploading files and tracking their state in the student response. Right now these are two separate operations where first the file is uploaded, then an entry for it is saved in the student response. 
3. Removal of files that are deleted prior to submission.  Right now the uploaded file stays in the bucket folder, only the reference to it is removed from the student response in the part attempt.

### DIFF
--- a/assets/src/components/activities/file_upload/utils.ts
+++ b/assets/src/components/activities/file_upload/utils.ts
@@ -1,0 +1,50 @@
+import { DEFAULT_PART_ID } from 'components/activities/common/utils';
+import { FileUploadSchema, FileSpec } from 'components/activities/file_upload/schema';
+import { Responses } from 'data/activities/model/responses';
+import { GradingApproach, makeHint, makeStem, ScoringStrategy } from '../types';
+
+export const defaultModel: () => FileUploadSchema = () => {
+  return {
+    stem: makeStem(''),
+    fileSpec: createDefaultFileSpec(),
+    authoring: {
+      parts: [
+        {
+          id: DEFAULT_PART_ID,
+          scoringStrategy: ScoringStrategy.average,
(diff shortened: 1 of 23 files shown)


assets/src/components/activities/file_upload/utils.ts
```diff
@@ -0,0 +1,50 @@
...
+}
+
+export function toDate(ms: number) {
+  const date = new Date(ms);
+  return date.toLocaleString();
+}
+
+export function getReadableFileSizeString(fileSizeInBytes: number) {
+  let i = -1;
+  const byteUnits = [' kB', ' MB', ' GB', ' TB', 'PB', 'EB', 'ZB', 'YB'];
```
bsparks, line 43: lol, well I feel like > TB is an unlikely scenario :D, actually I hope we don't even reach GB honestly

eliknebel (approved):

assets/src/components/activities/file_upload/utils.ts
darrensiegel: Yes indeed.  I "cloned and owned" this impl from somewhere, and simply left it to be quite scalable! lol

PR #2621, opened 2022-06-06
[BUGFIX] [MER-1155] Improve factories creation

mcass19:
We used to see this error somtimes: [MER-1155](https://eliterate.atlassian.net/browse/MER-1155)

Took the opportunity to follow what `ex_machina` proposes, which is: use [build for associations](https://github.com/thoughtbot/ex_machina#ecto-associations) and [delay the evaluation of attributes](https://github.com/thoughtbot/ex_machina#delayed-evaluation-of-attributes) to avoid falling into the scenario shown by the above issue.


test/support/factory.ex
```diff
@@ -82,8 +82,8 @@ defmodule Oli.Factory do
 
   def community_admin_account_factory() do
     %CommunityAccount{
-      community: insert(:community),
-      author: insert(:author),
+      community: fn -> build(:community) end,
```
jrissler, line 85: I think you could just do `community: build(: community)`

mcass19:
@julmontesdeoca oooh 😱 , I wasn't able to replicate it again after the changes. Let me continue looking then, thanks !!!

test/support/factory.ex
julmontesdeoca: Don't want to answer for Mauricio by I believe the anon function is to force a lazy evaluation, which is the fix suggested by the people at ExMachina. That should prevent associating the same `Community` in each `CommunityAccount` when using `insert_list()`.

julmontesdeoca (approved):
It's working as expected! Thanks for double checking.
As discussed offline it was an issue on my end 🤦‍♂️

jrissler (approved):
Awesome

gastonabella (approved):

test/support/factory.ex
```diff
@@ -472,7 +471,15 @@ defmodule Oli.Factory do
...
       result: :success,
       attempt_number: 1,
-      resource_access: insert(:resource_access)
+      resource_access: anonymous_build(:resource_access)
     }
   end
+
+  # HELPERS
+
+  defp anonymous_build(entity_name),
```
jrissler, line 480: Should we allow a map to be passed in, so specific attributes can be updated (like the original insert/build functions)?

mcass19: Definitely 💯

PR #2623, opened 2022-06-07
PMP-2818 | JAN-1674

dtiwarATS:


assets/src/components/parts/janus-mcq/MultipleChoiceQuestion.tsx
```diff
@@ -32,7 +32,7 @@ const MCQItemContent: React.FC<any> = ({ nodes, state }) => {
   return (
     // Need to set {{ left: 18, position: 'relative' }}. checked it in SS as well. This gets set for all the MCQ and external CSS override change it
     //depending upon their needs
-    <div style={{ left: 18, position: 'relative' }}>
+    <div style={{ left: 18, position: 'relative', overflow: 'hidden' }}>
```
dtiwarATS, line 35: SS applies this on every MCQ. They override it with external CSS if in case this needs to be overridden.

PR #2624, opened 2022-06-07
PMP-2830 | JAN-1679

dtiwarATS:


assets/src/adaptivity/operators/math.ts
```diff
@@ -28,6 +28,7 @@ const operators = {
   notIsExactly: notExactlyMathOperator,
   isEquivalentOf: isEquivalentOfMathOperator,
   notIsEquivalentOf: notIsEquivalentOfMathOperator,
```
dtiwarATS, line 30: There was a type mistake. for precaution, I didn't remove 'notIsEquivalentOf' just in case if it was used any where which is very unlikely.

bsparks (approved):

PR #2641, opened 2022-06-08
PMP-2858 | JAN-1707

dtiwarATS:
@bsparks, we can talk about the approach in our today's TB call.


bsparks:
it isn't variables that would have this, none of them currently have any round function applied to them. I was thinking that we still would want to make it so that in order to round you would do it in printing to text. it would be as if we didn't apply any rounding in the code to 4, except we would leave 4 as the default and IF they added specific rounding to the code for printing, then it would not do the auto rounding to 4, which would allow for the finer control to "override" the default rounding of 4.

bsparks:
close in favor of another approach

PR #2656, opened 2022-06-09
[FEATURE] [MER-1199 MER-1200] Rendering formulas

marc-hughes:
Adds the ability to render formulas within torus pages & activities as first-class objects. 

There are 3 types of formulas:

- MathML
- Latex
- Richtext

The first two are rendered by the mathJax library, the last consists of regular html.

They can be rendered in either block or inline formats.

Example of formulas rendered on torus pages:
image

Example of formulas rendered in activities:
image

This PR does not include authoring (soon to come). I'm attaching a project-export you can ingest to try this out:
[export_formulas_and_links (1).zip](https://github.com/Simon-Initiative/oli-torus/files/8900063/export_formulas_and_links.1.zip)

If you open the authoring environment on a lesson with a formula, you will see an error.
If you open the JSON editor on the admin page, you should be able to edit & save these.


assets/package.json
```diff
@@ -101,6 +102,7 @@
     "@types/react-redux": "^7.1.7",
     "@types/react-transition-group": "~4.4.0",
     "@types/redux-logger": "^3.0.7",
+    "@types/sanitize-html": "^2.6.2",
```
marc-hughes, line 105: Added the sanitize-html npm package.

assets/src/components/common/MathJaxFormula.tsx
```diff
@@ -0,0 +1,87 @@
+import React, { useCallback } from 'react';
+import { sanitizeMathML } from '../../utils/mathmlSanitizer';
+
+/**
```
marc-hughes, line 4: A good place to test that these work is a the hints section. Without the additional rendering logic here, markup appeared instead of the rendered formula when revealing new hints.
hints

assets/src/data/content/model/elements/types.ts
```diff
@@ -78,6 +86,23 @@ export interface ImageInline extends BaseImage {
   type: 'img_inline';
 }
 
+interface FormulaChildren<typeIdentifier>
```
marc-hughes, line 89: Separated out the formula with children (richtext) vs. the formula with a source attribute (mathml / latex) so I could type it that they weren't both allowed at the same time.

bsparks (approved):

lib/oli/rendering/content/html.ex
```diff
@@ -155,6 +157,42 @@ defmodule Oli.Rendering.Content.Html do
...
+    ]
+  end
+
+  def formula(%Oli.Rendering.Context{} = _context, next, _, inline) do
+    # The catch-all formula will handle anything with a children property, which should always be richtext
+    ["<span class=\"#{formula_class(inline)}\">", next.(), "</span>\n"]
+  end
+
+  def formula_inline(context, next, map) do
+    formula(context, next, map, true)
```
darrensiegel, line 193: Delegating to the `formula` impl for inline Latex will end up rendering using the incorrect MathJax escape sequence.   I believe our MathJax is configured to use \\[ and \\] to delimit block level formula and \\( and \\) for inline formula.

assets/src/components/common/MathJaxFormula.tsx
```diff
@@ -0,0 +1,87 @@
...
+      className={cssClass(inline)}
+      ref={ref}
+      dangerouslySetInnerHTML={{ __html: sanitizeMathML(src) }}
+    />
+  );
+};
+
+export const MathJaxLatexFormula: React.FC<MathJaxFormulaProps> = ({ src, inline }) => {
+  const ref = useMathJax(src);
+  const wrapped = inline ? `\\(${src}\\)` : `\\[${src}\\]`;
```
darrensiegel, line 66: This is using the correct escape sequence for inline

darrensiegel (changes requested):

lib/oli/rendering/content/html.ex
marc-hughes: Good catch, new commit up.

image

PR #2657, opened 2022-06-09
[BUG-FIX] [MER-1126] Fix product image breaking card in instructor's new section view

yulnr:
[MER-1126](https://eliterate.atlassian.net/browse/MER-1126)

## What does it change?

- Fixes large vertical product images breaking the product card in the instructors' new section view

### Before fix

image

image

### After fix

Using images: 

https://upload.wikimedia.org/wikipedia/commons/thumb/4/4d/Ma_Lin_010.jpg/640px-Ma_Lin_010.jpg
https://images-na.ssl-images-amazon.com/images/I/91TrrZRURSS.jpg

image

It  now behaves just like the other pages that show the product image did:

image

image

### DIFF
--- a/assets/styles/delivery/layout/delivery.scss
+++ b/assets/styles/delivery/layout/delivery.scss
@@ -14,13 +14,18 @@ body {
 
   .card-title {
     text-overflow: ellipsis;
-    overflow: hidden;
+    overflow: auto;
     white-space: nowrap;
   }
 
   .fade-text {
-    height: 4em;
-    overflow: scroll;
+    overflow: auto;
+    max-height: 10vh;
+    text-overflow: ellipsis;
+    color: $body-color;
+    .card-text {
+      overflow: auto;
+    }
   }
 
   .small-date {
--- a/assets/styles/common/card-link.scss
+++ b/assets/styles/common/card-link.scss
@@ -86,8 +86,8 @@ a.card-link {
 
 .course-card-link {
   img.card-img-top {
-    height: 150px;
-    width: auto;
+    height: 20vh;
+    width: 100%;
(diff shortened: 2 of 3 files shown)


assets/styles/delivery/layout/delivery.scss
```diff
@@ -14,13 +14,18 @@ body {
...
-    overflow: hidden;
+    overflow: auto;
     white-space: nowrap;
   }
 
   .fade-text {
-    height: 4em;
-    overflow: scroll;
+    overflow: auto;
+    max-height: 10vh;
```
jrissler, line 23: is em or vh used more? think we should stick the most commonly used one - either way looks great :)

jrissler (approved):

assets/styles/delivery/layout/delivery.scss
julmontesdeoca: Honestly just thought about setting a max-height instead of a fixed one, and carried over the vh usage from another authoring style. 

I did use `vh` here though, because I also set the image in `vh`. 
Good thing with `vh/wh` in this cases is that it does take in account the side/scroll bars into the viewport total, unlike just setting a percentage. And we have many places (like this and the sections index) that are using scrolling a lot.(which I don't love lol). So I think in this case I would rather go for % instead if not vh.

Ultimately, for those places I believe we need to refactor all the elements into using more relative (at least %) sizing rather than fixed heights/widths and relying in scrolling if content doesn't fit.

Not the most styling-savvy so I can change these into em if it makes more sense :)

mcass19 (approved):

gastonabella (approved):

PR #2676, opened 2022-06-14
[MER-1231] Relax current user and author in authorize section plug

mcass19:
[MER-1231](https://eliterate.atlassian.net/browse/MER-1231)


julmontesdeoca (approved):
🧑‍🚒

PR #2682, opened 2022-06-15
[ENHANCEMENT] orient toolbar vertically

### LINKED ISSUE #2613: Insert content toolbar hides content and gets in the way when editing
**Describe the bug**
While editing a basic page, the toolbar has a tendency to cover up content a sentence or two above the active cursor, which can be annoying for a user who is trying to review what they just authored. We should find a way to make this toolbar less intrusive. I think this is a usability regression of the "always show" toolbar approach from the existing "only show when selected" approach or show after click but hide after keystroke.

We should also add some styling to the toolbar and its menus to make it feel more consistent, like its part of the overall system.

**To Reproduce**
1. Go to basic page
2. Author some paragraph content
3. See toolbar usability issues

**Expected behavior**
The toolbar should get out of your way when you don't need it. We can either take a more "static toolbar" approach that lives at the top of the paragraph but sticky scrolls with the view, or we can only show the toolbar on click at the position of the click, show when text is selected and hide after the first keystroke.

darrensiegel:
Closes #2613 

This PR introduces vertical orientation support for the popup toolbar infrastructure.  It uses this vertical orientation for the core toolbar, so that when it appears it never hides user content. 

I also detected a bug and fixed the ability to show the "context specific" portion of the toolbar.  For Lists, Headings and CodeBlocks the toolbar was supposed to show more buttons for those items.  That now works again. 

Screen Shot 2022-06-15 at 11 43 26 AM

### DIFF
--- a/assets/src/components/editing/editor/Editor.tsx
+++ b/assets/src/components/editing/editor/Editor.tsx
@@ -113,7 +113,11 @@ export const Editor: React.FC<EditorProps> = React.memo((props: EditorProps) => 
       >
         {props.children}
 
-        <EditorToolbar context={props.commandContext} insertOptions={props.toolbarInsertDescs} />
+        <EditorToolbar
+          orientation="vertical"
+          context={props.commandContext}
+          insertOptions={props.toolbarInsertDescs}
+        />
 
         <Editable
           style={props.style}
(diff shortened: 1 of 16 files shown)


eliknebel (approved):

eliknebel:
This is a huge usability improvement

PR #2689, opened 2022-06-16
PMP-2871 | JAN-1720

dtiwarATS:


assets/src/apps/delivery/components/ActivityRenderer.tsx
```diff
@@ -395,12 +395,21 @@ const ActivityRenderer: React.FC<ActivityRendererProps> = ({
...
+    let eventsToProcess = events.results;
     const actionsByType: any = {
       feedback: [],
       mutateState: [],
       navigation: [],
     };
-    events.results.forEach((evt: any) => {
+    const currentActivity = currentActivityTree[currentActivityTree.length - 1];
+    const combineFeedback = !!currentActivity?.content?.custom?.combineFeedback;
+    if (!combineFeedback) {
```
dtiwarATS, line 409: If 'combineFeedback'  is false then we process only the first result so we need to check if the first one has navigation to different screen. Currently we were checking in all the rules so if we get multiple rules in result, even though we were processing only the first one, here hasNavigation was getting set to true which shouldn't be the case.

PR #2695, opened 2022-06-17
[MER-1223] Improve publishing a project waiting time

julmontesdeoca:
[MER-1223](https://eliterate.atlassian.net/browse/MER-1223)

### What does it change

- Reduces significantly the user wait time when publishing a project. The triggers that updated the `part_mapping` materialized view are dropped, in favor of running them asynchronously through spawning in elixir.
 
This does not completely fix the performance issue in regards to scaling optimizations, since the update will still occur and take the same database execution time. However we decided to fix that separately by tackling how we handle the revision content. 
The user wait time for a publication of a single new graded page goes down from ~ 4000ms to ~ 5ms when ran locally.

### DIFF
--- a/lib/oli/publishing.ex
+++ b/lib/oli/publishing.ex
@@ -9,7 +9,7 @@ defmodule Oli.Publishing do
   alias Oli.Delivery.Sections.Section
   alias Oli.Authoring.Course.ProjectResource
   alias Oli.Resources.{Revision, ResourceType}
-  alias Oli.Publishing.{Publication, PublishedResource}
+  alias Oli.Publishing.{Publication, PublishedResource, PartMappingRefreshAdapter, PartMappingRefreshAsync}
   alias Oli.Institutions.Institution
   alias Oli.Authoring.Clone
   alias Oli.Publishing
@@ -529,7 +529,7 @@ defmodule Oli.Publishing do
     publication
       |> Publication.changeset(attrs)
       |> Repo.update()
-      |> maybe_spawn_view_refresh()
+      |> refresh_adapter().maybe_refresh_part_mapping()
   end
 
   @doc """
@@ -542,7 +542,7 @@ defmodule Oli.Publishing do
   """
   def delete_publication(%Publication{} = publication) do
     Repo.delete(publication)
-      |> maybe_spawn_view_refresh()
+      |> refresh_adapter().maybe_refresh_part_mapping()
   end
 
   def get_published_objective_details(publication_id) do
@@ -1131,15 +1131,18 @@ defmodule Oli.Publishing do
     Enum.map(results, fn [slug, title] -> %{slug: slug, title: title} end)
   end
 
-  defp maybe_spawn_view_refresh({:ok, _} = result) do
-    spawn(fn -> refresh_part_mapping() end)
-    result
-  end
-
-  defp maybe_spawn_view_refresh(result), do: result
-
-  defp refresh_part_mapping() do
+  @doc """
+    Refreshes the part_mapping materialized view.
+    Since this operation is expensive, do not use perform it synchrnously unless neccesary.
+  """
+  def refresh_part_mapping() do
     Repo.query("REFRESH MATERIALIZED VIEW CONCURRENTLY part_mapping")
     Logger.info("Refreshed part_mapping view")
+  end
+
+  @spec refresh_adapter() :: PartMappingRefreshAdapter
+  defp refresh_adapter() do
+    Application.get_env(:oli, Oli.Publishing)
+      |> Keyword.get(:refresh_adapter, PartMappingRefreshAsync)
   end
 end
--- a/lib/oli/publishing.ex
+++ b/lib/oli/publishing.ex
@@ -1133,7 +1133,7 @@ defmodule Oli.Publishing do
 
   @doc """
     Refreshes the part_mapping materialized view.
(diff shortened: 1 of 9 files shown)


lib/oli/utils/db_seeder.ex
```diff
@@ -729,12 +729,6 @@ defmodule Oli.Seeder do
       %Publication{published: nil} = p ->
         Oli.Publishing.update_publication(p, %{published: DateTime.utc_now()})
     end
-
```
julmontesdeoca, line 732: This is unnecessary now, and I believe was also unnecessary before (view was refreshed by the trigger regardless)

lib/oli/utils/db_seeder.ex
```diff
@@ -729,12 +729,6 @@ defmodule Oli.Seeder do
...
-    query = """
-    REFRESH MATERIALIZED VIEW part_mapping;
-    """
-
-    Oli.Repo.query!(query, [])
```
julmontesdeoca, line 737: Removed this because it's unnecessary now, and I believe was also unnecessary before (view was refreshed by the trigger regardless)

test/support/part_mapping_mock_adapter.ex
```diff
@@ -0,0 +1,20 @@
...
+  """
+  @impl Oli.Publishing.PartMappingRefreshAdapter
+  @spec maybe_refresh_part_mapping({:ok, %Publication{}} | {:error, %Ecto.Changeset{}}) ::
+          {:ok, %Publication{}} | {:error, %Ecto.Changeset{}}
+  def maybe_refresh_part_mapping({:ok, _publication} = operation_result) do
+    Oli.Publishing.refresh_part_mapping()
+    operation_result
+  end
+
+  @impl Oli.Publishing.PartMappingRefreshAdapter
```
mcass19, line 18: NIT: think this `@impl` is not necessary

lib/oli/publishing/part_mapping_refresh_adapter.ex
```diff
@@ -0,0 +1,13 @@
...
+  alias Oli.Publishing.Publication
+
+  @moduledoc """
+    Behaviour for spawning a function that freshes the part_mapping materialized view:
+  """
+
+  @doc """
+  Defines a callback to be used by refresh adapters
+  """
+  @callback maybe_refresh_part_mapping({:ok, %Publication{}} | {:error, %Ecto.Changeset{}}) ::
```
mcass19, line 11: NIT: maybe we can declare some type with `@type` to avoid writing this all the times, wdyt?

mcass19 (approved):
🚀 🚀  great job !!!

lib/oli/publishing/part_mapping_refresh_adapter.ex
julmontesdeoca: Gotcha. Let me know if there's any other way to define and import `@types` from what I did.

lib/oli/publishing.ex
```diff
@@ -539,6 +542,7 @@ defmodule Oli.Publishing do
   """
   def delete_publication(%Publication{} = publication) do
     Repo.delete(publication)
+      |> refresh_adapter().maybe_refresh_part_mapping()
```
jrissler, line 545: Lets chain publication into delete for readability

lib/oli/publishing.ex
```diff
@@ -1126,4 +1130,19 @@ defmodule Oli.Publishing do
...
     Enum.map(results, fn [slug, title] -> %{slug: slug, title: title} end)
   end
+
+  @doc """
+    Refreshes the part_mapping materialized view.
+    Since this operation is expensive, do not use perform it synchronously unless neccesary.
+  """
+  def refresh_part_mapping() do
+    Repo.query("REFRESH MATERIALIZED VIEW CONCURRENTLY part_mapping")
+    Logger.info("Refreshed part_mapping view")
```
jrissler, line 1140: I'm not sure we need to log this - seems like extra noise, WDYT?

lib/oli/publishing.ex
```diff
@@ -1126,4 +1130,19 @@ defmodule Oli.Publishing do
...
+  end
+
+  @spec refresh_adapter() :: PartMappingRefreshAdapter
+  defp refresh_adapter() do
+    Application.get_env(:oli, Oli.Publishing)
```
jrissler, line 1145: Lets chain :oli into this for readability

test/oli/publishing_test.exs
```diff
@@ -361,6 +365,18 @@ defmodule Oli.PublishingTest do
       assert published_revision.content == revision_with_activity.content
     end
 
+    test "publish_project/1 refreshes part_mapping materialized view with published information" do
+      %{activity: %{revision: revision}, project: project} = project_with_activity()
+
+      capture_log(fn ->
```
jrissler, line 371: If we remove logging here, wlll have to update this test.

jrissler (approved):
Great job @julmontesdeoca

darrensiegel (approved):

lib/oli/publishing.ex
julmontesdeoca: No strong opinion - used it for debugging mostly and figured we could keep it to see the periodicity with which this is being run in prod. Now, having AppSignal setup we can totally remove this.

gastonabella (approved):

PR #2700, opened 2022-06-20
[BUGFIX] make sure that references in layers are required for eval; JAN-1713

bsparks:


rmathis (approved):
Tested and this works for me

dsuriano (approved):

PR #2718, opened 2022-06-22
[MER-937] Improve timezones handling

gastonabella:
[MER-937](https://eliterate.atlassian.net/browse/MER-937)

### Context
Currently, there are different sources of truth in Torus for timezones (browser timezone, section timezone, institution timezone). Different datetimes might be displayed using different timezones depending on the page in Torus. This is causing confusion to some users since it is not clear how this setting is leveraged in the system today.

### Description
This PR adds the following changes:
- Add a new timezone setting to the user/author dropdown in the header, to update the timezone globally for the users during the session. (see screenshot 1)
- Update every page in Torus that renders a datetime, to properly use this global timezone. (see screenshot 1)
- Disambiguate the source of truth for timezones used when handling sections (remove timezones column from `sections` table). (see screenshot 2).
- Display the timezone as part of the rendered date for crucial datetimes (assignment due dates, section start end dates, etc) (see screenshot 3).

**1)**

https://user-images.githubusercontent.com/26532202/175106327-e8866491-c001-451d-b360-6317fff3303f.mov

**2)**
Screen Shot 2022-06-22 at 3 00 51 PM

**3)**
Screen Shot 2022-06-22 at 3 52 54 PM

### Notes
- I know this PR is big and not easy to review (even though most of the changes are similar and repetitive), but I think it makes sense to fix the timezones throughout the whole app at once, instead of having steps in between where the app is inconsistent and shows dates in different timezones.
- There are many improvements that can be done, but decided to do them in a separate story.

### Questions
- Both `institutions` and `pending_registrations` tables have a `timezone` field that is only displayed but not used in the logic. Should we remove them too?

### DIFF
--- a/lib/oli_web/common/select_timezone.ex
+++ b/lib/oli_web/common/select_timezone.ex
@@ -19,7 +19,7 @@ defmodule OliWeb.Common.SelectTimezone do
       <%= form_for @conn, Routes.static_page_path(@conn, :update_timezone), [id: "timezone-form"], fn f -> %>
         <%= hidden_input f, :redirect_to, id: "hidden-redirect-to" %>
         <div class="form-label-group">
-          <%= select f, :timezone, Predefined.timezones(), onchange: "submitForm()", selected: @selected || "Etc/Greenwich", prompt: "Select Timezone", class: "form-control dropdown-select", required: true %>
+          <%= select f, :timezone, Predefined.timezones(), onchange: "submitForm()", selected: @selected || "Etc/Greenwich", class: "form-control dropdown-select", required: true %>
         </div>
       <% end %>
     """
--- a/lib/oli_web/common/select_timezone.ex
+++ b/lib/oli_web/common/select_timezone.ex
@@ -8,10 +8,18 @@ defmodule OliWeb.Common.SelectTimezone do
 
   def render(assigns) do
     ~H"""
-      <%= form_for @conn, Routes.static_page_path(@conn, :update_timezone), fn f -> %>
-        <%= hidden_input f, :redirect_to, value: @conn.request_path %>
+      <script>
+        function submitForm(){
+          const relativePath = window.location.pathname+window.location.search;
+          $('#hidden-redirect-to').val(relativePath);
+          $('#timezone-form').submit()
+        }
+      </script>
+
+      <%= form_for @conn, Routes.static_page_path(@conn, :update_timezone), [id: "timezone-form"], fn f -> %>
+        <%= hidden_input f, :redirect_to, id: "hidden-redirect-to" %>
         <div class="form-label-group">
-          <%= select f, :timezone, Predefined.timezones(), onchange: "this.form.submit()", selected: @selected || "Etc/Greenwich", prompt: "Select Timezone", class: "form-control dropdown-select", required: true %>
+          <%= select f, :timezone, Predefined.timezones(), onchange: "submitForm()", selected: @selected || "Etc/Greenwich", prompt: "Select Timezone", class: "form-control dropdown-select", required: true %>
         </div>
       <% end %>
     """
--- a/lib/oli_web/common/select_timezone.ex
+++ b/lib/oli_web/common/select_timezone.ex
@@ -1,0 +1,19 @@
+defmodule OliWeb.Common.SelectTimezone do
+  use Phoenix.Component
+
+  import Phoenix.HTML.Form
+
+  alias Oli.Predefined
+  alias OliWeb.Router.Helpers, as: Routes
+
+  def render(assigns) do
+    ~H"""
+      <%= form_for @conn, Routes.static_page_path(@conn, :update_timezone), fn f -> %>
+        <%= hidden_input f, :redirect_to, value: @conn.request_path %>
+        <div class="form-label-group">
+          <%= select f, :timezone, Predefined.timezones(), onchange: "this.form.submit()", selected: @selected || "Etc/Greenwich", prompt: "Select Timezone", class: "form-control dropdown-select", required: true %>
+        </div>
+      <% end %>
+    """
+  end
+end
--- a/lib/oli_web/controllers/page_delivery_controller.ex
+++ b/lib/oli_web/controllers/page_delivery_controller.ex
@@ -22,6 +22,7 @@ defmodule OliWeb.PageDeliveryController do
   alias Oli.Publishing.DeliveryResolver, as: Resolver
   alias Oli.Resources.Revision
   alias Oli.Utils.BibUtils
+  alias OliWeb.Common.SessionContext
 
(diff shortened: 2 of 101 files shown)


lib/oli_web/common/select_timezone.ex
```diff
@@ -0,0 +1,27 @@
...
+
+  alias Oli.Predefined
+  alias OliWeb.Router.Helpers, as: Routes
+
+  def render(assigns) do
+    ~H"""
+      <script>
+        function submitForm(){
+          const relativePath = window.location.pathname+window.location.search;
+          $('#hidden-redirect-to').val(relativePath);
```
eliknebel, line 14: This redirect mechanism seems to work great for most pages. I did however run into one instance where is doesn't, and that is on the Create Section page for open and free which has a `source_id` param `https://localhost/admin/open_and_free/new?source_id=publication%3A3`. We'll want to make sure the hidden redirect to path includes any parameters that were in the original url

lib/oli_web/common/select_timezone.ex
```diff
@@ -0,0 +1,27 @@
...
+          $('#timezone-form').submit()
+        }
+      </script>
+
+      <%= form_for @conn, Routes.static_page_path(@conn, :update_timezone), [id: "timezone-form"], fn f -> %>
```
eliknebel, line 19: I think we should display somewhere here what the system has determined the timezone to be. Perhaps instead of the default "Select Timezone" we indicate the timezone is determined from the browser and render the automatically determined timezone as the default. e.g. "Browser: America/New_York". What do you think?

Screen Shot 2022-06-22 at 2 49 36 PM

eliknebel (changes requested):
Looks great, nice work. I just had a couple comments

eliknebel:
It might make sense to special case the rendering of the default timezone UTC to *always* render the UTC identifier, for the case when the system fails to get timezone information from the browser, we want to make sure a user understands the time being displayed is UTC and not assume that it is in their local timezone.

eliknebel:
When trying to create a new section via open and free, I am seeing the following error:

Screen Shot 2022-06-22 at 3 40 08 PM
Screen Shot 2022-06-22 at 3 39 25 PM

lib/oli_web/common/select_timezone.ex
gastonabella: Could you detail a little bit more what you see when updating the timezone for this case? This scenario works correctly in my local, and the search params are included in the refreshed url because of `window.location.search`.

lib/oli_web/common/select_timezone.ex
gastonabella: Regarding the default "Select Timezone" option that is causing the issue that you found, I forgot to remove that prompt from the options (good catch!), and it is not expected to return to the default behavior of determining my timezone automatically.
Currently this is how it works:
-  If the timezone is not set (`local_tz == nil` in the session), then the `timezone.ts` script is run to get the timezone from the browser and update it in the session.
- Once the timezone is set, the script is no longer run.
- If a user updates the timezone with this new feature, the `local_tz` is updated in the session so the original browser timezone is lost.
- If for some reason the timezone cannot be fetched from the browser, the system uses UTC by default.

Therefore, there is no way now to track if the timezone was obtained from the browser or not. We could store a map in the session instead like `%{timezone: "America/New_York", browser: true}` if we want to distinguish one timezone from the other. Anyway, is it necessary to make this distinction? I assumed once a user changed the default browser timezone to a new one, there was no need to go back.

eliknebel: I see, I remember now we originally discussed this option would only be for the current session. I still think users will eventually expect this to behave like a preference, but that would require storing in the user record. So in order to restore auto selection, one just has to clear their session.

I think this is okay for now and if needed we can followup with a more user preference focused approach that stores the selection in the db.

lib/oli_web/common/select_timezone.ex
eliknebel: This seems to be resolved with the latest changes. I think it may have been related to the empty "Select timezone" timezone selection issue

eliknebel (approved):

lib/oli_web/common/select_timezone.ex
gastonabella: I agree this should behave like a preference, but wanted to keep it simple for this first version. I'll create a followup story to store it in the db for the user. Thanks!

lib/oli_web/templates/project/publish.html.eex
```diff
@@ -10,11 +10,11 @@
           </div>
         <% {false, _} -> %>
           <div class="my-3">
-            Last published <strong><%= @latest_published_publication.published |> date(@conn) %></strong>.
+            Last published <strong><%= Utils.render_date(@latest_published_publication, :published, @context) %></strong>.
             There are <strong>no changes</strong> since the latest publication.
           </div>
         <% {true, changes} -> %>
-          <div class="my-3">Last published <strong><%= @latest_published_publication.published |> date(@conn) %></strong>.
+          <div class="my-3">Last published <strong><%= Utils.render_date(@latest_published_publication, :published, @context) %></strong>.
```
mcass19, line 17: Something that came to mind is that we could also accept a `conn` in the `render_date` function, initialize the context there, and avoid the init context call in phoenix controllers since we'll always have the connection in the template. Wdyt? 

I'm pretty good with what we have, not necessary to change but wanted to share the thought 😄

mcass19 (approved):
This work is reeeeeally awesome, thanks for taking care of all dates !!! 👏🏼 👏🏼 🚀 🚀 

I tested locally and looks awesome! Left some comments that aren't necessary to do at all

lib/oli_web/controllers/page_delivery_controller.ex
```diff
@@ -342,6 +343,8 @@ defmodule OliWeb.PageDeliveryController do
          _,
          _
        ) do
+    session_context = SessionContext.init(conn)
```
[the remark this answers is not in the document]
gastonabella: Actually I decided to use `session_context` because this function also receives a `%PageContext{}` under a variable named `context`, and I thought it would be cleaner to differentiate them.

lib/oli_web/templates/project/publish.html.eex
gastonabella: I think it's a good idea 👍
I'll add it to the followup ticket and we can do it together with the other improvements if that works for you

lib/oli_web/controllers/page_delivery_controller.ex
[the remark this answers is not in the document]
mcass19: Ooooh missed that `context` var sorry. Let's pretend i didn't say anything 😃

PR #2725, opened 2022-06-23
[MER-1201] Formula editing

marc-hughes:
Adds the ability to author formulas in the slate editor.

formula-edit

This PR removes the "richtext" formula type and adds a new "callout" element that can be used for the same purpose. This was to make the code less complex and allowed a nicer authoring experience since we can now edit callouts inline in the main editor instead of in a popup window.

To edit formulas:

- Open a core-authoring lesson, place your cursor in the editor
- Choose the insert formula option under either the inline or the block menu
- Click the placeholder formula that appears
- Edit the formula in the popup window in either latex or mathml

To edit callouts

- Open a core-authoring lesson, place your cursor in the editor
- Choose the insert callout option under either the inline or the block menu
- Type and optionally use formatting options within the callout box that appears.

Here is a project export with various examples of formulas and callouts in it:
[export_formulas_and_links (2).zip](https://github.com/marc-hughes/oli-torus-1/files/8967303/export_formulas_and_links.2.zip)

### DIFF
--- a/assets/styles/common/elements.scss
+++ b/assets/styles/common/elements.scss
@@ -1,6 +1,7 @@
 @import 'bootstrap';
 
 .structured-content,
+.activity-editor,
 .content {
   blockquote {
     margin-top: 12px;
@@ -119,6 +120,7 @@
     @extend .embed-responsive-item;
   }
 
+  .callout-inline,
   .formula-inline {
     display: inline-block;
     background-color: darken($body-bg, 5%);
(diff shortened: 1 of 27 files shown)


assets/styles/common/elements.scss
```diff
@@ -1,6 +1,7 @@
 @import 'bootstrap';
 
 .structured-content,
+.activity-editor,
```
marc-hughes, line 4: There was a bug where none of this formatting was being applied in the authoring environment within activities, adding activity-editor in here fixed that.

The formatting does appear in the lesson itself, so presumably we'd want the author to see it too.

darrensiegel (approved):

PR #2729, opened 2022-06-23
[Feature] MER-992 adaptive manual grading

bsparks:
This will enable activities within adaptive lessons to be marked as requiring manual grading by means of marking parts as such.
In the manual grading interface page it will render the activity (screen) in review mode with student responses.

image
image
image

Known Issues:

* Refreshing or loading the manually graded selection from the query string causes it to fail to render, clicking on and off that item fixes it
* All parts will be listed in manual grading even those that are not set to a manual grading approach
* Layers will not be rendered, only the direct screen. This will be a problem if the thing requiring manual grading lives on a layer. (this one we might need to resolve sooner than later, but I think for the immediate need it is OK)


jrissler:
@mcass19 @gastonabella @julmontesdeoca there is a lot here, can y'all help me code review / offer suggestions?

darrensiegel (approved):

PR #2738, opened 2022-06-27
[MER-1236] Rename fixed amount label where missing

mcass19:
[MER-1236](https://eliterate.atlassian.net/browse/MER-1236)


mcass19:
@gastonabella yeah you are right, I thought the same when doing the change but I didn't just because don't know if there's something behind that value. If we all agree is better we can override it, should be an easy change.

marc-hughes (approved):
This looks good to me. It seems weird to me to update database fields to satisfy UI design.

jrissler:
Maybe create a tech debt ticket to rename field?

PR #2740, opened 2022-06-28
[FEATURE] Dynamic question infrastructure

darrensiegel:
This PR adds the core infrastructure to support OLI Legacy style "Dynamic Questions".  

The key pieces of this work:

1. Introduce a new transformation operation type: `variable_substitution`.  An activity that is a "Dynamic Question" will include an instance of this transformation operation type in its `transformations` collection.  
2. Rework the transformations infrastructure to introduce a "batch context" creation step.  The idea here is that if we have a page which contains 15 dynamic questions, we do not want to serially execute the external call to evaluate their variables (whether it is to the Node bridge, Lambda, or a REST impl of the `authoring-eval`).  Rather, we want to bundle all 15 of these requests and send them to the evaluator as a batch operation.  This new batch context creation step appears in the transformer behavior as callback `provide_batch_context`
3. Implement a dynamic dispatch for the specific implementation of the variable substitution transformer.  There are four impls included: `LambdaImpl`, `NodeImpl`, `RestImpl` and `NoOpImpl`.  The NoOp impl is a stub meant for unit testing and for situations where dynamic question support is disabled (usually a dev environment).  The REST Impl allows use of the legacy `authoring-eval` REST JSON implementation.  Both Lambda and Node impls are stubbed out and unsupported at the moment.  

While there is no authoring support yet, one can test this out by creating a digest with the test migration course, ingesting that, and then visiting the "Graded Assessment with Dynamic Question" page founding within Unit 1 / Module 3.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -68,7 +68,7 @@ config :oli, :rule_evaluator,
 
 variable_substitution_provider =
   case System.get_env("VARIABLE_SUBSTITUTION_PROVIDER") do
-    nil -> Oli.Activities.Transformers.VariableSubstitution.RestImpl
+    nil -> Oli.Activities.Transformers.VariableSubstitution.NoOpImpl
     provider -> Module.concat([Oli, Activities, Transformers, VariableSubstitution, provider])
   end
 
--- a/config/config.exs
+++ b/config/config.exs
@@ -52,7 +52,8 @@ config :oli,
       ),
     favicons: System.get_env("BRANDING_FAVICONS_DIR", "/favicons")
   ],
-  payment_provider: System.get_env("PAYMENT_PROVIDER", "none")
+  payment_provider: System.get_env("PAYMENT_PROVIDER", "none"),
+  node_js_pool_size: String.to_integer(System.get_env("NODE_JS_POOL_SIZE", "2"))
 
 rule_evaluator_provider =
   case System.get_env("RULE_EVALUATOR_PROVIDER") do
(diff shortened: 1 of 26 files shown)


config/config.exs
```diff
@@ -62,10 +63,22 @@ rule_evaluator_provider =
 
 config :oli, :rule_evaluator,
   dispatcher: rule_evaluator_provider,
-  node_js_pool_size: String.to_integer(System.get_env("NODE_JS_POOL_SIZE", "2")),
   aws_fn_name: System.get_env("EVAL_LAMBDA_FN_NAME", "rules"),
   aws_region: System.get_env("EVAL_LAMBDA_REGION", "us-east-1")
 
+variable_substitution_provider =
+  case System.get_env("VARIABLE_SUBSTITUTION_PROVIDER") do
```
eliknebel, line 70: We probably don't need all this logic in the compile-time `config.exs`. We could simplify here by just setting `variable_substitution_provider` to its default, `Oli.Activities.Transformers.VariableSubstitution.NoOpImpl` and remove the `config :oli, :variable_substitution`. As long as the tests and NoOpImpl doesn't expect these configs to exist, this should cover the default configuration and test configuration, while still allowing a runtime configuration to be provided via `releases.exs`

lib/oli/activities/transformers.ex
```diff
@@ -1,46 +1,161 @@
...
+        revision_id_transform_pairs ->
+          run_one_operation(model_repository, module, revision_id_transform_pairs, errors_by_id)
+      end
+    end)
+  end
+
+  # Run one batch of transformers, for a specific operation.
+  #
+  # Takes the model repository, the module of the transformer impl, and a list of
+  # tuples of revision ids and transformer.  The larevision_id_transform_pairsst argument is really our
```
eliknebel, line 91: Should this say, "The revision_id_transform_pairs argument ..."?

lib/oli/authoring/editing/page_editor.ex
```diff
@@ -330,16 +330,22 @@ defmodule Oli.Authoring.Editing.PageEditor do
                         activity_type_id: activity_type_id,
                         content: content,
                         graded: graded
-                      } ->
+                      } = revision ->
          # To support 'test mode' in the editor, we give the editor an initial transformed
```
eliknebel, line 334: I guess it makes sense to update the transformed data here since its already part of the editor payload, though I'm not sure it is being used anymore since we nixed the editor test-mode

eliknebel (changes requested):

config/config.exs
darrensiegel: I'm not sure I understand what to do here.   I want to ensure that a developer has the ability switch provider, without having the touch any Elixir code.

lib/oli/activities/transformers.ex
darrensiegel: It indeed should.  Changes pushed

eliknebel:
Should we add an example with some documentation to `oli.example.env` to clarify how to configure `VARIABLE_SUBSTITUTION_PROVIDER`?

eliknebel (approved):

PR #2741, opened 2022-06-28
[BUGFIX] [MER-1296] Fixed automated e2e test setup

marc-hughes:
Commit 542e45e90bd488b92ede9e3c5e49922e3a2c5aad broke this API call, this fixes it.

This call is only used in our automated e2e testing framework to set up test data and therefore this should be low-risk to apply.


julmontesdeoca (commented):
I'm not 100% sure why calling the refresh manually, fixes it. Since the project publication ends up calling `Publishing.publish_project` which does spawn a process that calls the refresh function. 
That said I'm ok with this change if it helps unblocking you, and we can then separately check why it's really needed. Perhaps the fact that we're running it against a real server, and there's some issue with the asynchronicity.

pd: The build failed because of a flaky test (PR open with the fix)

jrissler (approved):

marc-hughes:
> I'm not 100% sure why calling the refresh manually, fixes it. Since the project publication ends up calling `Publishing.publish_project` which does spawn a process that calls the refresh function. That said I'm ok with this change if it helps unblocking you, and we can then separately check why it's really needed. Perhaps the fact that we're running it against a real server, and there's some issue with the asynchronicity.


Could it be that it's being called within a transaction and the process spawns off and tries to update the view before the transaction has completed?

julmontesdeoca:
Let's not merge this until we solve the issue or revert the changes that caused this test to fail :)

marc-hughes:
I think we still need a change like this. Currently, the only way to make the cypress tests complete successfully is to add a multi-minute wait so the refresh has time to finish.

By adding this into the API call, the call won't return until that's done and the test can continue as soon as it gets a result.

This change would have the undesirable effect of running the refresh twice, once synchronously in the call, and once async in the separate task that https://github.com/Simon-Initiative/oli-torus/pull/2750 starts.

Tomorrow, I'll look at only running it once by skipping the async version when called via this api, so...

# Don't merge this yet

julmontesdeoca:
On a second thought..

We did implement a behaviour to differentiate how to run the `refresh_part_mapping()` function.
We have a `Oli.PublishingTest.PartMappingMockAdapter` which does run the refresh function but synchronously, for tests. Which I think is pretty much this scenario as well. 

Do you think we can utilize that to run the refreshing synchronously like you're doing now, but without having the `update_publication` with an option to not refresh?

In `Oli.Publishing`, `refresh_adapter()` returns an adapter depending on the environment:

```
 @spec refresh_adapter() :: PartMappingRefreshAdapter
  defp refresh_adapter() do
    :oli
      |> Application.get_env(Oli.Publishing)
      |> Keyword.get(:refresh_adapter, PartMappingRefreshAsync)
  end
```

If in this scenario we can use `Application.put_env` to setup the `refresh_adapter` = `PartMappingMockAdapter`, we wouldn't need to refresh manually, as the MockAdapter would do it synchronously. 

How we set it up in `config/test.exs`
```
# Config adapter for refreshing part_mapping
config :oli, Oli.Publishing, refresh_adapter: Oli.PublishingTest.PartMappingMockAdapter
```

Alternatively, if updating the env (even temporarily) is not possible in this setup or would introduce side effects, we could have `Publishing.update_publication` take a parameter "adapter" which would be either the `PartMappingRefreshAsync` or the `PartMappingMockAdapter`. 
Having something like:
- If adapter is given, use it.
- If adapter is not given use by default the one from the environment (just use `referesh_adapter()`).
   something like `update_publication(%Publication{} = publication, attrs, refresh_adapter \\ refresh_adapter())`

Speaking more broadly, the difference would be to have a flag which determines if it refreshes async or sync, instead of refreshing or not.

Does that make sense? What do you think?
Let me know if this is clear enough :)

marc-hughes:
> Does that make sense? What do you think?
> Let me know if this is clear enough :)

I don't think any solution that relies on environment is going to work here. These tests run against servers in real environments, QA right now, but running them against production is in the future plans.

The second option of passing in an adapter sounds good to me.

config/test.exs
```diff
@@ -57,7 +57,7 @@ config :oli, OliWeb.Endpoint,
   server: false
 
 # Config adapter for refreshing part_mapping
-config :oli, Oli.Publishing, refresh_adapter: Oli.PublishingTest.PartMappingMockAdapter
+config :oli, Oli.Publishing, refresh_adapter: Oli.Publishing.PartMappingRefreshSync
```
marc-hughes, line 60: Renamed this one since it's not used only in tests anymore.

julmontesdeoca: 💯

julmontesdeoca:
Thanks for taking the time and making the changes 🙌

PR #2750, opened 2022-06-28
[MER-1297] Attempt fix for part mapping refresh not being executed

julmontesdeoca:
[MER-1297](https://eliterate.atlassian.net/browse/MER-1297)

### What does it change?

- Attempts to fix the part_mapping view not being refreshed on project publication updates

This replaces `Task.async` for `Task.start`
From the [elixir docs](https://hexdocs.pm/elixir/1.13/Task.html#module-dynamically-supervised-tasks): 

> ..async tasks link the caller and the spawned process. This means that, if the caller crashes, the task will crash too and vice-versa. This is on purpose: if the process meant to receive the result no longer exists, there is no purpose in completing the computation.
>

One reason for the refresh not being executed could be that process running the publication update is finishing before the spawned process that runs the refresh. 

From the[ Task.start docs](https://hexdocs.pm/elixir/1.13/Task.html#start/1) 

>This should only used when the task is used for side-effects (like I/O) and you have no interest on its results nor if it completes successfully.
>
> If the current node is shutdown, the node will terminate even if the task was not completed. For this reason, we recommend to use [Task.Supervisor.start_child/2](https://hexdocs.pm/elixir/1.13/Task.Supervisor.html#start_child/2) instead, which allows you to control the shutdown time via the :shutdown option
>

This fits our use case, since we don't want to depend (or limit) the child process lifetime from the current process. We also don't need to supervise it or shut it down manually, so no need to use `Task.Supervisor.start_child`

Another option would be to use `Task.asnyc_nolink` but it requires a supervisor and we have no intention on supervising the process.

The `Process.sleep/1` call is added as an extra safety measure for cases when the `Publishing.update_publication` is executed inside another transaction. Before refreshing the view, we want all the changes to be persisted.

I didn't manage to replicate the original issue locally, so we agreed on running this in QA to test if this fixes the part_mapping view not being refreshed.

### DIFF
--- a/lib/oli/publishing/part_mapping_refresh_async.ex
+++ b/lib/oli/publishing/part_mapping_refresh_async.ex
@@ -8,7 +8,10 @@ defmodule Oli.Publishing.PartMappingRefreshAsync do
   @impl PartMappingRefreshAdapter
   @spec maybe_refresh_part_mapping(ecto_publication_operation) :: ecto_publication_operation
   def maybe_refresh_part_mapping({:ok, _publication} = operation_result) do
-    Task.async(fn -> Oli.Publishing.refresh_part_mapping() end)
+    Task.start(fn ->
+      Process.sleep(5_000)
+      Oli.Publishing.refresh_part_mapping()
+    end)
     operation_result
   end
 


jrissler (approved):

darrensiegel (changes requested):
There are a couple of edge cases that worry me here. 

1. What if many users publish their projects at the same time, over and over. Maliciously. What happens here?  We get numerous concurrent executing refresh_part_mapping calls.   

2. A publication is made and then three seconds later the system restarts. The spawned task never executes. And on restart the system is in an inconsistent state until the next publication.

Granted, the previous approach of the inline REFRESH was subject to the first problem. But I think there is a better solution here that addresses both of these if we leverage the `Oban` job scheduler.  An Oban worker can be configured in a way that basically allows one job to be executing (because it uses a queue configured to have only 1 concurrent job) and at most one job queued behind that job.  Any request to enqueue an additional job beyond that one gets ignored.  This approach gives us the rate limiting solution that we want, and since Oban is database backed these jobs survive system restarts - thus it solves the consistency on restart issue. 

Check the Oban docs, specifically the section on uniqueness and detecting duplicates for how to configure this.

julmontesdeoca:
@darrensiegel thanks for suggesting Oban, agree on all ends that it would be a better rounded solution. 
Would you be ok with us moving forward with this change for the time being? We're looking to deploy tonight, and we should either merge to test this fix in QA, or revert the previous changes entirely. In that case we'd start working on the Oban solution right after.

darrensiegel:
@julmontesdeoca For sure, this implementation can land right now and we can follow it up with the more robust Oban solution.

darrensiegel (approved):

PR #2755, opened 2022-06-29
[MER-1288] Add project ids to snapshots

julmontesdeoca:
[MER-1288](https://eliterate.atlassian.net/browse/MER-1288)

### What does it change?

- Updates snapshots table to add `project_id`s

For reference, we're dropping indices and re-creating them after the update, which is faster than having them update on each row update.
However the biggest improvement comes from removing an unnecessary self join on the update query when we first introduced it.

**Previous query**

```
UPDATE snapshots 
SET project_id = sect.base_project_id 
FROM snapshots snap 
LEFT JOIN sections sect ON snap.section_id = sect.id;
```
Screen Shot 2022-06-29 at 15 26 39

**New query**
```
UPDATE snapshots
SET project_id = sect.base_project_id
FROM sections sect WHERE sect.id = snapshots.section_id;
```

Screen Shot 2022-06-29 at 15 27 06

For a production dump with 14.235.739 records it took ~8 min locally.

Screen Shot 2022-06-29 at 16 08 26

### DIFF
--- a/priv/repo/migrations/20220629134948_update_snapshots_add_project_ids.exs
+++ b/priv/repo/migrations/20220629134948_update_snapshots_add_project_ids.exs
@@ -1,0 +1,28 @@
+defmodule Oli.Repo.Migrations.UpdateSnapshotsAddProjectIds do
+  use Ecto.Migration
+
+  def up do
+    drop index(:snapshots, [:objective_id])
+    drop index(:snapshots, [:activity_id])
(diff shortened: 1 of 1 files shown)


mcass19 (approved):
👏🏼 👏🏼 👏🏼

gastonabella (approved):
🚀

eliknebel (approved):

jrissler (approved):

PR #2766, opened 2022-06-30
[FEATURE] Dynamic questions editing

darrensiegel:
This PR adds dynamic question editing support via the following:

1. It migrates the Echo React components for both first gen and second gen editors.
2. Embeds within all activities (except ImageCoding, Adaptive and OLIEmbedded) the ability to enable and disable and edit dynamic question scripts. 
3. Replaces the ACE editor with Monaco in the migrated Echo components
4. Adds an endpoint for the client to request variable evaluation, which then is a passthrough to the configured variable substitution strategy. 

One needs the `dynamic-questions-editing` branch of both `course-digest` and `authoring-eval` in place to test this end to end.  

1. Start `authoring-eval`
2. Create a digest from the embedded test course via `course-digest`
3. Ensure RESTImpl is the configured variable substitution strategy in Torus. 
4. Ingest the course. 
5. As author, go to the Activity Bank to see two Dynamic Questions.  Test that they run in the "test mode"
6. Publish the course, create a section, verify that graded assessment selects both questions, evaluates, and renders them correctly.

### DIFF
--- a/assets/src/components/activities/check_all_that_apply/CheckAllThatApplyAuthoring.tsx
+++ b/assets/src/components/activities/check_all_that_apply/CheckAllThatApplyAuthoring.tsx
@@ -22,13 +22,15 @@ import { configureStore } from 'state/store';
 import { Maybe } from 'tsmonad';
 import { AuthoringElement, AuthoringElementProps } from '../AuthoringElement';
 import { AuthoringElementProvider, useAuthoringElementContext } from '../AuthoringElementProvider';
+import { VariableEditorOrNot } from '../common/variables/VariableEditorOrNot';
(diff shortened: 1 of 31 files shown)


eliknebel (changes requested):
When trying to preview the test migration course, I am running into the following error on the dynamic questions page:

`https://localhost/authoring/project/migration_test_course/preview/graded_assessment_with_selecti`
Screen Shot 2022-07-01 at 9 04 23 AM

This appears to only be a preview problem, as it works fine if I publish the course

eliknebel (approved):
The error was related to an empty Activity Bank Selection activity in the page, which is unrelated and filed in another issue

PR #2767, opened 2022-06-30
PR3 Survey Delivery

### LINKED ISSUE #2579: Survey PR3 Delivery: CMU-510
- [x] Add rendering support for survey element, using group as a template
    - Pass in contextId, hideFeedback and disableGrading flags to any activities rendered within a survey container
    - Ensure that the logic for calculating and rolling up grades within a graded page submission ignores any “non-scoreable” attempt. This might already be in place, give that the notion of “non-scoreable” activity attempts exists.  But let’s make sure, and let’s make sure that these survey attempts are indeed being created as “non-scorable”. 
- [x] Add group_id and survey_id to activity_attempt ecto schema and create database migration
- [x] Introduce survey_uid and group_id flags to activity contexts
- [x] Figure out how to determine the state of a survey (not_started/in_progress/completed) based on activity attempts and show/hide "Submit Survey" button accordingly. If survey is already submitted, instead show a thank you for feedback message.

eliknebel:
Adds delivery support for survey elements. A survey can now be authored in a basic page and delivered to students. Survey ids are captured as part of the activity attempt record.

A survey will hide any submit buttons an activity has as well as hints and feedback. The activities will respond to the event generated by the "Submit Survey" button by cleaning up their state and submitting a response. If any of these activity submit requests fail, the Submit Survey button will reappear next time a student accesses the page so that it can be clicked again.
Screen Shot 2022-06-30 at 3 14 17 PM

Once a survey has been submitted, it will no longer be interactive (expected to be handled by each individual activity via the coming SDK) and a "Response has been submitted" message is shown with the ability to reset the survey. The reset will reset all activities inside a survey and again display the "Submit Survey" button.
Screen Shot 2022-06-30 at 3 14 48 PM

Since survey functionality now exists for both authoring and delivery, the corresponding feature flag has been removed and surveys will become available to all project authors.

Closes #2579


lib/oli/resources/page_content.ex
```diff
@@ -69,9 +86,67 @@ defmodule Oli.Resources.PageContent do
...
+  ## Examples
+      iex> activity_parent_groups(content)
+      %{
+        10 => %{group: "44325", survey: nil},        # inside a group
+        11 => %{group: nil, survey: "92374"},        # inside a survey
+        23 => %{group: "53222", survey: "96366"},    # inside a group and a survey
+        26 => %{group: nil, survey: nil},            # not inside a group or survey
+      }
+  """
+  def activity_parent_groups(%{"model" => _model} = content) do
```
darrensiegel, line 111: Can we move `activity_parent_groups` and `survey_activities` out of this module and into something that is more activity and survey specific?  It would be good to keep this module small and focused as an almost "Swiss army knife" collection of low level utilities that can be used to build task specific things like these two functions.

lib/oli/features.ex
```diff
@@ -49,11 +45,22 @@ defmodule Oli.Features do
     do: update_feature_state(get_by_label(label).id, state)
 
   def bootstrap_feature_states() do
-    Enum.map(@features, fn %Feature{id: id} ->
-      case Repo.get(FeatureState, id) do
-        nil -> create_feature_state(%{state: :disabled})
-        e -> e
-      end
+    Repo.transaction(fn ->
+      # delete any feature flags that are no longer defined
```
darrensiegel, line 49: This is great.

darrensiegel (approved):
Just that one comment about moving some functions, but we can do that in a future PR.  

This is excellent work.  Easy to use and looks fantastic in preview and in delivery!

PR #2774, opened 2022-07-01
[BUGFIX] JAN-1711 resolves issues with dragging components and panning

bsparks:
only pan the editing area if the user clicks on the screen or the background JAN-1711


rmathis (approved):

marc-hughes (approved):
This is working a ton better now, way less annoying to use the editor.

PR #2775, opened 2022-07-01
[BUGFIX] adaptive author can't type in mutate action scripts

bsparks:
triggering a save on every key press is too much, can't finish validation

### DIFF
--- a/assets/src/apps/authoring/components/AdaptivityEditor/ActionMutateEditor.tsx
+++ b/assets/src/apps/authoring/components/AdaptivityEditor/ActionMutateEditor.tsx
@@ -45,7 +45,6 @@ const ActionMutateEditor: React.FC<ActionMutateEditorProps> = (props) => {
       return;
     }
     setValue(val);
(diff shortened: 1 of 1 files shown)


rmathis (approved):

assets/src/apps/authoring/components/AdaptivityEditor/ActionMutateEditor.tsx
```diff
@@ -173,6 +172,7 @@ const ActionMutateEditor: React.FC<ActionMutateEditorProps> = (props) => {
           id={`action-mutate-value-${uuid}`}
           value={value}
           onChange={(e) => handleValueChange(e)}
+          onBlur={(e) => setIsDirty(true)}
```
marc-hughes, line 175:
```suggestion
          onBlur={() => setIsDirty(true)}
```

There an eslint unused var warning on `e`

marc-hughes (approved):
I was able to reproduce the original bug and this fixes it.

PR #2780, opened 2022-07-05
[MER-1293] More improvements on timezones handling

gastonabella:
[MER-1293](https://eliterate.atlassian.net/browse/MER-1293)

Some improvements on the timezones handling:
- Remove timezone fields from institutions and pending registrations.
- Store user/author's selected timezone as a preference in the database.
- Refactoring.


mcass19 (approved):
🚀 🚀 🚀

eliknebel (changes requested):
Looks great, works as expected.

I think the only thing missing is the option to return to using the browser specified timezone (clear the preference). While it probably wont be used often, it is good practice to allow the user to return to the default state if they desire. It may also be more dynamic (e.g. change automatically when traveling) so we wouldn't want to lock a user out of this option if they manually set the timezone at some point in the past. One approach here would simply be to always have the first option in the select to be "Use Browser Timezone"

eliknebel:
I also noticed this option is missing from the Guest user dropdown. It would be nice if we could support this for guest users
Screen Shot 2022-07-13 at 5 01 59 PM

eliknebel:
> > Looks great, works as expected.
> > I think the only thing missing is the option to return to using the browser specified timezone (clear the preference). While it probably wont be used often, it is good practice to allow the user to return to the default state if they desire. It may also be more dynamic (e.g. change automatically when traveling) so we wouldn't want to lock a user out of this option if they manually set the timezone at some point in the past. One approach here would simply be to always have the first option in the select to be "Use Browser Timezone"
> 
> @eliknebel I tried to capture the essence of this idea in the current functionality, by specifying which one of the options was fetched from the browser (see screnshot). I could move this option to the top as you suggested and rename it to `"Use Browser Timezone"`, but I think it should also specify which timezone it is. Maybe something like `"Use Browser Timezone - (GMT-07:00) Arizona"`?
> 
> Screen Shot 2022-07-14 at 10 13 53 AM

Ah, yes I must have missed it! I do think this would be better at the top as a separate option so that it doesn't get lost in the other options (and someone can select that specific timezone even though it is their browser one too), and even better would be to show the actual value like you mentioned! 👍

eliknebel:
Great thank you! For some reason though, I still can't see the option in the dropdown. I reset my database, cleared my cookies and started from scratch, it seems to set Greenwich time (not my browser default) and I can't see an option the change to my browser. Any thoughts?
Screen Shot 2022-07-14 at 6 17 49 PM
Screen Shot 2022-07-14 at 6 17 43 PM

gastonabella:
> Great thank you! For some reason though, I still can't see the option in the dropdown. I reset my database, cleared my cookies and started from scratch, it seems to set Greenwich time (not my browser default) and I can't see an option the change to my browser. Any thoughts? Screen Shot 2022-07-14 at 6 17 49 PM Screen Shot 2022-07-14 at 6 17 43 PM

Yes, that's something I noticed some time ago. When the timezone isn't set in the session it seems that the `timezone.js` hook makes the request to Torus, the request is successfully according to the logs, but even after refreshing the page the cookie doesn't update and you still see the local timezone unset in your session. I believe that when that happened to me I just did some navigation through different views and at some point it gets set. Anyway I think we should dedicate some effort to fix this maybe in a separate story since it would require some important changes in the way the local timezone is fetched from the browser. [This post](https://thepugautomatic.com/2020/05/persistent-session-data-in-phoenix-liveview/) follows a similar approach, and according to the comments at the bottom some people experienced similar issues with this approach.

eliknebel:
Agreed, thanks for the thorough analysis

eliknebel (approved):

PR #2785, opened 2022-07-06
[FEATURE] [MER-1073] Instructor adaptive preview

mcass19:
[MER-1073](https://eliterate.atlassian.net/browse/MER-1073)

Add instructor preview for adaptive lessons. For that, wired an `isInstructor` boolean to the react app and from there look for the correct endpoints to retrieve the activities. Also, changed the format of the response so that it retrieves what the frontend is expecting to work.

[Small example](https://recordit.co/FFhccTCLGd)

### Other changes
- Fix searching by some special characters in the admin sections index view
- Disable admin section preview for now
  - The units and pages preview are breaking the server when accessing as admin because the logic doesn't fully contemplate this scenario
  - Will create a follow up ticket to tackle this in particular
- Hide manage section button when previewing the course as an instructor


lib/oli/delivery/sections/browse.ex
```diff
@@ -7,6 +7,8 @@ defmodule Oli.Delivery.Sections.Browse do
   alias Oli.Repo
   alias Oli.Repo.{Paging, Sorting}
 
+  @chars_to_replace_on_search [" ", "&", ":", ";", "(", ")", "|", "!", "'", "<", ">"]
```
gastonabella, line 10: Should we include more special characters to this list?

lib/oli_web/controllers/api/activity_controller.ex
```diff
@@ -439,6 +405,46 @@ defmodule OliWeb.Api.ActivityController do
...
   end
 
+  defp document_to_delivery_result(is_preview_mode, %{
+    title: title,
+    activity_type_id: activity_type_id,
+    content: content,
+    resource_id: resource_id
+  }) do
+    authoring = Map.get(content, "authoring")
+    content = Map.delete(content, "authoring")
```
gastonabella, line 415: Super nit but you could do it in one line:
```elixir
{authoring, content} = Map.pop(content, "authoring")
```

gastonabella (approved):
Nice work! I left some small comments that are not blockers at all

lib/oli/delivery/sections/browse.ex
mcass19: Is mainly because we shouldn't collide with the characters [postgres text search](https://www.postgresql.org/docs/current/datatype-textsearch.html) uses to build the lexemes

julmontesdeoca (approved):
Looks good to me! 🦫 
Should we create a ticket/issue for adding the admin's adaptive preview?

darrensiegel (approved):

PR #2793, opened 2022-07-11
[MER-1216] Add duplicate page action in project curriculum

julmontesdeoca:
[MER-1216](https://eliterate.atlassian.net/browse/MER-1216)

### What does it change?

- Adds ability to duplicate a page within a project, in the curriuculum overview page.

We're running a deepy copy of resources revisoins of the page (and of its embedded activities). 
The option is only enababled for pages, not including adaptive ones, which are going to be worked on separately. 

#### Screenshots

image

### DIFF
--- a/lib/oli/authoring/editing/container_editor.ex
+++ b/lib/oli/authoring/editing/container_editor.ex
@@ -16,6 +16,8 @@ defmodule Oli.Authoring.Editing.ContainerEditor do
   alias Oli.Authoring.Editing.PageEditor
   alias Oli.Repo
   alias Oli.Authoring.Broadcaster
+  alias Oli.Authoring.Editing.ActivityEditor
+  alias Oli.Activities
 
   @spec edit_page(Oli.Authoring.Course.Project.t(), any, map) :: any
   def edit_page(%Project{} = project, revision_slug, change) do
@@ -263,6 +265,102 @@ defmodule Oli.Authoring.Editing.ContainerEditor do
         e
     end
   end
+
+  @doc """
+    Duplicates a page within its project.
+    A true deep-copy of all the content is made. Resources and Revisions are created instead of simply referenced.
+  """
+  def duplicate_page(
+        %Revision{} = container,
+        page_id,
+        %Author{} = author,
+        %Project{} = project
+      ) do
+    original_page =
+      Resources.get_revision!(page_id)
+      |> Map.from_struct()
+
+    new_page_attrs =
+      original_page
+      |> Map.delete(:slug)
+      |> Map.delete(:inserted_at)
+      |> Map.delete(:updated_at)
+      |> Map.delete(:updated_at)
+      |> Map.delete(:resource_id)
+      |> Map.delete(:resource)
+      |> Map.put(:title, "Copy of #{original_page.title}")
+      |> Map.put(:content, nil)
+
+    Repo.transaction(fn ->
+      with {:ok, created_revision} <- add_new(container, new_page_attrs, author, project),
+           {:ok, duplicated_content} <-
+             duplicate_content_activities(
+               original_page.content,
+               project.slug,
+               author
+             ),
+           {:ok, updated_revision} <-
+             Resources.update_revision(created_revision, %{content: duplicated_content}) do
+        updated_revision
+      else
+        {:error, e} -> Repo.rollback(e)
+      end
+    end)
+  end
+
+  defp duplicate_content_activities(content, project_slug, author) do
+    case deep_copy_activities(content["model"], project_slug, author) do
+      {:ok, duplicated_activities} ->
+        {:ok, %{content | "model" => Enum.reverse(duplicated_activities)}}
+
+      {:error, error} ->
+        {:error, error}
+    end
+  end
+
+  defp deep_copy_activities([], _project_slug, _author), do: {:ok, []}
+
+  defp deep_copy_activities(model, project_slug, author) do
+    Enum.reduce_while(model, {:ok, []}, fn activity, {:ok, acc} ->
+      case deep_copy_activity(activity, project_slug, author) do
+        {:ok, new_activity} -> {:cont, {:ok, [new_activity | acc]}}
+        {:error, _} = err -> {:halt, err}
+      end
+    end)
+  end
+
+  defp deep_copy_activity(%{"type" => "activity-reference"} = item, project_slug, author) do
+    activity_revision = AuthoringResolver.from_resource_id(project_slug, item["activity_id"])
+
+    activity_type = Activities.get_registration(activity_revision.activity_type_id)
+
+    objectives =
+      Enum.reduce(activity_revision.objectives, [], fn {_key, ids}, acc -> acc ++ ids end)
+
+    case ActivityEditor.create(
+           project_slug,
+           activity_type.slug,
+           author,
+           activity_revision.content,
+           objectives,
+           "embedded",
+           activity_revision.title
+         ) do
+      {:ok, {revision, _}} ->
+        {:ok,
+         %{
+           "id" => item["id"],
+           "type" => "activity-reference",
+           "children" => [],
+           "activity_id" => revision.resource_id
+         }}
+
+      {:error, error} ->
+        {:error, error}
+    end
+  end
+
+  defp deep_copy_activity(item, _project_slug, _author), do: {:ok, item}
 
   defp remove_from_container(container, project_slug, revision, author) do
     # Create a change that removes the child
--- a/lib/oli/authoring/editing/container_editor.ex
+++ b/lib/oli/authoring/editing/container_editor.ex
@@ -282,43 +282,26 @@ defmodule Oli.Authoring.Editing.ContainerEditor do
 
     new_page_attrs =
       original_page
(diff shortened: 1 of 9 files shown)


lib/oli/authoring/editing/container_editor.ex
```diff
@@ -264,6 +266,102 @@ defmodule Oli.Authoring.Editing.ContainerEditor do
...
+  defp deep_copy_activities([], _project_slug, _author), do: {:ok, []}
+
+  defp deep_copy_activities(model, project_slug, author) do
+    Enum.reduce_while(model, {:ok, []}, fn activity, {:ok, acc} ->
+      case deep_copy_activity(activity, project_slug, author) do
+        {:ok, new_activity} -> {:cont, {:ok, [new_activity | acc]}}
+        {:error, _} = err -> {:halt, err}
+      end
+    end)
+  end
```
julmontesdeoca, line 330: Traverses the content's model, `deep_copy_activity` will return a replacement of the activity reference, which references the new created activity. Could rename `deep_copy_activity` to `maybe_deep_copy_activity`, as it's pattern matches on activity references to determine if it should copy.

test/oli_web/live/curriculum/container_test.exs
```diff
@@ -42,10 +44,60 @@ defmodule OliWeb.Curriculum.ContainerLiveTest do
...
+      conn =
+        recycle(conn)
+        |> Pow.Plug.assign_current_user(author, OliWeb.Pow.PowHelpers.get_pow_config(:author))
+        |> get("/authoring/project/#{project.slug}/curriculum/")
+
+      {:ok, view, _html} = live(conn)
+
+      adaptive_page_revision = Map.get(map, :adaptive_page_revision)
+
+      assert view
```
mcass19, line 90: NIT: you can just use [has_element?/3](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveViewTest.html#has_element?/3) here

lib/oli/authoring/editing/container_editor.ex
```diff
@@ -264,6 +266,102 @@ defmodule Oli.Authoring.Editing.ContainerEditor do
...
+        %Author{} = author,
+        %Project{} = project
+      ) do
+    original_page =
+      Resources.get_revision!(page_id)
+      |> Map.from_struct()
+
+    new_page_attrs =
+      original_page
+      |> Map.delete(:slug)
```
mcass19, line 285: [Map.drop](https://hexdocs.pm/elixir/Map.html#drop/2) ?

lib/oli/authoring/editing/container_editor.ex
```diff
@@ -264,6 +266,102 @@ defmodule Oli.Authoring.Editing.ContainerEditor do
...
+    case deep_copy_activities(content["model"], project_slug, author) do
+      {:ok, duplicated_activities} ->
+        {:ok, %{content | "model" => Enum.reverse(duplicated_activities)}}
+
+      {:error, error} ->
+        {:error, error}
+    end
+  end
+
+  defp deep_copy_activities([], _project_slug, _author), do: {:ok, []}
```
mcass19, line 321: Is this pattern match necessary? Seems like `reduce_while` below will just return `{:ok, []}` if the list is empty

lib/oli/authoring/editing/container_editor.ex
```diff
@@ -264,6 +266,102 @@ defmodule Oli.Authoring.Editing.ContainerEditor do
...
+             Resources.update_revision(created_revision, %{content: duplicated_content}) do
+        updated_revision
+      else
+        {:error, e} -> Repo.rollback(e)
+      end
+    end)
+  end
+
+  defp duplicate_content_activities(content, project_slug, author) do
+    case deep_copy_activities(content["model"], project_slug, author) do
```
mcass19, line 312: Seems like we could just call `deep_copy_activities` inside the transaction and avoid one function definition, wdyt?

lib/oli/authoring/editing/container_editor.ex
```diff
@@ -264,6 +266,102 @@ defmodule Oli.Authoring.Editing.ContainerEditor do
...
+    end)
+  end
+
+  defp deep_copy_activity(%{"type" => "activity-reference"} = item, project_slug, author) do
+    activity_revision = AuthoringResolver.from_resource_id(project_slug, item["activity_id"])
+
+    activity_type = Activities.get_registration(activity_revision.activity_type_id)
+
+    objectives =
+      Enum.reduce(activity_revision.objectives, [], fn {_key, ids}, acc -> acc ++ ids end)
```
mcass19, line 338: Is the reduce necessary or we just link the same objectives that are in `activity_revision`?

mcass19 (commented):
Great job 🚀 🚀 

Left some minor comments but feel free to change what you think is worth it, looks great to me!

lib/oli/authoring/editing/container_editor.ex
julmontesdeoca: Oh, neat.

lib/oli/authoring/editing/container_editor.ex
julmontesdeoca: Completely right, unneccesary.

lib/oli/authoring/editing/container_editor.ex
darrensiegel: In certain cases, this approach for handling the objectives is not going to lead to correct duplication, given that the `activity_editor.create/7` implementation takes the list of supplied objectives and applies them to *all* parts of the activity.  In the original activity, each part may have had a different objective.  For example, consider the following `objectives` value for the activity that is being duplicated:

```
%{
  "1" => [100, 101, 102],
 "2" => [200, 201, 202],
}
```

In the above, we see this activity has two parts, identified by `"1"` and `"2"` respectively.  Each part has three objectives attached to them.  As-is, the current duplication code would reduce this structure down to a list of `[100, 101, 102, 200, 201, 202]` and the `attach_objectives_to_all_parts/2` impl within `activity_editor` would create the following:

```
%{
  "1" => [100, 101, 102, 200, 201, 202],
 "2" => [100, 101, 102, 200, 201, 202]
}
```

darrensiegel (changes requested):

lib/oli/authoring/editing/container_editor.ex
julmontesdeoca: I get your point, I'm updating with some changes to perform the `deep_copy_activities` call inside the transaction, the less verbose way I came up with was to add another line in the `with`, let me know if that makes sense. 
I could also switch `%{original_page.content | "model" =>` for a `Map.put`, just went with what looked shorter.

julmontesdeoca:
@darrensiegel thanks for the catch with the objectives. Will look into that to keep the objectives on each part as it should be.

lib/oli/authoring/editing/container_editor.ex
```diff
@@ -264,6 +266,102 @@ defmodule Oli.Authoring.Editing.ContainerEditor do
...
+
+      {:error, error} ->
+        {:error, error}
+    end
+  end
+
+  defp deep_copy_activities([], _project_slug, _author), do: {:ok, []}
+
+  defp deep_copy_activities(model, project_slug, author) do
+    Enum.reduce_while(model, {:ok, []}, fn activity, {:ok, acc} ->
```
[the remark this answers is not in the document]
julmontesdeoca: Got it, although just to clarify, this implementation does not duplicate adaptive pages. And this is one of the reasons why we are implementing an adaptive page duplication separately, alongside different content formats, making it easier to debug/extend, etc.

You probably have more context on the amount data currently handled in production, do you think it's still necessary to implement some sort of streamlined/batch creation for normal graded and non-graded pages?

Initially thought the ROI might not be super high, however learning how the objectives need to be handled I probably have to avoid using `ActivityEditor.create` altogether. Which means I can probably go for that either way, which makes sense.

lib/oli/authoring/editing/container_editor.ex
mcass19: Yeah that seems fine 👍🏼

lib/oli/authoring/editing/container_editor.ex
julmontesdeoca: @darrensiegel to further understand this (trying to cover all scenarios);
Am I right to say that non-adaptive activities will only have one part in the content? Didn't manage to replicate a scenario where I could assign an objective to a specific part for normal pages. It also seems like `ActivityEditor.create` will always receive an empty objectives list for non-adaptive lessons. Since the objectives are attached on an update and not when first adding the activity to the page. Perhaps that's not needed when creating from the UI?

julmontesdeoca:
@darrensiegel pushed some changes to not attach objectives to all parts in duplication, while keeping the `ActivityEditor.create` working in the same way. I wanted to avoid duplicating that function, or sending an empty list and having an extra query to update objectives. 

Regarding the performance concerns, could we tackle that in a different PR? This does not duplicate adaptive pages, and we have another ticket for implementing that. We're also going to be working on module and unit duplication, perhaps in that work we can improve part of this and orchestrate a bigger bulk creation. What do you think?

PR #2798, opened 2022-07-12
[BUGFIX] [MER-1324] Replace get_user to get_author

mcass19:
[MER-1324](https://eliterate.atlassian.net/browse/MER-1324)

After locking the page and broadcasting the `lock_acquired` message, the `handle_info` in `ContainerLive` was going to the users table with the author id, so if you were other author looking at the same project, you were seeing the user's name that corresponds to the author id. Doesn't even make sense to have that there since it's an author view.


darrensiegel (approved):

gastonabella (approved):
🚀

PR #2800, opened 2022-07-12
[FEATURE] track session identifier in part attempt for datashop

eliknebel:
Adds the concept of a datashop session identifier to part attempts.

- [ ] Generate a unique datashop session id on every new session and track in the part attempt record
- [ ] Export datashop session identifier as part of the part attempts in the datashop export
- [ ] Unit tests

### DIFF
--- a/lib/oli/accounts.ex
+++ b/lib/oli/accounts.ex
@@ -167,6 +167,26 @@ defmodule Oli.Accounts do
   def get_user_by(clauses), do: Repo.get_by(User, clauses)
 
   @doc """
+  Gets a single user with platform roles and author preloaded
+  Returns `nil` if the User does not exist.
+  ## Examples
+      iex> get_user_with_roles(123)
+      %User{}
+      iex> get_user_with_roles(456)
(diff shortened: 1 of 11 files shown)


lib/oli/analytics/datashop.ex
```diff
@@ -78,6 +78,7 @@ defmodule Oli.Analytics.Datashop do
         date: hd(part_attempts).activity_attempt.resource_attempt.inserted_at,
         email: email,
         sub: sub,
+        datashop_session_id: hd(part_attempts).datashop_session_id,
```
darrensiegel, line 81: For existing part attempt records, this `datashop_session_id` will be `null`.  It seems like we should handle this case here and generate a UUID, otherwise the `<meta>` element will have a null session id.

darrensiegel (changes requested):

lib/oli/analytics/datashop.ex
eliknebel: fixed

PR #2811, opened 2022-07-14
[MER-1206] Display only two decimal points in grading scores

julmontesdeoca:
[MER-1206](https://eliterate.atlassian.net/browse/MER-1206)

### What does it change?

- Formats score for limiting to two decimal points in the displaying of scores in the following pages:
    - Student progress view
    - Student resource progress view
    - Section's Gradebook view
    - Attempt review (student and instructor)

When editing a resoure's score manually, a number with N decimals numbers is accepted, but following the same logic as everywhere else, the number get's rounded up and limited to two significant decimals for display.

### DIFF
--- a/lib/oli_web/live/common/utils.ex
+++ b/lib/oli_web/live/common/utils.ex
@@ -66,8 +66,7 @@ defmodule OliWeb.Common.Utils do
   """
   @spec format_score(float) :: float
   def format_score(score) when is_float(score) do
-    formatted_score = Float.round(score, 2)
-    if score - formatted_score != 0.0, do: formatted_score, else: score
+    Float.round(score, 2)
   end
 
   defp has_value(v) do
--- a/lib/oli_web/live/common/utils.ex
+++ b/lib/oli_web/live/common/utils.ex
@@ -36,12 +36,39 @@ defmodule OliWeb.Common.Utils do
     render_date_with_opts(item, attr_name, opts)
   end
 
+  @spec render_precise_date(map, any, any) :: binary
   def render_precise_date(item, attr_name, context) do
     opts = [context: context, precision: :minutes]
     render_date_with_opts(item, attr_name, opts)
   end
 
   def render_date_with_opts(item, attr_name, opts), do: date(Map.get(item, attr_name), opts)
+
+  @doc """
+    Rounds up a grading score to two significant figures.
+    For numbers with no decimals, or non-significant zeros after the comma, it keeps only one zero
+
+    ## Examples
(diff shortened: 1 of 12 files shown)


test/oli_web/live/student_view_live_test.exs
```diff
@@ -0,0 +1,162 @@
...
+      student: student
+    } do
+      redirect_path = "/sections/#{section.slug}/enroll"
+
+      {:error, {:redirect, %{to: ^redirect_path}}} =
+        live(conn, live_view_student_view_route(section.slug, student.id))
+    end
+  end
+
+  describe "student cannot access when is logged in as an author but is not a system admin" do
```
mcass19, line 34: Typo: `user` or `author` instead of `student`?

lib/oli_web/live/common/utils.ex
```diff
@@ -35,6 +35,32 @@ defmodule OliWeb.Common.Utils do
...
+    iex> format_score(78.479)
+    78.48
+
+    iex> format_score(0.0)
+    0.0
+  """
+  @spec format_score(float) :: float
+  def format_score(score) when is_float(score) do
+    formatted_score = Float.round(score, 2)
+    if score - formatted_score != 0.0, do: formatted_score, else: score
```
gastonabella, line 61: I might be missing something, but is this line necessary? If `score - formatted_score` is `0` it means they are equal, so you can just return always the `formatted_score`

lib/oli_web/live/progress/student_resource_view.ex
```diff
@@ -57,8 +57,14 @@ defmodule OliWeb.Progress.StudentResourceView do
...
               case resource_access do
-                nil -> nil
-                _ -> ResourceAccess.changeset(resource_access, %{})
+                nil ->
+                  nil
+
+                _ ->
+                  ResourceAccess.changeset(resource_access, %{
+                    # limit score decimals to two significant figures, rounding up
+                    score: Float.round(resource_access.score, 2)
```
gastonabella, line 66: Do we actually want to round the score in the database as well? I thought we wanted to store it as is, and then format the value only when it's displayed.

gastonabella (commented):
Great job with the tests 👏

lib/oli_web/live/progress/student_resource_view.ex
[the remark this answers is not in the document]
julmontesdeoca: There's not anymore, updating.

darrensiegel (approved):

PR #2825, opened 2022-07-19
PMP-2978 | JAN-1827 init state race condition

bsparks:
sometimes the check complete (context change) would happen before all the parts got initialized, so the part init was wiping out any init state that might have been applied, or it was simply not registering.


assets/src/apps/delivery/store/features/groups/actions/deck.ts
```diff
@@ -13,6 +13,7 @@ import {
   writePageAttemptState,
 } from 'data/persistence/state/intrinsic';
 import { ResourceId } from 'data/types';
+import thunk from 'redux-thunk';
```
bsparks, line 16:
```suggestion

```

PR #2846, opened 2022-07-21
[MER-1310] New Video content element.

marc-hughes:
Allows an author to upload video to be displayed in core authoring pages & activities.

Demo of feature:

https://user-images.githubusercontent.com/333265/180283928-5dc589ea-a22a-4bd0-98bf-bdba29da08c2.mp4


assets/package.json
```diff
@@ -77,6 +77,7 @@
     "typedoc-plugin-markdown": "^3.12.1",
     "typescript": "^4.6.4",
     "uuid": "^8.3.2",
+    "video-react": "^0.15.0",
```
marc-hughes, line 80: Using the video-react player to support advanced video features in the future.

assets/src/apps/app.tsx
```diff
@@ -5,19 +5,14 @@ import { configureStore } from 'state/store';
 import { Maybe } from 'tsmonad';
```
marc-hughes, line 5: Made a change to registerApplication so we could have a more lightweight application that doesn't get wrapped with a redux store. Used that for both the new VideoPlayer and the existing DarkModeSelector (but didn't do a full inventory of all apps to see if others could exist outside redux)

assets/src/components/editing/toolbar/FullScreenModal.tsx
```diff
@@ -1,7 +1,7 @@
 import React, { PropsWithChildren } from 'react';
 
 type Props = PropsWithChildren<{
-  children: JSX.Element;
+  children: React.ReactNode;
```
marc-hughes, line 4: This lets us have multiple children of FullScreenModal so you don't need an extra container element
```
<FullScreenModal>
  <button>1</button> 
  <button>2</button>
</FullScreenModal>
```

instead of

```
<FullScreenModal>
<div>
  <button>1</button> 
  <button>2</button>
</div>
</FullScreenModal>
```

assets/src/components/media/manager/MediaManager.tsx
```diff
@@ -121,6 +122,7 @@ export interface MediaManagerState {
   showDetails: boolean;
   error: Maybe<string>;
   filteredMimeTypes: string[] | undefined;
+  uploading: boolean;
```
marc-hughes, line 125: Added a new "Uploading File" message during uploads. Before, the user could still interact with the dialog during uploads, which could cause inconsistent results.

config/dev.exs
```diff
@@ -151,3 +151,7 @@ config :appsignal, :config, active: false
 # Configure AWS
 config :ex_aws,
   region: System.get_env("AWS_REGION", "us-east-1")
+
+config :ex_aws, :hackney_opts,
+  follow_redirect: true,
+  recv_timeout: 200_000
```
marc-hughes, line 157: Had to raise the timeout for uploading files to S3, only did it for dev, since hopefully live servers in the datacenter won't take long at all.

bsparks (approved):

darrensiegel (approved):

PR #2848, opened 2022-07-21
[MER-1321] Fix stripe charging incorrect amount on section payment

julmontesdeoca:
[MER-1321](https://eliterate.atlassian.net/browse/MER-1321)

### What does it change?

When we worked on #2726 and switched from calculating costs from product to always using the section cost, we missed updating the same logic which was duplicated in the Stripe provider. 

In the reported bug, when an admin updated the cost of a section that parts from a product:
- User would see e.g. 25 USD when paying, although Stripe would charge 50 USD
- This happened because `PaymentController.make_payment` used `Section.amount`, and `Stripe.init_payment` was still doing the determine_product + cost calculation to determine the amount.

### DIFF
--- a/lib/oli/delivery/paywall.ex
+++ b/lib/oli/delivery/paywall.ex
@@ -200,7 +200,7 @@ defmodule Oli.Delivery.Paywall do
   @doc """
     Given a section (blueprint) calculate the cost to use it for
     a specific institution, taking into account any product-wide and product-specific discounts
-    this instituttion has.
+    this institution has.
 
       Returns {:ok, %Money{}} or {:error, reason}
   """
@@ -211,21 +211,15 @@ defmodule Oli.Delivery.Paywall do
   ) do
     discounts =
       from(d in Discount,
-        where:
-          (is_nil(d.section_id) and d.institution_id == ^institution_id) or
-            (d.section_id == ^id and d.institution_id == ^institution_id),
+        where: d.institution_id == ^institution_id and (is_nil(d.section_id)) or (d.section_id == ^id),
         select: d
       )
       |> Repo.all()
 
     # Remove any institution-wide discounts if an institution and section specific discount exists
-    discounts =
-      case Enum.any?(discounts, fn d -> !is_nil(d.section_id) end) do
-        true ->
-          Enum.filter(discounts, fn d -> !is_nil(d.section_id) end)
-
-        false ->
-          discounts
+    discounts = case Enum.filter(discounts, fn d -> !is_nil(d.section_id) end) do
+        [] -> discounts
+        filtered_discounts -> filtered_discounts
       end
 
     # Now calculate the product cost, taking into account a discount
--- a/lib/oli/delivery/paywall.ex
+++ b/lib/oli/delivery/paywall.ex
@@ -198,28 +198,17 @@ defmodule Oli.Delivery.Paywall do
   end
 
   @doc """
-  Given a section (blueprint or enrollable), calculate the cost to use it for
-  a specific institution, taking into account any product-wide and product-specific discounts
-  this instituttion has.
+    Given a section (blueprint) calculate the cost to use it for
+    a specific institution, taking into account any product-wide and product-specific discounts
+    this instituttion has.
 
-  Returns {:ok, %Money{}} or {:error, reason}
+      Returns {:ok, %Money{}} or {:error, reason}
   """
-  def calculate_product_cost(
-        %Section{requires_payment: false},
-        _
-      ),
-      do: {:ok, Money.new(:USD, 0)}
-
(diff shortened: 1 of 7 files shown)


lib/oli/delivery/paywall.ex
```diff
@@ -198,28 +198,17 @@ defmodule Oli.Delivery.Paywall do
...
-        nil
-      ),
-      do: {:ok, amount}
-
-  def calculate_product_cost(
-        %Section{requires_payment: true, id: id, amount: amount},
-        %Institution{id: institution_id}
-      ) do
+  @spec section_cost_from_product(%Section{}, %Institution{}) :: {:ok, %Money{}} | {:error, any}
+   def section_cost_from_product(
```
julmontesdeoca, line 208: Renamed for clarity, from now on this will only be used when creating a section from a product. Since the correct amount is set at the section creation level, the amount is then always taken from the section directly.

lib/oli/delivery/paywall/providers/stripe.ex
```diff
@@ -68,40 +71,43 @@ defmodule Oli.Delivery.Paywall.Providers.Stripe do
...
+             "https://api.stripe.com/v1/payment_intents",
+             body,
+             headers.(private_secret)
+           ),
+         {:ok, intent} <- Poison.decode(body),
+         {:ok, _} <- Oli.Delivery.Paywall.create_pending_payment(user, section, attrs.(intent)) do
+      {:ok, intent}
+    else
+      {:ok, %HTTPoison.Response{} = response} ->
+        Logger.error("Failed request to Stripe payment intents, response: #{inspect(response)}")
```
julmontesdeoca, line 106: Re-organized this to be defensive when using `Application.fetch_env`, if the previous `Application.fetch_env!` did raise an exception, it wouldn't be handled properly.

mcass19 (approved):
LGTM, nice work 👍🏼 🚀

lib/oli/delivery/paywall.ex
```diff
@@ -198,28 +198,17 @@ defmodule Oli.Delivery.Paywall do
   end
 
   @doc """
-  Given a section (blueprint or enrollable), calculate the cost to use it for
-  a specific institution, taking into account any product-wide and product-specific discounts
-  this instituttion has.
+    Given a section (blueprint) calculate the cost to use it for
+    a specific institution, taking into account any product-wide and product-specific discounts
+    this instituttion has.
```
gastonabella, line 203: Typo: `institution`

lib/oli/delivery/paywall.ex
```diff
@@ -198,28 +198,17 @@ defmodule Oli.Delivery.Paywall do
...
-
-  def calculate_product_cost(
-        %Section{requires_payment: true, id: id, amount: amount},
-        %Institution{id: institution_id}
-      ) do
+  @spec section_cost_from_product(%Section{}, %Institution{}) :: {:ok, %Money{}} | {:error, any}
+   def section_cost_from_product(
+    %Section{requires_payment: true, id: id, amount: amount},
+    %Institution{id: institution_id}
+  ) do
```
gastonabella, line 211: A couple of comments about the logic inside this function:
1. Nit but I think we could simplify the `where` condition for getting the discounts
```elixir
where: d.institution_id == ^institution_id) and (is_nil(d.section_id) or d.section_id == ^id),
...
```
2. The logic to filter the discounts could be either an `if` instead of a `case`, or something like
```elixir
case Enum.filter(discounts, fn d -> !is_nil(d.section_id) end) do
  [] -> discounts
  filtered_discounts -> filtered_discounts
end
```

lib/oli/delivery/paywall/providers/stripe.ex
```diff
@@ -68,40 +71,43 @@ defmodule Oli.Delivery.Paywall.Providers.Stripe do
...
+        provider_type: :stripe,
+        section_id: section.id
+      }
+    end
+
+    with {:ok, [_, {:private_secret, private_secret}]} <-
+           Application.fetch_env(:oli, :stripe_provider),
+         {:ok, %{status_code: 200, body: body}} <-
+           http().post(
+             "https://api.stripe.com/v1/payment_intents",
```
gastonabella, line 97: Wdyt about moving this url to a module attribute?

gastonabella (commented):
Looks very good!

lib/oli/delivery/paywall.ex
julmontesdeoca: Gotcha, love the `case` on `Enum.filter` to avoid the unnecessary extra traversal

gastonabella (approved):
Thanks for the changes!

eliknebel (approved):

PR #2849, opened 2022-07-21
log-deployment.yml to allow for Jira integration

Charles-Jones-Unicon:
This parameterized file allows a Jira webhook to integrate deployment info into Jira tickets and dashboards


eliknebel:
@Charles-Jones-Unicon What's the intent of this action? It looks like a manual initiation is required, do we want to drive this logging on when releases are created? Or is the intent to have this be a manual step?

Charles-Jones-Unicon:
The action is called automatically by our deployment process.

eliknebel (approved):

PR #2862, opened 2022-07-25
[BUGFIX] none is not a valid phx-update value; MER-1332

bsparks:
this resolves refreshing or loading with the query string, still other issues persist


bsparks:
closing to work on alt approach

PR #2865, opened 2022-07-26
[FEATURE] [MER-1157] Improve lms grade sync failures handling

mcass19:
[MER-1157](https://eliterate.atlassian.net/browse/MER-1157)

## What's changed?
- Add a view to list the grades that failed to sync within a section
- Add an option to retry a specific one or retry all at once
  - For the bulk retry, did a reduce + specific insert (instead of insert_all) to preserve job uniqueness. If db fail to store a job, execution is halted
- Add `retryable` state to be compared for uniqueness, so we avoid having duplicates job if one it's running 
- Fix observer updates view to list all updates happening
- Fix `gettext` warning after some version upgrades

### DIFF
--- a/lib/oli/delivery/attempts/core.ex
+++ b/lib/oli/delivery/attempts/core.ex
@@ -6,6 +6,7 @@ defmodule Oli.Delivery.Attempts.Core do
   alias Oli.Repo
   alias Oli.Repo.{Paging, Sorting}
 
+  alias Oli.Accounts.User
   alias Oli.Delivery.Sections.Section
   alias Oli.Publishing.PublishedResource
   alias Oli.Resources.Revision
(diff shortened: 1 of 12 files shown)


lib/oli/delivery/attempts/core.ex
```diff
@@ -259,6 +260,57 @@ defmodule Oli.Delivery.Attempts.Core do
...
+        on: section_project_publication.publication_id == published_resource.publication_id,
+        join: revision in Revision,
+        on: revision.id == published_resource.revision_id and revision.resource_id == resource_access.resource_id,
+        where:
+          section.slug == ^section_slug and
+          section.status == :active and
+          revision.deleted == false and
+          revision.graded == true and
+          (not is_nil(resource_access.last_grade_update_id) and
+            resource_access.last_grade_update_id != resource_access.last_successful_grade_update_id),
```
julmontesdeoca, line 295: Could checking `last_grade_update_id` against `last_successful_grade_update_id` make it so that it does not retrieve a grade update failure when it's the fist one on the resource?

Tried making the test in `test/oli/delivery/attempts_test.exs:465` only insert a failing update, without a previous successful one, and `Attempts.get_failed_grade_sync_resource_accesses_for_section/1` does not return that failure.

Does that make sense, or do we always have a successful grade update to begin with?

mcass19: You are right, not taking into account that scenario. Changing it...

Thanksss !

julmontesdeoca (approved):
Looks great! 👏👏 
Thanks for looking into that query 🙌

gastonabella (approved):
LGTM!

eliknebel (approved):

PR #2870, opened 2022-07-27
Don't Merge - [Refactor] Clean up hooks in DeckLayoutView

marc-hughes:
@bsparks I was able to semi-reliably reproduce the race-condition bug on my local machine. Did a bunch of debugging, and saw that here, inside handleActivityReady sometimes getLocalizedStateSnapshot ends up returning NaN for one or more of the stage.XXXX.value and that trickles through the system eventually causes the input field to be empty.

I went through this component and fixed all the dependencies in all the useCallback and useEffect calls.

After doing that, I haven't been able to reproduce the bug at all.

I'm unclear on specifically why this (maybe) fixes the bug, and hate fixes that can't be clearly explained as to what was wrong.

Ways I was able to reproduce the bug:

1. Every time I went through keeping balance lesson, there would be some missing "starting C02" values sprinkled semi-randomly through the pages
2. On the `Where is the Carbon Dioxide coming from?` page, I'd see the bug in both number inputs and on the sliders:
   - Reloading it would sometimes cause the bug
   - Going backwards a screen and forwards would sometimes cause the bug
   - Quickly going backwards 2 screens and then forward 2 would very often cause the bug

I'm leaving this as draft / don't merge for now since I'd like to refactor these hooks to make them more readable, but thought seeing this might spark insight into the underlying cause.

### DIFF
--- a/assets/src/apps/delivery/layouts/deck/DeckLayoutView.tsx
+++ b/assets/src/apps/delivery/layouts/deck/DeckLayoutView.tsx
@@ -1,7 +1,7 @@
 /* eslint-disable react/prop-types */
 import chroma from 'chroma-js';
 import { ActivityState, PartResponse, StudentResponse } from 'components/activities/types';
-import React, { CSSProperties, useCallback, useEffect, useState } from 'react';
+import React, { CSSProperties, useCallback, useEffect, useMemo, useState } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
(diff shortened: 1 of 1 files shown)


bsparks:
I think if anything is the culprit it would be the save/submit part callbacks. the rest looks good to do but probably wouldn't make a difference there.

bsparks (approved):

PR #2871, opened 2022-07-27
[FEATURE] Add line items domain to registration

mcass19:
## What's changed?
- Add `line_items_domain` field to `regsitrations`
  - Is not required
- Update version library that will take into account this new field when configured


gastonabella (approved):

julmontesdeoca (approved):

lib/oli_web/templates/lti/register.html.eex
```diff
@@ -121,6 +121,7 @@
                     $('.auth_token_url').val(full_url(config, 'auth_token'));
                     $('.auth_login_url').val(full_url(config, 'auth_login'));
                     $('.auth_server').val(full_url(config, 'auth_server'));
+                    $('.line_items_service_domain').val(full_url(config, 'line_items_service_domain'));
```
eliknebel, line 124: Is it necessary to require this from an admin when registering? Or can we somehow update it from the lti launch params.

My concerns are 2, first we are requiring the admin to correctly provide another key piece of information and there appears to be no way to change/update this by the LMS admin. If we can automate this, I think that would be the best approach

eliknebel:
Is there a reasonable default we can set for this on registration like we do for auth_server? I imagine, esp. since this is not technically part of the LTI standard, some folks might be confused as to what to put here.

Screen Shot 2022-07-27 at 5 03 58 PM

ehilfer:
In most cases it will be left blank and the claims will be used to construct the line items service URL.  So the prompt could indicate that it can be skipped if not needed.  We've bee hand setting up LTI integrations with each of the institutions to work through these types of questions, so it doesn't need to be completely explained at this stage.

eliknebel (approved):

PR #2873, opened 2022-07-27
[MER-1358] Add support for array aud claims

gastonabella:
[MER-1358](https://eliterate.atlassian.net/browse/MER-1358)

### What's changed?
As the LTI 1.3 protocol states, when Torus `launch` endpoint is called by an LMS with an `id token` as a parameter, the `aud` field in the token can be either an array of strings or a string ([it is highly recommended to be a one-element-array](https://www.imsglobal.org/spec/security/v1p1#id-token)). Currently Torus only supports string `aud`s, and Schoology is sending an array so the app fails when launching from this LMS.

1. Accept `aud` claim with array format.
2. Bump [lti_1p3](https://github.com/Simon-Initiative/lti_1p3) library to version 0.4.4.


eliknebel (approved):
Looks good, tested locally and works as expected with Canvas

PR #2875, opened 2022-07-28
[JAN-1827] Fixes one cause a race condition

marc-hughes:
Additional debugging in this issue has come up with the following observations:

When I'm seeing the bug occur, in defaultGlobalEnv from scripting.ts, the value of the input is set to NaN, and it continues to be NaN long after the initial render, so it doesn't seem to be a case of the component not getting an updated value.

I put some debugging statements into evalScript and we're running this script:
`let {q:1461614198663:472|stage.VolcanicCChange.value} = NaN;`

When the bug occurs, it runs after this script which sets it to the correct value.
`let {q:1461614198663:472|stage.VolcanicCChange.value} = 4.7;`

So something is explicitly setting the value to NaN

Sometimes that script runs before, and sometimes it runs after the debug message:
DECK HANDLE READY (ALL ACTIVITIES DONE INIT) 

I've traced up where the `let {q:1461614198663:472|stage.VolcanicCChange.value} = NaN;` script gets run from, and it appears to be generated from the getAssignStatements in savePart.ts around line 48

This leads me to believe this might not be a systemic problem in adaptive lessons, but more a bug in individual components within them. We've seen similar problems on number inputs and in sliders, but it may be more widespread if the same patterns were followed in other components.

Sometimes, the `let {q:1461614198663:472|stage.VolcanicCChange.value} = NaN;` script runs with 2 other assignments, and sometimes it's alone. This corresponds to two different symptoms of the bug:

1. the input will be empty & enabled - I think this is caused by an init coming from that component after we've set the value
2. the input will be empty but disabled - This is caused by a debounced onSave coming through with an empty value after the initial value was set

This PR should fix the second source of the bug. There's a very slim chance it fixes the first, since I also cleaned up the dependencies around initialize.

This does not fix the bug in any component except the NumberInput, if this works, we'll have to do a similar refactor to each affected component

I've been unable to positively identify the first cause of the bug, so it probably is still possible.

### DIFF
--- a/assets/src/adaptivity/scripting.ts
+++ b/assets/src/adaptivity/scripting.ts
@@ -222,6 +222,9 @@ export const getExpressionStringForValue = (
     // val = convertExponentialToDecimal(val);
     val = parseFloat(val);
     if (val === '') {
+      // TODO: Figure out the intent here.
+      // If val == '' then parseFloat will return NaN before it
+      // hits this block, so we'll never get here.
       val = 'null';
     }
   }
--- a/assets/src/components/parts/janus-input-number/InputNumber.tsx
+++ b/assets/src/components/parts/janus-input-number/InputNumber.tsx
@@ -1,5 +1,12 @@
 import debounce from 'lodash/debounce';
-import React, { CSSProperties, ReactEventHandler, useCallback, useEffect, useState } from 'react';
+import React, {
+  CSSProperties,
+  ReactEventHandler,
+  useCallback,
+  useEffect,
+  useMemo,
+  useState,
+} from 'react';
 import { CapiVariableTypes } from '../../../adaptivity/capi';
 import {
   NotificationType,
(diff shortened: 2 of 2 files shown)


assets/src/adaptivity/scripting.ts
```diff
@@ -222,6 +222,9 @@ export const getExpressionStringForValue = (
     // val = convertExponentialToDecimal(val);
     val = parseFloat(val);
     if (val === '') {
+      // TODO: Figure out the intent here.
```
marc-hughes, line 225: This is why the script sets the value to NaN, the non-functional code would make me think we should set it to null instead, but left it as-is for now.

assets/src/components/parts/janus-input-number/InputNumber.tsx
```diff
@@ -11,25 +18,93 @@ import { PartComponentProps } from '../types/parts';
...
+  id,
+  model: inputModel,
+  notify,
+  onSave,
+  onInit,
+  onReady,
+  onResize,
+}) => {
+  //const debugRef = useRef(props);
+  const model = useMemo(() => maybeParseJson(inputModel), [inputModel]);
```
marc-hughes, line 51: We were using an awkward useState / useEffect combo before for this derived value that would have been causing extra renders and had some weird timing.

assets/src/components/parts/janus-input-number/InputNumber.tsx
```diff
@@ -11,25 +18,93 @@ import { PartComponentProps } from '../types/parts';
...
 
+  const [ready, setReady] = useState<boolean>(false);
   const [inputNumberValue, setInputNumberValue] = useState<string | number>('');
   const [enabled, setEnabled] = useState(true);
-  const [cssClass, setCssClass] = useState('');
```
marc-hughes, line 22: This was never used.

assets/src/components/parts/janus-input-number/InputNumber.tsx
```diff
@@ -120,81 +198,35 @@ const InputNumber: React.FC<PartComponentProps<InputNumberModel>> = (props) => {
...
-        setState(pState);
-      } catch (err) {
-        // bad json, what do?
-      }
-    }
-    if (!pModel) {
-      return;
-    }
-    initialize(pModel);
-  }, [props]);
```
marc-hughes, line 167: Passing [props] causes this to re-run every time any prop changes, probably not what we wanted.
I switched to dereferencing the attributes of props and using them directly in dependencies to more cleanly resolve that sort of issue.

assets/src/components/parts/janus-input-number/InputNumber.tsx
```diff
@@ -120,81 +198,35 @@ const InputNumber: React.FC<PartComponentProps<InputNumberModel>> = (props) => {
...
+  }, [id, notify]);
 
   useEffect(() => {
-    if (!ready) {
-      return;
-    }
-    props.onReady({ id, responses: [] });
```
marc-hughes, line 173: Didn't see a reason we couldn't just call onReady after calling setReady(true) above, fewer useEffect calls seems better than more.

assets/src/components/parts/janus-input-number/InputNumber.tsx
```diff
@@ -204,50 +236,40 @@ const InputNumber: React.FC<PartComponentProps<InputNumberModel>> = (props) => {
...
-          type: CapiVariableTypes.NUMBER,
-          value: val,
-        },
-      ],
-    });
-  };
-
   const handleOnChange: ReactEventHandler<HTMLInputElement> = (event) => {
-    const val = (event.target as HTMLInputElement).value;
+    const val = sanitizeValue((event.target as HTMLInputElement).value);
```
marc-hughes, line 268: This is the main point of this PR.
Before, the user would type, we would call setInputNumberValue
Later, an effect would see that value changed and call debounceSave 
Eventually that would trigger and the value would get passed up.

Unfortunately, we call setInputNumberValue for several other reasons that are not user-input. Each of those would cause a value change, effect trigger, and a debounceSave

We would also always get a debounceSave on the initial rendering since the value was new.

Now, we directly call debounceSave from the change handler, so only user input trigers it.

assets/src/components/parts/janus-input-number/InputNumber.tsx
```diff
@@ -60,19 +135,26 @@ const InputNumber: React.FC<PartComponentProps<InputNumberModel>> = (props) => {
...
+
     //Instead of hardcoding REVIEW, we can make it an global interface and then importa that here.
     if (initResult.context.mode === contexts.REVIEW) {
       setEnabled(false);
     }
-    setReady(true);
-  }, []);
+    setReady((currentValue) => {
+      if (currentValue) {
+        console.warn(
```
marc-hughes, line 145: I think, initialize() getting called multiple times, is the source of the bug where the input will be empty & enabled.

I wasn't able to figure out why that might be happening. Added this console warning so if we see the bug happen, and this message is in the console, it will point to that being the cause.

assets/src/components/parts/janus-input-number/InputNumber.tsx
```diff
@@ -11,25 +18,93 @@ import { PartComponentProps } from '../types/parts';
 import './InputNumber.scss';
 import { InputNumberModel } from './schema';
 
-const InputNumber: React.FC<PartComponentProps<InputNumberModel>> = (props) => {
-  const [state, setState] = useState<any[]>(Array.isArray(props.state) ? props.state : []);
```
marc-hughes, line 15: We never read the value of state, so dropped it.

dtiwarATS (approved):

PR #2879, opened 2022-07-29
[BUGFIX] history rendering breaks when returning to the present

bsparks:


assets/src/apps/delivery/layouts/deck/DeckLayoutView.tsx
```diff
@@ -451,12 +451,47 @@ const DeckLayoutView: React.FC<LayoutProps> = ({ pageTitle, pageContent, preview
...
+      }
+      return currentActivityTree.map((activity) => ({
+        ...activity,
+        activityKey: historyModeNavigation
+          ? `${activity.id}_${currentActivity.id}_history`
+          : activity.id,
+      }));
+    });
+  }, [currentActivityTree, historyModeNavigation]);
+
```
marc-hughes, line 488: The useState/useEffect pattern there scares me a little, there's going to be a bunch of extra render cycles baked in because of it that makes understanding the timing hard.

This doesn't perfectly replicate the behavior because the previous one had a currentLocalActivity from the previous render. I'm not sure if that was actually required or not, we couple replicate it with a ref it if it is, but it'll avoid that issue

```suggestion
   const localActivityTree = useMemo(() => {
    if (!currentActivityTree) {
      return null;
    }

    const currentActivity = currentActivityTree[currentActivityTree.length - 1];

    return currentActivityTree
      ? currentActivityTree.map((activity) => ({
          ...activity,
          activityKey: historyModeNavigation
            ? `${activity.id}_${currentActivity.id}_history`
            : activity.id,
        }))
      : null;
  }, [currentActivityTree, historyModeNavigation]);
```

marc-hughes (approved):

PR #2882, opened 2022-08-02
[MER-1368] Persistent session keep alive improvements

julmontesdeoca:
[MER-1368](https://eliterate.atlassian.net/browse/MER-1368)

### What does it change?

Improves the calls to the keep alive endpoint by:
- Reducing frequency from 1 minute to 15 (default Pow refresh time)
- Moving keep-alive script to liveviews only, instead of loading them on every page.

### DIFF
--- a/assets/src/phoenix/keep-alive.ts
+++ b/assets/src/phoenix/keep-alive.ts
@@ -16,11 +16,15 @@
  * More Info: https://github.com/danschultzer/pow/issues/271
  * https://elixirforum.com/t/get-user-id-with-pow-from-session-for-live-view/24206/4
  */
+
+// Wait time set to 15 min which is the deafult time set by Pow to renew sessions (session_ttl_renewal)
+const waitTime = 900000 // 15 minutes in ms
+
 if (!window.keepAlive) {
   window.keepAlive = () => {
-    const scriptTag = document.getElementById('keep-alive');
-    const userType = scriptTag ? scriptTag.getAttribute('data-type') : 'user';
-    const keepAliveUrl = `${userType == 'author' ? '/authoring' : ''}/keep-alive`;
+    // Check for which template (delivery or authoring) we're in
+    const is_authoring = $("#layout-id").data("layout-id") === "authoring";
+    const keepAliveUrl = `${is_authoring ? '/authoring' : ''}/keep-alive`;
 
     const wait = (ms: number) => {
       return () =>
@@ -30,7 +34,7 @@ if (!window.keepAlive) {
     };
 
     fetch(keepAliveUrl)
-      .then(wait(60 * 1000 /*ms*/))
+      .then(wait(waitTime))
       .then(window.keepAlive);
   };
 
--- a/assets/src/phoenix/keep-alive.ts
+++ b/assets/src/phoenix/keep-alive.ts
@@ -17,14 +17,14 @@
  * https://elixirforum.com/t/get-user-id-with-pow-from-session-for-live-view/24206/4
  */
 
-// Wait time set to 15 min which is the deafult time set by Pow to renew sessions (session_ttl_renewal)
-const waitTime = 900000 // 15 minutes in ms
+// Wait time to renew session set to 10 min. Default Pow session rewnew is at 15 min (session_ttl_renewal)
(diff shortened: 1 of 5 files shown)


assets/src/phoenix/keep-alive.ts
```diff
@@ -16,11 +16,15 @@
...
+// Wait time set to 15 min which is the deafult time set by Pow to renew sessions (session_ttl_renewal)
+const waitTime = 900000 // 15 minutes in ms
+
 if (!window.keepAlive) {
   window.keepAlive = () => {
-    const scriptTag = document.getElementById('keep-alive');
-    const userType = scriptTag ? scriptTag.getAttribute('data-type') : 'user';
-    const keepAliveUrl = `${userType == 'author' ? '/authoring' : ''}/keep-alive`;
+    // Check for which template (delivery or authoring) we're in
+    const is_authoring = $("#layout-id").data("layout-id") === "authoring";
```
mcass19, line 26: NIT: camel case? 😬

lib/oli_web/views/layout_view.ex
```diff
@@ -26,6 +26,7 @@ defmodule OliWeb.LayoutView do
   alias OliWeb.Breadcrumb.BreadcrumbTrailLive
   alias Oli.Delivery.Paywall.AccessSummary
   alias Oli.Delivery.Sections
+  alias OliWeb.Router.Helpers, as: Routes
```
mcass19, line 29: Was this necessary to add? Thought that we already has access to `Routes` in the layout

assets/src/phoenix/keep-alive.ts
```diff
@@ -16,11 +16,15 @@
  * More Info: https://github.com/danschultzer/pow/issues/271
  * https://elixirforum.com/t/get-user-id-with-pow-from-session-for-live-view/24206/4
  */
+
+// Wait time set to 15 min which is the deafult time set by Pow to renew sessions (session_ttl_renewal)
```
mcass19, line 20: Reading pow [session expiration](https://hexdocs.pm/pow/Pow.Plug.Session.html#module-session-expiration) again, wondering if we have to set it exactly to 15 minutes, or would be safer to do it a few seconds/minute earlier so we don't hit an edge case where the session expires for not having a request in between the 15 min of `session_ttl_renewal` and the `ttl` of the session itself. Anyway, we can discuss offline if it's easier (or if I'm thinking it wrong 😄 )

mcass19 (commented):
Great job!!! Thank you 🚀 

Just really small comments, and would be great if we can fix the `lint` failing because of minor things

assets/src/phoenix/keep-alive.ts
gastonabella, line 20: Nit: `default`

gastonabella: Same concern here, and I believe we should be fine with just reducing the time

assets/src/phoenix/keep-alive.ts
julmontesdeoca: JS 🐫

assets/src/phoenix/keep-alive.ts
julmontesdeoca: For what I read we could even set it up to 29 min and it wouldn't expire. The example it gives where it would invalidate the session in 14 minutes, is if we explicitly set up the expiration for 15 minutes. I ended up choosing 15 because is the time Pow uses to refresh, so we should be safe using the same wait time as them.
I also tried this locally setting the keep-alive to call at 25 minutes and it did work as expected.
Either way, we can move it to 10 - 14 min.

bsparks:
I wonder if while we're in here if we shouldn't add this to the adaptive author (advanced.html.eex)

julmontesdeoca:
> I wonder if while we're in here if we shouldn't add this to the adaptive author (advanced.html.eex)

Good question! Quickly checked and it seems to me that that template it's always rendered inside either `authoring.html` or `delivery.html`. Also looked at the page source code when editing/viewing adaptive lessons and a `<input type="hidden" id="layout-id" data-layout-id="...">` is present.

mcass19 (approved):

bsparks (approved):

PR #2883, opened 2022-08-02
[MER-1389] Fix discounts being applied to all institutions

julmontesdeoca:
[MER-1389](https://eliterate.atlassian.net/browse/MER-1389)

### What does it change?

Fixes the Paywall.section_cost_from_product/2 query retrieving all discounts for a product instead of limiting to a specific institution.

### DIFF
--- a/lib/oli/delivery/paywall.ex
+++ b/lib/oli/delivery/paywall.ex
@@ -211,8 +211,9 @@ defmodule Oli.Delivery.Paywall do
   ) do
     discounts =
       from(d in Discount,
-        where: (d.institution_id == ^institution_id and is_nil(d.section_id))  # Institution-wide discounts
-          or (d.section_id == ^id and d.institution_id == ^institution_id), # Section-specific discounts for the given institution
+        where: d.institution_id == ^institution_id and
+          (is_nil(d.section_id) or # Institution-wide discounts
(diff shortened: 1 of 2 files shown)


lib/oli/delivery/paywall.ex
```diff
@@ -211,7 +211,8 @@ defmodule Oli.Delivery.Paywall do
   ) do
     discounts =
       from(d in Discount,
-        where: d.institution_id == ^institution_id and (is_nil(d.section_id)) or (d.section_id == ^id),
+        where: (d.institution_id == ^institution_id and is_nil(d.section_id))  # Institution-wide discounts
+          or (d.section_id == ^id and d.institution_id == ^institution_id), # Section-specific discounts for the given institution
```
gastonabella, line 215: This works, but I think the bug is happening because of the lack of parentheses:

```elixir
where: d.institution_id == ^institution_id and (is_nil(d.section_id) or d.section_id == ^id),
```

gastonabella (approved):

lib/oli/delivery/paywall.ex
julmontesdeoca: Exactly! Kept the long form for readability, double-checked and if we keep the comments I believe this is more idiomatic. Updating.

bsparks (approved):

mcass19 (approved):

Issue #2744, opened 2022-06-28
Support advanced table style rendering and authoring MER-1309


darrensiegel:
Add support for the following advanced formatting options:

- [x] Row span
- [x] Column span
- [x] Cell alignment
- [x] Row style 
- [x] Overall border style

The first four items come from the legacy OLI implementation.  Their definition follows. In addition, tables need a setting to control the overall "cell border style".  Specifically we need two options "full borders" and "hidden".  The former is what is implemented right now, the latter would render a table with *all* cell and row borders hidden (essentially allowing a table to be used as a layout mechanism)

```
<!-- =================== Table ============================================ -->

<!-- TABLE CELL ATTRIBUTES -->
<!ENTITY % olcml.cell.alignments "left | right | center">

<!ENTITY % olcml.table.align.attr
 "align			(%olcml.cell.alignments;)	#IMPLIED"
>

<!ENTITY % olcml.table.colspan.attr
 "colspan		CDATA						#IMPLIED"
>

<!ENTITY % olcml.table.rowspan.attr
 "rowspan		CDATA						#IMPLIED"
>

<!ENTITY % table.cell.attrs
 "%olcml.core.title.attr;
  %olcml.i18n.attrs;
  %olcml.table.align.attr;
  %olcml.table.colspan.attr;
  %olcml.table.rowspan.attr;"
>


<!-- TABLE HEAD -->
<!ENTITY % olcml.th.qname "%olcml.pfx;th">
<!ELEMENT %olcml.th.qname; (%olcml.inline.content;)*>
<!ATTLIST %olcml.th.qname;
  %table.cell.attrs;
>


<!-- TABLE DATA -->
<!ENTITY % olcml.td.qname "%olcml.pfx;td">
<!ELEMENT %olcml.td.qname; (%olcml.inline.content;)*>
<!ATTLIST %olcml.td.qname;
  %table.cell.attrs;
>


<!-- TABLE ROW -->
<!ENTITY % olcml.row.styles "alternating | plain">

<!ENTITY % table.rowstyle.attr
 "rowstyle		(%olcml.row.styles;)		#IMPLIED"
>

<!ENTITY % olcml.tr.qname "%olcml.pfx;tr">
<!ELEMENT %olcml.tr.qname; (%olcml.th.qname; | %olcml.td.qname;)+>


<!-- TABLE -->
<!ELEMENT %olcml.table.qname; ((%olcml.labeled.content;), %olcml.tr.qname;+)>
[...]

PR #2894, opened 2022-08-04
[MER-1309] New formatting options for tables.

marc-hughes:
Implements github issue #2744 

This adds several new formatting options to tables and the ability to merge/spit the cells within a table.

-  Row span
-  Column span
-  Cell alignment
-  Row style
-  Overall border style

Demo of new options:
https://user-images.githubusercontent.com/333265/182847666-c0b2538f-58d5-4773-9163-1fa0bc7b5cf9.mp4


assets/package.json
```diff
@@ -13,7 +13,7 @@
     "lint": "eslint --quiet \"src/**/*.ts*\"",
     "format": "eslint --fix --quiet \"src/**/*.ts*\"",
     "test": "jest",
-    "test:watch": "jest --watch -o",
```
marc-hughes, line 16: Found it really annoying we had a coverage report during test:watch that would scroll results off screen while working on tests, hope nobody minds me removing this here.

assets/src/components/editing/editor/normalizers/normalizer.ts
```diff
@@ -38,13 +38,12 @@ export function installNormalizer(editor: Editor, context: NormalizerContext = {
           Transforms.wrapNodes(editor, Model.p(), { at: path });
           return;
         }
```
marc-hughes, line 40: These normalizers were not running on top level elements. Pretty sure they weren't meant to be in the else block.

I had this change in it's own PR and had asked for comments in case I was missing something big there:
https://github.com/Simon-Initiative/oli-torus/pull/2890

assets/src/components/editing/editor/normalizers/tables.ts
```diff
@@ -27,16 +73,17 @@ export const normalize = (editor: Editor, node: ModelElement | FormattedText, pa
     });
 
     if (anyDiffer) {
-      node.children.forEach((row, index: number) => {
-        const children = row.children;
-        let count = children.length;
+      console.warn('Normalizing table');
```
marc-hughes, line 76: What do we think of having console messages during the normalization process when it makes changes? (this is a wider question than this PR)

A couple times during dev, I ran into problems where the normalizer ran over and over and there was no debug output for why that was happening.

assets/src/components/editing/elements/table/TableDropdownMenu.tsx
```diff
@@ -1,24 +1,38 @@
...
 interface Props {
   editor: Editor;
-  model: TableData | TableHeader;
+  model: TableCell;
 }
-export const DropdownMenu = (props: Props) => {
-  const ref = useRef();
-  const { editor, model } = props;
-
-  // There has to be a better way to do this.
```
marc-hughes, line 16: IMO it's better to use a reference function instead of a reference + useEffect for stuff like this. Should be safer and result in one less render call.

assets/src/components/editing/elements/table/table-util.ts
```diff
@@ -0,0 +1,60 @@
...
+    className: attrs.align ? `text-${attrs.align}` : undefined,
+    colSpan: attrs.colspan || undefined,
+    rowSpan: attrs.rowspan || undefined,
+  };
+};
+
+export const getColspan = (cell: ContentModel.TableCell): number => cell.colspan || 1;
+export const getRowspan = (cell: ContentModel.TableCell): number => cell.rowspan || 1;
+
+/**
```
marc-hughes, line 19: I tried a few implementations of the whole merge/split trying to just use the slate location's, that turned out to be hard to get right and really hard to read. Came up with a solution to just calculate the visual grid and that made everything a lot easier.

assets/styles/authoring/model-editors.scss
```diff
@@ -16,13 +16,19 @@
       word-wrap: break-word;
       overflow: visible;
 
+      td {
```
marc-hughes, line 19: This moves that table dropdown button into the corner of the cell instead of the corner of the table. It seemed like a nice visual improvement to me.

darrensiegel (approved):

gjbunyea:
Issue found: Toggle header removes options button from that cell, making it impossible to untoggle header, apply other formatting options (center align text), etc...

If you'd prefer I file this as a separate bug ticket, let me know.

toggleheader

PR #2897, opened 2022-08-04
[MER-1340] Accept string based widths on images

marc-hughes:
img elements have a schema that accepts strings or numbers for width.

✔️ Both strings and numbers worked fine when rendered via elixir.
❌ The client side rendering only accepted numbers, and effectively ignored strings.

After this change, string or number widths work in both cases.

### DIFF
--- a/assets/src/components/editing/elements/image/block/ImageElement.tsx
+++ b/assets/src/components/editing/elements/image/block/ImageElement.tsx
@@ -11,11 +11,18 @@ import { ImageSettings } from 'components/editing/elements/image/ImageSettings';
 
 interface Props extends EditorProps<ContentModel.ImageBlock> {}
 
+const percentageWidth = /^[0-9]+%$/;
+
 const cssWidth = (width: number | string | undefined) => {
   if (!width) return undefined;
-  if (typeof width === 'number') return width; // Numbers are fine, they imply pixels
+  if (typeof width === 'number') return width;
+  if (typeof width !== 'string') return undefined;
 
-  const pixels = parseInt(width, 10);
+  if (percentageWidth.test(width)) {
+    return width;
+  }
+
+  const pixels = parseFloat(width);
   if (isNaN(pixels)) return undefined;
   return pixels;
 };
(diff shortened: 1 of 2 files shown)


assets/src/components/editing/elements/image/block/ImageElement.tsx
```diff
@@ -10,6 +10,16 @@ import { ImagePlaceholder } from 'components/editing/elements/image/block/ImageP
 import { ImageSettings } from 'components/editing/elements/image/ImageSettings';
 
 interface Props extends EditorProps<ContentModel.ImageBlock> {}
+
+const cssWidth = (width: number | string | undefined) => {
+  if (!width) return undefined;
```
bsparks, line 15: what if width === 0 ?

assets/src/components/editing/elements/image/block/ImageElement.tsx
```diff
@@ -10,6 +10,16 @@ import { ImagePlaceholder } from 'components/editing/elements/image/block/ImageP
...
+  if (typeof width === 'number') return width; // Numbers are fine, they imply pixels
+
+  const pixels = parseInt(width, 10);
```
bsparks, line 18: do we not allow `50%` ?

marc-hughes: oh! I don't know. Do we allow percentages and/or units?

I guess either of these could be valid:

50%
100px
10em
2vw

I'm going to go fix that.

marc-hughes: The phoenix/elixir side only does numbers and percent, so I updated this to match.

assets/src/components/editing/elements/image/block/ImageElement.tsx
```diff
@@ -10,6 +10,24 @@ import { ImagePlaceholder } from 'components/editing/elements/image/block/ImageP
...
+
+const percentageWidth = /^[0-9]+%$/;
+
+const cssWidth = (width: number | string | undefined) => {
+  if (!width) return undefined;
+  if (typeof width === 'number') return width; // Numbers are fine, they imply pixels
+  if (typeof width !== 'string') return undefined; // Numbers or strings only
+
+  if (percentageWidth.test(width)) {
+    /* It's a percent based width */
```
bsparks, line 22: seems a little self documenting above IMO, but this is a nitpick

bsparks (approved):

PR #2903, opened 2022-08-04
[FEATURE] [MER-1355] Instructor preview refactor

mcass19:
[MER-1355](https://eliterate.atlassian.net/browse/MER-1355)

Refactor instructor preview to make it work for all pages and add couple of tests to cover that functionality.

### DIFF
--- a/lib/oli_web/plugs/authorize_section_preview.ex
+++ b/lib/oli_web/plugs/authorize_section_preview.ex
@@ -17,15 +17,17 @@ defmodule Oli.Plugs.AuthorizeSectionPreview do
     cond do
       Sections.is_instructor?(user, section_slug) ->
         conn
+
       not is_nil(user) and Sections.is_enrolled?(user.id, section_slug) ->
         redirect_path =
           conn
           |> current_path()
-          |> String.replace("preview/", "")
+          |> String.replace("/preview/", "/")
 
         conn
         |> redirect(to: redirect_path)
         |> halt()
+
       true ->
         conn
         |> put_view(OliWeb.PageDeliveryView)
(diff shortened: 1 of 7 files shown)


mcass19:
@darrensiegel and/or @eliknebel can you please take a look at this one? Thank you!

lib/oli_web/plugs/authorize_section_preview.ex
```diff
@@ -0,0 +1,37 @@
...
+    section_slug = conn.path_params["section_slug"]
+
+    cond do
+      Sections.is_instructor?(user, section_slug) ->
+        conn
+      not is_nil(user) and Sections.is_enrolled?(user.id, section_slug) ->
+        redirect_path =
+          conn
+          |> current_path()
+          |> String.replace("preview/", "")
```
darrensiegel, line 24: This would be a problematic way to redirect away student attempts in the case where the section slug ends with `"preview"`. Consider the following example:

`"https://proton.oli.cmu.edu/sections/biology_preview/preview/page/intro"`



The `String.replace` would strip out a portion of the section slug.

darrensiegel (changes requested):

lib/oli_web/plugs/authorize_section_preview.ex
ehilfer: @darrensiegel, while @mcass19 is out, could you suggest a more ironclad search/replace pattern that would work, or should we look at a different strategy altogether?

mcass19: Yeah good catch @darrensiegel, wasn't completely sure if that would work for all cases.

Did the plug in this way to not have to write the redirect logic in each preview method. Will think if there's something safer we could do here or will re write it to be in each route specific.

darrensiegel: Yeah i think including the leading `/` is good enough.  That will avoid replacing any other component of the URL

simonchoxx: @darrensiegel the last changes were pushed, please take a look whenever you can.

PR #2904, opened 2022-08-05
[CHORE] Super activity files

rgachuhi:
This PR switches OLI superactivity support file references to use flattened hash-based S3 media files generated by the course digest tool.


config/dev.exs
```diff
@@ -55,12 +55,12 @@ config :oli, OliWeb.Endpoint,
     host: System.get_env("HOST", "localhost"),
     port: String.to_integer(System.get_env("PORT", "443"))
   ],
-  https: [
-    port: String.to_integer(System.get_env("HTTPS_PORT", "443")),
-    otp_app: :oli,
-    keyfile: System.get_env("SSL_KEY_PATH", "priv/ssl/localhost.key"),
-    certfile: System.get_env("SSL_CERT_PATH", "priv/ssl/localhost.crt")
-  ],
+  # https: [
```
eliknebel, line 58: is leaving this commented intentional? If so maybe we should just remove it altogether

eliknebel (changes requested):

config/dev.exs
eliknebel: I think https is necessary in the dev environment due to CSRF issues, this config should guarantee that. I'm wondering if removing this might cause other issues

rgachuhi: urgh. Nope. Removed it.

rgachuhi: Commented it out just for local purpose

eliknebel (approved):

PR #2906, opened 2022-08-05
[Bugfix] PMP-2988 JAN-1836 user state queue

bsparks:
this helps with an everapp race condition. everapps (observer) that are constantly polling the data previously might have executed a get request while a set request was still processing.


dsuriano (approved):
Have you seen this new `Unchanged files with check annotations` section during review? Neato.

image

PR #2908, opened 2022-08-05
[PERFORMANCE] Optimize resource_access fetch with gradebook LV

darrensiegel:
The query to retrieve the resource access records for all graded pages, for a window of 25 users was taking long enough time to trip the 10 second Phoenix socket timeout during the Gradebook LV init. 

Analyzing the query showed that it had:

```
"Startup Cost": 153.25,
"Total Cost": 270.65
```

This was primarily due to how the query accounted for the specific set of users, given the join through section slug to SPP to PR to revision. 

I simplified and reworked the join approach to lower that to:

```
"Startup Cost": 8.53,
"Total Cost": 41.99
```

### DIFF
--- a/lib/oli/delivery/attempts/core.ex
+++ b/lib/oli/delivery/attempts/core.ex
@@ -224,37 +224,31 @@ defmodule Oli.Delivery.Attempts.Core do
 
   `[%ResourceAccess{}, ...]`
   """
-  def get_graded_resource_access_for_context(section_slug) do
+  def get_graded_resource_access_for_context(section_id) do
     Repo.all(
-      from(a in ResourceAccess,
-        join: s in Section,
-        on: a.section_id == s.id,
-        join: spp in SectionsProjectsPublications,
-        on: s.id == spp.section_id,
+      from(spp in SectionsProjectsPublications,
         join: pr in PublishedResource,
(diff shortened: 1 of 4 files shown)


lib/oli/delivery/attempts/core.ex
```diff
@@ -224,37 +224,31 @@ defmodule Oli.Delivery.Attempts.Core do
 
   `[%ResourceAccess{}, ...]`
   """
-  def get_graded_resource_access_for_context(section_slug) do
+  def get_graded_resource_access_for_context(section_id) do
```
mcass19, line 227: Wdyt about writing a single query and reusing it in both methods since they are exactly the same? And in the below one, we can add the extra where condition dynamically

mcass19 (approved):

lib/oli/delivery/attempts/core.ex
darrensiegel: Oh yeah good call.  @mcass19 take another look, I reworked this into a base, composable query that adds in the user id condition only for the 2-arity instance of the function.

mcass19: Looks great!

rgachuhi (approved):

PR #2913, opened 2022-08-10
PMP-2990 | JAN-1838

dtiwarATS:
This should unblock user and they can continue on further screens. We already have another ticket (https://jira.unicon.net/jira/browse/PMP-2998) to fix some use cases.


bsparks:
closed to redo for update

PR #2914, opened 2022-08-10
MER-1401 bringing fix from #2900 for nil check over to hotfix branch

bsparks:


Charles-Jones-Unicon (approved):

mcass19 (commented):
Maybe we should include this one as well https://github.com/Simon-Initiative/oli-torus/pull/2900/files#diff-de3c500aa015e6ad547f266149fc64c940b34aa69419ff94c893e556f993f0b3R66

bsparks:
> Maybe we should include this one as well https://github.com/Simon-Initiative/oli-torus/pull/2900/files#diff-de3c500aa015e6ad547f266149fc64c940b34aa69419ff94c893e556f993f0b3R66

hmm yes good catch, we probably should

PR #2920, opened 2022-08-11
[BUG] [MER-1403] [#2918] MathJax config fix.

### LINKED ISSUE #2918: Latex rendering not configured correctly to render all Latex commands
1. Create a Latex Formula in a page.
2. Use the following source:

```
\require{cancel}\mathrm{mol}\;\mathrm{HCl}\;\mathrm{produced}\;=\;3\;\cancel{\mathrm{mol}\;{\mathrm H}_2}\;\times\;\frac{2\;\mathrm{mol}\;\mathrm{HCl}}{1\;\cancel{\mathrm{mol}\;{\mathrm H}_2}}\;\;=\;6\;\mathrm{mol}\;\mathrm{HCl}
```

In authoring, preview and delivery this formula incorrectly renders as:

Screen Shot 2022-08-11 at 8 41 35 AM

marc-hughes:
Fixes #2918

A tweak to our mathjax configuration to support the \require command which lets the \cancel command work in this sample formula.

```
\require{cancel}\mathrm{mol}\;\mathrm{HCl}\;\mathrm{produced}\;=\;3\;\cancel{\mathrm{mol}\;{\mathrm H}_2}\;\times\;\frac{2\;\mathrm{mol}\;\mathrm{HCl}}{1\;\cancel{\mathrm{mol}\;{\mathrm H}_2}}\;\;=\;6\;\mathrm{mol}\;\mathrm{HCl}
```

After fix applied:

image

While at it, I put this config in a phoenix component since it was copy/pasted in 3 different html files.

### DIFF
--- a/lib/oli_web/common/mathjax_script.ex
+++ b/lib/oli_web/common/mathjax_script.ex
@@ -1,0 +1,26 @@
+defmodule OliWeb.Common.MathJaxScript do
+  use Phoenix.Component
+
+  def render(assigns) do
+    ~H"""
+    <script>
+    window.MathJax = {
+      tex: {
+        inlineMath: [ ["\\(","\\)"] ],
+        displayMath: [ ['$$','$$'], ["\\[","\\]"] ],
+        processEscapes: true,
+        packages: ['base', 'ams', 'noerrors', 'noundefined', 'require']
+      },
+      options: {
+        ignoreHtmlClass: 'tex2jax_ignore',
+        processHtmlClass: 'tex2jax_process'
+      },
+      loader: {
+        load: ['[tex]/noerrors']
+      }
+    };
+    </script>
+    <script src="https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-mml-chtml.js" id="MathJax-script"></script>
+    """
+  end
+end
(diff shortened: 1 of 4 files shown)


lib/oli_web/common/mathjax_script.ex
```diff
@@ -0,0 +1,26 @@
...
+
+  def render(assigns) do
+    ~H"""
+    <script>
+    window.MathJax = {
+      tex: {
+        inlineMath: [ ["\\(","\\)"] ],
+        displayMath: [ ['$$','$$'], ["\\[","\\]"] ],
+        processEscapes: true,
+        packages: ['base', 'ams', 'noerrors', 'noundefined', 'require']
```
marc-hughes, line 12: The only real change to the config was adding 'require' here

bsparks (approved):

marc-hughes:
🚨 Don't merge this in yet.

Found a bug where if you're typing \require{cancel} instead of copy/pasting it in, you'll hit an error because it tries to load the "C" extension that doesn't exist.

marc-hughes:
> Found a bug where if you're typing \require{cancel} instead of copy/pasting it in, you'll hit an error because it tries to load the "C" extension that doesn't exist.

Supplying the list of valid extensions to the require extension fixed this.

darrensiegel (approved):

PR #2930, opened 2022-08-15
[CHORE] Update ex_aws dependency

eliknebel:
This PR updates the `ex_aws` dependency in order to try and resolve an issue with aws REGION configuration in production. The latest version of this library seems to include a fix for how this config get applied.

This is a high risk PR and should be tested thoroughly before being deployed to production. I had to force the dependency with `override: true` because other dependencies such as `bamboo_ses` require the older version.


eliknebel:
The root issue turned out to be insufficient permissions on the access key.

PR #2931, opened 2022-08-16
[FEATURE] Add ability to browse all pages in a course project

darrensiegel:
This PR adds a new view for authors which allows browsing, searching and filtering across all pages in a course, regardless of where (and if) they live inside of containers.  

This is really just an initial, base implementation.  I suspect we will want to do some/all of the following in future (small) PRs:

1. Add ability to create a new page from this view to allow authors to create pages outside of their curriculum.
2. Add ability to search "content" and not just titles of pages.
3. Add ability to filter by and see "deleted" pages, and undelete them.
4. Add ability to see which author made the most recent edit for all pages.
5. Show which objectives a page has attached to it.  
6. Indicate whether the page does not exist in the curriculum (aka "standalone pages")

### DIFF
--- a/lib/oli_web/live/resources/pages_view.ex
+++ b/lib/oli_web/live/resources/pages_view.ex
@@ -1,0 +1,283 @@
+defmodule OliWeb.Resources.PagesView do
+  use Surface.LiveView, layout: {OliWeb.LayoutView, "live.html"}
+  use OliWeb.Common.Modal
+
+  import OliWeb.DelegatedEvents
+  import OliWeb.Common.Params
+  import Oli.Authoring.Editing.Utils
+  alias Oli.Accounts
+  alias OliWeb.Router.Helpers, as: Routes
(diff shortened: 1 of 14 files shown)


lib/oli_web/live/resources/pages_view.ex
```diff
@@ -0,0 +1,321 @@
...
+  data project, :any
+  data breadcrumbs, :list
+  data author, :any
+  data pages, :list
+
+  @limit 25
+
+  defp limit, do: @limit
+  defp graded_opts, do: [{true, "Graded"}, {false, "Practice"}]
+  defp type_opts, do: [{true, "Regular"}, {false, "Adaptive"}]
```
bsparks, line 28: in the table it says "Advanced" here in the filter it says "Adaptive"

eliknebel (changes requested):
Agree with @bsparks comment, type should probably say Adaptive in the table to match the filter.

Otherwise, everything else looks good

eliknebel (approved):

PR #2932, opened 2022-08-16
PMP-2999 | JAN-1847

dtiwarATS:


assets/src/adaptivity/scripting.ts
```diff
@@ -723,7 +723,8 @@ export const checkExpressionsWithWrongBrackets = (value: string) => {
     const actualExpression = expression;
     let result = expression.match(/{([^{^}]+)}/g) || [];
     result = result.filter(
-      (expression) => expression.search(/app\.|variables\.|stage\.|session\./) !== -1,
+      (expression) =>
+        expression.search(/^[^\s]+app\.|^[^\s]+variables\.|^[^\s]+stage\.|^[^\s]+session\./) !== -1,
```
bsparks, line 727: I think we need to check for spaces on the back instead of the front

bsparks (approved):

PR #2937, opened 2022-08-18
[MER-1409] [GH#2929] Increase max upload size

### LINKED ISSUE #2929: Raise media library upload limit to support up to 500MB files
We need to allow folks to upload video files up to 500MB in size.

marc-hughes:
This increases our max file upload size from 20mb to 350mb, which is the maximum size we can currently do in chrome with our current architecture. Going beyond this would require us to upload the file in multiple parts to get around a string size-limitation in V8.

Mostly Fixes #2929

### DIFF
--- a/assets/src/components/media/manager/MediaManager.tsx
+++ b/assets/src/components/media/manager/MediaManager.tsx
@@ -280,7 +280,13 @@ export class MediaManager extends React.PureComponent<MediaManagerProps, MediaMa
           })
           .then(() => this.setState({ error: Maybe.nothing<string>() }));
       })
-      .catch((e: Error) => this.setState({ error: Maybe.just(e.message) }))
+      .catch((e: Error | string) => {
+        if (typeof e === 'string') {
+          this.setState({ error: Maybe.just(e) });
+        } else {
+          this.setState({ error: Maybe.just(e.message) });
+        }
+      })
       .finally(() => this.setState({ uploading: false }));
   }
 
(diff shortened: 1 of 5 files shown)


assets/src/components/media/manager/MediaManager.tsx
```diff
@@ -280,7 +280,13 @@ export class MediaManager extends React.PureComponent<MediaManagerProps, MediaMa
           })
           .then(() => this.setState({ error: Maybe.nothing<string>() }));
       })
-      .catch((e: Error) => this.setState({ error: Maybe.just(e.message) }))
+      .catch((e: Error | string) => {
```
marc-hughes, line 283: This error comes in as a string sometimes and was not always displaying previously.

assets/src/components/media/manager/MediaManager.tsx
```diff
@@ -572,7 +578,9 @@ export class MediaManager extends React.PureComponent<MediaManagerProps, MediaMa
     return error.caseOf({
       just: (error) => (
         <div className="alert alert-danger fade show" role="alert">
-          {error}
+          Error: Could not upload file.
+          <br />
+          <i>{error}</i>
```
marc-hughes, line 583: Before, the message would just say something like "Failed to fetch", now we'll get "Error: Could not upload file." to start with every time so the user knows what happened.

darrensiegel (approved):

PR #2940, opened 2022-08-19
[FEATURE] Add new activity type Vlab activity

dhbrown:
Added support for a new activity type - Vlab Activity
  - A modified version of MultiInput (assets/src/components/activities/vlab)
  - VlabDelivery includes iframe w/ CMU developed Virtual Chemistry Lab (from static/)
  - Question can be one of the exiting MultiInput types or new VlabInput type
  - If input type is VlabInput, Torus compares target parameter to VlabState for eval
  - Config of Vlab via author-composed JSON on model, or pre-built configs from static


dhbrown:
Having some trouble debugging this type check error.  Passes test locally, but not on GitHub.

darrensiegel (changes requested):
I pushed some changes that fix a problem with the MultiInput activity - it was showing "Vlab" as an input type choice, but when chosen would break the activity.

Also, when I render the Vlab activity as an author (after inserting one into a page) I see a missing image: `images/vlab.png`.  

Also, there appear to be some typing problem in `assets/src/components/activities/vlab/actions.ts`.  For example, 

```
 setVlabParameter(id: string, param: string) {
    return (model: VlabSchema) => {
      const input = getByUnsafe(model.inputs, (x) => x.id === id);
      (input as VlabInput).parameter = param;
    };
  },

  setVlabSpecies(id: string, species: string) {
    return (model: VlabSchema) => {
      const input = getByUnsafe(model.inputs, (x) => x.id === id);
      (input as VlabInput).species = species;
    };
  },
```

In the above, both `parameter` and `species` do not exist on type `VlabInput`. Maybe the intent here was to cast to `VlabValue`?  Take a look and see if you can make all of the typing errors in this file get resolved.

dhbrown:
Yes, input was being incorrectly cast to VlabInput, rather than VlabValue. Updated vlab/actions.ts. When I run 'yarn run test' and on GitHub, I am not seeing any additional type errors.  Is there another test I should be running?

Also added vlab.png to /assets/static/images

darrensiegel (approved):

PR #2942, opened 2022-08-22
[ENHANCEMENT] Toolbar improvements

eliknebel:
This PR contains a few toolbar improvements. Most notably, it changes the toolbar from vertical to horizontal and adds some smart positioning so that the toolbar is always in view (bounded by the content being edited) while ensuring it never covers up the material being edited. The toolbar will start at the top of the textview but as the user scrolls, the toolbar will scroll with the view to remain in view. This new approach follows a more common toolbar layout and approach that new users may be more familiar with.

Also included are style refinements and insertion item reordering and adding commonly used items like tables and images to be one click instead of two.
Screen Shot 2022-08-21 at 9 52 27 PM

Screen Shot 2022-08-21 at 9 52 42 PM

Screen Shot 2022-08-21 at 9 52 55 PM

Screen Shot 2022-08-21 at 9 53 20 PM

These toolbar style changes fully support dark mode as well
Screen Shot 2022-08-21 at 10 02 13 PM


darrensiegel (approved):
This is fantastic!

PR #2946, opened 2022-08-23
[MER-1413] [GH# 889] Adding Definition Element

marc-hughes:
New definition element that can be added to pages.

A couple examples:
image

The editor:
image

Demo:

https://user-images.githubusercontent.com/333265/186228718-36c4099b-1d33-4efa-a1eb-27c79acf1e27.mp4


assets/src/components/editing/elements/common/settings/CaptionEditor.tsx
```diff
@@ -1,44 +1,29 @@
...
   model: ModelElement & { caption?: Caption };
   commandContext: CommandContext;
 }
-export const CaptionEditor = (props: Props) => {
-  const editor = useSlate();
-  const editMode = getEditMode(editor);
 
+export const CaptionEditor: React.FC<Props> = ({ commandContext, model, onEdit }) => {
   return (
-    <div contentEditable={false}>
```
marc-hughes, line 23: Needed an editor for sub-parts of the definition.
Caption here was already doing this.
Refactored out the editor part of that to be usable in both situations.

priv/schemas/v0-1-0/content-element.schema.json
```diff
@@ -886,6 +889,92 @@
...
+      ]
+    },
+
+    "definition-pronunciation": {
+      "properties": {
+        "type": {
+          "const": "pronunciation"
+        },
+        "src": {
+          "type": "string"
```
darrensiegel, line 907: The type here should be the same as the type for translation and meaning.  In other words, authors must be able to embed "rich text" into the pronunciation field.

```
<!ELEMENT %olcml.pronunciation.qname; (%olcml.inline.content;)*>
<!ATTLIST %olcml.pronunciation.qname;
  %olcml.core.attrs;
  %olcml.i18n.attrs;
  src			%URI;						#IMPLIED
  type			%ContentType;				#IMPLIED
>

<!ELEMENT %olcml.translation.qname; (%olcml.inline.content;)*>
<!ATTLIST %olcml.translation.qname;
  %olcml.core.attrs;
  %olcml.i18n.attrs;
>
```

darrensiegel (changes requested):
Looks great, just one change is needed to expand the type of "pronunciation". See me inline comment

priv/schemas/v0-1-0/content-element.schema.json
marc-hughes: @darrensiegel Right now:

Meaning: accepts block elements
Translation & Pronunciation: accepts text blocks (and those can have inline rich text content in them)

The src attribute for pronunciation is for the audio file. I'm not a fan of the naming since that is confusing, but matched the name of src in the dtd.

darrensiegel: Ah yes! sorry, i saw `"type": "string"` and assumed that was the pronunciation content.

darrensiegel (approved):

PR #2959, opened 2022-08-30
[MER-1416] [GH#889] New dialog element

marc-hughes:
Adds a new conversation/dialog element.

Demo:

https://user-images.githubusercontent.com/333265/187427499-55b3e58a-6158-4bdf-80d6-b2f8ce7997ab.mp4

### DIFF
--- a/assets/src/components/Dialog.tsx
+++ b/assets/src/components/Dialog.tsx
@@ -1,0 +1,47 @@
+import React from 'react';
+import * as ContentModel from 'data/content/model/elements/types';
+import { HtmlContentModelRenderer } from '../data/content/writers/renderer';
+import { WriterContext } from '../data/content/writers/context';
+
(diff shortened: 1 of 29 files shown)


darrensiegel (commented):
This all looks good, it isn't evident to me from just reviewing the code, but what happens if I have a speaker that has one or more lines associated with it, and I then delete that speaker?  It seems like those lines might be associated with `null` (see line 190 of DialogInlineEditor)

marc-hughes:
> This all looks good, it isn't evident to me from just reviewing the code, but what happens if I have a speaker that has one or more lines associated with it, and I then delete that speaker? It seems like those lines might be associated with `null` (see line 190 of DialogInlineEditor)

Good call, right now everything works reasonably in the editor & client preview, it shows up blank for a speaker and an author can cycle out of it to a new speaker.

But the phoenix render fails with an error. I'll go fix that up.

marc-hughes:
More reasonable behavior on deleting speakers now.

image

PR #2967, opened 2022-09-01
[BUG FIX] Remove z-index from typeahead

### LINKED ISSUE #2951: Z-index is too low for toolbar menu
**Describe the bug**
Toolbar menu items should show above all items on the page. The zindex should be adjusted to be higher than any of the indices on page.

**Screenshots**
Screen Shot 2022-08-22 at 10 20 31 AM

 - OS: macOS
 - Browser: Chrome
 - Version: 0.21.4

darrensiegel:
Closes #2951 

I tried adding `z-index` to places in the toolbar that didn't specify it, tried increasing the `z-index` value for those that already did have it specified, but nothing would fix this issue.  The issue only seems to be a problem with the React Typeahead input text.  Since we had forked the CSS for that, I went in and removed the `z-index` specified there. It fixes the problem and in my testing I do not see any negative impacts of removing it.

### DIFF
--- a/assets/styles/authoring/react-typeahead.scss
+++ b/assets/styles/authoring/react-typeahead.scss
@@ -198,11 +198,6 @@
   flex: 1;
 }
 
-.input-group > .rbt .rbt-input-hint,
-.input-group > .rbt .rbt-aux {
-  z-index: $z-content-2;
-}
-
 .input-group > .rbt:not(:first-child) .form-control {
   border-top-left-radius: 0;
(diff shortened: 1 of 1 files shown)


eliknebel (changes requested):
I still seem to be having an issue with this where the placeholder text and remove button are still showing through

Screen Shot 2022-09-01 at 4 22 46 PM
Screen Shot 2022-09-01 at 4 22 36 PM

PR #2973, opened 2022-09-01
[MER-411] Bump lti_1p3 version to 0.4.6

mcass19:
[MER-411](https://eliterate.atlassian.net/browse/MER-411)

Bump lti_1p3 version to contemplate https://github.com/Simon-Initiative/lti_1p3/pull/26


mix.exs
```diff
@@ -158,7 +158,7 @@ defmodule Oli.MixProject do
       {:jason, "~> 1.3"},
       {:joken, "~> 2.2.0"},
       {:jose, "~> 1.10"},
-      {:lti_1p3, "~> 0.4.5"},
+      {:lti_1p3, "~> 0.4.6"},
       {:lti_1p3_ecto_provider, "~> 0.3.2"},
```
eliknebel, line 162: We need to bump lti_1p3_ecto_provider here as well to 0.3.3

eliknebel (changes requested):

mix.exs
mcass19: Good catch @eliknebel , updated! Thanks

eliknebel (approved):

PR #2974, opened 2022-09-02
PMP-3023 | JAN-1866

dtiwarATS:


assets/src/adaptivity/scripting.ts
```diff
@@ -50,7 +50,7 @@ export const getExpressionStringForValue = (
       canEval = test?.result !== undefined && !test.result.message;
       if (
         test?.result &&
-        test?.result?.length > 1 &&
+        test?.result?.length &&
```
dtiwarATS, line 53: This was the bug that was causing the issue in which if we only add 1 observable and then go back to another lesson, it was treated as regular array object and the observation variable was getting converted into array object.

bsparks (approved):

PR #2985, opened 2022-09-09
[FEATURE] All activities view

darrensiegel:
This PR adds a new, admin only, view: the ability to view all activities within a course project.  This view is designed to allow rapid review and troubleshooting of the rendering of the authoring component of all activities. 

This PR includes an extremely important fix: the removal of the use of `DEFAULT_PART_ID` in the various authoring and delivery components.  Course digests can contain legacy activities whose part ids are the original OLI part ids, and the default part id that Torus assumes. 

This PR also fixes a problem with `Oli.Publishing.determine_parent_pages`.  The JSON query inside this was only finding activity references at the top level of the page content model.


rgachuhi (commented):
I cannot seem to get the Open Sync View to do anything. Not sure about its usage. The sync view page opens fine but nothing happens when I enter either local path to my svn checkout or the full svn url to the course package

darrensiegel:
> I cannot seem to get the Open Sync View to do anything. Not sure about its usage. The sync view page opens fine but nothing happens when I enter either local path to my svn checkout or the full svn url to the course package

The only thing one needs to do to get the Sync view to work is to select an activity from the table in the main "All Activities" view.

rgachuhi:
> 



> The link at the bottom of the "All Activities" page opens the following page. 

Screen Shot 2022-09-09 at 5 43 01 PM

darrensiegel:
> > 
> 
> > The link at the bottom of the "All Activities" page opens the following page. Should it be displaying more info than that?
> 
> Screen Shot 2022-09-09 at 5 43 01 PM

Do the following, exactly:
1. Go to the "All Activities" course project that has been published at least once, and that has one or more activities in it.
2. Click the "Open Sync View" to open in a new Window.  Now you should have the "All Activities" window open in one view and the "Activity Review" (aka Sync) view in another window.
3. In the "All Activities" view, click to select an item from the table.  You should then see the full details of your selected activity in the Sync view.

rgachuhi (approved):

rgachuhi:
> 

That did the job

PR #2987, opened 2022-09-09
[FEATURE] Explanation feedback backend

eliknebel:
Add explanation backend code and rendering for feedback.

This PR completes the full support of explanation (multi-feedback) in torus.

Once an explanation is authored for one or more activities, the explanation strategy can be set in page options. The strategies will behave a bit differently for graded and ungraded pages, most notably graded pages will use the page (resource) attempt count whereas ungraded pages will use the activity attempt count.
Screen Shot 2022-09-08 at 7 18 17 PM

When an explanation condition is met, the part explanation will be displayed alongside the part feedback:
Screen Shot 2022-09-09 at 11 13 45 AM

This works for multiple part feedbacks as well such as multi-input:
Screen Shot 2022-09-09 at 11 24 04 AM

This PR also fixes an issue where math input in multi-input showed "Input (Text)" instead of "Input (Math)".
Screen Shot 2022-09-09 at 11 15 39 AM

This PR includes some new testing infrastructure as well with the intent to simply the process of setting up test state. The new test seeder code is defined in new domain-specific modules that related to the state being created (Seeder.Project, Seeder.Attempt). The ideas here build upon some of the patterns used in the current DBSeeder, such as tagging items that are created and adding them to an accumulation map, while also introducing some new niceties like being able to specify a reference to a tagged item in the map using `ref(:some_tagged_item)` so that a pipeline can make use of previous actions without have to break out of the pipeline and use the map from a previous result. The `test/oli/delivery/evaluation/explanation_test.exs` test file showcases this new functionality, ideas/improvements are welcomed here as the intent is to create an easily composable framework from which we can build higher-level state creator functions from lower level ones, while keeping the overall pipeline legible and easy to write when crafting new tests.


lib/oli/activities/parse_utils.ex
```diff
@@ -85,15 +85,4 @@ defmodule Oli.Activities.ParseUtils do
         true
     end
   end
-
```
eliknebel, line 88: Moved to `Feedback.from_text`

lib/oli/resources/explanation_strategy.ex
```diff
@@ -0,0 +1,38 @@
...
+      :after_set_num_attempts
+    ]
+  end
+
+  @derive Jason.Encoder
+  @primary_key false
+  embedded_schema do
+    field :type, Ecto.Enum,
+      values: [
+        :after_max_resource_attempts_exhausted,
```
darrensiegel, line 18: It seems like `:none` be included here?

darrensiegel (changes requested):

lib/oli/resources/explanation_strategy.ex
eliknebel: Actually I had just forgotten to remove `:none` from the above `types()`. I opted to model no policy as `nil` so that this wouldn't require a migration.

If you think that explicitly modeling this and writing a migration is a better way to go though, I can do that

darrensiegel (approved):

PR #2990, opened 2022-09-12
[MER-879] Add non empty lti score comment

mcass19:
[MER-879](https://eliterate.atlassian.net/browse/MER-879)

Fixing some issues Schoology was having for syncing grades:

- Do not send an empty comment when posting score
  - Fixed error: `[error] Error encountered posting score for user 118752747::f9e5b6fb327dad106e4cd6445cc2cc1f for line item 'New Assessment 1' {:ok, %HTTPoison.Response{body: "{\"error\":\"Invalid parameter \\\"body.comment\\\": \\\"comment\\\" is not allowed to be empty\"}", headers: [{`

Tested that Moodle and Canvas continue working as expected.

[Related PR](https://github.com/Simon-Initiative/lti_1p3/pull/27)

Leaving it as draft for now, since it's not fully working yet due to an error from the Schoology side. See ticket in Jira for further information. Although I agree if we want to proceed with these changes anyway.

### DIFF
--- a/lib/oli/grading.ex
+++ b/lib/oli/grading.ex
@@ -94,7 +94,7 @@ defmodule Oli.Grading do
       timestamp: timestamp,
       scoreGiven: resource_access.score,
       scoreMaximum: resource_access.out_of,
-      comment: "",
+      comment: "Score default comment",
       activityProgress: "Completed",
       gradingProgress: "FullyGraded",
       userId: sub
--- a/test/oli/grading_test.exs
+++ b/test/oli/grading_test.exs
@@ -1,7 +1,10 @@
 defmodule Oli.GradingTest do
   use Oli.DataCase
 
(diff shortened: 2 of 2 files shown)


lib/oli/grading.ex
```diff
@@ -94,7 +94,7 @@ defmodule Oli.Grading do
       timestamp: timestamp,
       scoreGiven: resource_access.score,
       scoreMaximum: resource_access.out_of,
-      comment: "",
+      comment: "Score default comment",
```
mcass19, line 97: Not sure if this is fine or if we should set something different here, or think in take another approach

mcass19:
Closing this as the integration seems to not be compliant with the standard. Making this change would force torus to always include a comment, which is not actually a required field

PR #2994, opened 2022-09-14
[BUGFIX] JAN-1857 manual grading everapp

bsparks:
this will send page level (extrinsic) state into an activity when it is rendered. this is currently only needed for standalone activity review. the adaptive lessons will store various pieces of information in the extrinsic state of the attempt, but most importantly they will store what is essentially a snapshot of user state (EverApp) in this area for an attempt. It is desirable to be able to review relevant user state within the context of an attempt.

### DIFF
--- a/assets/src/components/activities/DeliveryElement.ts
+++ b/assets/src/components/activities/DeliveryElement.ts
@@ -75,11 +75,6 @@ export interface DeliveryElementProps<T extends ActivityModelSchema> {
   state: ActivityState;
 
   /**
-   * The state of the page attempt. Usually only necessary for standalone activity review.
-   */
-  pageState?: any;
-
-  /**
    * The larger context that this activity operates within.
    */
   context: ActivityContext;
--- a/assets/src/components/activities/DeliveryElement.ts
+++ b/assets/src/components/activities/DeliveryElement.ts
@@ -56,6 +56,7 @@ export interface ActivityContext {
   surveyId: string | null;
   bibParams: any;
   pageAttemptGuid: string;
+  pageState?: any;
 }
 
 /**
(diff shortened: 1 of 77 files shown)


assets/src/components/activities/DeliveryElement.ts
```diff
@@ -73,6 +73,11 @@ export interface DeliveryElementProps<T extends ActivityModelSchema> {
    */
   state: ActivityState;
 
+  /**
```
darrensiegel, line 76: Recently I did some work to cleanup the `DeliveryElement` interface.  It had gotten out of hand, with a whole bunch of top level properties having been added.  I think it makes sense to strive to continue to keep that interface small.  So instead of adding a new top level attr for `page-state` I'd recommend adding `pageState` as an attribute of the top level `activityContext instead.

darrensiegel (changes requested):

assets/src/components/activities/DeliveryElement.ts
bsparks: ah yes, good point. thanks

assets/src/components/activities/DeliveryElement.ts
```diff
@@ -73,6 +74,11 @@ export interface DeliveryElementProps<T extends ActivityModelSchema> {
...
+   * The state of the page attempt. Usually only necessary for standalone activity review.
+   */
+  pageState?: any;
```
darrensiegel, line 80: I think you meant to remove this as well.

darrensiegel (changes requested):

darrensiegel (approved):

PR #2996, opened 2022-09-14
[BUG FIX] fix an issue where the activity delivery component expects an authoring property

eliknebel:
Fixes an issue where the activity delivery component expects an authoring property to exist but it does not


assets/src/components/activities/multiple_choice/MultipleChoiceDelivery.tsx
```diff
@@ -41,9 +42,11 @@ export const MultipleChoiceComponent: React.FC = () => {
   const { surveyId } = context;
 
   useEffect(() => {
+    console.log(model, activityState);
```
bsparks, line 45: probably forgot to remove

eliknebel: thanks, good catch

darrensiegel (approved):

PR #3002, opened 2022-09-20
PMP-2988 | Handle SIM Event Flow

dtiwarATS:
DO NOT MERGE it before @bsparks  reviews it.


dtiwarATS:
I will fix this with another approach

PR #3016, opened 2022-09-22
PMP-2988 | JAN-1836

dtiwarATS:


assets/src/data/persistence/extrinsic.ts
```diff
@@ -52,10 +52,35 @@ export const readGlobalUserState = async (
     if (lastSet()) {
       await lastSet();
     }
-    const serverUserState = await readGlobal(keys);
-    // merge server state with result
-    if ((serverUserState as any).type !== 'ServerError') {
-      result = serverUserState;
+    let cacheMissingRequestedKey = false;
+    if (keys) {
+      const lastchacheTimeStamp = userStateCache['timestamp'];
```
bsparks, line 57: typo, plus you could do like:
`const { timestamp: lastCacheTimeStamp } = userStateCache;`

assets/src/data/persistence/extrinsic.ts
```diff
@@ -52,10 +52,35 @@ export const readGlobalUserState = async (
...
+        if ((serverUserState as any).type !== 'ServerError') {
+          result = serverUserState;
+        }
+      }
+    } else {
+      result = userStateCache;
+    }
+    if (cacheMissingRequestedKey) {
+      //if cache does not have any of the requested keys, we should make the server call
+      const serverUserState = await readGlobal(keys);
```
bsparks, line 79: since this is repeated, probably better to just manage a single flag; 
maybe rename `cacheMissingRequestedKey` to `refreshFromServer` and then if it's because of the time or because it was missing a key, just do it once at the bottom (here) `if (refreshFromServer) `

assets/src/data/persistence/extrinsic.ts
```diff
@@ -100,7 +125,19 @@ export const updateGlobalUserState = async (
   updates: { [topKey: string]: { [key: string]: any } },
   useLocalStorage = false,
 ) => {
-  /* console.log('updateGlobalUserState called', { updates, useLocalStorage }); */
+  //Lets update the cache with latest changes.
+  const topLevelKeys = Object.keys(updates);
+  userStateCache['timestamp'] = Date.now();
+  topLevelKeys.forEach((topKey: any) => {
```
bsparks, line 131: this logic is basically repeated from the internal one, so probably should factor it out into a shared method

bsparks (approved):

PR #3018, opened 2022-09-23
[FEATURE] New Conjugation content element [MER-1417]

marc-hughes:
Adds a new element that can be authored into lessons for the conjugation of verbs.

image

Demo:
https://user-images.githubusercontent.com/333265/192010391-66a40148-159f-4848-acea-af49c8e081b9.mp4


assets/src/components/common/Pronunciation.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
```
marc-hughes, line 1: Refactored DefinitionPronunciation -> Pronunciation since both Definition and Conjugation could use it.

assets/src/components/editing/editor/Editor.tsx
```diff
@@ -35,6 +35,7 @@ export type EditorProps = {
   placeholder?: string;
```
marc-hughes, line 35: Getting the sub-editor that only worked on this special kind of table was a little tricky to get working. The stuff in here lets us override the configuration of the editor object.

assets/src/components/editing/editor/normalizers/normalizer.ts
```diff
@@ -14,15 +16,62 @@ export interface NormalizerContext {
 
 const restrictedElements = new Set(['input_ref']);
 
-export function installNormalizer(editor: Editor, context: NormalizerContext = {}) {
```
marc-hughes, line 17: Added the ability to configure what normalizers were configured, only needed 2 of these options for this, but made them all to be complete.

assets/src/components/editing/editor/normalizers/tables.ts
```diff
@@ -8,7 +8,7 @@ import { FormattedText } from 'data/content/model/text';
  * col span
  **/
 export const getColumnSpan = (cells: TableCell[]): number =>
-  cells.reduce((acc, curr) => acc + (curr.colspan || 1), 0);
```
marc-hughes, line 11: Found a previously created bug here with empty tables.

assets/src/components/editing/elements/common/AudioClipPicker.tsx
```diff
@@ -0,0 +1,76 @@
+import React, { useCallback, useRef } from 'react';
```
marc-hughes, line 1: Refactored out the audio pick from the pronunciation component since it's also can be used in the <tc> components now.

assets/src/components/editing/elements/conjugation/ConjugationToolbar.tsx
```diff
@@ -0,0 +1,47 @@
...
+interface SettingsProps {
+  commandContext: CommandContext;
+  model: ContentModel.Conjugation;
+  toggleEdit: () => void;
+  editing: boolean;
+}
+
+export const ConjugationSettings = (props: SettingsProps) => {
+  return (
+    <Toolbar context={props.commandContext}>
```
marc-hughes, line 19: Should we add a delete option to all of these toolbars? I started adding one, then realized most (all?) don't so it wouldn't fit in and didn't add it, but it seems like it would be a good usability addition.

assets/src/components/editing/elements/definition/DefinitionInlineEditor.tsx
```diff
@@ -1,40 +1,33 @@
-import React, { useCallback, useMemo, useRef } from 'react';
```
marc-hughes, line 1: This is all refactoring audio picker & pronunciation editor out.

assets/src/data/content/model/elements/factories.ts
```diff
@@ -73,6 +81,20 @@ export const Model = {
   calloutInline: (text = '') =>
     create<CalloutInline>({ type: 'callout_inline', children: [{ text }] }),
 
+  conjugationTable: () =>
+    Model.table([
```
darrensiegel, line 85: Can we change this so that the default conjugation table has four rows, and three columns, and reads:

```
<empty>   | Singular | Plural
1st Person | <empty> | <empty>
2nd Person | <empty> | <empty>
3rd Person | <empty> | <empty>
```

darrensiegel (changes requested):
Works very well locally, only comment was to change the default "new" conjugation model when one is inserted so that it is more immediately useful.

assets/src/data/content/model/elements/factories.ts
marc-hughes: New default: 
image

darrensiegel (approved):

Issue #3017, opened 2022-09-23
Crash on copy/paste in table editing


marc-hughes:
**Describe the bug**
The editor crashes if you copy & paste from a cell into a new cell of a table

**To Reproduce**
1. Edit a practice page
2. Insert a table
3. Type some text into a cell
4. Select all of that text
5. Copy it
6. Paste it into an empty cell

Observe: Editor crashes

**Expected behavior**
The text is pasted

**Screenshots**
![tablecrash](https://user-images.githubusercontent.com/333265/191999993-71618ee8-a3ac-43b0-adde-49b959b50909.gif)

PR #3025, opened 2022-09-27
Bug/copy paste table crash 3017

marc-hughes:
While investigating issue #3017 , I found there were 3  related issues here. 

## Copying from within a table cell would copy a table element instead of just the highlighted part.

I couldn't get to a simple solution for this one, and it's not too horrible an experience so I'm just leaving it as-is in interest of time to get on to other work. I might come back to it at some point. Some notes for if someone ever does:

Down in the slate-react copy code it calls
```
ReactEditor.setFragmentData(
                    editor,
                    event.clipboardData,
                    'copy'
                  )
```
https://github.com/ianstormtaylor/slate/blob/main/packages/slate-react/src/components/editable.tsx#L1134

You can create an override for setFragmentData where we create the editor (one of the with** functions)
https://github.com/Simon-Initiative/oli-torus/blob/master/assets/src/components/editing/editor/Editor.tsx#L63

Then we could selectively detect this case of copying within a cell and override the functionality, but that potentially requires exposing some currently private slate functions and I fear it may be brittle in future slateJS upgrades (another reason I'm not sure the cost::benefit is there to fix this if nobody but me is complaining about it).

## The overall normalizer would try to do too much in one pass

If one of our normalizers triggered, we were previously still running all of them. You might have normalizer A put the doc in a state that normalizer B couldn't deal with, the specific example here was deleting the element the cursor was in and then a later normalizer dying because of that.

Now, whenever we make a normalization change, we exit out and start a new loop which fixes that sort of issue, and that fixes the crash reported in the issue #3017 

## The block normalizer would completely remove invalid nodes instead of unwrapping them to preserve potentially valid content.

If I tried to paste something like `<table><td><td>HI!</td></tr></table>` into a table cell, our normalizer over in block.ts would see that as an invalid child and delete it. To the user, nothing happened when they tried to paste (well, actually, the editor previously crashed, but ignore that).

A more friendly behavior might be to unwrap the node until we get to one that is valid, the text node "HI!" in that example. This way the user gets content pasted in instead of just nothing seeming to happen.


assets/src/components/editing/editor/normalizers/block.ts
```diff
@@ -16,12 +20,12 @@ export const normalize = (editor: Editor, node: ModelElement | FormattedText, pa
...
           Transforms.removeNodes(editor, { at: parentPath });
           console.warn('Normalizing content: Special case code, removing node', node.type);
-          return;
+          return true;
         }
 
-        Transforms.removeNodes(editor, { at: path });
-        console.warn('Normalizing content: Removing non block node', node.type);
-        return;
+        Transforms.unwrapNodes(editor, { at: path });
```
marc-hughes, line 26: This is the part where we unwrap instead of remove the invalid children

assets/src/components/editing/editor/normalizers/code.ts
```diff
@@ -14,13 +18,14 @@ export const normalize = (editor: Editor, node: ModelElement | FormattedText, pa
       if (Text.isText(node)) {
         Transforms.wrapNodes(editor, { type: 'code_line', id: guid(), children: [] }, { at: path });
         console.warn('Normalizing content: wrapping code_line in code block');
-        return;
+        return true;
```
marc-hughes, line 21: Now, all these normalizers return true if they made a modification.

assets/src/components/editing/editor/normalizers/normalizer.ts
```diff
@@ -91,14 +91,14 @@ export function installNormalizer(
         }
       }
 
-      options.spacesNormalize && spacesNormalize(editor, node, path);
-      options.blockNormalize && blockNormalize(editor, node, path);
-      options.codeNormalize && codeNormalize(editor, node, path);
-      options.listNormalize && listNormalize(editor, node, path);
-      options.tableNormalize && tableNormalize(editor, node, path);
+      if (options.spacesNormalize && spacesNormalize(editor, node, path)) return;
```
marc-hughes, line 94: These if/return bits are where we return early if a normalization change was made so a new loop starts.

eliknebel (approved):
Works great, fixes a lot of the normalization issues

PR #3030, opened 2022-09-27
[BUG FIX] Fix page link issues

eliknebel:
This PR fixes the following issues related to page links:

- Styling when a page link is in a group with a purpose
Screen Shot 2022-09-27 at 3 14 51 PM

- Implements missing client side rendering of page links for activities
Screen Shot 2022-09-27 at 3 10 23 PM
Screen Shot 2022-09-27 at 3 10 18 PM


darrensiegel (approved):

darrensiegel (changes requested):
I just realized that the page_link implementation "freezes" the original title of the page into the page_link element.  This content element should not store the page title of the linked page, rather, at rendering time (for both authoring and delivery) it should resolve the current title.  Take this example:

1. I create a page titled "Fuzzy Wuzzy"
2. I create another page and place a page_link in it to "Fuzzy Wuzzy"
3. I then rename the original page from "Fuzzy Wuzzy" to "Wuz a Bear"

The current impl will render the page link as "Fuzzy Wuzzy" since the title was caked into the page element in the second page.

assets/src/data/content/writers/html.tsx
```diff
@@ -388,6 +390,31 @@ export class HtmlParser implements WriterImpl {
...
+      internalHref = `/sections/${context.sectionSlug}/page/${revisionSlug}`;
+      targetAndRel = { target: '_blank', rel: 'noreferrer' };
+    } else {
+      internalHref = '#';
+      targetAndRel = {};
+    }
+
+    return (
+      <div className={classNames('content-page-link content-purpose', purpose)}>
+        <div className="content-purpose-label">Checkpoint</div>
```
darrensiegel, line 408: Should the purpose here be hardcoded as "Checkpoint"?

darrensiegel (changes requested):

eliknebel:
Good point, I was trying to avoid adding another map for page names into the render context by just storing the title, however I didn't consider the case where pages are renamed. I will address this

darrensiegel (changes requested):
I didn't correctly understand what the value of `ref` was in the page_link model. I assumed it was a resource_id.  The fact that it is a link, which then this code parses out the slug and uses the `pageTitles` to retrieve the title is problematic for the case when the slug of a page changes (due to its title being changed).  Upstream, the `Oli.Resources.page_titles` builds a map of the "latest" revision slug to its title.  But there can be "old" slugs littered throughout content which will break title rendering with this approach. 

Moreover, the cost of building the page titles map is not small.  It fetches all page revisions (e.g. there are 511 pages in Chemistry I), just to get to the titles and build that map.  It does that on every page render, even when there are no page_title instances present.

Instead, I think we need to do this:

1. Change the model of the `page_link` from `ref` to `idref`.  Simply embed into `idref` the resource_id of the page linked to.
2. Leverage a good assumption that at most a page will contain one or two `page_link` instances. Replace the page_titles map with a targeted function that takes a resource_id and returns the title of the resolved revision.  Call this when needed during rendering.  
3. Let's just cut the cord and remove the ability to embed `page_link` instances within Activities. That simplifies a bunch of this.

darrensiegel (approved):

PR #3035, opened 2022-09-29
[Feature] Added learning language attribute and foreign element

marc-hughes:
Adds a new "foreign" element that can be inserted into pages, and adds a new project-attributes section to the project details where you can set the default language.

For learners, the foreign tag renders out a "lang=xxx" attribute on the selected text for screen readers to correctly pronounce the word, there's no visual indicator to learners.

Demo:
https://user-images.githubusercontent.com/333265/193059154-0be74c7d-e423-4ff6-b1f0-6829fef03064.mp4


lib/oli_web/live/projects/projects_live.ex
```diff
@@ -37,7 +37,8 @@ defmodule OliWeb.Projects.ProjectsLive do
 
   data author, :any
   data is_admin, :boolean, default: false
-  data changeset, :any, default: Project.changeset(%Project{title: ""})
+
+  data changeset, :any, default: Project.new_project_changeset(%Project{title: ""})
```
marc-hughes, line 41: I had a problem here, and I don't think this is the best way to do this, but wasn't sure how else to solve it.

As soon as I added the cast_embed to the normal Project.changest, I started getting this compile-time error on this line:

```
== Compilation error in file lib/oli_web/live/projects/projects_live.ex ==
** (ArgumentError) cannot escape #Function<19.44846763/2 in Ecto.Changeset.on_cast_default/2>. The supported values are: lists, tuples, maps, atoms, numbers, bitstrings, PIDs and remote functions in the format &Mod.fun/arity
    (elixir 1.13.2) src/elixir_quote.erl:185: :elixir_quote.argument_error/1
    (elixir 1.13.2) src/elixir_quote.erl:340: :elixir_quote.do_quote/3
    (elixir 1.13.2) src/elixir_quote.erl:471: :elixir_quote.do_quote_simple_list/4
    (elixir 1.13.2) src/elixir_quote.erl:471: :elixir_quote.do_quote_simple_list/4
    (elixir 1.13.2) src/elixir_quote.erl:386: :elixir_quote.do_escape/3
    (elixir 1.13.2) src/elixir_quote.erl:340: :elixir_quote.do_quote/3
...
```

My elixir knowledge hit it's limit here and I'm not sure how to properly solve that. So instead, I made two different changeset functions, one with the embed, and one without (since this didn't need to actually set any of those embedded attributes)

assets/src/components/editing/elements/foreign/ForeignEditor.tsx
```diff
@@ -0,0 +1,70 @@
...
+                icon: 'edit',
+                description: 'Change Language',
+                execute: onChangeLang,
+              })}
+            />
+          </Toolbar.Group>
+        </Toolbar>
+      }
+    >
+      <span className="foreign-editor" lang={props.model.lang} {...props.attributes}>
```
marc-hughes, line 65: Note: In the editor, we don't currently get the project-default setting for lang, but in the preview we do.

assets/src/data/content/writers/html.tsx
```diff
@@ -203,6 +204,10 @@ export class HtmlParser implements WriterImpl {
     return <Dialog dialog={dialog} context={context} />;
   }
 
+  foreign(ctx: WriterContext, next: Next, element: Foreign) {
+    return <span lang={element.lang || ctx.learningLanguage}>{next()}</span>;
```
darrensiegel, line 208: The server-side render includes a `class="foreign"` but this client-side render does not.  I see that there actually isn't a `foreign` CSS class that we define, which is okay, but let's keep the client and server renderers in sync and have them both either render that class name or both not render it.  I'd vote for having them both render it, even it if is unspecified.

assets/src/components/editing/elements/foreign/ForeignModal.tsx
```diff
@@ -0,0 +1,46 @@
...
+          <p className="mb-4">
+            Set this text as belonging to a foreign language for screen readers to pronounce
+            properly. This has no visual effect for the learner. You can set a project-wide default
+            in your project settings.
+          </p>
+          <div className="popup__modalContent">
+            <form onSubmit={() => {}} id="popup__trigger_mode">
+              <div className="form-group">
+                <label>Target Language</label>
+                <input type="text" className="form-control" onChange={onLangChange} value={lang} />
```
darrensiegel, line 38: Can you easily replace this text input with a select whose options are the ISO 639-1 language codes?

Perhaps the rendered label could be "Language [code]", for example: "French [fr]", and of course the option value would simply be the code.

[**ISO 639-1 Language Codes**](https://www.w3docs.com/learn-html/html-language-codes.html)

lib/oli_web/templates/project/overview.html.eex
```diff
@@ -37,10 +38,32 @@
...
+      Project wide configuration, not all options may be relevant for all
+      subject areas.
+    </div>
+  </div>
+  <div class="col-md-8">
+
+    <%= inputs_for f, :attributes, fn fp -> %>
+      <div class="form-label-group mb-3">
+        <%= label fp, :learning_language, "Learning Language", class: "control-label" %>
+        <%= text_input fp, :learning_language, class: "form-control", required: false %>
```
darrensiegel, line 57: We'd want the same "<select>" construct here as well

lib/oli_web/live/projects/projects_live.ex
darrensiegel: Im not entirely sure, but I think if you used the original changeset and supplied a default value for attributes like this, that it would work:

```
data changeset, :any, default: Project.new_project_changeset(%Project{title: "", attributes: %{learning_language: "en"})
```

darrensiegel (changes requested):

lib/oli_web/live/projects/projects_live.ex
marc-hughes: I tried several variations (including exactly that) on that idea with no luck.

I don't think having a separate changeset for creating a project here is a horrible solution, it just doesn't feel necessary.

assets/src/components/editing/elements/foreign/ForeignModal.tsx
marc-hughes: image

Added that in.

lib/oli_web/templates/project/overview.html.eex
marc-hughes: Got this in too

image


Note: I did not change the schema to restrict the value to that list, do you think we should?

lib/oli_web/live/projects/projects_live.ex
darrensiegel: Agreed.  After it lands as-is perhaps we can figure out a way to remove the need for it.

assets/src/components/editing/elements/foreign/ForeignModal.tsx
darrensiegel: Nice!

assets/src/components/editing/elements/foreign/ForeignEditor.tsx
darrensiegel: I don't think we need it in the pure authoring context.

darrensiegel (approved):

PR #3040, opened 2022-09-30
AOPS-997 | Session resume causing scoring issue

dtiwarATS:


assets/src/apps/delivery/store/features/groups/actions/deck.ts
```diff
@@ -270,32 +270,38 @@ const getSessionVisitHistory = async (
     }));
 };
 
-export const navigateToNextActivity = createAsyncThunk(
-  `${GroupsSlice}/deck/navigateToNextActivity`,
-  async (_, thunkApi) => {
+export const findNextSequenceId = createAsyncThunk(
+  `${GroupsSlice}/deck/findNextSequenceId`,
+  async (payload: { sequenceType: string; sequenceId: string }, thunkApi) => {
```
bsparks, line 275: hmm seems like you only need sequenceId, instead of using `'-1'` you could just put 'next' in there. makes the sequenceType not needed

bsparks (approved):

PR #3046, opened 2022-10-03
[Feature] Audio support on popups. [MER-1470]

marc-hughes:
This adds the ability to add an audio clip to popup elements. That audio clip will play when the popup is displayed, and stop if the popup is dismissed.

Due to some recent changes in the modal infrastructure, I had to modify the audio-picker control to browse the audio files within the already-open popup config modal instead of opening a new one.

popup

### DIFF
--- a/assets/src/phoenix/app.ts
+++ b/assets/src/phoenix/app.ts
@@ -68,10 +68,27 @@ $(() => {
   ($('.ui.dropdown') as any).dropdown();
   ($('.ui.dropdown.item') as any).dropdown();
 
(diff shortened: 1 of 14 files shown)


assets/src/phoenix/app.ts
```diff
@@ -68,10 +68,27 @@ $(() => {
   ($('.ui.dropdown') as any).dropdown();
   ($('.ui.dropdown.item') as any).dropdown();
```
marc-hughes, line 70: This is what actually plays the audio during preview/delivery

darrensiegel (approved):
Tested, works great.

PR #3047, opened 2022-10-03
[ENHANCEMENT] [MER 1468] provide an option to create a new page from the all pages view

eliknebel:
Adds the ability to create a new page directly in the new "All Pages" view
Screen Shot 2022-10-03 at 3 58 41 PM

As well as several other related enhancements:
- Ability to move a page from the All Pages view into the curriculum
- Ability to remove a page from the curriculum (without necessarily deleting the page)
Screen Shot 2022-10-03 at 4 10 52 PM

- Updates the breadcrumb to show the "All Pages" link (instead of curriculum) if the page being edited does not exist in the curriculum
Screen Shot 2022-10-03 at 4 01 55 PM

- Ability to navigate directly from a Page link to the linked page in authoring, so that authors can seamlessly jump into an unconnected page to edit it
Screen Shot 2022-10-03 at 4 01 01 PM

### DIFF
--- a/lib/oli_web/live/resources/pages_view.ex
+++ b/lib/oli_web/live/resources/pages_view.ex
@@ -388,10 +388,7 @@ defmodule OliWeb.Resources.PagesView do
   def handle_event("create_page", _, socket) do
     %{
       project: project,
-      author: author,
-      table_model: table_model,
-      options: options,
-      offset: offset
+      author: author
     } = socket.assigns
 
     attrs = %{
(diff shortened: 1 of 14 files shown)


lib/oli_web/live/resources/pages_view.ex
```diff
@@ -272,6 +285,167 @@ defmodule OliWeb.Resources.PagesView do
...
+             author,
+             project
+           ) do
+        {:ok, _} ->
+          socket
+
+        {:error, %Ecto.Changeset{} = _changeset} ->
+          socket
+          |> put_flash(:error, "Could not create new page")
+      end
```
darrensiegel, line 426: After I clicked the button to create a page (in the Chemistry course, which has 500 pages), I saw nothing happen.  No updates, no modal.  So I clicked again, and again.  Then, I realized it was creating the pages, but there were just not in my current "view" of the 25 rows of pages. 

I think a better UX would be to take the user into the "page editor" for the new page immediately after it is created.  I suspect the common use case here is "I am creating a new page because I want to author a new page"

darrensiegel (changes requested):
Works well, Just had one comment about a confusing part of the UX that I think we need to address somehow.

lib/oli_web/live/resources/pages_view.ex
eliknebel: Good point, I agree. I will add this in

eliknebel: Added the ability to create practice, graded and adaptive (when enabled), same as container live view.
Screen Shot 2022-10-04 at 2 01 09 PM

darrensiegel (approved):

PR #3049, opened 2022-10-04
[FEATURE] [MER-1456] [MER-1457] [MER-1458] Reuse sub objectives

mcass19:
[MER-1456](https://eliterate.atlassian.net/browse/MER-1456) & [MER-1457](https://eliterate.atlassian.net/browse/MER-1457) & [MER-1458](https://eliterate.atlassian.net/browse/MER-1458)

- Add the ability to reuse an existing sub objective 
- Rework the delete logic to account for this new scenario
- Rewrite the entire view


darrensiegel (changes requested):
Functionally this PR accomplishes the task of allowing sub-objectives to be reused.  I saw two significant usability issues with this when I tested it with a course that contains hundreds of top-level objectives:

1. Creating a new sub objective triggers a reordering of the top-level objectives upon the next render, which is confusing when there are many top-level objectives.  The UI should have a stable ordering of the list of top-level objectives. 

2. The “Select existing Sub-Objective” modal will be difficult to use when there are many sub-objectives (like 100).   I would recommend adding a text input at the top of the modal that auto filters the sub objectives to help the user to find the sub objective that they wish to select.

mcass19:
Great @darrensiegel, let's hold merging this until we have the rewrite of the live view in place then. Thanks!

lib/oli_web/live/objectives/listing.ex
```diff
@@ -0,0 +1,77 @@
...
+              </div>
+            </div>
+
+            <div id={"collapse#{index}"} class={"collapse" <> if item.slug == @selected, do: " show", else: ""} aria-labelledby={"heading#{index}"} data-parent="#accordion">
+              <div class="card-body">
+                <div class="mb-3">
+                  <u>Sub-Objectives</u>
+                  <ul class="list-group list-group-flush">
+                    {#for sub_objective <- item.children}
+                      <li class="list-group-item p-2 text-info d-flex sub-obj">
```
mcass19, line 38: Not sure if `text-info` is a nice color to use here but liked more than primary or secondary (specially in dark mode). Open to any suggestion

lib/oli_web/live/objectives/listing.ex
```diff
@@ -0,0 +1,77 @@
...
+                <div><i class="fa fa-cubes c0183 mr-1"></i>Sub-Objectives {item.sub_objectives_count}</div>
+                <div><i class="far fa-file c0183 mr-1"></i>Pages {item.page_attachments_count}</div>
+                <div><i class="fa fa-list mr-1"></i>Activities {item.activity_attachments_count}</div>
+              </div>
+            </div>
+
+            <div id={"collapse#{index}"} class={"collapse" <> if item.slug == @selected, do: " show", else: ""} aria-labelledby={"heading#{index}"} data-parent="#accordion">
+              <div class="card-body">
+                <div class="mb-3">
+                  <u>Sub-Objectives</u>
```
mcass19, line 35: We could also make a collapse accordion here and just show `Sub-objectives` or `Pages` based on what the user wants to see (instead of showing all the info at once)

lib/oli_web/live/objectives/objectives.ex
```diff
@@ -1,253 +1,329 @@
...
-      <div class="mb-2 row">
-        <div class="col-12">
-          <p>
-            Learning objectives help you to organize course content and determine appropriate assessments and instructional strategies.
-            <br/>Refer to the <a href="https://www.cmu.edu/teaching/designteach/design/learningobjectives.html" target="_blank">CMU Eberly Center guide on learning objectives</a> to learn more about the importance of attaching learning objectives to pages and activities.
-          </p>
-        </div>
-      </div>
+    pid = self()
+    if first_load do
```
mcass19, line 115: We are fetching the attachments just on mount and then use it from the assigns, so we don't make all these queries on every update of objectives. That info is not going to change in this view, and we want to just display it

mcass19 (commented):
@darrensiegel implemented the new view in the last commits.

Some other notes apart from below comments
- Bump jquery because collapsing was not working with `3.5.0`
- Refactored everything to be written in Surface
- Fetching all objectives on mount and iterate from there when sorting, paging, building the tree, etc

Feel free to play with it + I'm really open to any suggestion you have in terms of UI/UX, usability, etc

image

lib/oli_web/live/objectives/objectives.ex
```diff
@@ -1,431 +1,447 @@
...
+      |> Enum.map(& &1.revision)
+
+    all_children =
+      all_objectives
+      |> Enum.reduce([], fn rev, acc -> rev.children ++ acc end)
+      |> Enum.uniq()
+
+    objectives =
+      Enum.reduce(all_objectives, [], fn rev, acc ->
+        case Enum.find(all_children, fn child_id -> child_id == rev.resource_id end) do
```
darrensiegel, line 93: We should take this opportunity to remove this `find` called within `reduce`, given that this is n^2 behavior.

lib/oli_web/live/objectives/objectives.ex
```diff
@@ -1,431 +1,447 @@
...
-      <%= live_component CreateNew, changeset: @changeset, project: @project %>
+    {:ok, table_model} = TableModel.new(objectives)
 
-      <hr class="my-4" />
+    pid = self()
+    if first_load do
+      Task.async(fn ->
+        objectives_attachments = Enum.map(objectives, fn obj ->
+          obj.resource_id
+          |> ObjectiveEditor.preview_objective_detatchment(project)
```
darrensiegel, line 119: Calling `preview_objective_detatchment` for each top-level objective is going to invoke a lot of database calls.  Internally this function issues 3 db calls. The chemistry course has 91 top level objectives, which means opening the Objectives view will throw 3 * 91 = 273 queries at the database.  

I would say that it is a hard requirement that this view has a constant number of database queries that get executed when it opens (regardless of whether they are synchronous within `mount` or async here in this impl).  To do that, we may have to rework how the page attachments and activity attachments are determined.  There might need to be introduced a new function that returns "page attachments" for all objectives, and one that returns "activity attachments" for all objectives - as opposed to the per objectives ones that are being used here. 

If it isn't feasible to rework some aspects of the data collection into a constant number of queries, then perhaps we can remove it from the table model, and request that data on a per objective basis when the user first clicks to expand a specific objective.

lib/oli_web/live/objectives/listing.ex
```diff
@@ -0,0 +1,77 @@
...
+                      </li>
+                      <div class="border border-light w-75"></div>
+                    {/for}
+                  </ul>
+                </div>
+                <div class="mb-3">
+                  <u>Pages</u>
+                  <ul class="list-group list-group-flush">
+                    {#for page <- item.page_attachments}
+                      <li class="list-group-item p-2 text-info">{page.title}</li>
```
darrensiegel, line 63: It would be really great here to render these page titles as links that route to the editor view for that page.

darrensiegel (changes requested):
Overall this is fantastic work.  See my comments inline for a couple of minor things to change, and one significant thing.

lib/oli_web/live/objectives/objectives.ex
mcass19: Yeah totally agree. I just used what we had but generates a lot of queries.

Will try to write a query to retrieve page + activities attachments for all objectives and see how it goes. Thanks !

darrensiegel (approved):

PR #3054, opened 2022-10-05
[FEATURE] [MER 1469] [MER 1487] provide backwards compatible precision support for numeric

eliknebel:
Adds support for numeric precision for legacy courses. Configuring a precision will tell the evaluator to check to a specific number of digits in the input response.
Screen Shot 2022-10-05 at 5 16 26 PM

This PR also includes a fix to add support for legacy numeric range format: `input like {[151.0,151.3]}`. Though, this addition will result in support for two different range formats in the system. Any creation/edit of numeric input activities will result in the new (or legacy) format as it is more concise and a bit easier to parse.

Finally, there is a bit of refactor around the way torus escapes/unescapes input. These functions were starting to leak into the codebase beyond the parsing layer into activities, it makes much more sense to keep them scope to the parsing step so that activities do not have to worry about this implementation detail. The addition of precision and ranges with inclusive/exclusive support required much of the client side parsing to be refactored. I also took the opportunity to fix a few bugs I stumbled across as well as ensure there are no valid states persisted (such as an empty `input {}`). I added some more tests around this to ensure the change isn't breaking anything.

Still to do:
- [x] Update course-digest to produce range rules as `input = {[151.0,151.3]}` (instead of `input like {[151.0,151.3]}`). The `=` operator is considered numeric in the system while `like` is used for string regular expressions. https://github.com/Simon-Initiative/course-digest/pull/101

### DIFF
--- a/.yarnrc
+++ b/.yarnrc
@@ -1,0 +1,1 @@
+--cwd assets
--- a/assets/src/components/activities/AuthoringElementProvider.tsx
+++ b/assets/src/components/activities/AuthoringElementProvider.tsx
@@ -4,6 +4,7 @@ import { Maybe } from 'tsmonad';
 import { ActivityModelSchema, MediaItemRequest, PostUndoable } from './types';
 import { ModalDisplay } from 'components/modal/ModalDisplay';
 import { AuthoringElementProps } from './AuthoringElement';
+import { ErrorBoundary } from 'components/common/ErrorBoundary';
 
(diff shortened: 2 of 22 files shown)


eliknebel:
I've identified another legacy format that will have to be handled: `input like {(59.99,80.01)#4}`

darrensiegel (approved):
Works very nicely.  Tested with existing authored Numeric range activities, works correctly.

PR #3057, opened 2022-10-11
[Feature] Adding a command button element [MER-1449]

marc-hughes:
This adds a new command button element which can send commands to other components on the page.

It also adds support for accepting those commands in the video player to start playing at a specific cue point and then pausing at an optional end point.

Demo:
https://user-images.githubusercontent.com/333265/195158159-c92c69e5-c568-419b-9583-fa10650a0108.mp4

### DIFF
--- a/assets/src/phoenix/app.ts
+++ b/assets/src/phoenix/app.ts
@@ -14,6 +14,7 @@ import { showModal } from './modal';
 import { enableSubmitWhenTitleMatches } from './package_delete';
 import { onReady } from './ready';
 import 'react-phoenix';
+import { commandButtonClicked } from '../components/editing/elements/command_button/commandButtonClicked';
 
 const csrfToken = (document as any)
   .querySelector('meta[name="csrf-token"]')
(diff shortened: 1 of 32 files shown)


assets/src/phoenix/app.ts
```diff
@@ -68,6 +69,8 @@ $(() => {
   ($('.ui.dropdown') as any).dropdown();
   ($('.ui.dropdown.item') as any).dropdown();
 
+  $('[data-action="command-button"]').on('click', commandButtonClicked);
```
marc-hughes, line 72: This is the sketchiest part of this PR that annoys me.

We've got ~3 ways of getting stuff on a page during delivery.

1. Just having phoenix render html
2. Having phoenix instantiate a react component to render
3. phoneix renders html, but it gets augmented jquery style (popovers, this, maybe others?)


I prefer #1 or #2, but this is a component with JS requirements *and* it has children rendered. I don't think #2 is appropriate in those cases?

darrensiegel (approved):
Nice work.  I tested that this works in Preview and in Delivery. Also demonstrated that it works when the button is in an Activity.  It works correctly even in the case where the Button is in Feedback for an activity (which gets dynamically added to the page after evaluation).

PR #3062, opened 2022-10-13
AOPS-1021 | Unable to submit/edit blue planet report

dtiwarATS:
Currently, the system does not allow students to submit additional attempts if their current attempt 'lifecycle_state' is in 'Active' or 'Submitted' state. 

As per the conversation with the team, if the 'lifecycle_state' is 'Submitted', we would like the student to be able to submit additional attempts even if the first attempt isn't evaluated or graded by the author yet.

### DIFF
--- a/lib/oli_web/controllers/page_delivery_controller.ex
+++ b/lib/oli_web/controllers/page_delivery_controller.ex
@@ -337,8 +337,7 @@ defmodule OliWeb.PageDeliveryController do
       |> Enum.reduce(%{}, fn {survey_id, activity_ids}, acc ->
         survey_state =
           Enum.all?(activity_ids, fn id ->
-            context.activities[id].lifecycle_state === :submitted ||
-              context.activities[id].lifecycle_state === :evaluated
+            context.activities[id].lifecycle_state === :evaluated
           end)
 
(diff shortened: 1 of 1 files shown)


darrensiegel (changes requested):
I do not understand why this change is being made:  this is in a bit of code that I believe only is called for Basic pages, not Adaptive.  This change would affect how surveys work in basic pages.

darrensiegel:
I think instead you want to change line 91 of `lib/oli/delivery/attempts/activity_lifecycle/evaluate.ex` from:

```
Enum.filter(part_attempts, fn pa -> pa.lifecycle_state == :active end)
```

To:

```
Enum.filter(part_attempts, fn pa -> pa.lifecycle_state == :active or pa.lifecycle_state == :submitted end)
```

dtiwarATS:
Hey @darrensiegel, I tried the solution that you mentioned but it's not working. I think that code check the lifecycle_state of the part attempts and I am trying to filter the resource attempt. 

It's tricky for me because I don't know the flow :)

dtiwarATS:
Something strange happening now. Even my original PR is also not fixing the issue now. I don't know. Closing this for now and will create another PR later.

PR #3066, opened 2022-10-14
[BUGFIX] Command button was submitting MCQ question [MER-1496]

marc-hughes:
Before, if you had a command button within an MCQ choice, and you clicked it, it would submit the MCQ question. Now, only the command is exectuted.

From bug:

```
Steps to reproduce:

1. Create a multiple choice question
2. Embed a video within the question stem.
3. In one of the choices, embed a command button that sends a message to the video.
4. Preview the question (or deploy it within a course section)
5. Click the button to start/stop the video.

Expected: the video starts and stops.
Actually: the video starts and stops, but the click event bubbles up to the choice and triggers a submission of the activity with that choice selected
```

### DIFF
--- a/assets/src/components/activities/common/choices/delivery/ChoicesDelivery.tsx
+++ b/assets/src/components/activities/common/choices/delivery/ChoicesDelivery.tsx
@@ -1,7 +1,7 @@
 import { Choice, ChoiceId } from 'components/activities/types';
 import { WriterContext } from 'data/content/writers/context';
 import { HtmlContentModelRenderer } from 'data/content/writers/renderer';
-import React from 'react';
+import React, { useCallback } from 'react';
 import { classNames } from 'utils/classNames';
 import styles from './ChoicesDelivery.modules.scss';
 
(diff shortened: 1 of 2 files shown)


assets/src/components/activities/common/choices/delivery/ChoicesDelivery.tsx
```diff
@@ -24,13 +24,27 @@ export const ChoicesDelivery: React.FC<Props> = ({
   selectedIcon,
 }) => {
   const isSelected = (choiceId: ChoiceId) => !!selected.find((s) => s === choiceId);
+
+  const onClicked = useCallback(
+    (choiceId: ChoiceId) => (event: React.MouseEvent) => {
+      if (event.isDefaultPrevented()) {
```
marc-hughes, line 30: Went with a preventDefault solution over stopPropagation to be a little more light-handed to let other components decide how to handle events.

darrensiegel (approved):

PR #3068, opened 2022-10-17
[BUGFIX] Rendering issues inside MCQ/CATA choices [MER-1502]

marc-hughes:
There were a series of rendering-issues with certain elements inside the choices of MCQ/CATA questions related to how they would resize vs. the width of the choice container element. Some of these, like Video, Youtube, and Webpage made the component unusable. Some, like in Dialog, introduced minor visual problems.

# Video

Before:
image

After:
image

# Youtube:

Before:
image

After:
image

# Webpages:

Before:
image

After:
image

# Dialogs

Before:
image

After:
image

# Callouts & Block Level Formulas:

Before:
image

After:
image
image

Tables:

Before:
image

After:
image

### DIFF
--- a/assets/src/components/activities/common/choices/delivery/ChoicesDelivery.tsx
+++ b/assets/src/components/activities/common/choices/delivery/ChoicesDelivery.tsx
@@ -49,7 +49,7 @@ export const ChoicesDelivery: React.FC<Props> = ({
         >
           <div className={styles.choicesChoiceWrapper}>
             <label className={styles.choicesChoiceLabel} htmlFor={`choice-${index}`}>
-              <div className="d-flex align-items-center">
+              <div className="d-flex align-items-center col">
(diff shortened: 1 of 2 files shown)


assets/src/components/activities/common/choices/delivery/ChoicesDelivery.tsx
```diff
@@ -49,7 +49,7 @@ export const ChoicesDelivery: React.FC<Props> = ({
         >
           <div className={styles.choicesChoiceWrapper}>
             <label className={styles.choicesChoiceLabel} htmlFor={`choice-${index}`}>
-              <div className="d-flex align-items-center">
+              <div className="d-flex align-items-center col">
```
marc-hughes, line 52: `col` here causes this part to grow in the flex layout `flex-grow:1` so any % based width children start laying out correctly.

darrensiegel (approved):

PR #3083, opened 2022-10-26
[FEATURE] Add description list element [MER-1507]

marc-hughes:
Adds a definition list element to core authoring. They consist of a title followed by a series of terms and definitions which get rendered out in html dl,dd,dt tags.

Example:
image

A note on implementation: Originally, in the json format, I had the dd/dt tags in the children property of the dl element. I liked that since you had just one editor and it flowed a little nicer. However, I ran into a couple slate JS specific issues with that method that I was unable to resolve with the core problem being that many times, invalid children of the DL would get added and eventually invalid HTML would be rendered.

We have that same issue with our normal lists. For example, if you try to insert an image into a list item, then go preview, you end up with a structure like this which isn't properly formatted html.  This may be a bug we need to address someday in the future.
image

### DIFF
--- a/assets/src/components/common/DescriptionList.tsx
+++ b/assets/src/components/common/DescriptionList.tsx
@@ -1,4 +1,4 @@
-import React, { MouseEventHandler, ReactNode } from 'react';
+import React from 'react';
 import * as ContentModel from 'data/content/model/elements/types';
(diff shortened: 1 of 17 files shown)


darrensiegel (approved):
Looks great, i pushed a commit just to tweak the rendering styles a bit to allow it to look closer to how it looks in legacy chaperone theme

PR #3086, opened 2022-10-31
[Chore] Webpack optimization [MER-1512]

marc-hughes:
This PR tweaks our webpack config for more efficient builds. Some results:

| Measurement |  Before PR | After PR |
| --------------- | ------------ | --------- |
| Production Build Time | 91 seconds | 88 seconds |
| Development incremental Build Time after modifying JS file | 10s | 2s |
| Development incremental Build Time after modifying scss file | 10s | 10s |
| Total JS File size in build dir | 118M | 31M |
| Total js resources loaded on core-authoring page | 76.9M | 30.5 |
| Total js resources loaded on development version of core-authoring page | 176M | 36M |
| Total js resources loaded on development version of adaptive-authoring page | 57M | 41M |
| Total js resources loaded on adaptive-authoring page | 25.8M | 23M |
| Total js resources loaded on preview of a core-authoring page | 76.9 | 27.9 |
| Total js resources loaded on preview of an adaptive-authoring page | 14.5M | 16.9M |
| Total JS resources on delivery-mode of the of a core authoring page.| 29.6M | 21.2M |
| Total JS resources on delivery-mode of the Blue Planet keeping balance adaptive lesson | 17.4M | 19.8M |

Two lines stand out there as being worse than before, namely the preview & delivery of adaptive lessons.  Those numbers hide the benefit that now, all our vendor code is shared on all pages, which accounts for 11M of that figure. If we assume these learners will also hit another page (sign in, the section list, etc.) than a bulk of that loading will already be done and cached in their browser and they'll have a better overall experience.

I kept a detailed log of everything I tried in the comment section of
https://eliterate.atlassian.net/browse/MER-1512

There are more ideal solutions, but it would require bigger changes and more vigilance in our development going forward because we have so many separate entry-points and code is shared heavily between those entry points.

### DIFF
--- a/.github/actions/torus-builder/entrypoint.sh
+++ b/.github/actions/torus-builder/entrypoint.sh
@@ -10,8 +10,8 @@ mix deps.get --only prod
 MIX_ENV=prod SHA=$RELEASE_SHA mix compile
 
 yarn --cwd ./assets
-npm run deploy --prefix ./assets
-npm run deploy-node --prefix ./assets
+NODE_ENV=production npm run deploy --prefix ./assets
+NODE_ENV=production npm run deploy-node --prefix ./assets
 
 MIX_ENV=prod mix phx.digest
 MIX_ENV=prod SHA=$RELEASE_SHA mix release
--- a/assets/package.json
+++ b/assets/package.json
@@ -5,6 +5,7 @@
   "repository": {},
   "license": "MIT",
   "scripts": {
+    "analyze-bundle": "BUNDLE_ANALYZER_ENABLED=true node --max-old-space-size=8192 node_modules/webpack/bin/webpack.js --mode production",
     "deploy": "node --max-old-space-size=8192 node_modules/webpack/bin/webpack.js --mode production",
     "deploy-node": "webpack --mode production --config webpack.config.node.js",
     "watch": "webpack --mode development --watch",
@@ -142,6 +143,7 @@
     "ts-loader": "^9.2.1",
     "type-coverage": "^2.18.0",
     "webpack": "^5.36.0",
+    "webpack-bundle-analyzer": "^4.7.0",
     "webpack-cli": "^4.7.0"
   },
   "typeCoverage": {
--- a/assets/webpack.config.js
+++ b/assets/webpack.config.js
@@ -6,8 +6,12 @@ const { ESBuildMinifyPlugin } = require('esbuild-loader');
 const CopyWebpackPlugin = require('copy-webpack-plugin');
 const MonacoWebpackPlugin = require('monaco-editor-webpack-plugin');
 
+const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin;
+
 const MONACO_DIR = path.resolve(__dirname, './node_modules/monaco-editor');
 const SHADOW_DOM_ENABLED = [path.resolve(__dirname, './src/components/parts/janus-fill-blanks')];
+
+const BUNDLE_ANALYZER_ENABLED = process.env.BUNDLE_ANALYZER_ENABLED === 'true';
 
 // Determines the entry points for the webpack by looking at activity
 // implementations in src/components/activities folder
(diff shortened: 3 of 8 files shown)


.github/actions/torus-builder/entrypoint.sh
```diff
@@ -10,8 +10,8 @@ mix deps.get --only prod
 MIX_ENV=prod SHA=$RELEASE_SHA mix compile
 
 yarn --cwd ./assets
-npm run deploy --prefix ./assets
-npm run deploy-node --prefix ./assets
+NODE_ENV=production npm run deploy --prefix ./assets
+NODE_ENV=production npm run deploy-node --prefix ./assets
```
marc-hughes, line 14: Webpack recommends setting NODE_ENV=production for prod builds since some modules look for that env var.

assets/package.json
```diff
@@ -5,6 +5,7 @@
   "repository": {},
   "license": "MIT",
   "scripts": {
+    "analyze-bundle": "BUNDLE_ANALYZER_ENABLED=true node --max-old-space-size=8192 node_modules/webpack/bin/webpack.js --mode production",
```
marc-hughes, line 8: If you run this, you'll get a browser window that you can explore what's in each .js chunk.

assets/webpack.config.js
```diff
@@ -121,8 +125,23 @@ module.exports = (env, options) => ({
   },
   devtool: 'source-map',
   optimization: {
-    minimize: process.env.NODE_ENV == 'production',
+    minimize: options.mode == 'production',
```
marc-hughes, line 128: Because we weren't setting NODE_ENV in our build, we weren't actually minimizing our .js files, for example, right now, this isn't minimized:
https://tokamak.oli.cmu.edu/js/oli_multi_input_authoring-dd7471757314849a789170675c6adb4f.js?vsn=d

While I did set NODE_ENV in the build, I also moved this to use options.mode like was probably intended.

assets/webpack.config.js
```diff
@@ -213,7 +232,7 @@ module.exports = (env, options) => ({
           {
             loader: 'css-loader',
             options: {
-              sourceMap: true,
+              sourceMap: options.mode !== 'production',
```
marc-hughes, line 235: This was weird, we were getting CSS sourcemaps embedded inside the .js chunks.

assets/webpack.config.js
```diff
@@ -287,10 +307,19 @@ module.exports = (env, options) => ({
...
     new MiniCssExtractPlugin({
       filename: '../css/[name].css',
     }),
-    new CopyWebpackPlugin({ patterns: [{ from: 'static/', to: '../' }] }),
+    new CopyWebpackPlugin({
+      patterns: [
+        {
+          from: 'static/',
+          to: '../',
+          filter: async (path) => path.indexOf('.map') == -1, // these were causing a duplicate file error on a production build
```
marc-hughes, line 319: Once minimization was actually running, webpack would fail here, so filtering out the .map files.

lib/oli_web/templates/layout/chromeless.html.eex
```diff
@@ -36,6 +36,7 @@
 
     <!-- Janus Part Component external CSS Files -->
     <link rel="stylesheet" href="/css/janus_image_carousel_delivery.css"/>
+    <link id="vendor-css" rel="stylesheet" href="/css/vendor.css" />
```
marc-hughes, line 39: Should these be using Routes.static_path ?

eliknebel: Probably, but idt its a huge deal either way since its unlikely we would move the static assets location. Maybe a good idea for consistency if nothing else

eliknebel (approved):
Great analysis. I agree there are probably more invasive optimizations that can be made but this seems like a good tradeoff between short term improvements and a massive overhaul of styles, etc.

lib/oli_web/templates/layout/chromeless.html.eex
marc-hughes: That's what I figured.

One use-case of moving the static assets would be plopping them on a CDN to speed things up a bit, so if we ever get there we should audit all the links / scripts

PR #3087, opened 2022-10-31
[MER-1463] Add tests GradesLive

simonchoxx:
[MER-1463](https://eliterate.atlassian.net/browse/MER-1463)

This PR adds tests for the GradesLive view.

Connection to the LMS, downloading the gradebook, updating LMS line items and synchronizing grades are tested.

Files are added to be able to handle the execution of the **_fetch_access_token_** function depending on the environment in which it is running. This change has been made as it was failing in the test environment due to the time it takes to execute.

### DIFF
--- a/test/oli_web/live/grades_live_test.exs
+++ b/test/oli_web/live/grades_live_test.exs
@@ -1,0 +1,142 @@
+defmodule OliWeb.ObjectivesLiveTest do
+  use ExUnit.Case
+  use OliWeb.ConnCase
+
+  import Phoenix.LiveViewTest
+  import Phoenix.ConnTest
+  import Oli.Factory
+  import Mox
+
+  alias Oli.Test.MockHTTP
+  alias OliWeb.Router.Helpers, as: Routes
+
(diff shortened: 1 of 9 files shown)


test/oli_web/live/grades_live_test.exs
```diff
@@ -0,0 +1,329 @@
...
+      |> render_click()
+
+      assert has_element?(view, "div.alert.alert-danger", "Error getting LMS access token")
+    end
+  end
+
+  describe "sync grades" do
+    setup [:admin_conn, :section_with_assessment]
+
+    test "shows the results to synchronization successfully", %{
```
mcass19, line 231: Same as above, what if we have some grades to sync?

mcass19 (commented):
Awesome work 👏🏼 👏🏼 👏🏼 
Left some small comments but is great to add tests for this view in particular

mcass19 (commented):
Aaaand how about squashing that first `wip` commit with another one? So we don't have that "I don't know what I'm changing" commit there 😄

mcass19 (approved):

darrensiegel (approved):

PR #3088, opened 2022-11-01
[FEATURE] Ability to customize Unit, Module, Section labels [MER-1481]

rgachuhi:
This PR adds support for the ability to customize various labels such as those supported in the curriculum. (Unit, Module, and Section). This feature works such that an author can set up default labels for a project, which are then inherited by a course instance that is created from this course project.  It also allows course section creators the ability to override these package-level defaults. This feature is set up to be flexible enough to support additional kinds of labels as the project needs expand.


lib/oli_web/common/links.ex
```diff
@@ -80,6 +80,17 @@ defmodule OliWeb.Common.Links do
           title =
             if numbering do
               Numbering.prefix(numbering) <> ": " <> revision.title
+              # case project.customizations do
```
darrensiegel, line 83: Remove the commented out code

lib/oli_web/live/progress/resource_title.ex
```diff
@@ -11,13 +11,15 @@ defmodule OliWeb.Progress.ResourceTitle do
...
     length = length(assigns.node.ancestors)
+    numbering = assigns.node.numbering
 
     ~F"""
     <div>
       <div>
         <small class="text-muted">
           {#for {ancestor, index} <- Enum.with_index(@node.ancestors)}
-            <span>{Oli.Resources.Numbering.container_type(ancestor.section_resource.numbering_level)} {ancestor.section_resource.numbering_index}</span>
+            {n = %Oli.Resources.Numbering{numbering | level: ancestor.section_resource.numbering_level, index: ancestor.section_resource.numbering_index}}
```
eliknebel, line 21: According to the liveview docs, it is bad practice to assign variables in the template due to the way phoenix does change tracking, I'm not sure but I imagine its probably similar for surface views https://hexdocs.pm/phoenix_live_view/assigns-eex.html#pitfalls

eliknebel, line 21: According to the liveview docs, it is bad practice to assign variables in the template due to the way phoenix does change tracking, I'm not sure but I imagine its probably similar for surface views https://hexdocs.pm/phoenix_live_view/assigns-eex.html#pitfalls

lib/oli_web/live/common/custom_labels_form.ex
```diff
@@ -0,0 +1,24 @@
+defmodule OliWeb.Common.CustomLabelsForm do
+  use Surface.Component
+  alias Surface.Components.Form
+  alias Surface.Components.Form.{Field, TextInput, Label}
+
+  prop labels, :map, default: %{unit: "unit", module: "module", section: "section"}
```
darrensiegel, line 6: Are these the correct defaults? Shouldn't these values be capitalized?

darrensiegel (changes requested):
The work here updates the original ingest code, but not the V2 ingest code.  There are some changes necessary in both:

```
lib/oli/interop/ingest/processor/project.ex
lib/oli/interop/ingest/processor/products.ex
```

To apply custom labels to both the project and any products.

lib/oli_web/live/sections/edit_view.ex
```diff
@@ -92,6 +104,23 @@ defmodule OliWeb.Sections.EditView do
     end
   end
 
+  def handle_event("save_labels", %{"view" => params}, socket) do
+    socket = clear_flash(socket)
+
+    params = Map.merge(%{"unit" => "unit", "module" => "module", "section" => "section"}, params, fn _k, v1, v2 ->
```
eliknebel, line 110: I noticed when changing only a single value of these three, the other ones are reset to the default value. This is most likely because the text feild value is empty with the placeholder being set to the value. I think an easy fix here would be to set the value of the text field to the field value when the textfield renders, reserving the placeholder for only when the value is actually nil

lib/oli_web/common/links.ex
eliknebel, line 83: Should this code be removed?

rgachuhi: Done

lib/oli_web/live/common/custom_labels_form.ex
rgachuhi: Done

lib/oli_web/common/links.ex
rgachuhi: Done

lib/oli_web/live/sections/edit_view.ex
rgachuhi: Set empty text field to default value

rgachuhi:
>

rgachuhi:
> 

Interesting. I added the label support in both the old ingest.ex and in the new performant ingest. I have tested both ingests from the admin panel and they both seem to work

darrensiegel (approved):

PR #3093, opened 2022-11-02
[FEATURE] [MER 1450] alternatives alternative support

### LINKED ISSUE #2190: Provide Legacy Alternatives/Alternative compatibility
Legacy OLI provides the `alternatives` and `alternative` elements, which allow an author to provide two or more different representations of a piece of content, one of which is displayed at run-time based on a a course section / organization setting. This feature is used to allow a Statistics course, for example, to support multiple statistical software packages in their worked examples.  For every worked example in the course, an alternative would exist implemented in Excel, one in Minitab, and one in R.  Then, at a org level, a preference is set to determine which alternative to use across then entire course.

Torus needs to provide the same capability, but needs to do so at a more general level so that eventually the same infrastructure can power A/B testing and perhaps other features.

eliknebel:
This PR adds support for alternative content rendering, including user defined alternative groups which can then be selected as a preference by a learner. This implementation includes the ability for future extensibility by providing an alternatives strategy framework which more strategies can be implemented in the future to support other features such as A/B testing etc. The 2 strategies that are included in the PR are a UserPreference strategy (full legacy alternatives/alternative support) and a "select all" which simply selects all alternatives (not used, but a simple example included for the sake of completeness)

Alternative groups can be created project-wide by navigation from the Overview page
Screenshot 2022-11-02 at 5 55 19 PM

Alternative content can be authored according to the created groups
Screenshot 2022-11-02 at 5 55 09 PM

Editor includes helpful messages for when an alternative option no longer exists or is a duplicate of another
Screenshot 2022-11-01 at 3 59 34 PM
Screenshot 2022-11-01 at 3 59 51 PM
Screenshot 2022-11-01 at 3 59 56 PM

A dropdown menu is rendered during delivery for students to select their alternatives preference for a particular group
Screenshot 2022-11-02 at 5 59 55 PM

Closes #2190


assets/src/apps/page-editor/PageEditor.tsx
```diff
@@ -338,9 +339,17 @@ export class PageEditor extends React.Component<PageEditorProps, PageEditorState
 
     if (item !== undefined) {
       if (item.undoable.type === 'PageUndoable') {
-        this.update({
-          content: this.state.content.insertAt(item.undoable.index, item.undoable.item),
-        });
+        if (this.state.content.find(item.contentKey)) {
```
eliknebel, line 342: In order to support pushing an undoable of a nested item, this was modified to check if the undoable being restored exists with a certain unique key. If it does, then the item is replaced with the undoable (restore a nested change/deletion). If not, then the item is inserted (restore a top level deletion)

assets/src/apps/page-editor/PageEditor.tsx
```diff
@@ -551,32 +560,34 @@ export class PageEditor extends React.Component<PageEditorProps, PageEditorState
...
-                activityContexts={this.state.activityContexts}
-                onRemove={(key: string) => this.onRemove(key)}
-                onEdit={onEdit}
-                onEditActivity={this.onEditActivity}
-                onPostUndoable={this.onPostUndoable}
-                content={this.state.content}
-                onAddItem={onAddItem}
-                resourceContext={props}
-              />
+              <AlternativesContextProvider projectSlug={projectSlug}>
```
eliknebel, line 563: Alternative options are loaded at the top-level so that there is only a single network request and when the request resolves (or fails) all components are up to date at the same time.

assets/src/components/delivery/AlternativesPreferenceSelector.tsx
```diff
@@ -0,0 +1,88 @@
+import React, { useEffect, useState } from 'react';
```
eliknebel, line 1: Delivery react component that allows a learner to select their alternatives preference. When a selection is made/changed, an event goes out to update all other selector of the same alternative group on the page so they are all in sync.

Alternative options are passed to the component as part of rendering (no async call)

assets/src/components/editing/elements/page_link/PageLinkEditor.tsx
```diff
@@ -38,10 +38,22 @@ export const PageLinkEditor = ({ model, commandContext, attributes, children }:
   const showModal = () =>
     window.oliDispatch(
       modalActions.display(
-        <PageLinkModal
-          model={model}
-          commandContext={commandContext}
-          onDone={({ idref }: Partial<ContentModel.PageLink>) => {
+        <SelectModal
```
eliknebel, line 41: I refactored all disparate selection models to a single reusable component SelectModal

assets/src/components/resource/editors/ActivityBankSelectionEditor.tsx
```diff
@@ -134,3 +134,22 @@ export const ActivityBankSelectionEditor = (props: ActivityBankSelectionEditorPr
     </div>
   );
 };
+
```
eliknebel, line 137: I refactored all editor outline components to be part of their corresponding editor file for better organization and colocation of similar logic

assets/src/components/resource/editors/GroupBlock.tsx
```diff
@@ -1,94 +1,26 @@
 import React, { PropsWithChildren } from 'react';
```
eliknebel, line 1: Editor blocks were also refactored into their corresponding editor files, where applicable

assets/src/components/resource/editors/OutlineItem.tsx
```diff
@@ -0,0 +1,208 @@
+import React, { PropsWithChildren, useRef } from 'react';
```
eliknebel, line 1: Common outline item components and functions were broken out into this file and are imported into the different editor files as needed

assets/src/components/resource/editors/PurposeGroupEditor.tsx
```diff
@@ -0,0 +1,188 @@
+import React, { PropsWithChildren } from 'react';
```
eliknebel, line 1: Since we are starting to have more and more groups, I renamed this specific group to be PurposeGroupEditor to avoid confusion that this could be more general purpose that it actually is. The components and logic in here are specific to "purpose groups"

assets/src/components/resource/editors/dragndrop/handlers/drop.ts
```diff
@@ -30,7 +28,7 @@ export const dropHandler =
 
         const sourceIndex = content.findIndex((k) => k.id === droppedContent.id);
 
-        if (sourceIndex === []) {
+        if (sourceIndex.length === 0) {
```
eliknebel, line 31: Strict comparison of a list to [] will never be true, since they are different objects. Best way to check if list is empty is to use `length`

assets/src/data/content/resource.ts
```diff
@@ -24,33 +24,35 @@ export type ResourceContent =
...
+// Items that can be present as elements in a resource content array and contain
+// other resource content as children
+export type ResourceGroup =
+  | PurposeGroupContent
+  | SurveyContent
+  | AlternativesContent
+  | AlternativeContent;
 
-export const isNestableContainer = (content: ResourceContent) => {
+export const isResourceGroup = (content: ResourceContent) => {
```
eliknebel, line 41: Renamed "NestableContainer" to ResourceGroup for better clarity

assets/src/data/content/resource.ts
```diff
@@ -215,6 +249,21 @@ export interface GroupContent {
   children: Immutable.List<ResourceContent>;
 }
 
+export interface AlternativesContent {
+  type: 'alternatives';
+  id: string;
+  strategy: AlternativesStrategy;
+  children: Immutable.List<AlternativeContent>;
+  alternatives_id: number;
```
eliknebel, line 257: Unfortunately the clash of schema/server-side conventions means that this variable `alternatives_id` must be snake-cased instead of camel

assets/src/data/content/utils.tsx
```diff
@@ -127,9 +128,9 @@ export const positionRect = (
     // here we check to see if the view is scrolled past the top of the editor and if so
     // then we adjust the position the toolbar relative to the top of the window so it is
     // always in view
-    const topAdjustedScrollPos = parentRect.top - TOP_SCROLL_OFFSET;
+    const topAdjustedScrollPos = parentRect.top - topScrollOffset;
```
eliknebel, line 131: Added a bit more padding to the toolbar since it was uncomfortably close to the cursor

lib/oli/authoring/editing/page_editor.ex
```diff
@@ -203,7 +203,12 @@ defmodule Oli.Authoring.Editing.PageEditor do
            Publishing.get_published_objective_details(publication.id) |> trap_nil(),
          {:ok, objectives_with_parent_reference} <-
            construct_parent_references(objectives) |> trap_nil(),
-         {:ok, tags} <- Oli.Authoring.Editing.TagEditor.list(project_slug, author),
+         {:ok, tags} <-
+           Oli.Authoring.Editing.ResourceEditor.list(
```
eliknebel, line 207: Refactored the "TagEditor" into a more general "ResourceEditor" that can take a generic resource and do a limited set of operations which can be performed on any resource. This allowed me to reuse this module for alternatives edits, as well as any future version-able resources we might need in the future as well

lib/oli/publishing.ex
```diff
@@ -1235,6 +1236,28 @@ defmodule Oli.Publishing do
     Enum.map(results, fn [slug, title] -> %{slug: slug, title: title} end)
   end
 
+  def find_alternatives_group_references_in_pages(alternatives_resource_id, publication_id) do
```
eliknebel, line 1239: Using JSONB to directly query all resources for alternatives group references, same as objectives above

lib/oli/resources/alternatives.ex
```diff
@@ -0,0 +1,21 @@
...
+  alias Oli.Resources.Alternatives.AlternativesStrategyContext
+  alias Oli.Resources.Alternatives.SelectAllStrategy
+  alias Oli.Resources.Alternatives.UserSectionPreferenceStrategy
+
+  @doc """
+  Selects one or more alternatives using the element's specified strategy.
+
+  Returns a list of `Oli.Resources.Alternatives.Selection` structs.
+  """
+  def select(
```
eliknebel, line 11: Determines which strategy to use based on the element's "strategy" field. Really after this PR the only strategy that is possible to create in the editor is "user_section_preference"

lib/oli/resources/alternatives/user_section_preference_strategy.ex
```diff
@@ -0,0 +1,61 @@
...
+           section_slug,
+           MapSet.new([pref_key])
+         ) do
+      {:ok, %{^pref_key => pref}} ->
+        # return all children with display: :none except for the alternative that matches the selected preference
+        Enum.map(children, fn alt ->
+          if alt["value"] == pref do
+            %Selection{alternative: alt}
+          else
+            %Selection{alternative: alt, hidden: true}
```
eliknebel, line 37: User preference strategy actually returns all alternatives wrapped in a `%Selection{}` struct. This struct has a `hidden` field which tells the renderer which items to render with a `hidden` class so that only the user preferred (or default/first) alternative is initially shown. The preference selector adds/removes the `hidden` class as necessary when it is changed

darrensiegel (approved):

PR #3096, opened 2022-11-03
[BUGFIX] Crash on editing adaptive lesson popup json [MER-1471]

marc-hughes:
Fixes bug:

1. Create or open up any adaptive page
2. Add a popup
3. Click the edit button in the right hand side panel under the component tab that opens up the JSON
4. Scroll down and click save in the JSON popup

Watch the lesson crash. 

The issue here, was the json editor was editing the model after it had passed through `transformSchemaToModel` when it should have been using the original model value. 

For most components, this doesn't matter, because the user is only editing the id & custom fields. Popup would crash because it's version of `transformSchemaToModel` overwrote `custom`.

### DIFF
--- a/assets/src/apps/authoring/components/RightMenu/RightMenu.tsx
+++ b/assets/src/apps/authoring/components/RightMenu/RightMenu.tsx
@@ -3,7 +3,7 @@ import { UiSchema } from '@rjsf/core';
 import { updatePart } from 'apps/authoring/store/parts/actions/updatePart';
 import { JSONSchema7 } from 'json-schema';
 import { debounce, isEqual } from 'lodash';
-import React, { useCallback, useEffect, useState } from 'react';
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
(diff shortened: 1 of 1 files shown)


assets/src/apps/authoring/components/RightMenu/RightMenu.tsx
```diff
@@ -88,6 +88,12 @@ const RightMenu: React.FC<any> = () => {
   const [questionBankData, setBankData] = useState<any>();
   const [questionBankSchema, setBankSchema] = useState<JSONSchema7>();
   const [showConfirmDelete, setShowConfirmDelete] = useState<boolean>(false);
+
+  const selectedPartDef = useMemo(
+    () => findPartByIdInActivity(currentActivity, currentPartSelection),
+    [currentActivity, currentPartSelection],
+  );
+
   useEffect(() => {
```
marc-hughes, line 97: There are a lot of useState / useEffect combos in RightMenu that could be refactored to useMemo to make them simpler to understand and reduce the number of render cycles the component needs

bsparks (approved):

PR #3101, opened 2022-11-03
[FEATURE] [MER-1514] Collaborative Spaces Infra

mcass19:
[MER-1514](https://eliterate.atlassian.net/browse/MER-1514)

Add some infra for collaborative spaces feature.

### DIFF
--- a/lib/oli/resources/collaboration.ex
+++ b/lib/oli/resources/collaboration.ex
@@ -2,11 +2,14 @@ defmodule Oli.Resources.Collaboration do
   alias Oli.Authoring.Course
   alias Oli.Authoring.Course.Project
   alias Oli.Publishing
-  alias Oli.Publishing.Publication
-  alias Oli.Resources
-  alias Oli.Resources.{Collaboration.Post, ResourceType, Revision}
-  alias Oli.Repo
+  alias Oli.Publishing.{DeliveryResolver, Publication}
+  alias Oli.Delivery
+  alias Oli.Delivery.{DeliverySetting, Sections}
   alias Oli.Delivery.Sections.{Section, SectionResource}
+  alias Oli.Resources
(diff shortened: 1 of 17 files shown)


lib/oli/resources/collaboration.ex
```diff
@@ -0,0 +1,286 @@
...
+  def create_collaborative_space(attrs, %Project{} = project, page_slug, author_id) do
+    attrs =
+      Map.merge(attrs, %{
+        author_id: author_id,
+        resource_type_id: ResourceType.get_id_by_type("collabspace")
+      })
+
+    Repo.transaction(fn ->
+      with {:ok, %{resource: cs_resource, revision: cs_revision}} <-
+            Course.create_and_attach_resource(project, attrs),
```
darrensiegel, line 53: It seems like the collab space is being created in a new resource, then this new resource is referenced from a page revision. We don't need the new resource.  To create/enable a collab space in a page, all that should be necessary to do is to create a new revision of that page, with the collab_space_config set to `:enabled`

lib/oli/resources/collaboration.ex
```diff
@@ -0,0 +1,286 @@
...
+      from(
+        section in Section,
+        join: section_resource in SectionResource,
+        on: section.id == section_resource.section_id,
+        join: cs_revision in Revision,
+        on:
+          cs_revision.resource_id == section_resource.resource_id and
+            cs_revision.resource_type_id == ^collab_space_type_id,
+        join: page_revision in Revision,
+        on: page_revision.collab_space_id == cs_revision.resource_id,
```
darrensiegel, line 136: This will change after the collab space resource is eliminated

darrensiegel (changes requested):

darrensiegel (approved):

PR #3104, opened 2022-11-04
[BUG FIX] fix failing pattern match [MER-1547]

darrensiegel:
Fixes the returned list to add a fourth string to match properly the expected pattern


lib/oli/rendering/content/html.ex
```diff
@@ -179,7 +179,7 @@ defmodule Oli.Rendering.Content.Html do
   def tc(%Context{} = _context, next, attrs) do
     [click_class, audio_element, play_code, _] =
       case attrs["audioSrc"] do
-        nil -> ["", "", ""]
+        nil -> ["", "", "", ""]
```
marc-hughes, line 182: Oops, the audio_player added a return value a while back, guess I missed this one.

marc-hughes (approved):

PR #3112, opened 2022-11-08
[FEATURE] Add media-browser support to adaptive video & audio elements [MER-1528, MER-1527]

marc-hughes:
Use the torus media-manager to select audio and video URL's for the audio and video components in adaptive lessons. 

Audio editor allows you to preview audio from the editor.

Example Audio Editor:

image

Example Video Editor:

image

### DIFF
--- a/assets/src/components/hooks/useAudio.tsx
+++ b/assets/src/components/hooks/useAudio.tsx
@@ -15,6 +15,7 @@ import { useCallback, useRef } from 'react';
 export const useAudio = (src?: string) => {
   const audioRef = useRef<HTMLAudioElement>(null);
   const [isPlaying, setIsPlaying] = useState(false);
+
   const playAudio = useCallback((event?: any) => {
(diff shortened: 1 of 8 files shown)


assets/src/components/hooks/useAudio.tsx
```diff
@@ -15,16 +15,17 @@ import { useCallback, useRef } from 'react';
...
     event?.preventDefault && event?.preventDefault(); // Fixes https://eliterate.atlassian.net/browse/MER-1503
 
     const audio = audioRef.current;
     if (audio) {
       if (audio.paused) {
         audio.currentTime = 0;
-        setIsPlaying(true);
-        audio.currentTime = 0;
-        audio.play();
+        audio.play().then(() => setIsPlaying(true));
```
marc-hughes, line 26: Found a bug here, where if the audio url wasn't a valid audio file, and you tried to play it, it would get stuck in the playing state.

bsparks (approved):

PR #3129, opened 2022-11-15
[FEATURE] Allow components to be resized in advanced editor canvas [MER-1530]

marc-hughes:
Adds the ability to resize elements via on-screen controls in the adaptive lesson editor instead of having to type in width/height numbers in the right hand panel.

Demo:

https://user-images.githubusercontent.com/333265/201970814-c5b56980-8fef-4f6a-ba68-c7ce87f632b6.mp4


assets/package.json
```diff
@@ -57,12 +57,12 @@
     "react-bootstrap-typeahead": "^5.1.4",
     "react-color": "^2.19.3",
     "react-dom": "^17.0.2",
-    "react-draggable": "^4.4.3",
     "react-keyed-file-browser": "^1.13.1",
     "react-phoenix": "file:../deps/react_phoenix",
     "react-popper": "^2.3.0",
     "react-quill": "^1.3.5",
     "react-redux": "^7.2.4",
+    "react-rnd": "^10.3.7",
```
marc-hughes, line 65: react-rnd is based off of react-draggable, but adds the resizing support.

assets/src/components/activities/adaptive/components/authoring/LayoutEditor.tsx
```diff
@@ -37,97 +43,86 @@ const defaultHandler = async () => {
...
+  const [isDragging, setIsDragging] = useState(false);
 
   const [parts, setParts] = useState(props.parts);
-  const [selectedPartId, setSelectedPartId] = useState(props.selected || '');
+
   const [configurePartId, setConfigurePartId] = useState('');
-  const [selectedPart, setSelectedPart] = useState<any>(null);
-  const [isDragging, setIsDragging] = useState(false);
+
+  const selectedPartId = props.selected || '';
```
marc-hughes, line 84: Some more useState refactors, the currently selected part is really controlled higher up in the component chain, and this value would sometimes be out of sync here. 

I saw at least one bug with a component that was on a containing layer could stay selected when going into a screen. This fixes that sort of bug.

assets/src/components/activities/adaptive/components/authoring/LayoutEditor.tsx
```diff
@@ -37,97 +43,86 @@ const defaultHandler = async () => {
...
-      setSelectedPart(null);
-      /* console.log('PART SELECTION CHANGED', {
-        selectedPartId,
-        selectedPart: null,
-      }); */
-    }
-    // any time selection changes we need to stop editing
-    setConfigurePartId('');
-  }, [selectedPartId, parts]);
+  const toolbarPosition = { x: 0, y: 0 };
```
marc-hughes, line 100: toolbar position is wholly dependent on the current selected part, so it doesn't need state either.

assets/src/components/activities/adaptive/components/authoring/LayoutEditor.tsx
```diff
@@ -37,97 +43,86 @@ const defaultHandler = async () => {
...
+    toolbarPosition.y = y;
+  }
 
   // this effect is to cover the case when the user is clicking "off" of a part to deselect it
   useEffect(() => {
     const handleHostClick = (e: any) => {
       const path = e.path || (e.composedPath && e.composedPath());
       const pathIds =
         path?.map((node: HTMLElement) => node.getAttribute && node.getAttribute('id')) || [];
+      const classes = path?.map((node: HTMLElement) => node.className) || [];
```
marc-hughes, line 114: After replacing react-draggable with react-rnd, I ended up running into an odd case when trying to select a component with a shadow-dom. This check for isDraggable fixed that.

assets/src/components/activities/adaptive/components/authoring/LayoutEditor.tsx
```diff
@@ -139,7 +134,7 @@ const LayoutEditor: React.FC<LayoutEditorProps> = (props) => {
         props.hostRef.removeEventListener('click', handleHostClick);
       }
     };
-  }, [props, parts, configurePartId]);
+  }, [props.id, props.hostRef, parts, configurePartId]);
```
marc-hughes, line 137: I could be wrong, but I think adding `props` as a dependency would trigger this effect every single render cycle.

assets/src/components/activities/adaptive/components/authoring/LayoutEditor.tsx
```diff
@@ -148,39 +143,75 @@ const LayoutEditor: React.FC<LayoutEditorProps> = (props) => {
...
-    async (payload: any) => {
-      // console.log('AUTHOR PART DRAG', payload);
-      if (payload.dragData.deltaX === 0 && payload.dragData.deltaY === 0) {
+  /**
+   * Given a part ID, this will clone and modify values on part.custom and post the changes up the chain.
+   *
+   * Example:
+   *   modifyPartCustomProp("part-123", { x: 100, y: 200 });
+   */
+  const modifyPartCustomProp = useCallback(
```
marc-hughes, line 157: We had 3 places that made modifications to part properties that all replicated similar logic. I refactored them out to this helper (and now we have 4 since resizing was added)

assets/src/components/activities/adaptive/components/authoring/LayoutEditor.tsx
```diff
@@ -220,14 +251,14 @@ const LayoutEditor: React.FC<LayoutEditorProps> = (props) => {
 
   const handlePartDelete = useCallback(async () => {
     // console.log('AUTHOR PART DELETE', { selectedPart }, selectedPartId);
-    const filteredParts = parts.filter((part) => part.id !== selectedPart.id);
+    if (!selectedPartAndCapabilities) return;
```
marc-hughes, line 254: After adding some typing to the capabilities part, there was a potential-null compiler warning that popped up, that's the reason for this early return.

assets/src/components/activities/adaptive/components/authoring/LayoutEditor.tsx
```diff
@@ -511,22 +527,53 @@ const LayoutEditor: React.FC<LayoutEditorProps> = (props) => {
             onSubmit: defaultHandler,
             onResize: defaultHandler,
           };
+
           const disableDrag =
-            selectedPartId !== part.id ||
```
marc-hughes, line 515: I'm allowing an unselected part to be dragged/resized now. It seems like better UX to just drag an element instead of having to select it, and then start dragging.

PR #3137, opened 2022-11-18
MER-1510

dtiwarATS:


assets/src/data/persistence/extrinsic.ts
```diff
@@ -51,7 +52,27 @@ export const readGlobalUserState = async (
     if (lastSet()) {
       await lastSet();
     }
+    let refreshFromServer = false;
     if (keys) {
+      const cacheTimeThreshold = 500000;
```
bsparks, line 57: I don't think this is 5 minutes
```
const second = 1000;
const minute = 60 * second;
const threshold = 5 * minute; // 300,000
```

bsparks:
I'm a little concerned that even 5 minutes is somewhat arbitrary. It seems like no matter what the time is, there is always a risk of it happening at the same time as a write... I suppose however that this *does* cut down on network traffic and database queries, so maybe it will help.

PR #3139, opened 2022-11-21
[Bugfix] adaptive data bounce [MER-1561]

marc-hughes:
### Description of bug:

This bug is for the right-hand panel, but in a lot of places in the adaptive lesson editor, if you:

1.  Edit something
2.  Then edit something else before 1 finishes saving
3.  Your edit to 2 will be reset to data before 1.

**Easiest way I can reproduce this:**

Let’s say I want to change my dimensions from 700x500 to 800x600

image

I might type in 800, hit tab, type in 600 and expect those values to be there.

In reality what usually happens is I type in 800, hit tab, start typing 600, maybe get one or two characters in, and then it resets to 500 and I finish typing. So on each keypress I might get:

- 500 <- starting
- 6     <- typed 6
- 60   <- typed 0
- 500 <- it resets to the original on it’s own
- 5000 <- I type my last 0

 
**A similar symptom of the same type of issue:**

- You edit a value.
- You briefly see the old value.
- You then see your new value.

### Cause of bug

The main cause of this sort of bug is trying to debounce the saving of data to torus too soon.

We have a few redux-thunk actions (such as saveActivity) that actually do two things:

- Save the data to torus
- Update our internal data model

Previously, the entire saveActivity was getting wrapped in a debounce, which means our internal data update was also getting debounced, effectively delaying a change by a minimum of half a second. It's that half-second that was causing this bug since you might make other changes during it, and then that time elapses and reverts your change.

This fix moves the debounce from around saveActivity to around just the torus api call. That way our internal state immediately updates.

This PR only fixes the bug from happening in the right-hand panel, where it was most obvious. We may want similar changes in the future if other areas of the app are experiencing similar issues.

### DIFF
--- a/assets/src/apps/authoring/BottomPanel.tsx
+++ b/assets/src/apps/authoring/BottomPanel.tsx
@@ -64,7 +64,8 @@ export const BottomPanel: React.FC<BottomPanelProps> = (props: BottomPanelProps)
   const debounceSaveChanges = useCallback(
     debounce(
       (activity) => {
-        dispatch(saveActivity({ activity, undoable: true }));
+        // TODO - we could probably refactor this component to debounce inside saveActivity instead of here.
+        dispatch(saveActivity({ activity, undoable: true, immediate: true }));
       },
       500,
       { maxWait: 10000, leading: false },
--- a/assets/src/apps/authoring/components/PropertyEditor/schemas/lesson.ts
+++ b/assets/src/apps/authoring/components/PropertyEditor/schemas/lesson.ts
@@ -216,7 +216,7 @@ export const transformSchemaToModel = (schema: any) => {
     additionalStylesheets,
     title: schema.Properties.title,
     customCss: schema.Properties.customCSS || '',
-    customScript: schema.CustomLogic.customScript,
+    customScript: schema.CustomLogic.customScript || '',
   };
 };
 
--- a/assets/src/apps/authoring/components/RightMenu/RightMenu.tsx
+++ b/assets/src/apps/authoring/components/RightMenu/RightMenu.tsx
@@ -2,7 +2,7 @@
 import { UiSchema } from '@rjsf/core';
 import { updatePart } from 'apps/authoring/store/parts/actions/updatePart';
 import { JSONSchema7 } from 'json-schema';
-import { debounce, isEqual } from 'lodash';
+import { isEqual } from 'lodash';
 import React, { useCallback, useEffect, useMemo, useState } from 'react';
 import { Button, ButtonGroup, ButtonToolbar, Tab, Tabs } from 'react-bootstrap';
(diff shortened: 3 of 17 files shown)


assets/src/apps/authoring/BottomPanel.tsx
```diff
@@ -64,7 +64,8 @@ export const BottomPanel: React.FC<BottomPanelProps> = (props: BottomPanelProps)
   const debounceSaveChanges = useCallback(
     debounce(
       (activity) => {
-        dispatch(saveActivity({ activity, undoable: true }));
```
marc-hughes, line 67: By default, saveActivity now debounces the torus save.
I added an immediate mode for all areas of the app using saveActivity that I wasn't changing in this PR.

assets/src/apps/authoring/components/PropertyEditor/schemas/lesson.ts
```diff
@@ -216,7 +216,7 @@ export const transformSchemaToModel = (schema: any) => {
     additionalStylesheets,
     title: schema.Properties.title,
     customCss: schema.Properties.customCSS || '',
-    customScript: schema.CustomLogic.customScript,
+    customScript: schema.CustomLogic.customScript || '',
```
marc-hughes, line 219: There was a bug where you couldn't completely erase the customScript, this fixes that.

assets/src/apps/authoring/components/RightMenu/RightMenu.tsx
```diff
@@ -147,27 +147,9 @@ const RightMenu: React.FC<any> = () => {
           dispatch(updateSequenceItem({ sequence: cloneSequence, group: currentGroup }));
           dispatch(savePage({ undoable: true }));
         }
-        //debounceSaveBankSettings(currentGroup, currentSequence, bankShowCount, bankEndTarget);
       }
     },
-    [currentSequence],
-  );
-
-  const debounceSaveBankSettings = useCallback(
```
marc-hughes, line 156: This was dead code not used anywhere.

assets/src/apps/authoring/store/page/actions/savePage.ts
```diff
@@ -90,14 +91,14 @@ export const savePage = createAsyncThunk(
         advancedDelivery,
         displayApplicationChrome,
         custom: customUpdate,
-        customCss: payload.customCss || currentPage.customCss,
-        customScript: payload.customScript || currentPage.customScript,
-        additionalStylesheets: payload.additionalStylesheets || currentPage.additionalStylesheets,
+        customCss: payload.customCss ?? currentPage.customCss,
```
marc-hughes, line 94: Another bug where you couldn't fully delete these properties.

bsparks (approved):

PR #3151, opened 2022-11-30
[BUG FIX] MER-1583 re-enable instructor preview for admins

eliknebel:
This PR re-enables instructor preview for administrators.


darrensiegel:
Can you confirm that this preview mode works for admins when there is an adaptive page in the course hierarchy?

mcass19:
I think we'll need to adapt some preview endpoints to make this work. For instance containers and adaptive pages are throwing exceptions

eliknebel:
Can confirm, container and adaptive previews were broken. To unblock this feature for courses with basic pages, I added some logic to detect this cases and handle them gracefully. Containers now work in preview when there is no user and author is an admin and adaptive activities simply render an "unsupported" message.

Screenshot 2022-12-01 at 12 01 06 PM

mcass19 (approved):

darrensiegel (approved):

PR #3152, opened 2022-12-01
[BUG FIX] Allow grade passback to occur after manual scoring [MER-1460]

darrensiegel:
There's a lot of formatting changes here, but the only change of substance is line 328 of `manual_grading.ex` where I corrected the pattern that was attempting to be matched in the function definition of `maybe_initiate_grade_passback`.  

The results of `Oli.Delivery.Attempts.PageLifecycle.Graded.roll_up_resource_attempts_to_access` were being piped into `maybe_initiate_grade_passback`, yet `maybe_initiate_grade_passback` was trying to match `{:ok, %ResourceAttempt{}}` as the first argument when in actuality it is a `{:ok, %ResourceAccess{}}` that is being passed in.   Since there was a catch all function definition also, this case was never being hit, and thus passback never initiated.

### DIFF
--- a/lib/oli/delivery/attempts/manual_grading.ex
+++ b/lib/oli/delivery/attempts/manual_grading.ex
@@ -12,19 +12,28 @@ defmodule Oli.Delivery.Attempts.ManualGrading do
   alias Oli.Resources.Revision
   alias Oli.Delivery.Attempts.ActivityLifecycle.Evaluate
   alias Oli.Delivery.Attempts.Core
+
   alias Oli.Delivery.Attempts.Core.{
     ResourceAccess,
     ResourceAttempt,
     ActivityAttempt
   }
+
   alias Oli.Activities.Model
 
   def count_submitted_attempts(%Section{} = section) do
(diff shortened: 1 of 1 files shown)


lib/oli/delivery/attempts/manual_grading.ex
```diff
@@ -260,32 +283,52 @@ defmodule Oli.Delivery.Attempts.ManualGrading do
...
         |> maybe_initiate_grade_passback(section)
-      e -> e
+
+      e ->
+        e
     end
   end
 
-  defp maybe_initiate_grade_passback({:ok, %ResourceAttempt{resource_access_id: resource_access_id}}, %Section{id: section_id, grade_passback_enabled: true}) do
+  defp maybe_initiate_grade_passback({:ok, %ResourceAccess{id: resource_access_id}}, %Section{
```
darrensiegel, line 328: This is the only real change.  Corrects the pattern match to target ResourceAccess.

eliknebel (approved):

PR #3154, opened 2022-12-02
[BUG FIX] Move save button [MER-1586]

darrensiegel:
The save button got accidentally moved to the wrong location.

### DIFF
--- a/lib/oli_web/templates/project/overview.html.eex
+++ b/lib/oli_web/templates/project/overview.html.eex
@@ -38,6 +38,8 @@
         <%= select f, :publisher_id, Enum.map(@publishers, &{&1.name, &1.id}), class: "form-control", required: true %>
       </div>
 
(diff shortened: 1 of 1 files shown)


lib/oli_web/templates/project/overview.html.eex
```diff
@@ -59,15 +61,13 @@
...
       <% end %>
+    </div>
 
-    <%= submit "Save", class: "float-left btn btn-md btn-primary mt-2" %>
-      <div class="d-flex flex-row">
-        <div class="flex-grow-1"></div>
-        <%= submit "Save", class: "float-left btn btn-md btn-primary mt-2" %>
-      </div>
+    <div>
+      <%= submit "Save", class: "btn btn-md btn-primary mt-2" %>
```
eliknebel, line 67: It looks like there are still 2 save buttons that both belong to the same form. Is this intentional?

eliknebel (approved):

PR #3161, opened 2022-12-06
MER-1560 | Time Out Notification

dtiwarATS:
Hey @bsparks & @darrensiegel, I am not sure if there already exists a session time out key that can be passed to delivery. I added 'SCREEN_IDLE_TIMEOUT' in 'oli.env' file that holds the screen time out value at the moment. If we already have a key, then I can modify the code to pass that existing session timeout key.


lib/oli_web/controllers/page_delivery_controller.ex
```diff
@@ -287,7 +287,8 @@ defmodule OliWeb.PageDeliveryController do
             section.slug,
             context.page.slug,
             resource_attempt.attempt_guid
-          )
+          ),
+        screenIdleTimeOut: Application.fetch_env!(:oli, :screen_idle_timeout)
```
dtiwarATS, line 291: This is the property that delivery is expecting to get from server. I am not sure if there already exists a key that can be passed to delivery. I added 'SCREEN_IDLE_TIMEOUT' in oli.env file that holds the screen time out value.

dtiwarATS:
I will create another PR later

PR #3162, opened 2022-12-06
MER-1560 | Time-Out Notification

dtiwarATS:
Hey @bsparks and @darrensiegel, I am not sure if there already exists a session time out key that can be passed to delivery. I added 'SCREEN_IDLE_TIMEOUT' in 'oli.env' file that holds the screen time out value at the moment. If we already have a key, then I can modify the code to pass that existing session timeout key.


config/releases.exs
```diff
@@ -89,7 +89,8 @@ config :oli,
       ),
     favicons: System.get_env("BRANDING_FAVICONS_DIR", "/favicons")
   ],
-  node_js_pool_size: String.to_integer(System.get_env("NODE_JS_POOL_SIZE", "2"))
+  node_js_pool_size: String.to_integer(System.get_env("NODE_JS_POOL_SIZE", "2")),
+  screen_idle_timeout: String.to_integer(System.get_env("SCREEN_IDLE_TIMEOUT", "30"))
```
darrensiegel, line 93: All of the other timeout variables express their values in milliseconds.  It might make sense to follow that convention and convert `30` to `30000`, or alternatively to rename the variable to include the units: `SCREEN_IDLE_TIMEOUT_IN_SECONDS`

darrensiegel: I cannot find any other variable that designates type type of timeout, so I think it is fine to introduce a new variable for this.

darrensiegel (changes requested):

config/releases.exs
bsparks: ok, but ideally this should match the session timeout that is set for the whole system or else it's possible that the time could be wrong and the session will still timeout regardless of this helper. 
I assume there *is* a session timeout configured somewhere, perhaps it's just not configured in this way and is part of some other env script? I think we need to keep these 2 values in sync. @eliknebel do you know?

dtiwarATS: The requested variable name changes are done.

config/releases.exs
```diff
@@ -89,7 +89,9 @@ config :oli,
       ),
     favicons: System.get_env("BRANDING_FAVICONS_DIR", "/favicons")
   ],
-  node_js_pool_size: String.to_integer(System.get_env("NODE_JS_POOL_SIZE", "2"))
+  node_js_pool_size: String.to_integer(System.get_env("NODE_JS_POOL_SIZE", "2")),
+  screen_idle_timeout_in_seconds:
+    String.to_integer(System.get_env("SCREEN_IDLE_TIMEOUT_IN_SECONDS", "1,800,000"))
```
bsparks, line 94: I think it was an either or, this would imply that it's 1.8 million seconds. instead this would be in milliseconds. the rename to "IN_SECONDS" would mean that you leave it at `30*60` (for 30 minutes); might even make more sense to rename it to "IN_MINUTES" since it eases the configuration and it's not likely that anyone would configure this for less than 1 minute anyway

config/releases.exs
eliknebel: It sort of depends on whether or not a user is in a persistent session. In either case though I think it uses the POW default. For persistent sessions it looks like this is 30 days https://hexdocs.pm/pow/pow_persistent_session.html#configuration

bsparks: hmm, well I would think the entire reason for even having a feature like this is because students are getting timed out of their sessions while actively working on an adaptive page which is mostly client side and *not* doing a heart beat keep alive. so there must be something somewhere that is configuring this

bsparks: @darrensiegel @eliknebel do you agree? w/ ^^

eliknebel: If the adaptive page is making requests and passing the authenticated cookie along then the session should be extended and follow the defaults configured in pow. We use the keepalive heartbeat as a work around for liveviews which get upgraded to a websocket and do not make these requests and therefore were timing out.

I dont think torus has the concept of a session idle timeout anywhere else currently, so it seems reasonable to add a new config for one and perhaps default it to some reasonable value.

dtiwarATS:
Closing this PR as I have to target the 'post-release-integration' branch. I will create another one.

PR #3163, opened 2022-12-06
working hotspot delivery w/temp authoring

andersweinstein:
Adds an image-hotspot activity with working delivery, suitable for use with ingested courses. Activity may be defined as single selection (behaving like a MCQ) or multiple selection (behaving like CATA).

For authoring, there is only a provisional interface to allow hand-entering and editing of hotspot coordinates in text boxes as comma-separated lists of numbers as follows:

For circle: cx, cy, radius    Example:   150,150, 50
For rect: x1, y1, x2, y2.      Example:    100,100, 150,150
For polygon: x1, y1, x2, y2, x3, y3 [...]  Example: 10,10, 100, 10, 50, 50

Desired hotspot shape is derived from the number of coordinates given. This authoring method was included solely for to allow testing during development, but should be robust enough (though user-unfriendly) to be safely deployed: Unparseable fields in the comma-separated list are ignored and invalid coordinate lists should result in an inactive hotspot that has no effect until corrected.

Note also that changing question type between single and multi-selection re-initializes the question so that the correct choice reverts to the default of first choice in list.

### DIFF
--- a/assets/src/components/activities/image_hotspot/ImageHotspotDelivery.tsx
+++ b/assets/src/components/activities/image_hotspot/ImageHotspotDelivery.tsx
@@ -1,0 +1,195 @@
+import React, { useEffect, useReducer, useRef } from 'react';
+import ReactDOM from 'react-dom';
+import { DeliveryElement, DeliveryElementProps } from '../DeliveryElement';
+import { getShape, Hotspot, ImageHotspotModelSchema } from './schema';
+import * as ActivityTypes from '../types';
+import { StemDelivery } from '../common/stem/delivery/StemDelivery';
+import { configureStore } from 'state/store';
+import {
+  activityDeliverySlice,
+  ActivityDeliveryState,
+  initializeState,
+  setSelection,
+  resetAction,
+  PartInputs,
+  isEvaluated,
+  listenForParentSurveySubmit,
+  listenForParentSurveyReset,
+  listenForReviewAttemptChange,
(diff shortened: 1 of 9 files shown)


assets/src/components/activities/image_hotspot/ImageHotspotDelivery.tsx
```diff
@@ -0,0 +1,195 @@
...
+import { HintsDeliveryConnected } from '../common/hints/delivery/HintsDeliveryConnected';
+import { EvaluationConnected } from '../common/delivery/evaluation/EvaluationConnected';
+import { GradedPointsConnected } from '../common/delivery/graded_points/GradedPointsConnected';
+import { ResetButtonConnected } from '../common/delivery/reset_button/ResetButtonConnected';
+import { useDeliveryElementContext, DeliveryElementProvider } from '../DeliveryElementProvider';
+import { castPartId } from '../common/utils';
+import { toSimpleText } from 'components/editing/slateUtils';
+import guid from 'utils/guid';
+import { Choices } from '../common/choices/authoring/ChoicesAuthoring';
+import { ConsoleLogger } from 'typedoc/dist/lib/utils';
```
eliknebel, line 32: Should these unused imports be removed?

eliknebel (changes requested):
Is example_content.json being used anywhere or should it be removed?

eliknebel (approved):

PR #3166, opened 2022-12-06
[ENHANCEMENT] [MER-1153] PR1 Auto push publication to products and sections

### LINKED ISSUE #3134: Publication updates to product created course sections can violate intent of product
This issue really points to some holes in our logic for applying publication updates to course sections, specifically those created from a product.  

Keeping in mind that a key use case of products is to allow a product creator to provide an "open and free" version of their materials, where all graded pages are removed, consider the following steps:

1. Author creates a new project, adding a practice page and a graded quiz.  The project is published.
2. The author then creates a Product, customizing its content only by removing the graded quiz.
3. An instructor creates a course section *from the product*.  They now have a course section that contains just the practice page. 
4. The author adds to their project a second graded page, a "Final Exam", and issues a new publication.  The author does not update the product. 
5. The instructor sees this new publication is available and applies it to their open and free section.   

The problem here is that the course section now contains the practice page and the final exam.

eliknebel:
This PR adds the capability to automatically push a publication update to course sections and products from the publish page according to the approach outlined in the first section of the FDD https://docs.google.com/document/d/1nGmt1UokwrV3ss3HXfeEbOXpNhrTG59aaAZSB55o1R4/edit?pli=1#.

It also greatly improves the update "intelligence" by utilizing a custom diffing algorithm "AIRRO" in order to determine what updates to resources to apply depending on 3 different common cases.

Followup work will involve tracking changes within a product in order to better apply structural updates to products. This PR will only apply minor updates to a product, requiring a manual remix step in order to bring in new resources.

Finally this PR includes a refactoring of `alias Oli.Publishing.Publication` to a more proper location `alias Oli.Publishing.Publications.Publication`.

https://eliterate.atlassian.net/browse/MER-1153

Closes #3134


darrensiegel (approved):
Nice work. The Agent based diff sharing was an interesting approach.  I was worried when I saw it that it would not be robust to server restarts, but then I saw you have a JIT diff calculation that takes place if the agent does not have that state.

PR #3168, opened 2022-12-07
[ENHANCEMENT] Streamline saving of parts [MER-1597]

marc-hughes:
This will cause the adaptive player to batch attempts to save parts together in a single request and throttle them to a reasonable rate.

From a user-perspective, the only thing you might notice is a browser warning if you try to navigate away from a page while there is still a queue of requests outstanding.

The easiest way to see the problem is to enter an adaptive lesson with a slider component. If you slide that up and down, before this change you'll see hundreds of server requests. After this change, you'll see far fewer. (max of one per 2-seconds after releasing the slider, or every 5 seconds while actively manipulating it)

I've tested this on the inspark gas-laws and concentration lessons, as well as some simple ones I authored.

If you have a screen with multiple layers, you may see multiple requests since each layer may need to be updated individually, but the throttling rules will apply to each of those.


assets/src/components/parts/janus-slider/Slider.tsx
```diff
@@ -240,7 +240,7 @@ const Slider: React.FC<PartComponentProps<SliderModel>> = (props) => {
       const options = [];
       const numberOfTicks = (maximum - minimum) / snapInterval;
       for (let i = 0; i <= numberOfTicks; i++) {
-        options.push(<option value={i * snapInterval}></option>);
+        options.push(<option key={i} value={i * snapInterval}></option>);
```
marc-hughes, line 243: unrelated react warning here.

darrensiegel (approved):
Looks good.

bsparks (approved):
lgtm

PR #3172, opened 2022-12-09
[BUG FIX] Left arrow cursor navigation does not work properly through all inline types [MER-1603]

marc-hughes:
Fixes a cursor related bug and also inserting callouts now doesn't clear the selection.

https://user-images.githubusercontent.com/333265/206767114-9cac6aae-fe0f-4f35-9c20-956adcf20b17.mov

### DIFF
--- a/assets/src/components/editing/elements/callout/calloutActions.ts
+++ b/assets/src/components/editing/elements/callout/calloutActions.ts
@@ -22,6 +22,6 @@ export const insertInlineCallout = createButtonCommandDesc({
     const at = editor.selection;
     if (!at) return;
 
-    Transforms.insertNodes(editor, Model.calloutInline(), { at, select: true });
+    Transforms.wrapNodes(editor, Model.calloutInline(), { at, split: true });
   },
 });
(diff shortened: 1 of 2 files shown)


assets/src/components/editing/elements/callout/calloutActions.ts
```diff
@@ -22,6 +22,6 @@ export const insertInlineCallout = createButtonCommandDesc({
     const at = editor.selection;
     if (!at) return;
 
-    Transforms.insertNodes(editor, Model.calloutInline(), { at, select: true });
```
marc-hughes, line 25: This was the issue with callouts clearing selected content.

assets/src/components/editing/toolbar/HoverContainer.tsx
```diff
@@ -32,13 +32,18 @@ export const HoverContainer = (props: PropsWithChildren<Props>) => {
     [],
   );
 
-  const children = <span style={{ ...props.style }}>{props.children}</span>;
-
-  if (!isOpen) return children;
```
marc-hughes, line 37: This early return ended up being the problem.

If the structure of the react-tree changes, then slate does some weird things. The solution was to always render the `<Popover>` so the hierarchy doesn't change, but selectively tell it to be open or not.

darrensiegel (approved):

PR #3175, opened 2022-12-09
[BUG FIX] Videos, youtube unusable when placed in hints or feedback [MER-1602]

marc-hughes:
This fixes two bugs

1. If you inserted a video into a hint, the video would be sized based on other content in that hint. If there was no other content, it would be very tiny and unusable.
2. If you inserted ONLY a video into a single hint, the "request a hint" link would not show up.

The second one was caused by us filtering out "empty" hints with an improper definition of "empty"

### DIFF
--- a/assets/src/utils/common.ts
+++ b/assets/src/utils/common.ts
@@ -1,3 +1,5 @@
+import { ModelTypes, schema } from '../data/content/model/schema';
+
 /**
  * Returns the given value if it is not null or undefined. Otherwise, it returns
  * the default value. The return value will always be a defined value of the type given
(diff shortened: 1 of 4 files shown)


assets/src/utils/common.ts
```diff
@@ -31,6 +33,14 @@ function hasContent(item: any): boolean {
     if (!item) return false;
     if (typeof item?.type === 'string' && item.type !== 'p') return true;
     if (Array.isArray(item)) return item.some(hasContent);
+
+    if (item?.type) {
+      const s = schema[item.type as ModelTypes];
+      const sc = schema;
+      // We'll assume void elements are content.
+      if (s?.isVoid) return true;
+    }
```
marc-hughes, line 42: On the client, I could rely on the element schemas, but over in serverland, I ended up with a static list of elements since there's no concept of void-elements there.

darrensiegel (approved):

PR #3176, opened 2022-12-10
[BUG FIX] Part attempt creation fails during periods of system inconsistency [MER-1606]

darrensiegel:
# Background

## The First Performance Problem

The successful creation of activity and part attempts (which takes place when a student visits a page with activities for the first time) requires  knowing all of the part ids that exist within the JSON body of the activities found within the page.  Retrieving these part ids from the JSON of these activities via a postgres JSON query proved to be a performance bottleneck for the adaptive page, due to the sheer number of activities and parts for which attempts had to be created (a large adaptive activity can have hundreds of activities that total more than a thousand parts). 

To optimize this creation, a postgres materialized view (named `part_mapping`) was introduced that effectively pre-computed and stored the activity revision id and part id mappings.  This allowed the bulk part attempt creation INSERT query to JOIN over this simple mapping instead of having to query through the JSON - a huge performance improvement. 

## Another Problem Arises

Initially the `part_mapping` materialized view was refreshed every time a publication record changed, via a database trigger.  This had the net effect of refreshing this view when an author clicked "Publish" to make a new publication available.  This solution worked fine (i.e. quickly) on small datasets in dev environments, but when run against production sized datasets it took a long time (20 seconds).  

To address this problem, an asynchronous execution of the part mapping refresh was introduced, with some rate limiting so that there could be only one active and one pending refresh job.  This allowed the actual publication of a specific project to execute in a reasonably short time (under a second). By the time that an instructor would apply this new update, the part mapping async job had finished and everything in the system worked fine.

## Introduction of "auto publish"

Recent work introduced a feature that when an author publishes a course that they can opt-in to forcing all course sections to automatically move forward to that publication.  

# The Inconsistency Problem

The introduction of the "auto publish" feature significantly worsened a problem that had been present and lurking in the system: given the asynchronous nature of the part mapping refresh, a Torus system would go through periods of inconsistency where a new author-issued publication that was quickly applied to a course section by an instructor could lead to failures to create part attempt records.  

Prior to the "auto-publish" feature, here is typically how this publish and apply update timeline has been working:
1. Author adds an activity to a page and issues a new publication. (System enters inconsistent state for that publication)
2. The async part mapping refresh kicks off
3. 40-80 seconds later, that part refresh completes (System re-enters consistent state for that publication)
4. An instructor applies the course section update to their course section
5. A student views the page with that new activity, triggering code that JOINs through the part_mapping view to successfully create the part_attempt record

With the "auto-publish" feature present, here is now how this easily leads to failure:
1. Author adds an activity to a page and issues a new publication. (System enters inconsistent state for that publication) *AND* they select "auto-publish"
2. The async part mapping refresh kicks off
3. The auto-publish feature applies the new publication to a course section
4. A student visits the page with the new activity.  While the course section knows that the page has been updated and now contains this new activity, the part_mapping hasn't been refreshed yet, so the JOIN through it to create the part_attempt fails
5. 40-80 seconds later, that part refresh completes (System re-enters consistent state for that publication)

* Note: Even prior to the "auto-publish" feature an author that issued a new publication and then quickly applied that update to a course section could trigger a part attempt creation failure. 

# The Solution

The proper way to fix this is to completely remove the `part_mapping` materialized view and replace it with an actual table, one that can be updated efficiently and synchronously during author issued publication.  This approach is optimal for a couple of reasons:

1. It completely removes the "inconsistent" periods that the system had been entering, relative to each newly issued publication
2. It is far more efficient, form a db perspective, than the part mapping refresh.  That part_mapping refresh is a long running query (taking 40 seconds on argos prod). 

## Components of the solution

1. A database migration creates the new `revision_parts` table directly from the shape and data contained in the current `part_mapping` view.   That same migration drops the view and its related indices. 
2. `Oli.Publishing.publish_project` has been updated to synchronously insert any new {revision_id, part_id, grading_approach} tuples into the `revision_parts` table.  
3. The query that creates part attempts now JOINs through the `revision_parts` table instead of the `part_mapping`
4. The async part mapping refresh infrastructure has been removed

I also saw the `AIRRO` and three way merge tests had the wrong file extension and weren't being run.


lib/oli/publishing.ex
```diff
@@ -869,6 +864,42 @@ defmodule Oli.Publishing do
...
+  def insert_revision_part_records(publication_id) do
+    query = """
+      INSERT INTO revision_parts(part_id, grading_approach, revision_id)
+      SELECT DISTINCT t.parts->>'id' as part_id, t.parts->>'gradingApproach' as grading_approach, t.revision_id as revision_id FROM (
+        SELECT jsonb_path_query(r.content, '$."authoring"."parts"[*]') as parts,
+          r.id as revision_id
+        FROM published_resources pr
+          LEFT JOIN publications p ON p.id = pr.publication_id
+          LEFT JOIN revisions r ON r.id = pr.revision_id
+        WHERE pr.publication_id = $1 AND r.resource_type_id = 3) t
```
mcass19, line 887: Maybe we could get the `resource_type_id` dynamically

mcass19 (approved):

eliknebel (approved):

lib/oli/publishing.ex
darrensiegel: Ah yes indeed we could.  Specifying a param here and supplying it with `Oli.Resources.ResourceType.get_id_by_type("activity")` would do the trick

darrensiegel: I filed https://eliterate.atlassian.net/browse/MER-1609 for tracking this work

PR #3181, opened 2022-12-13
MER-1604

dtiwarATS:


assets/src/apps/delivery/layouts/deck/DeckLayoutFooter.tsx
```diff
@@ -444,6 +444,29 @@ const DeckLayoutFooter: React.FC = () => {
     }
   }, [lastCheckResults, isPreviewMode]);
 
+  const isActivityHistoryTimeStampIsZero = () => {
```
bsparks, line 447: the name of this function would imply the return type would be a boolean value, but it is more like `getCurrentActivityVisitTimestamp`

assets/src/apps/delivery/layouts/deck/DeckLayoutFooter.tsx
```diff
@@ -444,6 +444,29 @@ const DeckLayoutFooter: React.FC = () => {
...
+    const trackingStampKey = `session.visitTimestamps.${currentActivity?.id}`;
+    const globalSnapshot = getEnvState(defaultGlobalEnv);
```
bsparks, line 449: I guess this works, but aren't there methods also for just getting a single value?

bsparks (approved):

PR #3192, opened 2022-12-14
[BUGFIX] [MER-1627] ensure resolvers of type do not return deleted items

eliknebel:
Fixes an issue where alternatives options were still showing deleted alternatives groups as an option in the editor.

The underlying issue is that the resolvers `revisions_of_type` were not filtering out deleted revisions.


darrensiegel (changes requested):
I don't think we want to solve the alternatives issue in this manner.  There might be code in the system that is depending on the resolvers returning deleted items - if we take this change this late in regression testing we have a high risk of breaking something.

darrensiegel (approved):
Nevermind, this is a much safer change than what i originally realized.  The revisions_of_type is only used for fetching alternatives, at the moment

PR #3195, opened 2022-12-15
MER-1604 | Session Resume Issue

dtiwarATS:


assets/src/apps/delivery/store/features/adaptivity/actions/triggerCheck.ts
```diff
@@ -383,10 +386,11 @@ export const triggerCheck = createAsyncThunk(
       const hasFeedback = actionsByType.feedback.length > 0;
       const hasNavigation = actionsByType.navigation.length > 0;
       let expectedResumeActivityId = currentActivity.id;
+      const doesFirstEventHasNavigation = checkIfFirstEventHasNavigation(checkResult[0]);
       if (hasFeedback && hasNavigation) {
         const [firstNavAction] = actionsByType.navigation;
         const navTarget = firstNavAction.params.target;
-        if (navTarget !== expectedResumeActivityId) {
+        if (navTarget !== expectedResumeActivityId && doesFirstEventHasNavigation) {
```
dtiwarATS, line 393: Previously the idea was that, once user clicks on the 'Next' button, gets correct feedback and on clicking on the 'Next' button again, they would have taken to respective screen. If they choose to not click on the 'Next' button and instead leave the lesson, when they come back, we are redirecting them to the screen where they were supposed to if they have clicked when they left the lesson. This is because they already have a correct attempt for that screen.

During this process, I forgot to check if the first event (i.e., check results) has a navigation because if it's not then user should be redirected to the same screen when they reload. 

**I checked if the check results have any navigation instead of checking if first event in check results have navigation or not.**

bsparks (approved):

PR #3198, opened 2022-12-16
[MER-1613] Build hierarchy from previous_next_index on course index

mcass19:
[MER-1613](https://eliterate.atlassian.net/browse/MER-1613)

Analyzed the query to build the hierarchy itself and from a SQL perspective there isn't much else to improve as it goes through indices (basically primary keys). One minor improvement we can make is to not retrieve as much data that is being fetched now, since not all of it is needed but would require further investigation.

Instead we look for another approach using the `previous_next_index` field in the section. Calling `retrieve` from `PreviousNextIndex` should be safe enough to always retrieve the correct hierarchy at the moment.

The hierarchy query will only be called when that field is null, but any other time a student enters the index it will just use that data.

### DIFF
--- a/lib/oli_web/controllers/page_delivery_controller.ex
+++ b/lib/oli_web/controllers/page_delivery_controller.ex
@@ -1,30 +1,25 @@
 defmodule OliWeb.PageDeliveryController do
   use OliWeb, :controller
-  require Logger
 
   import OliWeb.Common.FormatDateTime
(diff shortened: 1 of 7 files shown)


darrensiegel (approved):

darrensiegel:
Tested locally, and works correctly.  Code changes look good.

PR #3201, opened 2022-12-19
[BUGFIX] Media browser crashing in popup editor [MER-1637]

marc-hughes:
Bug:

1. Create an adaptive page
2. Add a popup to it
3. Edit the content of the popup
4. Add an image, video, or audio element

Expected: the element is added
Actual: the browser crashes.

This is caused by that popup content editor being a part of the popup web component in it's own react context instead of being part of the main authoring react app.

After this fix, you will get the old text-input box to enter the urls for those 3 components. A future update will attempt to re-introduce the media browser.

popupmeia


bsparks (approved):
lgtm

PR #3203, opened 2022-12-20
[BUG FIX] Fix rendering of deeply nested course hierarchies [MER-1641]

darrensiegel:
The fix here was to greatly simplify the code that constructs the hierarchy and to add a unit test that demonstrates that it handles deeply nested containers.


mcass19 (approved):
Nice, looks easier with that recursion!

Tested with nested sections and works fine:
image

PR #3209, opened 2022-12-21
[BUG FIX] handle missing attrs in group rendering [MER-1644]

darrensiegel:
The issue here is that `group/3` specifies a pattern match that fails rendering if either the `id` or the `purpose` attributes are missing.  Handle this by generating and id and setting purpose to `"none"` when they are missing.

### DIFF
--- a/lib/oli/rendering/group/html.ex
+++ b/lib/oli/rendering/group/html.ex
@@ -18,6 +18,16 @@ defmodule Oli.Rendering.Group.Html do
     ]
   end
 
+  def group(%Context{} = context, next, params) do
+
+    id = Map.get(params, "id", UUID.uuid4())
+
+    params = Map.put(params, "id", id)
+    |> Map.put("purpose", "none")
+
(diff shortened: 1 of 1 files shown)


lib/oli/rendering/group/html.ex
```diff
@@ -18,6 +18,16 @@ defmodule Oli.Rendering.Group.Html do
     ]
   end
 
+  def group(%Context{} = context, next, params) do
+
+    id = Map.get(params, "id", UUID.uuid4())
+
+    params = Map.put(params, "id", id)
+    |> Map.put("purpose", "none")
```
mcass19, line 26: We should handle the `purpose` as the id. I mean, we can already have it in the params and use it from there, right?

darrensiegel: Ah yes, that is a great catch.  If only the id is missing this will reset the purpose

mcass19 (approved):

PR #3211, opened 2022-12-22
[FEATURE] [MER-1518] Collab space instructor view

simonchoxx:
[MER-1518](https://eliterate.atlassian.net/browse/MER-1518)

This PR adds the collab space view to the instructor preview. Here instructors, besides to have the same functionalities that students, will be able to: 

- Accept or reject posts and replies that need approval.
- Delete posts and replies.
- Archive and unarchive posts and replies.

Here is a video of the feature, but feel free to pull the branch and play with it. We're really open for any kind of feedback.

https://user-images.githubusercontent.com/16328384/209135094-31b9d9f6-4a9e-4781-bbb0-9c742be2b010.mov


lib/oli_web/live/collaboration_live/collab_space_view.ex
```diff
@@ -532,7 +532,4 @@ defmodule OliWeb.CollaborationLive.CollabSpaceView do
 
   defp is_archived?(:archived), do: true
   defp is_archived?(_), do: false
-
-  defp send_updated_posts(:submitted, _), do: send(self(), :updated_posts)
```
simonchoxx, line 409: Here, it's necessary broadcast the post that is sent always, since we may have instructors to broadcast that post in case approval is required. We have no way of differentiate when a post is broadcasted to a student or to an instructor. 
Are we okay with this?

lib/oli_web/live/collaboration_live/collab_space_config_view.ex
```diff
@@ -72,22 +79,23 @@ defmodule OliWeb.CollaborationLive.CollabSpaceConfigView do
...
-            <span class="badge badge-info ml-2" style="height: fit-content">{humanize(@collab_space_status)}</span>
+          <div class="d-flex flex-column">
+            <h3 class="card-title">Collaborative Space Config</h3>
+            <h6 class="d-flex align-items-center">Current status <span class="badge badge-info ml-2" style="height: fit-content">{humanize(@collab_space_status)}</span></h6>
           </div>
 
           {#case @collab_space_status}
             {#match :disabled}
-              <button class="btn btn-outline-primary" :on-click="enable">Enable</button>
+              <div><button class="btn btn-outline-primary" :on-click="enable">Enable</button></div>
```
mcass19, line 98: NIT: Seems like an unnecessary div 😄

mcass19 (approved):
Awesome work!!! 🚀 🚀

lib/oli_web/live/collaboration_live/collab_space_config_view.ex
simonchoxx: Yes, thank you mate! I saw it and I think it's necessary to wrap the button because if not this happens 👇

Captura de pantalla 2022-12-22 a la(s) 11 39 04

But if you know how to fix it, I appreciate if you let me know.

mcass19: You're right. Let's leave it then!

Something I noticed is that we are adding divs in every clause of the match instead of just adding one outside the div, something like:

```elixir
<div>
  {#case @collab_space_status}
    {#match :disabled}
      <button class="btn btn-outline-primary" :on-click="enable">Enable</button>
    {#match :enabled}
      <button class="btn btn-outline-primary" :on-click="archive">Archive</button>
      <button class="btn btn-outline-danger" :on-click="disable">Disable</button>
    {#match _}
      <button class="btn btn-outline-primary" :on-click="enable">Enable</button>
      <button class="btn btn-outline-danger" :on-click="disable">Disable</button>
  {/case}
</div>
```

What do you think?

simonchoxx: Yes! great! It works perfect! I'll fix it!

darrensiegel (changes requested):
There's an error thrown if someone logged in as an admin (an authoring account) tries to view a collab space.  The root cause is the pattern match in `OliWeb.CollaborationLive.CollabSpaceView` includes the `current_user_id` key from the session, which will not be there when it is simply an admin.  Removing that key from that match, and inserting something like this at least got me past that problem locally: 


```
{user, user_id} = case Map.get(session, "current_user_id") do
      nil ->
        id = Map.get(session, "current_author_id")
        {Accounts.get_author(id), id}
      id ->
        {Accounts.get_user_by(%{id: id}), id}
    end
```

I'm a little concerned about the pattern here where new post creation (and deletion, editing) triggers a query for all connected users.  This probably isn't bad in most common situations where five users are in the same room, but what kind of load are we putting into the system if there ever is a situation with hundreds of users in the same room?  In that case, one post creation triggers 100 near simultaneous DB queries to fetch all posts.  And if there are 100 users in the room, a good portion of them might be posting and replying to each other rapidly (especially in a chat room type configuration).  This could end up throwing hundreds of full post history queries at the DB, every second.    A delta based impl here would eliminated this problem.

mcass19:
> 

@darrensiegel Good catch about the admin, we only thought accessing the view as an instructor but it's true that an admin can do it too.

Related to the other concern, do you think a delta based implementation is worth working on? Or are we not that close now of having 100 active users at the same time?

darrensiegel:
Yes I think it is worth working on the delta implementation now.  We have no idea when a collab space may get 100 users.  It could be never, it could be on the second day after this feature is released.

lib/oli_web/live/collaboration_live/collab_space_view.ex
```diff
@@ -51,14 +51,21 @@ defmodule OliWeb.CollaborationLive.CollabSpaceView do
...
-    user = Accounts.get_user_by(%{id: current_user_id})
+    {user, user_id} = case Map.get(session, "current_user_id") do
+      nil ->
+        id = Map.get(session, "current_author_id")
+        {Accounts.get_author(id), id}
+      id ->
+        {Accounts.get_user_by(%{id: id}), id}
+    end
+
+    user = user
```
mcass19, line 68: 😱

test/oli_web/live/collaboration_live_test.exs
```diff
@@ -317,6 +317,24 @@ defmodule OliWeb.CollaborationLiveTest do
...
+    test "returns the collab space config", %{
+      conn: conn,
+      admin: admin,
+      project: project,
+      page_revision: page_revision
+    } do
+      conn =
+        conn
+        |> Pow.Plug.assign_current_user(admin, OliWeb.Pow.PowHelpers.get_pow_config(:author))
+        |> get(live_view_author_edit(project.slug, page_revision.slug))
```
mcass19, line 332: We should test that an admin can see the collab space accessing the preview route instead of this one

mcass19 (approved):

darrensiegel (approved):

PR #3212, opened 2022-12-22
MER-1630 - Alt Text on Video & Audio elements

marc-hughes:
This exposes alt-text as aria-label attributes on audio and video elements to support people using assistive devices.

I did a mini-redesign of the video settings to break the options up into tabs since that dialog was getting rather long. The new alt-text for that can be found on the accessibility tab.

This also fixes a bug where it was impossible to edit the audio clip after you inserted it because the settings dialog couldn't be opened.

image

Audio settings:
image

### DIFF
--- a/assets/src/components/editing/elements/audio/AudioElement.tsx
+++ b/assets/src/components/editing/elements/audio/AudioElement.tsx
@@ -4,7 +4,10 @@ import { EditorProps } from 'components/editing/elements/interfaces';
 import { updateModel } from 'components/editing/elements/utils';
 import { CaptionEditor } from 'components/editing/elements/common/settings/CaptionEditor';
 import { useSlate } from 'slate-react';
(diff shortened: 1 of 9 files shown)


darrensiegel (changes requested):
All looks good except in testing this out I noticed the columns in the table in the "Video Source" tab are in the wrong order with respect to the data.  "Content Type" and "URL" need to be swapped. 
Screenshot 2023-01-03 at 9 15 00 AM

marc-hughes:
> All looks good except in testing this out I noticed the columns in the table in the "Video Source" tab are in the wrong order with respect to the data. "Content Type" and "URL" need to be swapped. Screenshot 2023-01-03 at 9 15 00 AM

This should be all set now.

darrensiegel (approved):

PR #3213, opened 2022-12-28
[MER-1618] Fix bug for grades

augusto1024:


darrensiegel (approved):

test/oli_web/live/progress/student_resource_view_live_test.exs
```diff
@@ -0,0 +1,108 @@
...
+
+  describe "student resource view" do
+    setup [:instructor_conn]
+
+    test "score must be less or equal to out_of value", %{
+      conn: conn,
+      instructor: instructor
+    } do
+      [section_slug: section_slug, student_id: student_id, resource_id: resource_id] =
+        setup_section_resource_access(instructor)
```
mcass19, line 47: NIT: We can add this function to the describe `setup` since is something all tests will share, wdyt?

test/oli_web/live/progress/student_resource_view_live_test.exs
```diff
@@ -0,0 +1,108 @@
...
+
+      assert view
+             |> element("input[id='resource_access_score']")
+             |> render() =~ "2.5"
+
+      assert view
+             |> element("input[id='resource_access_out_of']")
+             |> render() =~ "2.5"
+
+      assert view
```
mcass19, line 103: NIT: we could also get the resource access here and assert if it has the score and out_of we expect

mcass19 (approved):
👏🏼 👏🏼

PR #3218, opened 2023-01-05
[BUG FIX] Apply [MER-1656] to hotfix branch

darrensiegel:
This simply applies the MER-1656 fix that has already landed to `master` to the `hotfix-v0.22.1` branch. It sets up the changelog and ticks the version in `mix.exs` as well


marc-hughes (approved):
I compared the original 1656 fix with this PR and verified only those changes came over.

PR #3219, opened 2023-01-05
[FEATURE] [MER-1636] Bulk update activity attempts all at once

mcass19:
[MER-1636](https://eliterate.atlassian.net/browse/MER-1636)

Intent to improve finalization of a page attempt by making only one query to update all activity attempts at once. Still need to update the part attempts within an iteration, as is needed to have that in place to have the correct values ​​to update the activity.

We should thoroughly test these changes, as I'm a bit worried this won't break anything else.

Feel free to ask questions or make any suggestions. In fact, we don't have to take these changes if you feel they are not necessary at all.

[MER-1636]: https://eliterate.atlassian.net/browse/MER-1636?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli/delivery/attempts/activity_lifecycle/evaluate.ex
+++ b/lib/oli/delivery/attempts/activity_lifecycle/evaluate.ex
@@ -377,8 +377,8 @@ defmodule Oli.Delivery.Attempts.ActivityLifecycle.Evaluate do
     {:ok, evaluations}
   end
 
-  def update_part_attempts_and_get_activity_attempts(resource_attempt_id, datashop_session_id) do
-    resource_attempt_id
+  def update_part_attempts_and_get_activity_attempts(resource_attempt, datashop_session_id) do
+    resource_attempt.id
     |> get_latest_activity_attempts()
     |> Enum.reduce_while(
       {0, [], [], []},
@@ -390,30 +390,10 @@ defmodule Oli.Delivery.Attempts.ActivityLifecycle.Evaluate do
            part_attempt_guids
          } = acc ->
         if activity_attempt.lifecycle_state != :evaluated and activity_attempt.scoreable do
-          # --- part attempts
-          activity_attempt = activity_attempt |> Repo.preload([:revision, :resource_attempt])
-          part_attempts = get_latest_part_attempts(activity_attempt.attempt_guid)
+          activity_attempt = Map.put(activity_attempt, :resource_attempt, resource_attempt)
 
-          part_inputs =
-            part_attempts
-            |> Enum.map(fn pa ->
-              {input, files} =
-                if pa.response,
-                  do: {Map.get(pa.response, "input"), Map.get(pa.response, "files", [])},
-                  else: {nil, nil}
-
-              %{
-                attempt_guid: pa.attempt_guid,
-                input: %StudentInput{input: input, files: files}
-              }
-            end)
-            |> filter_already_evaluated(part_attempts)
-
-          case activity_attempt
-               |> do_evaluate_submissions(part_inputs, part_attempts)
-               |> persist_evaluations(part_inputs, fn result -> result end, datashop_session_id) do
-            {:ok, _} ->
-              # --- activity attempt
+          case update_part_attempts_for_activity(activity_attempt, datashop_session_id) do
+            {:ok, part_inputs} ->
               part_attempts = get_latest_part_attempts(activity_attempt.attempt_guid)
 
               %{
@@ -424,7 +404,6 @@ defmodule Oli.Delivery.Attempts.ActivityLifecycle.Evaluate do
                 date_submitted: date_submitted
               } = determine_activity_rollup_attrs(part_inputs, part_attempts, activity_attempt)
 
-              # ---
               {:cont,
                {
                  i + 6,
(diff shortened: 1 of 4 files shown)


lib/oli/delivery/attempts/activity_lifecycle/evaluate.ex
```diff
@@ -377,44 +377,81 @@ defmodule Oli.Delivery.Attempts.ActivityLifecycle.Evaluate do
...
+      fn activity_attempt,
+         {
+           i,
+           activity_attempt_values,
+           activity_attempt_params,
+           part_attempt_guids
+         } = acc ->
+        if activity_attempt.lifecycle_state != :evaluated and activity_attempt.scoreable do
+          # --- part attempts
+          activity_attempt = activity_attempt |> Repo.preload([:revision, :resource_attempt])
```
mcass19, line 394: We can avoid two selects here if we preload these values on a join

lib/oli/delivery/attempts/activity_lifecycle/evaluate.ex
```diff
@@ -377,44 +377,81 @@ defmodule Oli.Delivery.Attempts.ActivityLifecycle.Evaluate do
...
+              }
+            end)
+            |> filter_already_evaluated(part_attempts)
+
+          case activity_attempt
+               |> do_evaluate_submissions(part_inputs, part_attempts)
+               |> persist_evaluations(part_inputs, fn result -> result end, datashop_session_id) do
+            {:ok, _} ->
+              # --- activity attempt
+              part_attempts = get_latest_part_attempts(activity_attempt.attempt_guid)
```
mcass19, line 417: We are getting the part attempts again here because we need them after being updated, but if we want to avoid this select, we could try something like returning more data on the above update and work from there

lib/oli/delivery/attempts/activity_lifecycle/persistence.ex
```diff
@@ -86,6 +85,24 @@ defmodule Oli.Delivery.Attempts.ActivityLifecycle.Persistence do
...
+        score = batch_values.score,
+        out_of = batch_values.out_of,
+        lifecycle_state = batch_values.lifecycle_state,
+        date_evaluated = batch_values.date_evaluated,
+        date_submitted = batch_values.date_submitted
+      FROM (VALUES #{values}) AS batch_values (activity_attempt_guid, score, out_of, lifecycle_state, date_evaluated, date_submitted)
+      WHERE activity_attempts.attempt_guid = batch_values.activity_attempt_guid
+    """
+
+    Ecto.Adapters.SQL.query(Oli.Repo, sql, params)
```
simonchoxx, line 103: I ask you to have a better understanding, this could have been written using Ecto expressions, right? Is there a more efficient way?

simonchoxx (approved):
Nice work mate!! 🚀🚀

lib/oli/delivery/attempts/activity_lifecycle/persistence.ex
darrensiegel: I'm unaware of a way to do bulk record updates directly with Ecto.  This `batch_values` approach here is used in a few other places within Torus.  It might be worthwhile to eventually create a "builder" module for more easily creating and maintaining these types of queries.  Something that takes this out of string literal space back into Elixir space. I could imagine something like:

```
BulkUpdate.update(:activity_attempts)
|> BulkUpdate.set([:score, :out_of, :lifecycle_state, :date_evaluated, :date_submitted])
|> BulkUpdate.from_values(values)
|> BulkUpdate.matching_on(:attempt_guid)
```

Not for this PR, but perhaps in the future.

darrensiegel (changes requested):
Overall this is looking good.  

I would ask you to pay careful attention to the readability and maintainability of `update_part_attempts_and_get_activity_attempts`.  It is getting a bit long, we may need to break it down and group what it does into some smaller functions to allow it to be more understandable and approachable in the future.

lib/oli/delivery/attempts/activity_lifecycle/evaluate.ex
```diff
@@ -377,44 +377,81 @@ defmodule Oli.Delivery.Attempts.ActivityLifecycle.Evaluate do
...
+                     out_of,
+                     Atom.to_string(lifecycle_state),
+                     date_evaluated,
+                     date_submitted
+                   ],
+                 part_attempt_guids ++
+                   Enum.map(part_attempts, fn part_attempt -> part_attempt.attempt_guid end)
+               }}
+
+            error -> {:halt, error}
```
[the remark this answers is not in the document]
mcass19: We're doing a `reduce_while` so we can stop the iteration returning a `halt`. 

Did that because if the part attempts fail to update and we continue, there will be invalid data to update the activity attempt (and then will be invalid data for the resource attempt, etc etc), so I'd rather stop the iteration and since everything is inside a transaction, roll back all the changes.

Anyway I'm really open to any other suggestions here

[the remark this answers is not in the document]
simonchoxx: Excellent. Very clear explanation. Thank you for it. 🙌

lib/oli/delivery/attempts/activity_lifecycle/persistence.ex
mcass19: Like it! Created this ticket to track the change for the future: https://eliterate.atlassian.net/browse/MER-1665

darrensiegel (approved):

PR #3221, opened 2023-01-06
[FEATURE] Allow Collab spaces to ingest [MER-1657]

darrensiegel:
This extends the ingest V2 to allow it to read the `collabSpace` key that can optionally exist as a top-level key in a digest "Page" JSON blob.  

Testing this, convert a course that contains discussions (for example, French 1), then ingest and verify that that end-to-end that a Discussion activity is converted to a practice page with an enabled Collab Space.  You will need this PR to land to `course-digest` first, though: https://github.com/Simon-Initiative/course-digest/pull/120

### DIFF
--- a/lib/oli/interop/ingest/processor/pages.ex
+++ b/lib/oli/interop/ingest/processor/pages.ex
@@ -1,6 +1,7 @@
 defmodule Oli.Interop.Ingest.Processor.Pages do
   alias Oli.Interop.Ingest.State
   alias Oli.Interop.Ingest.Processing.Rewiring
+  alias Oli.Resources.Collaboration.CollabSpaceConfig
   import Oli.Interop.Ingest.Processor.Common
 
   def process(%State{} = state) do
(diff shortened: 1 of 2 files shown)


lib/oli/interop/ingest/processor/pages.ex
```diff
@@ -22,6 +23,34 @@ defmodule Oli.Interop.Ingest.Processor.Pages do
...
+  # Read the collab space config in a robust manner, falling back to defaults if
+  # the entire collabSpace key is missing or if any sub keys are missing
+  defp read_collab_space(%{"collabSpace" => config}) do
+
+    # Read the status in a way that falls back to the proper default
+    # status if no status is specified or if an unsupported status is specified
+    default_status_str = CollabSpaceConfig.default_status() |> Atom.to_string()
+    candidate_status = Map.get(config, "status", default_status_str)
+
+    status = case CollabSpaceConfig.status_values()
```
mcass19, line 35: Leave other way of doing this if it's clearer, but anyway looks great 🚀 

image

mcass19 (approved):

PR #3229, opened 2023-01-18
Annotaiton proof of concept

marc-hughes:
# ⚠️ This is an example proof of concept and not meant to actually be merged at this time.

Upcoming work will be requesting an annotation / highlighting feature to be added to the torus discussion feature. This will let a student highlight a section of content on a page, and other students or educators to see that highlight and comment on it.

Some requirements to consider:

- Ability to point-annotate objects. Such as images, videos, or other content where the entire thing should be annotated.
- Ability to annotate ranges of elements. (start here, highlight through here… may include several torus elements.
- Ability to annotate ranges of text within an element.

Annotation in this context means identifying the items and showing a visual indicator around them. Additional text to actually “annotate” them will be in a separate panel with the

## There are three interesting challenges

This proof of concept shows a way of doing the first two, and comments in this PR will show some viable paths for the third.

1. Rendering the highlights on a page
2. Handling the user-input of specifying those highlights
3. Storing those highlights in a way that will be resilient to content changing.

Some things to remember:

- Elements can both be in page-content and within activities.
- Some content is server-rendered, some is react-rendered and annotation should work seamlessly across.
- Not all students get the same content. Things like question-banks exist.
- Not all students see the content in the same order. Example: MCQ options are randomized order.
- React (and probably live view???) doesn’t like an external script modifying the dom of elements it controls, so there’s render-challenges there.

## Out of scope for the first version:

- Annotating within shadow-dom based elements.
- Annotating within iframe content
- Adaptive lessons (for V1 annotation feature)
- Marking specific parts of elements (section of an image, cue point of a video, etc.)
- User input - we won’t annotate things the learner does / answers. Like _what_ option they selected in an MCQ, or what text they type in an input.

# Description of the solution

For rendering the highlights, there's two primary options.

1. Modifying the content to display highlights
2. Overlaying an element over the page to display highlights.

This POC takes the second option, primarily to solve issues with rendering the highlights across both client-rendered and server-rendered content without having to implement it twice.

For handling user-input, there are two potential API calls that come in useful for detecting what specific text ranges are at a point on the screen.

```
    document.caretPositionFromPoint(clientX, clientY);
    document.caretRangeFromPoint(clientX, clientY);
```

Most modern browsers support caretRangeFromPoint, however Firefox does not. So we'll need to use caretPositionFromPoint for Firefox.

Neither of these work when something interactable is overlaid on top of the content, so we need to make our highlight overlay `pointer-events: none;` and deal with document-level mouse events.

# Quick demo:

annotations

### DIFF
--- a/assets/src/apps/AnnotatorApp.tsx
+++ b/assets/src/apps/AnnotatorApp.tsx
@@ -1,0 +1,4 @@
+import { Annotator } from './annotator/Annotator';
+import { registerApplication } from './app';
+
+registerApplication('Annotator', Annotator);
--- a/assets/src/apps/annotator/Annotator.tsx
+++ b/assets/src/apps/annotator/Annotator.tsx
@@ -1,0 +1,212 @@
+import React, { useCallback, useMemo } from 'react';
+import { useToggle } from '../../components/hooks/useToggle';
+import useWindowSize from '../../components/hooks/useWindowSize';
+import { useDocumentMouseEvents } from './useDocumentMouseEvents';
+
+interface TextAnnotation {
+  id: string;
+  startOffset: number;
+  endOffset: number;
+}
+
+const getHeadingBlockHighlight = () => {
+  const heading = document.querySelector('#E236710925');
+  const rects = heading?.getClientRects();
+  // console.info(rects);
+  if (rects?.length) {
+    return rects![0];
+  }
+  return null;
+};
+
+const getHeadingParagraphTextHighlight = () => {
+  return getAnnotation('E236710926', 10, 19);
+};
+
+const getAnnotation = (id: string, startOffset: number, endOffset: number) => {
+  const element = document.querySelector('#' + id);
+  if (!element) return [];
+  if (!element.firstChild) return [];
+
+  try {
+    const range = document.createRange();
+    range.setStart(element.firstChild, startOffset);
+    range.setEnd(element.firstChild, endOffset);
+
+    const rects = range?.getClientRects();
+    if (rects?.length) {
+      return Array.from(rects);
+    }
+  } catch (e) {
+    console.info(e);
+  }
+
+  return [];
+};
+
+const getHeadingParagraphTextHighlight2 = () => {
+  return getAnnotation('E236710926', 99, 150);
+};
+
+const getIdFromTextNode = (node: Node | null): string | null => {
+  if (!node) return null;
+  if (node.nodeType === Node.TEXT_NODE) {
+    return getIdFromTextNode(node.parentElement);
+  }
+  if (node.nodeType === Node.ELEMENT_NODE) {
+    const id = (node as Element).attributes.getNamedItem('id')?.value;
+    if (id) return id;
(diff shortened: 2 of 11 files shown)


assets/src/apps/AnnotatorApp.tsx
```diff
@@ -0,0 +1,4 @@
+import { Annotator } from './annotator/Annotator';
+import { registerApplication } from './app';
+
+registerApplication('Annotator', Annotator);
```
marc-hughes, line 4: Implemented as a new top-level OLI application to be added to a delivery page.

Right now, it's self contained, but we'll need to make a public API that the discussion feature can use to talk with it. Either an object globally accessible or a series of events.

assets/src/apps/annotator/Annotator.tsx
```diff
@@ -0,0 +1,212 @@
+import React, { useCallback, useMemo } from 'react';
+import { useToggle } from '../../components/hooks/useToggle';
+import useWindowSize from '../../components/hooks/useWindowSize';
+import { useDocumentMouseEvents } from './useDocumentMouseEvents';
+
+interface TextAnnotation {
```
marc-hughes, line 6: Right now, it just stores an offset from an element's ID. To make this more resilient and work across child elements, we'll need to also store:

1. What child-element inside that ID it is
2. Either a short start/end snippet, or some sort of checksum to tell if the content has changed.

This POC only highlights stuff that is inside elements that have been marked as highlightable with a custom data-annote attribute. This would allow us to properly scope what is highlighted, especially on pieces that might change ordering or not be available for all students.

For example, each MCQ option should be a separate highlightable element. 

When we store the highlights, a single "user highlight" might span multiple data-annotate elements, so we'll store them as a list.

assets/src/components/activities/common/stem/delivery/StemDelivery.tsx
```diff
@@ -15,7 +15,10 @@ interface StemProps {
 
 export const StemDelivery: React.FC<StemProps> = (props) => {
   return (
-    <div className={`stem__delivery${props.className ? ' ' + props.className : ''}`}>
+    <div
```
marc-hughes, line 18: We will need to go in and add ID attributes throughout.

assets/src/data/content/writers/html.tsx
```diff
@@ -124,8 +124,12 @@ export class HtmlParser implements WriterImpl {
     );
   }
 
-  p(context: WriterContext, next: Next, _x: Paragraph) {
-    return <p>{next()}</p>;
+  p(context: WriterContext, next: Next, attrs: Paragraph) {
+    return (
+      <p id={'E' + attrs.id} data-annotate="text">
```
marc-hughes, line 129: We will need to go in and add ID and data-annotate attributes throughout where appropriate. 
Note: ID's can't start with a number, prepended "E" here.

marc-hughes:
Closing this PR as it's documentation and a proof-of-concept and not meant to be merged.

PR #3230, opened 2023-01-18
[FEATURE] [MER-1648] Account lookup cache

mcass19:
[MER-1648](https://eliterate.atlassian.net/browse/MER-1648)

Add a cache for account lookup using [Cachex](https://hexdocs.pm/cachex/Cachex.html):
- Get account from cache first in set current user plug, and put entry when reading from db
- Remove entry when logging out
- Update entry on context update functions

To make it work in a distributed scenario, we implemented a GenServer which will be the responsible for broadcasting and consuming changes in remote nodes. Also, did some logic to initialize the cache with the info from other node in the cluster if there is any.

[MER-1648]: https://eliterate.atlassian.net/browse/MER-1648?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/oli_web/account_lookup_cache.ex
```diff
@@ -0,0 +1,134 @@
...
+        PubSub.broadcast(Oli.PubSub, cache_topic(), {:delete, key})
+        {:reply, :ok, state}
+
+      _ ->
+        {:reply, :error, state}
+    end
+  end
+
+  def handle_call({:put, key, value}, _from, state) do
+    ttl = :timer.hours(24)
```
mcass19, line 69: This is an arbitrary number, we can change it to whatever we feel is right, or turn it into an env var

lib/oli_web/controllers/session_controller.ex
```diff
@@ -22,4 +25,15 @@ defmodule OliWeb.SessionController do
 
   defp session_data_to_delete(type),
     do: [String.to_atom("current_#{type}_id") | @shared_session_data_to_delete]
+
+  defp delete_cache_entry(conn, type) do
+    id =
+      conn.assigns
+      |> Map.get(String.to_existing_atom("current_#{type}"))
```
simonchoxx, line 32: To understand a little more... Here, are you getting an existing key in the assigns from the string you are passing? 🤔

simonchoxx (approved):
Awesome work you've done! Everything looks very clear!

lib/oli_web/controllers/session_controller.ex
mcass19: We are getting `:current_user` or `:current_author` from the assigns map in the connection `conn`

lib/oli_web/account_lookup_cache.ex
```diff
@@ -0,0 +1,134 @@
...
+
+    {:ok, [], {:continue, :init}}
+  end
+
+  def handle_continue(:init, state) do
+    nodes = Node.list()
+
+    if length(nodes) > 0 do
+      # just pick a random node in the cluster to request the dump
+      send({__MODULE__, Enum.random(nodes)}, {:request_dump, self()})
```
augusto1024, line 47: Is there a reason why it needs to be a random node and not just the first one?

lib/oli_web/account_lookup_cache.ex
augusto1024: Maybe this is too far-fetched, but if the nodes are in different timezones (which IDK if it happens or not), shouldn't this be set in the `handle_info`?

lib/oli_web/account_lookup_cache.ex
```diff
@@ -0,0 +1,134 @@
...
+        PubSub.broadcast(Oli.PubSub, cache_topic(), {:put, key, value, ttl})
+        {:reply, :ok, state}
+
+      _ ->
+        {:reply, :error, state}
+    end
+  end
+
+  def handle_call({:update, key, value}, _from, state) do
+    case Cachex.update(@cache_name, key, value) do
```
augusto1024, line 82: Don't we need to update the ttl here?

augusto1024 (commented):
Amazing job! I left some small comments and I was also wondering if we shouldn't use [`broadcast_from`](https://hexdocs.pm/phoenix_pubsub/Phoenix.PubSub.html#broadcast_from/5) instead of `broadcast` so we don't send the `update`, `put` and `delete` messages to the process that performed those actions

lib/oli_web/account_lookup_cache.ex
mcass19: Nop, we can choose always the first one if we like it more. 

Tbh I don't know if `Node.list()` always return in the same order but if that the case, a random one could be better to not load the same node always, wdyt?

lib/oli_web/account_lookup_cache.ex
mcass19: `:timer.hours(24)` just returns that time in miliseconds, is not based on any current time, so I assume each node will perform the `put` in the `handle_info` with that `ttl` but based on it's own time

lib/oli_web/account_lookup_cache.ex
augusto1024: Yeah, I agree 100%

mcass19:
@augusto1024 yeah, use `broadcast_from` sounds like a good idea, will try it! Thank you !

lib/oli_web/account_lookup_cache.ex
augusto1024: I don't have a strong opinion about it but since it's fresh data, I thought it might be useful to update the ttl. We can leave it like this and update it later if we need to

darrensiegel (changes requested):
This approach here is somewhat fractured in where it places cache usage:

In the Oli.Accounts module, cache updating is hidden within the calls that change users, which effectively makes updaing the cache an "implicit" operation from the standpoint of client code.   This impl also introduces an extra DB read with the calls to `get_user_with_roles` on every user write, which isn't ideal. 

Reading from the cache, however, is an "explicit" operation that client code is supposed to perform as seen in Oli.Plugs.SetCurrentUser.

We could make the cache usage entirely explicit.  That would involve removing the "updating" code from Oli.Accounts, and instead tracking down the places where these methods are called, and adding cache updating code there.    This opens up the risk to future code that introduces new user updating of not updating the cache.  So, I think a better approach would be change the implicit operation that is taking place within Oli.Accounts.  Instead of reading the user again and updating the cache, simply purge that cache entry.

darrensiegel (approved):

PR #3231, opened 2023-01-19
[MER-1643] Do not persist part attempts submission when evaluating an adaptive activity

mcass19:
[MER-1643](https://eliterate.atlassian.net/browse/MER-1643)

Don't persist in the db the part attempts submission when evaluating an activity in an adaptive page. Instead, update the list in memory and use that updated list in subsequent functions.

I understand this is only used to get a correct result when calling `determine_activity_rollup_fn`, so the change shouldn't be too risky, but I'd still spend some time testing that nothing is broken or that we don't need that data persisted elsewhere.

[MER-1643]: https://eliterate.atlassian.net/browse/MER-1643?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli/delivery/attempts/activity_lifecycle/evaluate.ex
+++ b/lib/oli/delivery/attempts/activity_lifecycle/evaluate.ex
@@ -32,7 +32,7 @@ defmodule Oli.Delivery.Attempts.ActivityLifecycle.Evaluate do
         evaluate_from_input(section_slug, activity_attempt_guid, part_inputs, datashop_session_id)
 
       {:ok, %Model{rules: rules, delivery: delivery, authoring: authoring}} ->
-        submit_active_part_attempts(part_attempts)
+        part_attempts_submitted = submit_active_part_attempts(part_attempts)
 
         custom = Map.get(delivery, "custom", %{})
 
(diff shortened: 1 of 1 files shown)


lib/oli/delivery/attempts/activity_lifecycle/evaluate.ex
```diff
@@ -501,7 +493,7 @@ defmodule Oli.Delivery.Attempts.ActivityLifecycle.Evaluate do
     case Oli.Activities.get_registration_by_slug(activity_registration_slug) do
       %Oli.Activities.ActivityRegistration{allow_client_evaluation: true} ->
         Repo.transaction(fn ->
-          part_attempts = get_latest_part_attempts(activity_attempt_guid)
+          part_attempts = part_attempts_input || get_latest_part_attempts(activity_attempt_guid)
```
simonchoxx, line 496: Here, the value of `part_attempts_input` received in the function will be used and, in case it's nil, the function `get_latest_part_attempts/1` will be called, I'm right?

simonchoxx (approved):
Nice work mate!!! 🚀🚀
I just left a comment to clear up a doubt!!

lib/oli/delivery/attempts/activity_lifecycle/evaluate.ex
mcass19: Exactly! 

The `||` operator accept any type (as opposite of `or`), and `nil` is falsy so evaluates to false, so as you mention, in that case will return what `get_latest_part_attempts/1` has

darrensiegel (approved):

PR #3232, opened 2023-01-20
[FEATURE] [MER-1646] Sections/Projects info in publish view

simonchoxx:
[MER-1646](https://eliterate.atlassian.net/browse/MER-1646)

This PR provides visibility for authors to know which and how many course sections and products are using their project materials.

This information is rendered in a paged and sortable table for better visualization.
In addition, it shows the number of course sections/projects that will be affected by force push update.

screenshot1

screenshot2

[MER-1646]: https://eliterate.atlassian.net/browse/MER-1646?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/oli_web/live/projects/table_model_active_sections.ex
```diff
@@ -0,0 +1,69 @@
+defmodule OliWeb.Projects.TableModelActiveSections do
```
mcass19, line 1: I'm pretty sure the convention is to use `TableModel` at the end 😄

test/oli_web/live/publish_live_test.exs
```diff
@@ -0,0 +1,374 @@
...
+      project = insert(:project)
+
+      assert conn
+             |> get(live_view_publish_route(project.slug))
+             |> html_response(302) =~
+               "You are being <a href=\"/authoring/session/new?request_path=%2Fauthoring%2Fproject%2F#{project.slug}%2Fpublish\">redirected</a>"
+    end
+  end
+
+  describe "user cannot access when is logged in as an author but is not a system admin" do
```
mcass19, line 146: Not sure if this case makes sense... the author can't access because is not author of the project more than because is not a system admin

lib/oli_web/live/projects/versioning_details.ex
```diff
@@ -0,0 +1,137 @@
...
+  def render(assigns) do
+    ~F"""
+      <div class="my-4 border-top pt-3">
+        <Form for={@changeset} submit={@publish_active}>
+
+          {#if @has_changes && @active_publication_changes}
+            <h5>Versioning Details</h5>
+            <h6 class="pb-3">The version number is automatically determined by the nature of the changes.</h6>
+            {#case @version_change}
+              {#match {change_type, _}}
```
mcass19, line 37: Seems we are doing just one `match` for this `case`

lib/oli_web/live/projects/versioning_details.ex
```diff
@@ -0,0 +1,137 @@
...
+    """
+  end
+
+  defp render_product_count(assigns, product_count, section_count) do
+    ~F"""
+      {#case product_count}
+        {#match 0}
+
+        {#match 1}
+          {#if section_count > 0}and{/if}<strong> {product_count} product</strong>
```
mcass19, line 112: We can just do something like

```
This force push update will affect:
- 2 sections
- 0 products
```

What do you think? So we don't need to do this `sections and products` logic? Same with the `s` at the end, we can just do `section(s)` always

lib/oli_web/live/projects/versioning_details.ex
```diff
@@ -0,0 +1,137 @@
...
+          <strong> {section_count} course section</strong>
+
+        {#match _}
+          <strong> {section_count} course sections</strong>
+      {/case}
+    """
+  end
+
+  defp render_version(edition, major, minor) do
+    "v#{edition}.#{major}.#{minor}"
```
mcass19, line 135: We have this render version somewhere else right? Seems like a great possibility to move this a common place like `Utils` and use it always from there

mcass19 (commented):
Awesome job man!!! 👏🏼 👏🏼 

I really like what you did re-writing to Surface and adding tests 🚀 🚀 🚀 🚀

lib/oli_web/live/projects/publish_view.ex
```diff
@@ -0,0 +1,251 @@
...
+  @table_filter_fn &__MODULE__.filter_rows/3
+  @table_push_patch_path &__MODULE__.live_path/2
+
+  def filter_rows(socket, _, _), do: socket.assigns.active_sections
+
+  def live_path(socket, params),
+    do: Routes.live_path(socket, __MODULE__, socket.assigns.project.slug, params)
+
+  def mount(
+        %{"project_id" => project_slug} = _params,
```
augusto1024, line 39: NIT: Since you're not using the `_params` assignment maybe we can remove it

lib/oli_web/live/projects/publish_view.ex
```diff
@@ -0,0 +1,251 @@
...
+              redirect_uris={@redirect_uris}
+              tool_url={@tool_url}
+            />
+          </div>
+        </div>
+      </div>
+    """
+  end
+
+  def handle_event("publish_active", %{"publication" => publication} = _params, socket) do
```
augusto1024, line 201: NIT: We can skip the `_params` assignment here

augusto1024 (commented):
LGTM! Left some NITs but feel free to ignore them

lib/oli/delivery/sections.ex
```diff
@@ -561,6 +561,50 @@ defmodule Oli.Delivery.Sections do
...
+      iex> get_push_force_affected_sections(project)
+      %{product_count: 1, section_count: 1}
+  """
+  def get_push_force_affected_sections(project) do
+    Repo.one(
+      from(
+        section in Section,
+        join: spp in SectionsProjectsPublications,
+        on: section.id == spp.section_id,
+        where: spp.project_id == ^project.id and section.status == :active,
```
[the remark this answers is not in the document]
simonchoxx: I changed it. I think it's now as it should work. 🙌

lib/oli/delivery/sections.ex
```diff
@@ -561,6 +561,58 @@ defmodule Oli.Delivery.Sections do
...
+  ## Examples
+      iex> get_push_force_affected_sections(project)
+      %{product_count: 1, section_count: 1}
+  """
+  def get_push_force_affected_sections(project_id) do
+    today = DateTime.utc_now()
+
+    Repo.one(
+      from(
+        section in Section,
```
darrensiegel, line 603: This will not return correct counts in all cases.  Force pushing an update will only affect course sections that are exactly on the most recent publication.  This function needs to take that into consideration.  To do this, you will need to pass in (or lookup in the query) the most recent publication id.  Take a look at `Oli.Publishing.fetch_products_and_sections_eligible_for_update/2` to see the logic for how the force publish feature identifies the sections that it can force update.

darrensiegel (changes requested):

lib/oli/delivery/sections.ex
simonchoxx: All right! I'll fix it then!

lib/oli/delivery/sections.ex
```diff
@@ -561,6 +561,58 @@ defmodule Oli.Delivery.Sections do
     |> Repo.all()
   end
 
+  @doc """
+  Gets all sections that use a particular project when their 'end_date' attribute is not nil and is later than the current date.
+
+  ## Examples
+      iex> get_active_sections_by_project(project, now_time)
```
mcass19, line 568: I think this doc needs an update 😄

lib/oli_web/live/projects/publication_details.ex
```diff
@@ -0,0 +1,48 @@
+defmodule OliWeb.Projects.PublicationDetails do
+  use Surface.Component
+
+  alias Oli.Utils
+  alias OliWeb.Common.{Utils}
```
mcass19, line 5: Are these two alias correct or just one of them?

lib/oli_web/live/projects/versioning_details.ex
```diff
@@ -0,0 +1,118 @@
+defmodule OliWeb.Projects.VersioningDetails do
+  use Surface.Component
+
+  alias OliWeb.Common.Utils
+
+  alias Surface.Components.{Form}
```
mcass19, line 6: NIT: remove `{}`

test/oli/sections_test.exs
```diff
@@ -338,6 +338,65 @@ defmodule Oli.SectionsTest do
       remixed_projects = Sections.get_remixed_projects(section.id, section.base_project_id)
       assert 2 = length(remixed_projects)
     end
+
+    test "get_active_sections_by_project/2 returns course sections actives for a project" do
```
mcass19, line 342: Maybe we can add a test for the cases where the two new functions in the context returns empty (or zero). Wdyt?

test/oli/sections_test.exs
```diff
@@ -338,6 +338,65 @@ defmodule Oli.SectionsTest do
...
+          base_project: project,
+          type: :enrollable,
+          start_date: yesterday(),
+          end_date: tomorrow()
+        )
+
+      {:ok, _sr} = Sections.create_section_resources(section1, publication)
+      {:ok, _sr} = Sections.create_section_resources(section2, publication)
+
+      assert length(Sections.get_active_sections_by_project(project.id)) == 2
```
mcass19, line 365: NIT: pipe instead of composite functions? 😄

mcass19 (approved):
Looks great! Left minor things but take what you want.

After fixing that count for the sections/products affected by the force push, for me is good to go 👍🏼 
Great job 🚀

lib/oli/delivery/sections.ex
simonchoxx: Sure, nice catch!

test/oli/sections_test.exs
simonchoxx: Sure! I definitely have to make friends with the pipes! 😂

mcass19 (approved):

darrensiegel (approved):

PR #3237, opened 2023-01-25
[FEATURE] [MER-1646] Add necessary fields for exploration pages

augusto1024:
[Link to the story](https://eliterate.atlassian.net/browse/MER-1671)
### NOTES
I'm not sure what's the best way to handle the update of application pages once a resource gets deleted (if a resource gets deleted and it's related to an application page, the `relates_to` of that page should be updated). Maybe we can ignore it and when we get the related resources of an application page, we can filter deleted resources.

### DIFF
--- a/lib/oli/resources.ex
+++ b/lib/oli/resources.ex
@@ -315,7 +315,9 @@ defmodule Oli.Resources do
           legacy: previous_revision.legacy |> convert_legacy,
           tags: previous_revision.tags,
           explanation_strategy: previous_revision.explanation_strategy,
-          collab_space_config: previous_revision.collab_space_config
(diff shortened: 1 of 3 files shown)


mcass19 (approved):
Wdyt about adding a link to the ticket in the description? Just for ease.

Related to what you mention about the deletion, I think it's worth do some logic now and not leave dead references but let's see what Darren thinks. We can also follow other approach such as block the deletion for the pages if some exploration is referencing it.

darrensiegel (approved):

darrensiegel:
I think we can handle referential integrity questions around deletion of linked pages at the UI/application level

darrensiegel (approved):

PR #3247, opened 2023-02-03
[FEATURE] [MER-1673] Infrastructure for exploration and relates_to resource access

simonchoxx:
[MER-1673 ](https://eliterate.atlassian.net/browse/MER-1673)

This PR adds two new functions to the delivery and authoring resolvers. 

- `get_by_purpose/2` - Returns the current versions of all page resources whose purpose type matches.

- `targeted_via_related_to/2` - Returns a list of Revisions of all pages that target the given resource id in their "relates_to" attribute.

### DIFF
--- a/lib/oli/publishing/authoring_resolver.ex
+++ b/lib/oli/publishing/authoring_resolver.ex
@@ -9,7 +9,7 @@ defmodule Oli.Publishing.AuthoringResolver do
   alias Oli.Resources.Revision
   alias Oli.Publishing.Publications.Publication
   alias Oli.Publishing.PublishedResource
-  alias Oli.Authoring.Course.Project
+  alias Oli.Authoring.Course.{Project, ProjectResource}
   alias Oli.Delivery.Hierarchy.HierarchyNode
   alias Oli.Resources.Numbering
   alias Oli.Authoring.Course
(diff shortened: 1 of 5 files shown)


lib/oli/publishing/authoring_resolver.ex
```diff
@@ -285,4 +285,65 @@ defmodule Oli.Publishing.AuthoringResolver do
...
+      iex> get_by_purpose(invalid_project_slug, invalid_purpose)
+      []
+  """
+
+  def get_by_purpose(project_slug, purpose) do
+    Repo.all(
+      from(
+        revision in Revision,
+        left_join: revision2 in Revision,
+        on: revision2.resource_id == revision.resource_id and revision.id < revision2.id,
```
darrensiegel, line 304: This query must instead leverage the "working publication" for the project to ensure it is picking up the most recent revision for this project, and not just the "most recent revision". See the fragment in this file `project_working_publication(project_slug)` and how it is used in other resolver functions. 

This will fail when a project is duplicated from an existing project.  Consider project A that contains page resource B.  If I create a duplicate project C, and in the context of C begin editing page B - there will be later revisions of page B that do not correspond to project A. 

We also should not be hardcoding the resource type id of `1` here.  Instead, use a call to `Oli.Resouurces.ResourceType.get_id_by_type("page")`

lib/oli/publishing/authoring_resolver.ex
```diff
@@ -285,4 +285,65 @@ defmodule Oli.Publishing.AuthoringResolver do
...
+      iex> targeted_via_related_to(invalid_project_slug, invalid_resource_id)
+      []
+  """
+
+  def targeted_via_related_to(project_slug, resource_id) do
+    Repo.all(
+      from(
+        revision in Revision,
+        left_join: revision2 in Revision,
+        on: revision2.resource_id == revision.resource_id and revision.id < revision2.id,
```
darrensiegel, line 335: Same comment above, we need to return the revision that is pinned to the project's working publication.

lib/oli/publishing/delivery_resolver.ex
```diff
@@ -279,4 +279,63 @@ defmodule Oli.Publishing.DeliveryResolver do
...
+  """
+
+  def get_by_purpose(section_slug, purpose) do
+    Repo.all(
+      from(
+        revision in Revision,
+        join: sect_res in SectionResource,
+        on: sect_res.resource_id == revision.resource_id,
+        join: sec in Section,
+        on: sec.id == sect_res.section_id,
```
darrensiegel, line 300: This implementation isn't correct. It is going to return *all* revisions for each resource in a course.  What we need instead is it to return the exact published revision for each resource in a course. 

Take a look at how the other functions in this resolver leverage `section_resource_revisions` fragment builder to query through the `SectionsProjectsPublications` and `PublishedResource` schemas.  You can almost certainly delegate to that function in both of these two new resolver functions.

Again, we don't want to hardcode the resource type.

darrensiegel (changes requested):

lib/oli/publishing/authoring_resolver.ex
simonchoxx: Good! I understand. I'll fix it

lib/oli/publishing/delivery_resolver.ex
simonchoxx: Excellent. I will pay attention to what you tell me!

augusto1024 (approved):
LGTM 🚀

darrensiegel (approved):

PR #3249, opened 2023-02-06
Wrong Trap State | Refactored New Attempt Creation Logic

dtiwarATS:
We should not create a new attempt if the check result contains a wrong trap state and has a navigation to another screen.

### DIFF
--- a/assets/src/apps/delivery/store/features/adaptivity/actions/triggerCheck.ts
+++ b/assets/src/apps/delivery/store/features/adaptivity/actions/triggerCheck.ts
@@ -328,8 +328,28 @@ export const triggerCheck = createAsyncThunk(
     attempt.dateEvaluated = Date.now();
 
     await dispatch(upsertActivityAttemptState({ attempt }));
+    let doesCheckResultContainsNavigationToDifferentScreen = false;
+    const actionsByType = processResults(checkResult);
+    const hasFeedback = actionsByType.feedback.length > 0;
+    const hasNavigation = actionsByType.navigation.length > 0;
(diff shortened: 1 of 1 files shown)


assets/src/apps/delivery/store/features/adaptivity/actions/triggerCheck.ts
```diff
@@ -328,8 +328,28 @@ export const triggerCheck = createAsyncThunk(
     attempt.dateEvaluated = Date.now();
 
     await dispatch(upsertActivityAttemptState({ attempt }));
+    let doesCheckResultContainsNavigationToDifferentScreen = false;
```
dtiwarATS, line 331: Just moved this logic up here so that same code can be used for determining if check result has a wrong trap state with navigation to a different screen.

assets/src/apps/delivery/store/features/adaptivity/actions/triggerCheck.ts
```diff
@@ -382,49 +402,42 @@ export const triggerCheck = createAsyncThunk(
...
-                const isVarVariable = key.startsWith('variables.');
-                const isEverAppVariable = key.startsWith('app.');
-                if (isSessionVariable || isVarVariable || isEverAppVariable) {
-                  acc[key] = latestSnapshot[key];
-                }
-                return acc;
-              },
-              {},
+        switch (navTarget) {
+          case 'next':
```
dtiwarATS, line 409: These are just prettier things. I haven't changed anything.

bsparks (approved):

PR #3259, opened 2023-02-10
Feature/storybook

marc-hughes:
Adds storybook support to the client side code.

You can run storybook locally with yarn storybook from within the assets folder. When you do, most changes will live-reload as you edit components. However, if you use a tailwind class that hadn't been used previously, you will have to reload the entire page to see the changes.

Check out the .mdx files for how to write documentation.

On a `master` build it gets deployed to https://simon-initiative.github.io/oli-torus/storybook-static/

Note: Pick that docs tab at the top, way more useful than the default canvas.

image


.github/workflows/documentation.yml
```diff
@@ -0,0 +1,58 @@
+name: Documentation
```
marc-hughes, line 1: I separated out the docs task from the package workflow so I didn't accidently run the real deploy while testing. It seemed nice to keep it separate.

assets/babel.config.js
```diff
@@ -4,5 +4,5 @@ module.exports = {
     ['@babel/preset-react', { runtime: 'automatic' }],
     '@babel/preset-typescript',
   ],
-  plugins: ['@babel/plugin-proposal-class-properties'],
```
marc-hughes, line 7: '@babel/plugin-proposal-class-properties' should be included in the '@babel/preset-env' preset and was causing storybook builds to fail. I'm pretty sure removing it is safe and haven't seen any negative effects.

darrensiegel (approved):

PR #3270, opened 2023-02-16
[ENHANCEMENT] Chemistry style fixes and modal refactor

eliknebel:
This PR addresses several delivery style issues for the chemistry course.

This PR also addresses some long standing modal issues. First, only a single modal is stored in redux to prevent a "stack" modal affect from occurring and also prevent in the future any ux anti-patterns of showing a modal within a modal. Second, the redux store was being instantiated in multiple places, possibly more than once for a given session. This PR refactors all components that utilize redux to choose which store to use, either the globalStore or adaptive authoring store. This approach is also flexible in that more stores can be created/configured in the future if necessary. Finally, the lockScroll function was not providing any benefit and causing issues in the modal component so it has also been removed from the client modal component.


assets/src/apps/app.tsx
```diff
@@ -11,13 +10,11 @@ import { Maybe } from 'tsmonad';
  * @param name A good debug-name to see in the react-dev-tools
  * @returns The wrapped component
  */
-const wrapWithRedux = (Component: FC, name: string): FC => {
-  const store = configureStore();
```
eliknebel, line 15: Instead of creating a new store for every component that is wrapped, the store is explicitly passed in

assets/src/apps/bibliography/Bibliography.tsx
```diff
@@ -326,39 +324,39 @@ const Bibliography: React.FC<BibliographyProps> = (props: BibliographyProps) =>
...
+      </ErrorBoundary>
+    </React.StrictMode>
   );
 };
 
 function defaultPaging() {
   return { offset: 0, limit: PAGE_SIZE };
 }
 
-const BibliographyApp: React.FC<BibliographyProps> = (props) => (
```
eliknebel, line 358: Some apps were being unnecessarily wrapped twice in redux provider. The call to registerApplication wraps these components with a specified redux store so this is unnecessary

darrensiegel (approved):

PR #3272, opened 2023-02-17
[FEATURE] [MER-1696] Student oriented summary view

simonchoxx:
[MER-1696](https://eliterate.atlassian.net/browse/MER-1696)

This PR aims to shows a "Discussion" link in the left navigation bar, which when clicked takes the student to a view showing two groups of publications.  

- Summary of all their publications in all collaboration spaces.

- Summary of other students most recent publications in all collaboration spaces. 

We only need to think about some kind of pagination in case we want to show more than x amount of posts. Do you think it could be possible to do it in another ticket maybe @darrensiegel?

Captura de pantalla 2023-02-17 a la(s) 10 57 10

Captura de pantalla 2023-02-17 a la(s) 10 58 07

[MER-1696]: https://eliterate.atlassian.net/browse/MER-1696?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli/resources/collaboration.ex
+++ b/lib/oli/resources/collaboration.ex
@@ -415,9 +415,8 @@ defmodule Oli.Resources.Collaboration do
         join: user in User,
         on: post.user_id == user.id,
         where:
-          post.section_id == ^section_id and
-            (post.status in [:approved, :archived] or
-               (post.status == :submitted and post.user_id == ^user_id)) and is_nil(rev2) and
+          post.section_id == ^section_id and post.user_id != ^user_id and
+            post.status in [:approved, :archived] and is_nil(rev2) and
             rev.deleted == false,
         select: %{
           id: post.id,
--- a/lib/oli/resources/collaboration.ex
+++ b/lib/oli/resources/collaboration.ex
@@ -357,22 +357,21 @@ defmodule Oli.Resources.Collaboration do
       from(
         post in Post,
         join: sr in SectionResource,
-        on: sr.resource_id == post.resource_id and sr.section_id == post.section_id,
+        on:
+          sr.resource_id == post.resource_id and sr.section_id == post.section_id and
+            sr.numbering_level > 0,
         join: spp in SectionsProjectsPublications,
         on: spp.section_id == post.section_id and spp.project_id == sr.project_id,
         join: pr in PublishedResource,
         on: pr.publication_id == spp.publication_id and pr.resource_id == post.resource_id,
         join: rev in Revision,
         on: rev.id == pr.revision_id,
-        left_join: rev2 in Revision,
-        on: rev.resource_id == rev2.resource_id and rev.id < rev2.id,
         join: user in User,
         on: post.user_id == user.id,
         where:
           post.section_id == ^section_id and post.user_id == ^user_id and
             (post.status in [:approved, :archived] or
-               (post.status == :submitted and post.user_id == ^user_id)) and is_nil(rev2) and
-            rev.deleted == false,
+               (post.status == :submitted and post.user_id == ^user_id)),
         select: %{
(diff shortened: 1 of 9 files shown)


lib/oli/resources/collaboration.ex
```diff
@@ -339,6 +340,97 @@ defmodule Oli.Resources.Collaboration do
...
+        on: rev.id == pr.revision_id,
+        left_join: rev2 in Revision,
+        on: rev.resource_id == rev2.resource_id and rev.id < rev2.id,
+        join: user in User,
+        on: post.user_id == user.id,
+        where:
+          post.section_id == ^section_id and post.user_id == ^user_id and
+            (post.status in [:approved, :archived] or
+               (post.status == :submitted and post.user_id == ^user_id)) and is_nil(rev2) and
+            rev.deleted == false,
```
darrensiegel, line 375: This is good thinking, to exclude posts from pages that were once part of the course but are now deleted, but it isn't necessary to do this.  When a page is deleted (by an author), and that new publication is applied to a course section, the section resource record that had corresponded to that page gets deleted from the course section.  So I think you can simplify this query by locating posts through only a join of SectionResource records.

lib/oli/resources/collaboration.ex
```diff
@@ -339,6 +340,97 @@ defmodule Oli.Resources.Collaboration do
...
+        join: rev in Revision,
+        on: rev.id == pr.revision_id,
+        left_join: rev2 in Revision,
+        on: rev.resource_id == rev2.resource_id and rev.id < rev2.id,
+        join: user in User,
+        on: post.user_id == user.id,
+        where:
+          post.section_id == ^section_id and post.user_id != ^user_id and
+            post.status in [:approved, :archived] and is_nil(rev2) and
+            rev.deleted == false,
```
darrensiegel, line 420: Same comment as above, this can be simplified to only join through SectionResource.

lib/oli_web/components/delivery/discussion_board.ex
```diff
@@ -0,0 +1,41 @@
+defmodule OliWeb.Components.Delivery.DiscussionBoard do
+  use Phoenix.LiveView
+
+  alias OliWeb.Components.Delivery.DiscussionPost
+  alias Oli.Resources.Collaboration
+
+  @posts_limit 5
```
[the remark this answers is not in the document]
darrensiegel: I think that this is perfect for now.  We show the student at most, their five most recent posts, and at most the five most recent posts by their classmates.  No need to make this a paged table and allow arbitrarily deep access.

darrensiegel (changes requested):

darrensiegel:
Also, we should remove the two buttons "Create Study Group" and "Add Discussion".  These represent features that a student cannot actually perform.

lib/oli/resources/collaboration.ex
```diff
@@ -357,30 +357,29 @@ defmodule Oli.Resources.Collaboration do
       from(
         post in Post,
         join: sr in SectionResource,
-        on: sr.resource_id == post.resource_id and sr.section_id == post.section_id,
+        on:
+          sr.resource_id == post.resource_id and sr.section_id == post.section_id and
+            sr.numbering_level > 0,
```
augusto1024, line 362: NIT: Once we allow posts at the course level, we might want to change this to be `sr.numbering_level == 0`. This is just something to keep in mind. We don't need to change it now 😁

augusto1024 (approved):
LGTM 🚀

darrensiegel (approved):

PR #3245, opened 2023-02-01
[FEATURE] Soft scheduling infra [DO NOT MERGE]

darrensiegel:
Adds infrastructure for accessing and updating section resources that can be scheduled.  Adds a new context `Oli.Delivery.Sections.Scheduling` and an endpoint controller for giving access to client-side code.

PR #3274, opened 2023-02-17
[FEATURE] Soft scheduling [MER-1699]

marc-hughes:
Adds a new ui for the soft scheduling of a section, plus the simplest use-case of hard scheduling where there's a single due date.

This PR includes the backend scheduling API, so we should be able to close https://github.com/Simon-Initiative/oli-torus/pull/3245

Access it from the "Schedule" link on the section overview page.

image

## Docs

There is some additional component documentation in the storybook, plus you can now see light & dark modes side by side for quicker development.

image


darrensiegel (approved):
I adjusted a couple of things: the router needed to allow pure admin usage.  Global store had to be set properly after latest chagnes on master

darrensiegel (approved):

PR #3282, opened 2023-02-23
[ENHANCEMENT] Remove tw-elements and replace with bootstrap 5 js

eliknebel:
This PR removes tailwind elements library in favor of a pure bootstrap js + tailwind styles approach. In making these changes, I noticed several issues with the adaptive author which required wrapping it in a shadow DOM and restoring the bootstrap 4 styles. I struggled to get bootstrap 4 to work properly within the shadow DOM so instead I refactored the places that utilize bootstrap to instead use react-bootstrap components.


augusto1024:
If we're just going to use bootstrap for the js, it might be better to remove the library which is kind of heavy and just load the lightweight js file from the CDN. Then, for the shadow DOM thing, we can load the CSS from bootstrap 4 just like you did on the authoring container. WDYT?

darrensiegel (approved):

PR #3287, opened 2023-02-24
Feature/mer 1693 change collab spaces design

augusto1024:
## STORY
[MER-1693](https://eliterate.atlassian.net/jira/software/c/projects/MER/boards/2?selectedIssue=MER-1693)
## TODO
- [ ] Fix tests
## NOTES
There were some inconsistencies in the designs (IMO). It's weird that you need a modal to create new posts but not for replying to existing ones, so I decided to implement a different way to create posts. Also, repeating the whole comment just to show it's a reply it's not really good for long posts, so I did what we had in the first place but with a little redesign. Finally, there were no designs for the sorting and active user features and I thought they were good features that shouldn't be removed.

Here are some screenshots of how it looks
Screen Shot 2023-02-24 at 16 46 04
Screen Shot 2023-02-24 at 16 51 55
Screen Shot 2023-02-24 at 16 52 42
Screen Shot 2023-02-24 at 16 53 37
Screen Shot 2023-02-24 at 16 53 54


lib/oli_web/live/collaboration_live/collab_space_view.ex
```diff
@@ -266,18 +249,16 @@ defmodule OliWeb.CollaborationLive.CollabSpaceView do
...
-      on_submit: "edit_post",
-      changeset: changeset
-    }
+    {is_edition_mode, post} =
+      case editing_post && Integer.to_string(editing_post.id) == post_id do
+        true -> {false, nil}
+        _ -> {true, Collaboration.get_post_by(%{id: post_id})}
+      end
 
-    display_post_modal(modal_assigns, assign(socket, editing_post: post))
```
nicocirio, line 280: If we are not going to use the modal, then we can remove the `OliWeb.CollaborationLive.Posts.Modal` alias

lib/oli_web/live/collaboration_live/collab_space_view.ex
```diff
@@ -266,18 +249,16 @@ defmodule OliWeb.CollaborationLive.CollabSpaceView do
...
+    {:noreply, assign(socket, editing_post: post, is_edition_mode: is_edition_mode)}
```
nicocirio, line 261: I noticed than when we enter the "edit" mode, we do not have focus set to the corresponding textArea. To do so, we can push an event:
```suggestion
    socket = assign(socket, editing_post: post, is_edition_mode: is_edition_mode)

    {:noreply, push_event(socket, "set_focus", %{id: "post_text_area_#{post_id}"})}
```

```
// textarea_listener.ts

export const TextareaListener = {
  mounted(): void {
   ...
   window.addEventListener(`phx:set_focus`, (e: any) => {
      const el = document.getElementById(e.detail.id);
      if (el) {
        el.focus();
      }
    });
  },
...
};
```

```
// show.ex

<TextArea id={"post_text_area_#{@post.id}"}
              opts={
                placeholder: "Reply",
                "data-grow": "true",
                "data-initial-height": 44,
                onkeyup: "resizeTextArea(this)"
              }
              class="torus-input border-r-0 collab-space__reply"
            />

```

WDYT?

nicocirio (commented):
LGTM!! 😄

nicocirio (commented):
LGTM, great work! 😄 
I left some small comments that may improve user's experience while editing or sorting posts

lib/oli_web/live/collaboration_live/collab_space_view.ex
augusto1024: Thanks for this ❤️

lib/oli_web/live/collaboration_live/collab_space_view.ex
```diff
@@ -120,53 +130,83 @@ defmodule OliWeb.CollaborationLive.CollabSpaceView do
 
   def render(assigns) do
     ~F"""
-      {#if show_collab_space?(@collab_space_config)}
-        {render_modal(assigns)}
-
-        <div class="card">
-          <div class="card-body d-flex align-items-center">
-            <h3 class="card-title">Collaborative Space</h3>
+    {render_modal(assigns)}
```
simonchoxx, line 133: Just to understand... If the collab space is disabled, will this be displayed anyway or that it will be displayed?  I think we should contemplate the possibility that the collab space is disabled, WDYT?

simonchoxx (approved):
Awesome work mate! 👏👏

lib/oli_web/live/collaboration_live/collab_space_view.ex
augusto1024: We already discussed this online but it's going to be done [here](https://github.com/Simon-Initiative/oli-torus/pull/3287/files#diff-ee74bc1fa45f0e681c16090db0f9135eb587164e4eea0fc999c97ae88fa5fa01R43) now 😀

lib/oli_web/templates/page_delivery/instructor_page_preview.html.heex
```diff
@@ -15,28 +15,4 @@
...
-      live_render(@conn, OliWeb.CollaborationLive.CollabSpaceConfigView,
-        session: %{
-          "collab_space_config" => @collab_space_config,
-          "section_slug" => @section_slug,
-          "page_slug" => @revision.slug,
-          "is_delivery" => true
-        })
-    %>
-  </div>
-
```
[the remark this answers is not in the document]
augusto1024: You're totally right. I thought I was removing `CollabSpaceView` 😅

darrensiegel (changes requested):
Looks great, only thing I saw was that the "No Posts Yet" message still displays even after I created a post.  See this screenshot:

Screenshot 2023-03-01 at 2 03 41 PM

darrensiegel (approved):

PR #3298, opened 2023-03-03
[Feature] Flowchart feature, stage 1

marc-hughes:
There are 3 categories of work here.

1. Modified the base advanced authoring app to have multiple modes to support the flowchart feature
2. Added in, and cleaned up some types
3. Feature-work on the flowchart view
 
This is the first PR of many for the flowchart-feature. The main goal of this PR is to get #1 and #2 merged in to keep things moving smoothly with other work that may be going on.

## Base app changes

The main advanced-authoring entry point Authoring.tsx has been refactored into a few smaller components with the aim of re-using some in the flowchart mode, and replacing some with logic for switching between them.

## Additional / cleaned up Types

The `IActivity` type has been expanded, and a bunch of places that were returning these have been changed from `any` to `IActivity` return types.

While doing that, I found both an `AdaptiveRule` and an `IRule` type that described the same thing in slightly different ways. Those have been condensed into a single type.

There is a new yarn command `yarn check-types` to do a full scan of all types across the app.  I fixed all errors that came up there (hard to say how many, as I would fix one and a couple more would pop up, but the initial count was around 60).

We had multiple "RootState" types for authoring & delivery, but they were sometimes used interchangeably and cast to `any` which became confusing. They've been renamed to DeliveryRootState and AuthoringRootState

## To use the flowchart feature:

Currently, there is no UI to enter the new flowchart mode.  It feature relies on a new custom.contentMode attribute set within the lesson json to enter, a minimal lesson would be:

```json
{
  "model": [],
  "advancedDelivery": true,
  "advancedAuthoring": true,
  "displayApplicationChrome": false,
  "custom": {
    "contentMode": "flowchart"
  }
}
```

## Flowchart feature progress

Once that is set up, and you attempt to edit the page, you get a preliminary flowchart editing interface. Demo:

https://user-images.githubusercontent.com/333265/222732041-773dd368-9a5b-49c4-972b-77a7e453532b.mp4


assets/.storybook/main.js
```diff
@@ -7,6 +7,7 @@ module.exports = {
     '@storybook/addon-links',
     '@storybook/addon-essentials',
     '@storybook/addon-interactions',
+    '@storybook/preset-scss',
```
marc-hughes, line 10: I was trying to get our scss modules working in storybook, but didn't have a lot of luck there. I want to re-visit this eventually

assets/package.json
```diff
@@ -11,14 +11,15 @@
     "watch": "webpack --mode development --watch",
     "build": "npm run clean && npm run lint && npm run test && npm run deploy",
     "clean": "del --force coverage/** ../priv/static",
-    "lint": "npx eslint src test",
+    "lint": "eslint src test",
```
marc-hughes, line 14: prettier & eslint are installed locally, not sure why we were running via npx. This should make it quicker

assets/src/apps/authoring/AuthoringExpertPageEditor.tsx
```diff
@@ -0,0 +1,61 @@
+import React, { useRef } from 'react';
```
marc-hughes, line 1: This is the refactored editor for the previous advanced authoring environment.

darrensiegel (approved):

PR #3301, opened 2023-03-07
[FEATURE] [MER-1774] Section does not show access to Explorations

simonchoxx:
[MER-1774](https://eliterate.atlassian.net/browse/MER-1774)

This PR adds functionality to not show access to "Explorations" when the section does not have any resource of this type. 
When this happens, both the access from the left sidebar and the Windowshade are hidden. 

Also, thanks to the contribution of @nicocirio, this PR includes the fix for the error that happens when the last resource is removed from the content of a section.

[MER-1774]: https://eliterate.atlassian.net/browse/MER-1774?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test/oli/delivery_test.exs
+++ b/test/oli/delivery_test.exs
@@ -89,5 +89,24 @@ defmodule Oli.DeliveryTest do
       assert delivery_setting.id == updated_delivery_setting.id
       assert updated_delivery_setting.collab_space_config.status == :archived
     end
+
+    test "maybe_update_section_contains_explorations/1 update contains_explorations field" do
+      {:ok,
+       project: _project,
+       section: section,
+       page_revision: _page_revision,
+       other_revision: other_revision} = project_section_revisions(%{})
+
+      author = insert(:author)
+
+      assert section.contains_explorations
+
+      Oli.Resources.update_revision(other_revision, %{purpose: :foundation, author_id: author.id})
+
+      Delivery.maybe_update_section_contains_explorations(section)
+      section_without_explorations = Oli.Delivery.Sections.get_section_by_slug(section.slug)
+
+      refute section_without_explorations.contains_explorations
+    end
   end
 end
--- a/test/oli_web/controllers/page_delivery_controller_test.exs
+++ b/test/oli_web/controllers/page_delivery_controller_test.exs
@@ -1549,6 +1549,40 @@ defmodule OliWeb.PageDeliveryControllerTest do
 
       assert html_response(conn, 200) =~ "<h6>There are no exploration pages available</h6>"
     end
+
+    test "do not show the 'exploration' access in the left navbar when the section has no explorations to show",
+         %{conn: conn} do
+      {:ok,
+       section: section, unit_one_revision: _unit_one_revision, page_revision: _page_revision} =
+        section_with_assessment(%{})
+
+      user = insert(:user)
+      Sections.enroll(user.id, section.id, [ContextRoles.get_role(:context_learner)])
+
+      conn =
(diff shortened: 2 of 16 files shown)


darrensiegel (changes requested):
I can see where it is updating `contains_explorations` in the Section after a Remix, but I do not see where it sets `contains_explorations` at the time of initial course section creation.  

There are couple of different places where course sections get created: From within the open_and_free_controller.ex and within `delivery.ex` (`create_from_product` and `create_from_publication`). 

Creating from a product can simply copy the value of `contains_explorations`, since that value will exist for the product, and will getting updated after a Remix in the product.

test/oli/delivery_test.exs
```diff
@@ -89,5 +89,24 @@ defmodule Oli.DeliveryTest do
       assert delivery_setting.id == updated_delivery_setting.id
       assert updated_delivery_setting.collab_space_config.status == :archived
     end
+
+    test "maybe_update_section_contains_explorations/1 update contains_explorations field" do
+      {:ok,
+       project: _project,
+       section: section,
+       page_revision: _page_revision,
+       other_revision: other_revision} = project_section_revisions(%{})
```
augusto1024, line 98: NIT: If we're not using `project` and `page_revision` we can remove them from the map

test/oli_web/controllers/page_delivery_controller_test.exs
```diff
@@ -1507,6 +1507,40 @@ defmodule OliWeb.PageDeliveryControllerTest do
 
       assert html_response(conn, 200) =~ "<h6>There are no exploration pages available</h6>"
     end
+
+    test "do not show the 'exploration' access in the left navbar when the section has no explorations to show",
+         %{conn: conn} do
+      {:ok,
+       section: section, unit_one_revision: _unit_one_revision, page_revision: _page_revision} =
```
augusto1024, line 1514: Same as above

test/oli_web/controllers/page_delivery_controller_test.exs
```diff
@@ -1507,6 +1507,40 @@ defmodule OliWeb.PageDeliveryControllerTest do
...
+        |> Pow.Plug.assign_current_user(user, OliWeb.Pow.PowHelpers.get_pow_config(:user))
+        |> get(Routes.page_delivery_path(conn, :index, section.slug))
+
+      refute html_response(conn, 200) =~ "<a>Exploration</a>"
+    end
+
+    test "do not show the 'exploration' access in the Windowshade when the section does not have explorations to show",
+         %{conn: conn} do
+      {:ok,
+       section: section, unit_one_revision: _unit_one_revision, page_revision: _page_revision} =
```
augusto1024, line 1531: Same as above

augusto1024 (commented):
Great work! I agree with what Darren said, and also added some small NITs

test/oli/delivery_test.exs
simonchoxx: Here, I think that should match the full map anyway because otherwise it doesn't match what the helper function I'm using returns.

nicocirio:
Great Work! LGTM 🚀

darrensiegel (changes requested):
There is one additional place to handle "course section creation".  See `open_and_free_controller.ex`, function `create_from_publication` (around line 355)

darrensiegel (approved):

PR #3302, opened 2023-03-07
[BUG FIX] rewire activity links

darrensiegel:
Adds a step to rewire links within activities.


andersweinstein:
In my testing: although ingested activity links on a page function in preview (clicking them opens the referenced page), activity links in question stems do not function in preview.  The latter go through translation in html.tsx (line 405) which converts them to no-op urls in case of no sectionSlug. Links in a workbook page evidently do not go through this code path.

In Delivery, the ingested link in a question stem has an href of a non-existent resource, so does not work. In my testing, the href picked up the id of the course package, rather than the resource in question.

andersweinstein (approved):

PR #3318, opened 2023-03-13
[BUGFIX] Avoid error msg on clicking empty multi-input [MER-1785]

andersweinstein:
Clicking empty multi-input in preview/delivery was causing error message to javascript console, though the error was harmless.  This adds undefined check to reduce noise in the console window.


darrensiegel (changes requested):
Looks good, just remove the `node_modules/.yarn-integrity` file from the PR

darrensiegel (approved):

PR #3320, opened 2023-03-13
[ENHANCEMENT] do not show page read by or time estimate when not provided

eliknebel:
This PR removes the page read by and time estimate by hiding those elements when they are not provided

### DIFF
--- a/lib/oli_web/components/delivery/page_delivery.ex
+++ b/lib/oli_web/components/delivery/page_delivery.ex
@@ -20,15 +20,16 @@ defmodule OliWeb.Components.Delivery.PageDelivery do
     """
   end
 
-  attr :read_by, Date, default: nil
+  attr :scheduling_type, :atom, values: [:read_by, :inclass_activity]
+  attr :end_date, Date, default: nil
   attr :est_reading_time, Timex.Duration, default: nil
 
   def details(assigns) do
     ~H"""
       <div class="flex flex-row my-2">
-        <%= if @read_by do %>
+        <%= if @end_date do %>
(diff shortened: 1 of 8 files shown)


lib/oli_web/components/delivery/page_delivery.ex
```diff
@@ -18,15 +20,22 @@ defmodule OliWeb.Components.Delivery.PageDelivery do
...
       <div class="flex flex-row my-2">
-        <div class="py-1.5 px-4 bg-gray-100 text-gray-700 rounded">
-          Read by 10-03-2022
-        </div>
-        <div class="py-1.5 px-4 bg-gray-100 text-gray-700 rounded ml-1">
-          Estimated reading time: 5 mins
-        </div>
+        <%= if @read_by do %>
+          <div class="py-1.5 px-4 bg-gray-100 text-gray-700 rounded">
+            Read by <%= format_date(@read_by) %>
```
darrensiegel, line 31: The soft scheduled `end_date`, which is what I believe is populating the `read_by` field here, isn't always "Read By".  On the `SectionResource` record, there is a `scheduling_type` attribute which can be either `:read_by` or `:inclass_activity`, which should be labeled "Ready by" or "In class activity"

darrensiegel (changes requested):

lib/oli_web/components/delivery/page_delivery.ex
eliknebel: Addressed, added case check for scheduling type and connected up the section's soft schedule

darrensiegel (approved):

PR #3329, opened 2023-03-15
[BUGFIX] fix bootstrap JS issues including modals, dropdowns and tooltips

eliknebel:
A lot of existing interfaces still rely on bootstrap javascript (in combination with tailwind styles) to function. There was 2 different bootstrap versions being loaded (v4 and v5) and multiple calls to initialize components were causing them to not work at all.

This PR addresses these issues by consolidating on a single version of bootstrap JavaScript (v5) and importing/initializing those components in the root `assets/src/phoenix/app.ts` file instead of loading them separately in layout templates.

Where possible, `react-bootstrap` is used in React components instead of directly accessing from the window. For all other server rendered views, the old `data-toggle` attributes have been upgraded to `data-bs-toggle` which fixes the version mismatch issues.


assets/src/apps/bank/EditButton.tsx
```diff
@@ -13,10 +14,6 @@ export const EditButton = (props: EditButtonProps) => {
       onClick={() => onChangeEditMode(!editMode)}
       type="button"
       className="edit-btn btn btn-sm btn-success mr-1"
-      aria-pressed="false"
-      data-toggle="tooltip"
-      data-placement="top"
-      title="Finish editing this activity"
```
eliknebel, line 19: These tooltips were more distracting that helpful, so they were just removed

darrensiegel (approved):

PR #3332, opened 2023-03-15
[BUG FIX] handle list (array) values in json [MER-1784]

darrensiegel:
Handles a case where variables are lists by converting to a pure string representation to allow substitution to succeed.


andersweinstein (approved):
Works on reported cases.

PR #3335, opened 2023-03-16
[MER-1735] Assignments view

augusto1024:
[Link to the story](https://eliterate.atlassian.net/browse/MER-1735)

image


darrensiegel (changes requested):
See inline comments

darrensiegel (approved):

PR #3361, opened 2023-03-24
[MER-1712] Instructor portal students view

nicocirio:
[Link to ticket](https://eliterate.atlassian.net/browse/MER-1712)

## Notes
The ticket says:

> Another key change that we need to the existing “Manage enrolled students” view is what view the user is routed to when they click on a student’s name.  Currently, instructors are routed to the “Student progress view”, which is a listing of all course pages and the progress for that student. Instead, we need to route the user to the “Student details view” for that student. 

The Students Details view has not yet been developed, so I left the link unchanged. 
When that view gets developed we will be able to update the link

### DIFF
--- a/lib/oli/delivery/metrics.ex
+++ b/lib/oli/delivery/metrics.ex
@@ -1,5 +1,4 @@
 defmodule Oli.Delivery.Metrics do
-
   import Ecto.Query, warn: false
 
   alias Oli.Repo
@@ -10,8 +9,8 @@ defmodule Oli.Delivery.Metrics do
   alias Oli.Delivery.Attempts.Core
 
   @doc """
-  Calculate the progress for a specific student, in all pages of a
-  specific container.
+  Calculate the progress for a specific student (or a list of students),
+  in all pages of a specific container.
 
   Ommitting the container_id (or specifying nil) calculates progress
   across the entire course section.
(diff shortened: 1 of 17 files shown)


lib/oli/delivery/metrics.ex
```diff
@@ -32,20 +44,28 @@ defmodule Oli.Delivery.Metrics do
...
+      |> where([cp], cp.section_id == ^section_id)
+      |> where(^filter_by_container)
+      |> select([cp], count(cp.id))
+      |> Repo.one()
+      |> max(1)
+
     query =
       ContainedPage
-      |> join(:left, [cp], ra in ResourceAccess, on: cp.page_id == ra.resource_id and cp.section_id == ra.section_id and ra.user_id == ^user_id)
+      |> join(:left, [cp], ra in ResourceAccess,
```
augusto1024, line 57: Since you're using a `left join` here, if you have a user that doesn't have a resource access for a given contained page, it will still take that contained page into account which will add a nil entry in the returned map (this just happened to me `%{2 => 0.75, nil => nil}`. Changing it to an `inner join` should do the trick 😁

lib/oli/delivery/metrics.ex
```diff
@@ -55,7 +75,11 @@ defmodule Oli.Delivery.Metrics do
   def progress_across(section_id, container_ids, user_id) do
     query =
       ContainedPage
-      |> join(:left, [cp], ra in ResourceAccess, on: cp.page_id == ra.resource_id and cp.section_id == ra.section_id and ra.user_id == ^user_id)
+      |> join(:left, [cp], ra in ResourceAccess,
+        on:
+          cp.page_id == ra.resource_id and cp.section_id == ra.section_id and
+            ra.user_id == ^user_id
+      )
```
augusto1024, line 82: Same as above

augusto1024 (commented):
Looks good so far but I wasn't able to review the whole thing yet. Left one comment regarding the function that fetches the progress

lib/oli/delivery/metrics.ex
nicocirio: Makes sense, good catch, thanks!

lib/oli/delivery/metrics.ex
nicocirio: In this case left join is needed because we do want to get the nil entry in order to calculate the total count over the ContainedPage table

darrensiegel (approved):

PR #3363, opened 2023-03-24
[Feature][MER-1732] Support enabling required survey for courses

augusto1024:
Uploading Screen Recording 2023-03-31 at 16.25.13.mov…

### NOTES
I transformed the project overview view into a liveview because we were rendering other liveviews in those components that were fetching the project when they didn't really need to (they still are, but it shouldn't be hard to transform those liveviews into livecomponents and pass the project as a prop and we should definitely do it).

Also, every time the survey option is enabled it creates a new resource and a revision instead of checking if there was already a revision of that type marked as deleted (because maybe the author has already enabled it and disabled it). Is that ok or should we just reuse any existing resource and revision if they exist?

Once a project gets the required survey option enabled (we do this by setting the `required_survey_resource_id`), it gets enabled for the sections as well. If an instructor enables the survey for a given section, it takes the same resource id for its own `required_survey_resource_id`. 

If an admin removes the `required_survey_resource_id` from the project, it will also be removed for the sections that were based on that project (maybe we should improve this behavior a little bit).

**ALSO, AS I MENTIONED, I WILL BE OUT UNTIL APRIL 10th SO I WON'T BE ABLE TO ADDRESS ANY COMMENTS UNTIL THAT DATE (FEEL FREE TO DO ANY NECESSARY CHANGES)**


darrensiegel (changes requested):
This PR currently introduces a new resource type "survey".  There is no need to do this, as the student required survey should instead just simply be a basic Torus page.

lib/oli/authoring/course.ex
```diff
@@ -338,4 +340,95 @@ defmodule Oli.Authoring.Course do
       description: "New family from project creation"
     }
   end
+
+  def get_project_survey(project_id) do
+    ProjectResource
+    |> join(:inner, [pr], rev in Revision, on: pr.resource_id == rev.resource_id)
+    |> join(:left, [_, rev], rev2 in Revision,
```
darrensiegel, line 347: This is not the correct way to retrieve the specific Revision for a resource from the authoring perspective.  Instead, you need to `join` through the `PublishedResource` schema on the `publication_id` for the unpublished publication for this project.  Look at the functions in `Oli.Publishing.AuthoringResolver` for examples of how to do this.

lib/oli/authoring/course.ex
```diff
@@ -338,4 +340,95 @@ defmodule Oli.Authoring.Course do
...
+
+  defp update_project_required_survey_resource_id(project_id, resource_id) do
+    Project
+    |> where([p], p.id == ^project_id)
+    |> Repo.one()
+    |> Project.changeset(%{required_survey_resource_id: resource_id})
+    |> Repo.update()
+  end
+
+  defp remove_survey_for_existing_project_sections(project_id, survey_id) do
```
darrensiegel, line 371: When a course author decides to remove or change the required survey in their project, we do not want to remove or change it from course sections that are using that project.

lib/oli/delivery/sections.ex
```diff
@@ -2175,4 +2174,94 @@ defmodule Oli.Delivery.Sections do
     ] ++
       get_flatten_hierarchy(rest, resources)
   end
+
+  defp get_parent_project_survey(section_id) do
```
darrensiegel, line 2178: Retrieving the correct revision for the required survey for a course section should be simpler than this code.  Instead, we only need to do two things:

1. Check to see if `required_survey_resource_id` on the `Section` is not `nil`.  If it is `nil`, there is no survey required.
2. With the value of `required_survey_resource_id`, make a call to `Oli.Publishing.DeliveryResolver.from_resource_id`, which resolves the correct revision of the resource for that section.

lib/oli/resources/resource_type.ex
```diff
@@ -29,6 +29,7 @@ defmodule Oli.Resources.ResourceType do
   def get_type_by_id(6), do: Map.get(@by_id, 6)
   def get_type_by_id(7), do: Map.get(@by_id, 7)
   def get_type_by_id(8), do: Map.get(@by_id, 8)
+  def get_type_by_id(9), do: Map.get(@by_id, 9)
```
darrensiegel, line 32: This should be removed?

darrensiegel (changes requested):
See inline comments

lib/oli/authoring/course.ex
augusto1024: Thanks for this! I'm still wrapping my head around all the relations 😅

darrensiegel (approved):

PR #3364, opened 2023-03-24
[MER-1833] delivery renders header twice for admin and instructors

eliknebel:
This PR fixes an issue where two different headers were being rendered in delivery. It also addresses the issue that some manage view do not have the proper header or breadcrumb:
- refactor layout pipeline to render proper delivery header for all instructor views
- render main layout in delivery layout instead of separately in each LiveView
- remove session context and replace with explicit current_user and other assigns
- restore delivery breadcrumb
- fix some dark mode and missing container mx-auto
- remove unnecessary delivery_breadcrumb: true assign

### DIFF
--- a/assets/css/app.css
+++ b/assets/css/app.css
@@ -46,3 +46,7 @@ p {
 .slate-editor p:last-child {
   @apply mb-0;
 }
+
+.text-muted {
+  @apply text-gray-500;
+}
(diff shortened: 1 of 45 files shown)


darrensiegel:
One thing that i noticed in testing this locally, clicking "Enter this course as a student" - as a pure instructor - no longer seems to work. It opens a new tab, but simply routes me back to the "Manage Section" view.

darrensiegel (approved):

PR #3369, opened 2023-03-27
[ENHANCEMENT] Unify instructor dashboard live view

nicocirio:
This PR improves the overall instructor dashboard performance reducing the amount of queries needed, by having a unique liveview that orchestrates (via url params) which tab should be rendered, instead of having one liveview per tab.
This change avoids having to remount common assigns (section, for example) every time we navigate between tabs. 
Now we mount the common assigns once, and fetch the additional needed assigns for a given tab when that tab gets rendered.
Depending on the use case and the need of handling events, some tabs ended up being live_components (students and discussion activity tabs, for example) and some others just plain components (content tab, for example).


lib/oli_web/live/delivery/instructor_dashboard/instructor_dashboard_live.ex
```diff
@@ -0,0 +1,146 @@
...
+        )
+      else
+        &Routes.page_delivery_path(OliWeb.Endpoint, :container, assigns.section_slug, &1)
+      end
+
+    assigns =
+      Map.merge(
+        assigns,
+        %{
+          hierarchy: Sections.build_hierarchy(assigns.section),
```
darrensiegel, line 72: This seems suspicious: Within the rendering step we are calling a fairly expensive `build_hierarchy/1`.  I see other places in this LV where DB calls are made within the rendering code.  Unless we have strong guarantees here, this code could get executed way more than what we would like.  

And we probably do not even want this DB code to execute when tabs are switched, correct?  Instead, I think what we want is a state holder within this LV that caches the state needed for each tab (live component), and that only executes the DB code that a tab needs *one* time, the very first time that a tab is displayed.

darrensiegel (changes requested):

lib/oli_web/live/delivery/instructor_dashboard/instructor_dashboard_live.ex
darrensiegel: Let me describe a scenario and you can let me know if what you have built here supports this or not:

Imagine this LV approach paired with four tabs.  Each tab has a corresponding live_component that displays a searchable, filterable table.  Each component tab must run an expensive query to fetch all the data that it needs to display.  When a user firsts visits this LiveView, there will of course be a default tab that is displayed.   Then, after the user clicks to select another tab, the query related to that other tab is run.  When the user clicks back to the original, default tab its query is **not re-run.**  The original data from its initial execution is preserved and displayed.  

What we really need also is to support two different types of live_component tab implementations:

The first implementation type is where the DB queries required to fetch the data to display in the table is so expensive that paging, filtering and sorting must be done across the entire dataset, in memory.  An example of this would be the "content" tab.  Another requirement is that the number of total records should be reasonably small (for instance, we know that a course will have at most around 50 or 60 modules, so the "content" tab qualifies against this criteria).

The other type of live_component tab provider is one where the query to fetch data is not as expensive, and perhaps the number of records can be so large that we must do the paging, sorting and filtering at the DB query level.  

For the first implementation type, we want a strong guarantee that the query to fetch all the data is run exactly one time for the entire lifetime of the parent LiveView.  So that means the first time that a user clicks on one of these types of tab the query is run then.  But afterwards, it is never run again.  The user can sort, filter, page and all that is being done in memory.  The user can click away to another tab, then click back and the query is not re-run.  The original full data results are cached in the parent LiveView.

darrensiegel (changes requested):

darrensiegel (approved):

PR #3382, opened 2023-03-30
[BUG FIX] Fix snapshot creation when objectives are duplicated

darrensiegel:
Snapshot creation fails when there is an activity that duplicates an objective for one part.  This can only happen when an objective is ingested in this state.  

This PR does the following:
1. Makes the snapshot generation code in `worker.ex` robust to duplicates, by deduping the part specific list of objectives, prior to snapshot insertion.
2. Fixes the activity ingest processor to dedupe objective ids
3. Adds a new LiveView "Manage Snapshot Records", which is Admin accessible only, that detects when there are snapshot records missing for a specific course section and then allows these snapshot records to be created. 

To test the LV out end-to-end, the easiest thing to do is to create a course section with a page that have a couple of MCQ questions.  Then, as a guest student, access that page and answer the questions.  This will generate snapshot records.  Open PgAdmin, and run `DELETE from snapshots;` to delete those records.   Then go to the LV and verify that it detects the missing snapshot records and after clicking "Generate Snaphot records" that indeed generates them (by looking in PgAdmin)

### DIFF
--- a/lib/oli/delivery/snapshots/worker.ex
+++ b/lib/oli/delivery/snapshots/worker.ex
@@ -88,9 +88,12 @@ defmodule Oli.Delivery.Snapshots.Worker do
           # If there are no attached objectives, create one record recoring nils for the objectives
           [] -> [to_attrs(result, nil, nil, project_id)]
 
-          # Otherwise create one record for each objective
+          # Otherwise create one record for each objective, careful to dedupe in the event that
+          # somehow a part has objectives duplicated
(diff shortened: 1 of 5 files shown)


rgachuhi (approved):
I noticed that deleting the records in the snapshot db does not automatically update the LV; manually refreshing the view works fine. It is able to list the missing snapshot records. "Generate Snapshot records" works as well. The missing records are successfully written back into the snapshots table.

PR #3383, opened 2023-03-31
[FEATURE] [MER-1711] Instructor Portal - Learning Objectives Tab

simonchoxx:
[MER-1711
](https://eliterate.atlassian.net/browse/MER-1711)
This PR adds a list of the objectives contained in the curriculum pages of a course section, within the "Learning Objectives" tab.

The information about the objectives is rendered in a table that allows to be sorted, paginated and also searchable through a search field. In addition, by means of a dropdown you can filter the objectives by the unit or module that contains them.

The columns "STUDENT MASTERY OBJ." and "STUDENT MASTERY(SUB OBJ.)" of the table contain dummy data as they cannot be calculated yet.

The "STUDENT ENGAGEMENT" column is obtained randomly in order to be able to differentiate the rows containing a value equal to "Low". All this can be modified if required.

Captura de pantalla 2023-03-31 a la(s) 11 00 50

Captura de pantalla 2023-03-31 a la(s) 11 01 15

### DIFF
--- a/lib/oli/delivery/sections.ex
+++ b/lib/oli/delivery/sections.ex
@@ -2359,19 +2359,44 @@ defmodule Oli.Delivery.Sections do
   end
 
   def get_units_and_modules_from_a_section(section_slug) do
-    Repo.all(
-      from([sr: sr, rev: rev, s: s] in DeliveryResolver.section_resource_revisions(section_slug),
-        join: cp in ContainedPage,
-        on: cp.container_id == rev.resource_id,
-        where: rev.deleted == false and s.slug == ^section_slug,
-        distinct: cp.container_id,
-        order_by: [asc: rev.title],
-        select: %{
-          container_id: cp.container_id,
-          title: rev.title
-        }
+    all =
+      Repo.all(
+        from(
+          [sr: sr, rev: rev, s: s] in DeliveryResolver.section_resource_revisions(section_slug),
+          join: cp in ContainedPage,
+          on: cp.container_id == rev.resource_id,
+          where: rev.deleted == false and s.slug == ^section_slug and rev.resource_type_id == 2,
+          group_by: [cp.container_id, rev.title, sr.numbering_level],
+          order_by: [asc: rev.title],
+          select: %{
+            container_id: cp.container_id,
+            title: rev.title,
+            level: sr.numbering_level
+          }
+        )
       )
-    )
+
+    Enum.map(all, fn elem ->
+      if elem.level == 2 do
+        Map.put(
+          elem,
+          :title,
+          Repo.one(
+            from(
+              [sr: sr, rev: rev, s: s] in DeliveryResolver.section_resource_revisions(
+                section_slug
+              ),
+              where:
+                rev.deleted == false and s.slug == ^section_slug and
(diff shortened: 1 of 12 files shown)


test/support/test_helpers.ex
```diff
@@ -683,6 +683,223 @@ defmodule Oli.TestHelpers do
...
+    Sections.rebuild_contained_pages(section)
+
+    [
+      project: project,
+      section: section,
+      publication: publication,
+      obj_revision_1: obj_revision_1,
+      obj_revision_2: obj_revision_2,
+      module_revision: module_revision
+    ]
```
nicocirio, line 900: What do you think about returning a map? With a map we can pattern match against a single key-value when we invoke this function inside a test (not necessarily in the setup), since we might not need all the fields in some cases.

lib/oli_web/live/common/paged_table.ex
```diff
@@ -24,7 +24,7 @@ defmodule OliWeb.Common.PagedTable do
         {/if}
 
         {#if @total_count > 0 and @total_count > @limit}
-          <Paging id="header_paging" total_count={@total_count} offset={@offset} limit={@limit} click={@page_change}/>
+          <Paging id="header_paging" total_count={@total_count} offset={@offset} limit={@limit} click={@page_change} additional_class="bg-white"/>
```
nicocirio, line 27: Can we avoid adding this class here? I think we may not want to change the background color of every paged_table the project renders.
I think having `<div class="mx-10 mb-10 bg-white">` in [this line](https://github.com/Simon-Initiative/oli-torus/blob/52eec16fa8ef7db223ed06955faef0487f55dc6b/lib/oli_web/components/delivery/learning_objectives/learning_objectives.ex#L58) would have the same result, but not affecting other places where we render the table.

nicocirio (commented):
Great job Santi! 😄  I left some NITs

lib/oli_web/live/common/paged_table.ex
simonchoxx: Nice catch 🙌

lib/oli/delivery/sections.ex
```diff
@@ -2113,4 +2113,133 @@ defmodule Oli.Delivery.Sections do
...
+          order_by: [asc: rev.title],
+          select: %{
+            container_id: cp.container_id,
+            title: rev.title,
+            level: sr.numbering_level
+          }
+        )
+      )
+
+    Enum.map(all, fn elem ->
```
darrensiegel, line 2193: This `Enum.map` appears to be prefixing the names of modules with their parent Unit.  It does so, however, in a way that triggers one query for every module in the course.  In a small course, this is probably fine.  We have courses, however, that contain 50, 60, or even 70+ modules.  This needs to be rewritten so that exactly one query is executed to get this Unit and Module information.

simonchoxx: I understand. You are right. I will refactor as necessary so that we call the `OliWeb.Components.Delivery.LearningObjectives.update/2` function only once and so that the filtering, searching, paging and sorting operations are done in memory.

I will also modify the query to get all the units with their modules at once.

darrensiegel (approved):

PR #3386, opened 2023-04-03
[MER-1713] Instructor dashboard content view

nicocirio:
This PR addresses[ this ticket](https://eliterate.atlassian.net/browse/MER-1713) and updates the students tab to handle filtering by page_id in the url params.

### DIFF
--- a/lib/oli_web/components/delivery/content/content_table_model.ex
+++ b/lib/oli_web/components/delivery/content/content_table_model.ex
@@ -55,8 +55,7 @@ defmodule OliWeb.Components.Delivery.ContentTableModel do
       Map.merge(assigns, %{
         progress: parse_progress(container.progress),
         title: container.title,
-        url_params: url_params,
-        section_slug: assigns.model.data.section_slug
+        url_params: url_params
       })
 
     ~H"""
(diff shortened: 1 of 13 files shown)


lib/oli_web/components/delivery/content/content_table_model.ex
```diff
@@ -0,0 +1,112 @@
...
+        "PAGES" -> %{page_id: container.id}
+        _ -> %{container_id: container.id}
+      end
+
+    assigns =
+      Map.merge(assigns, %{
+        progress: parse_progress(container.progress),
+        title: container.title,
+        url_params: url_params,
+        section_slug: assigns.model.data.section_slug
```
simonchoxx, line 59: I'm not entirely sure, but I think you could call 'assigns.section_slug' directly here, since you already defined section_slug in the data attribute of the SortableTableModel definition above.

simonchoxx (commented):
Great work mate 🚀🚀 !! I just leave you a couple of minor comments!!

darrensiegel (changes requested):
This all looks really good.  

The only thing that I'm now realizing that we need (that the UX did not specify) is a way to allow the instructor to sort by the order of the units and modules as they appear in the course hierarchy.   The most straightforward way to do this would be to include a new column, to the left of the container name, called "Order" which displays the ordinal position of the container within the hierarchy.  That value is already present in the SectionResource schema as `numbering_index`.

darrensiegel (approved):

PR #3389, opened 2023-04-04
 [BUGFIX] Config MathJax to autoload standard extensions[MER-1863][MER-1869]

andersweinstein:
Adjusts MathJax configuration to allow the autoload extension to be used to automatically load other tex extensions as needed. This fixes issue seen on the \boldsymbol macro, enabling the macro to trigger autoload the boldsymbol tex extension without use of \require{boldsymbol}. Also the \style macro which should trigger autoload the html extension. 

MathJax documentation suggests the tex-mml-chtml combined component we are using includes the autoload package with it so that \requires are not needed for the macros above. But it seems autoload must still be listed in the configured tex packages list before MathJax will use it.

### DIFF
--- a/lib/oli_web/common/mathjax_script.ex
+++ b/lib/oli_web/common/mathjax_script.ex
@@ -49,7 +49,7 @@ defmodule OliWeb.Common.MathJaxScript do
         processHtmlClass: 'tex2jax_process'
       },
       loader: {
-        load: ['[tex]/noerrors', '[tex]/autoload']
+        load: ['[tex]/noerrors']
       },
       renderMathML(math, doc) {
         math.typesetRoot = document.createElement('mjx-container');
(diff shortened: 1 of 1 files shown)


darrensiegel (approved):

andersweinstein:
The reason we are not getting require or autoload extension by default is that our MathJax configuration has (perhaps unknowingly) chosen to override the default tex packages list for the combined MathJax component we are using, tex-mml-chtml. Because of that we have to specify every tex package we want to include explicitly. MathJax defines a magic notation by which a user config can specify *adding* or *subtracting* a package to the default. If we used this notation
`tex: { "packages" : { "[+]" : "noerrors" }`
we could add the noerrors to the default list, and would not have to explicitly specify require or autoload, they would get loaded by default with tex-mml-chtml as docs suggest.

PR #3401, opened 2023-04-11
[MER-1703] Course content browser view

nicocirio:
[Link ](https://eliterate.atlassian.net/browse/MER-1703) to the ticket

This PR also added the possibility of navigating the hierarchy tree "up and down" through a breadscrumb

image


darrensiegel (changes requested):
This is fantastic work.  I tested it out locally with one of our largest courses "French 1" and it works very well.  Easy (and efficient) to use and navigate.  

The only thing that I noticed that we need to fix is that currently it only pulls in "soft scheduled" dates (the dates that appear in SectionResource records).  It also needs to take into account "hard scheduled" dates.  Those are the dates that appear in GatingCondition records of type `:schedule`.    

To do this, you will need to query all `GatingCondition` records of type `:schedule` and combine those results with the results of your new `get_resource_scheduled_dates`.  Any GatingCondition should take precedence over what is present in the SectionResource.  So if a SectionResource record for "Page 1" shows 4/12/23 as the `:read_by`, but there exists a GatingCondition record of type `:schedule` for that same page for 4/28/23, the Content Browser should ignore the SectionResource record and instead show "Due By: 4/28/23"

Also, there can be user specific (aka "Student Exceptions") for the GatingCondition records. If a user specific GatingCondition exists, that should be used instead of the non-user specific GatingCondition record.

nicocirio:
> This is fantastic work. I tested it out locally with one of our largest courses "French 1" and it works very well. Easy (and efficient) to use and navigate.
> 
> The only thing that I noticed that we need to fix is that currently it only pulls in "soft scheduled" dates (the dates that appear in SectionResource records). It also needs to take into account "hard scheduled" dates. Those are the dates that appear in GatingCondition records of type `:schedule`.
> 
> To do this, you will need to query all `GatingCondition` records of type `:schedule` and combine those results with the results of your new `get_resource_scheduled_dates`. Any GatingCondition should take precedence over what is present in the SectionResource. So if a SectionResource record for "Page 1" shows 4/12/23 as the `:read_by`, but there exists a GatingCondition record of type `:schedule` for that same page for 4/28/23, the Content Browser should ignore the SectionResource record and instead show "Due By: 4/28/23"
> 
> Also, there can be user specific (aka "Student Exceptions") for the GatingCondition records. If a user specific GatingCondition exists, that should be used instead of the non-user specific GatingCondition record.

Glad to hear it worked fine in a large course! 
I will add some logic to consider correctly hard scheduled dates, thanks for the feedback!

lib/oli_web/live/delivery/student_dashboard/course_content_live.ex
```diff
@@ -0,0 +1,349 @@
...
+      breadcrumbs_tree
+      |> Enum.reduce({length(breadcrumbs_tree) - 1, hierarchy["children"]}, fn b, acc ->
+        {steps_remaining, hierarchy} = acc
+        {_level, position, _text} = b
+
+        if steps_remaining == 0 do
+          {0, hierarchy}
+        else
+          {steps_remaining - 1, Enum.fetch!(hierarchy, position)["children"]}
+        end
```
simonchoxx, line 271: Just a minor writing detail, maybe this could be written in one line, with something like: `if steps_remaining == 0, do: {0, hierarchy}, else: {steps_remaining - 1, Enum.fetch!(hierarchy, position)["children"]}`

test/oli_web/live/delivery/student_dashboard/course_content_live_test.exs
```diff
@@ -0,0 +1,419 @@
...
+
+    test "with breadcrumbs initially set to Curriculum", %{
+      conn: conn,
+      user: user,
+      section: section
+    } do
+      {:ok, view, _html} = isolated_live_view_course_content(conn, section.slug, user.id)
+
+      assert breadcrumbs_length(view) == 1
+      assert has_element?(view, "button[phx-click=\"breadcrumb-navigate\"]", "Curriculum")
```
simonchoxx, line 49: I'm not sure, but maybe here you could make an assert with the container that could be obtained when using the "section_with_larger_hierarchy function"

test/oli_web/live/delivery/student_dashboard/course_content_live_test.exs
```diff
@@ -0,0 +1,419 @@
...
+    end
+
+    test "navigates left and right", %{conn: conn, user: user, section: section} do
```
simonchoxx, line 52: Same as above, but with the units and pages that are being created in section_with_larger_hierarchy

simonchoxx (commented):
You have done an incredible job Nico!! 👏👏 
I leave you just a few small comments, but you don't need to change them!

darrensiegel (approved):

PR #3405, opened 2023-04-12
[BUG FIX] Fixes an issue where popups were not rendering correctly

eliknebel:
Fixes an issue where popups were not rendering correctly due to previous removal of bootstrap. This PR fixes the issue by adding a ReactPhoenix component which can render any element client-side directly from the server-side rendering code.

- Add DeliveryElementRenderer to call client HTML renderer components directly from server rendering
- Fix toolbar theming


darrensiegel (approved):
Fantastic solution!

PR #3409, opened 2023-04-14
[FEATURE][MER-1730] Redirect students to course survey

augusto1024:
[Link to the ticket](https://eliterate.atlassian.net/browse/MER-1730)

### NOTES
In order to dev qa this ticket, you need to edit the survey page and add some activity to it. For this, you need to check the slug of the survey's revision, go to `/authoring/project/#{project_slug}/resource/#{survey_slug}` and add some multiple choice question or something. Then you need to publish the changes, apply them to the section and check if the user gets redirected to the survey.

### DIFF
--- a/lib/oli/delivery.ex
+++ b/lib/oli/delivery.ex
@@ -296,39 +296,42 @@ defmodule Oli.Delivery do
   def has_completed_survey?(section_slug, user_id) do
     survey = Sections.get_survey(section_slug)
 
-    survey_activities =
-      Enum.reduce(survey.content["model"], [], fn item, activities ->
-        if item["type"] == "activity-reference" do
-          [item["activity_id"] | activities]
-        else
-          activities
-        end
-      end)
-
     Section
     |> join(:inner, [s], spp in SectionsProjectsPublications, on: spp.section_id == s.id)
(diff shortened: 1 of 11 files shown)


lib/oli/delivery.ex
```diff
@@ -291,4 +292,43 @@ defmodule Oli.Delivery do
         {:ok, section}
     end
   end
+
+  def has_completed_survey?(section_slug, user_id) do
+    survey = Sections.get_survey(section_slug)
```
darrensiegel, line 297: This approach to checking if all of the activities present are evaluated is problematic because it does not take into account activity bank selections.   Furthermore, it only identifies `activity-reference` elements at the top level (they can exist nested within groups)

Instead of fetching the page revision, the correct approach is to simply fetch the `%ActivityAttempt` records for that student for the latest `%ResourceAttempt` for the survey resource id.  It is crucial that only the latest `%ResourceAttempt` record is considered.  For the latest resource attempt, if all activity records present are evaluated, then we consider this required survey page to be complete.   This approach properly takes into account activity bank selections (and will handle the upcoming audience filtered content feature).

darrensiegel (changes requested):
See inline comment

darrensiegel (approved):

PR #3413, opened 2023-04-18
MER-1438

dtiwarATS:


assets/src/apps/delivery/store/features/page/actions/loadInitialPageState.ts
```diff
@@ -173,14 +172,6 @@ export const loadInitialPageState = createAsyncThunk(
           }, 0);
           resumeSequenceId = sequence[resumeTarget].custom.sequenceId;
         }
-        // need to check the visitHistory to see if the resumeSequenceId is in there and is NOT the latest, then we need to set history mode to true
```
dtiwarATS, line 176: Previously, we used to check the visitHistory to see if the  resumeSequenceId is in there and if we found it then we used to set history mode to true.  

We don't need to that anymore because when we resume now, we always navigated to the current active screen.

assets/src/apps/delivery/store/features/page/actions/loadInitialPageState.ts
```diff
@@ -37,7 +36,7 @@ export const loadInitialPageState = createAsyncThunk(
     if (otherTypes.length) {
       groups.push({ type: 'group', layout: 'deck', children: [...otherTypes] });
     }
-    const isReviewMode = selectReviewMode;
+    const isReviewMode = selectReviewMode(getState() as DeliveryRootState);
```
dtiwarATS, line 39: This was a bug from review mode that I fixed.

bsparks (approved):

PR #3415, opened 2023-04-18
[MER-1723] Learning objectives tab for student dashboard

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-1723) to the ticket

### DIFF
--- a/lib/oli/delivery/sections.ex
+++ b/lib/oli/delivery/sections.ex
@@ -1117,19 +1117,21 @@ defmodule Oli.Delivery.Sections do
 
   def get_project_by_section_resource(section_id, resource_id) do
     Repo.one(
-      from s in SectionResource,
+      from(s in SectionResource,
         join: p in Project,
         on: s.project_id == p.id,
         where: s.section_id == ^section_id and s.resource_id == ^resource_id,
         select: p
+      )
     )
   end
 
   def get_section_resource(section_id, resource_id) do
     Repo.one(
-      from s in SectionResource,
+      from(s in SectionResource,
         where: s.section_id == ^section_id and s.resource_id == ^resource_id,
         select: s
+      )
     )
   end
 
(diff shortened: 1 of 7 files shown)


augusto1024 (commented):
Great work! I really liked the way you handled multiple routes with the `patch_url_type` 👏

lib/oli/delivery/sections.ex
```diff
@@ -2400,7 +2403,9 @@ defmodule Oli.Delivery.Sections do
     objectives_pages_map =
       DeliveryResolver.all_pages(section_slug)
```
darrensiegel, line 2404: I think we want a specialized query that will pull all attached objectives, instead of fetching the revisions of all pages.  In a large course (with more than 500 or 600 pages) this approach may introduce problems.

lib/oli/publishing/delivery_resolver.ex
```diff
@@ -135,6 +134,22 @@ defmodule Oli.Publishing.DeliveryResolver do
...
 
+  def pages_with_attached_objectives(section_slug) do
+    page_id = Oli.Resources.ResourceType.get_id_by_type("page")
+
+    fn ->
+      from([s: s, sr: sr, rev: rev] in section_resource_revisions(section_slug),
+        where:
+          rev.resource_type_id == ^page_id and
+            fragment("? != '[]'", rev.objectives["attached"]),
+        select: rev
```
darrensiegel, line 145: Instead of selecting the entire revision, can you change this to only select the page resource_id and the objectives (the only two things that the callers of this actually need).  Currrently this selects the entire revision, which contains the potentially large `content` field. I was hoping to avoid selecting that content field.

darrensiegel (changes requested):

darrensiegel (approved):

PR #3416, opened 2023-04-18
[FEATURE][MER-1731] Extract survey answers to user details view

augusto1024:
[Link to the story](https://eliterate.atlassian.net/browse/MER-1731?atlOrigin=eyJpIjoiNDZiNmUwODZiNjljNGU2M2E5N2M2NGRmZjNjODVhMmYiLCJwIjoiaiJ9)

https://user-images.githubusercontent.com/53446634/232901857-68a88be9-811f-49d9-878d-6f24fb2b9109.mov

### DIFF
--- a/lib/oli/delivery/attempts.ex
+++ b/lib/oli/delivery/attempts.ex
@@ -1,2 +1,47 @@
 defmodule Oli.Delivery.Attempts do
+  alias Oli.Delivery.Attempts.Core.{ResourceAttempt, ActivityAttempt, ResourceAccess, PartAttempt}
+  alias Oli.Resources.Revision
+
+  alias Oli.Repo
+  import Ecto.Query, warn: false
+
+  def summarize_survey(survey_resource_id, user_id) do
+    ActivityAttempt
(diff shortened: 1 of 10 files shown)


lib/oli/delivery/attempts.ex
```diff
@@ -1,2 +1,47 @@
...
+      r_acc.resource_id == ^survey_resource_id and r_acc.user_id == ^user_id and is_nil(a_att_2) and
+        a_att.lifecycle_state == :evaluated
+    )
+    |> select([a_att, r_att, r_acc, rev, a_att_2, p_att], %{
+      question: rev.title,
+      options: a_att.transformed_model["choices"],
+      answer: p_att.response
+    })
+    |> Repo.all()
+    |> Enum.map(fn %{question: question, options: options, answer: answer} ->
```
darrensiegel, line 33: We should add a step here to ensure that we only consider multiple choice questions.  This can be done simply by expanding the query to join thru the `activity_type_id` of the activity revision to the registered activity, then filtering the result in the where clause  by only those that are  multiple choice activities.

darrensiegel (changes requested):
This looks great, only one comment, see inline:

darrensiegel (approved):

PR #3419, opened 2023-04-20
[BUGFIX] hotspot error if questions use same choice IDs [MER-1905]

andersweinstein:
Image hotspot generated an ID for the image’s associated map element based on the id of the first choice in the question. However choice IDs do not have to be globally unique. Ran into a case in an ingested course of multiple questions using same image/hotspots, apparently copied and pasted, leading to bad behavior on delivery (mousing over any of the images always highlighted the first one.) This changes to use a guid.

### DIFF
--- a/assets/src/components/activities/image_hotspot/ImageHotspotDelivery.tsx
+++ b/assets/src/components/activities/image_hotspot/ImageHotspotDelivery.tsx
@@ -26,6 +26,7 @@ import { castPartId } from '../common/utils';
 import * as ActivityTypes from '../types';
 import { Hotspot, ImageHotspotModelSchema, getShape } from './schema';
 import { HS_COLOR, drawHotspotShape } from './utils';
+import guid from 'utils/guid';
 
 const ImageHotspotComponent: React.FC = () => {
   const {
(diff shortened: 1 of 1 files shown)


assets/src/components/activities/image_hotspot/ImageHotspotDelivery.tsx
```diff
@@ -110,7 +111,7 @@ const ImageHotspotComponent: React.FC = () => {
     }
   };
 
-  const mapName = 'map' + model.choices[0].id;
+  const mapName = 'map' + guid();
```
darrensiegel, line 114: With this approach, the `mapName` changes on every render.  That might pose a problem. Instead, you should capture the result of `guid()` in state or via `useRef`

darrensiegel (changes requested):

assets/src/components/activities/image_hotspot/ImageHotspotDelivery.tsx
andersweinstein: That's why I avoided generating a guid the first time around! Not clear to me this causes any problem, though.

andersweinstein: Fixed

PR #3421, opened 2023-04-21
[MER-1864] support audience targeted structured content and groups

eliknebel:
Adds support for audience attribute on groups. Filters out any activities during bulk attempt creation and filters the content during rendering so that only the intended audience will generate data and see content that is label by a corresponding audience.

I also noticed the font awesome js was unnecessary, and it results in the icon being swapped out for the svg version. It seems to be much cleaner to do without this and simply rely on the icon font set by the css imports.


darrensiegel (changes requested):
If I place a survey with an activity in it in an Instructor only group, then view it as a student I see: 

`key :lifecycle_state not found in: nil. If you are using the dot syntax, such as map.field, make sure the left-hand side of the dot is a map`

```
[lib/oli_web/controllers/page_delivery_controller.ex](https://localhost/sections/example_course_section/page/new_page)
    submitted_surveys =
      PageContent.survey_activities(context.page.content)
      |> Enum.reduce(%{}, fn {survey_id, activity_ids}, acc ->
        survey_state =
          Enum.all?(activity_ids, fn id ->
            context.activities[id].lifecycle_state === :submitted ||
              context.activities[id].lifecycle_state === :evaluated
          end)
        Map.put(acc, survey_id, survey_state)
      end)
```

lib/oli/delivery/activity_provider.ex
```diff
@@ -241,22 +254,30 @@ defmodule Oli.Delivery.ActivityProvider do
...
+  defp do_fulfill(
+         fulfillment_state,
+         %{"type" => type} = model_component,
+         group_id,
+         survey_id,
+         user,
+         section_slug
+       ) do
+    if PageContent.is_resource_group?(model_component) &&
+         Audience.is_intended_audience?(model_component["audience"], user, section_slug, false) do
```
darrensiegel, line 266: So I think this is likely the wrong place to do this filtering, as evidenced by the bug that appears when one places a survey with an activity in an instructor only group.

I think we must add an upfront, earlier step that prunes out the page content to remove audience groups that do not match.  Then, that "transformed" page content must be the content that is stored in the `resource_attempt` record in the `:content` field.  The principal reason for this `:content` value is that it stores the transformed page content after activity realization has occurred - where activity bank selections are turned into actual activity-references.  The results of audience filtering must be captured in this field.

darrensiegel (changes requested):

darrensiegel:
The latest round of changes updates the approach taken here to:
1. Perform an "upfront" filtering of the content of the page, in a way that preserves that filtered content in the resource_attempt's `content` attribute.  This was necessary to properly support activities and surveys within "instructor" only groups. 
2. Refine the audience filtering during rendering to only handle the case of showing or hiding feedback groups. This is now done in a way that requires no database calls per each content element.

darrensiegel (approved):

PR #3423, opened 2023-04-21
[FEATURE][MER-1702] Add "Up Next" view to student dashboard

augusto1024:
[Link to the story](https://eliterate.atlassian.net/browse/MER-1702)

### DIFF
--- a/lib/oli/delivery/metrics.ex
+++ b/lib/oli/delivery/metrics.ex
@@ -7,6 +7,9 @@ defmodule Oli.Delivery.Metrics do
   alias Oli.Delivery.Sections.SectionResource
   alias Oli.Delivery.Attempts.Core.{ResourceAccess, ActivityAttempt}
   alias Oli.Delivery.Attempts.Core
+  alias Oli.Delivery.Sections.{Enrollment, EnrollmentContextRole}
+  alias Oli.Accounts.User
+  alias Lti_1p3.Tool.ContextRoles
 
   @doc """
   Calculate the progress for a specific student (or a list of students),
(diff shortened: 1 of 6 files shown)


lib/oli/delivery/sections.ex
```diff
@@ -2272,58 +2272,128 @@ defmodule Oli.Delivery.Sections do
...
+    })
+    |> order_by([{:asc_nulls_last, fragment("end_date")}])
+  end
+
+  @doc """
+    Returns the activities that a student need to complete next.
+  """
+  def get_next_activities_for_student(section_slug, user_id) do
+    get_student_pages(section_slug, user_id)
+    |> Repo.all()
```
darrensiegel, line 2341: I think that it would make sense to move this step into the query that is executed in `get_student_pages`.  I think that it would involve adding a where clause comparing the end date against the current date and time, ordering by the date/time and then adding a `limit` of 2.   That will avoid this query (which has to run on every view of the course portal) from fetching all pages.

darrensiegel (changes requested):
See inline

lib/oli/delivery/metrics.ex
```diff
@@ -7,6 +7,9 @@ defmodule Oli.Delivery.Metrics do
   alias Oli.Delivery.Sections.SectionResource
   alias Oli.Delivery.Attempts.Core.{ResourceAccess, ActivityAttempt}
   alias Oli.Delivery.Attempts.Core
+  alias Oli.Delivery.Sections.{Enrollment, EnrollmentContextRole}
```
simonchoxx, line 10: Maybe we could add within the `Oli.Delivery.Sections.{Enrollment, EnrollmentContextRole}` aliases also the calls to the `SectionResource` and `ContainedPage` aliases found a few lines above, WDYT?

simonchoxx (commented):
Nice work mate!! 🚀🚀

PR #3481, opened 2023-05-04
MER-1930

dtiwarATS:
Looks like the new CSS that got applied across the Torus impacted the delivery parts rendering as well. 

In this case, the following CSS was affecting the rendering of the img in MCQ item.

```
 .content img {
    margin-bottom: 1.5em;
}
```

I have made the change so that 'auto' gets set to margin-bottom property. Again, this can be overridden from an external CSS file so it should not cause any issues.


rmathis (changes requested):
This doesn't seem optimal to solve the issue, as the classes you are affecting are defined elsewhere, so if they change there is no way for the person changing them to know it needs to change in other files.

One option would be to add it to the element via JS within the loop itself (like style.display is being added to the img parent), or to define this in the css file near where the classes are styled in the first place.

dtiwarATS:
Closing this PR as we can handle it from the external CSS file that contains the override styling for lesson. I will be making changes in inspark_Assets repository.

PR #3490, opened 2023-05-04
[BUG FIX] Enrolled users table model over constrains payment status pattern

darrensiegel:
** (CaseClauseError) no case clause matching: :instructor
Backtrace: lib/oli_web/live/users/user_enrolled_table_model.ex:105 in OliWeb.Users.UserEnrolledTableModel.render_payment_status/3

See the above in AppSignal.  Root case is that payment status is being reported as `:instructor` for users that are instructors in a paid section.  Added a catch all clause to make this future proof.


simonchoxx (approved):
LGTM!!

andersweinstein (approved):

PR #3514, opened 2023-05-11
HOTFIX - MER-1916 | MER-1940 

dtiwarATS:


assets/src/components/activities/adaptive/AdaptiveDelivery.tsx
```diff
@@ -193,6 +193,7 @@ const Adaptive = (props: DeliveryElementProps<AdaptiveModelSchema>) => {
             },
             env,
           });
+          partInitResponseMap.clear();
```
dtiwarATS, line 196: Every part call `props.onInit()` and send its default values. We save all the responses in `partInitResponseMap` and once every part calls its `props.onReady()`, we combine all the responses from each part and do a single save in database.

Once we save these part responses, we should have cleared the `partInitResponseMap` array.

Since, I didn't clear the `partInitResponseMap`, when we navigate to the nested sub-screens, default value of parent screen part was getting saved again even though they were not re-rendering and this happend because  they were present in `partInitResponseMap` array.

assets/src/apps/authoring/components/RightMenu/RightMenu.tsx
```diff
@@ -416,6 +416,7 @@ const RightMenu: React.FC<any> = () => {
               uiSchema={bankUiSchema}
               value={questionBankData}
               onChangeHandler={bankPropertyChangeHandler}
+              triggerOnChange={true}
```
dtiwarATS, line 419: MER-1916

rmathis (approved):
LGTM

Charles-Jones-Unicon (approved):

PR #3533, opened 2023-05-16
[MER-1988] add ability to reset practice page attempt

eliknebel:
Adds the ability to reset all activities on an ungraded page by finalizing the resource attempt and creating a new one on subsequent access, similar to how graded pages work except no scores are recorded.

Screenshot 2023-05-16 at 3 00 39 PM

### DIFF
--- a/lib/oli/delivery/attempts/page_lifecycle/ungraded.ex
+++ b/lib/oli/delivery/attempts/page_lifecycle/ungraded.ex
@@ -14,8 +14,7 @@ defmodule Oli.Delivery.Attempts.PageLifecycle.Ungraded do
     Hierarchy
   }
 
-  alias Oli.Delivery.Attempts.Core.{ResourceAttempt, ResourceAccess}
-  alias Oli.Delivery.Attempts.ActivityLifecycle.{Evaluate, Persistence}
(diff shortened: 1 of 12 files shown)


lib/oli/delivery/attempts/page_lifecycle/ungraded.ex
```diff
@@ -49,8 +55,50 @@ defmodule Oli.Delivery.Attempts.PageLifecycle.Ungraded do
...
 
   @impl Lifecycle
-  def finalize(%FinalizationContext{} = _context) do
-    {:error, {:unsupported}}
+  def finalize(%FinalizationContext{
+        resource_attempt: %ResourceAttempt{lifecycle_state: :active} = resource_attempt,
+        datashop_session_id: datashop_session_id
+      }) do
+    # Collect all of the part attempt guids for all of the activities that get finalized
+    with {:ok, part_attempt_guids} <-
```
darrensiegel, line 63: This finalization can be much simpler: all that it needs to do is mark the attempt as `:evaluated`.  It should not evaluate existing active activities.  That will end up polluting analytics.

darrensiegel (changes requested):

darrensiegel (approved):

PR #3535, opened 2023-05-17
[MER-1978] Block list items

marc-hughes:
This PR allows block elements inside list items.

One important change to note here, is I had to start wrapping text nodes inside list items in a paragraph. That will happen in a normalization pass for any existing content the first time it's edited. I haven't seen any visual differences, but it's a possibility.

Since pressing enter in a LI crates a new list item, an author can now create multiple paragraphs inside an LI by pressing shift-enter. 

I am not currently inserting leading and trailing paragraphs around block items. So if you have some text, then insert a block image, it will be difficult to type in the same list item after that image.  I think this best preserves the visual appearance we want in most scenarios. An author can work around this by typing in the content they want then inserting an image in the middle of that text to break it into two.

These interactions should all work:

- Pressing enter at the start of a list item creates a new empty list item "before" the current one.
- Pressing enter on an empty list item that is the last one of the list, gets rid of the empty LI and closes the list.
- Pressing enter at the end of a list item creates a new after it
- Pressing enter in the middle of a list item, splits the content into 2 different list items. Before, this was always 2 text nodes. Now it may be multiple block elements.
- Pressing shift-enter to create a new paragraph within a LI
- Pressing tab / shift-tab at the start of a LI indents or outdents the item
- The list item toolbar items all work
   - I changed the behavior of indent/outdent so the toolbar items do this no matter where they are in the list item, before they operated like the keyboard events and only worked at the start of the item.

### DIFF
--- a/assets/src/components/editing/editor/handlers/lists.ts
+++ b/assets/src/components/editing/editor/handlers/lists.ts
@@ -62,8 +62,9 @@ export function handleIndent(editor: SlateEditor, e?: KeyboardEvent) {
     const [current, listItemPath] = match;
     const start = SlateEditor.start(editor, listItemPath);
 
-    if (!Point.equals(editor.selection.anchor, start)) {
+    if (e && !Point.equals(editor.selection.anchor, start)) {
       // Only indent if the cursor is at the start of the list item.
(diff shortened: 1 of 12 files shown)


darrensiegel (approved):
Tested locally, works very nicely!

PR #3544, opened 2023-05-17
[FEATURE] Time limit, grace period, late start, late submit, auto submit and review submission settings

darrensiegel:
This PR implements give new assessment options:

1. Time Limit: the number of minutes available for a student to complete a graded page
2. Late submit: An option to either allow or disallow submissions past the deadline.  Allowing late submissions will mark the attempt and the resource access record as "late".  Not allowing late submissions will auto submit the attempt at the deadline.
3. Late start: An option to allow attempts to be started after a deadline.  Allow or disallow.  
4. Grace period: the number of minutes after a deadline (either time limit or due date) that a student can submit before the late submit policy is applied.
5. Review submission:  An option which either allows or forbids a student from reviewing their completed attempts.  The default is `:allow`.  

## Auto submit

The auto submit implementation is both server side and client side, as it needs to handle the following two cases:
1. Student is actively working on an attempt and they simply run out of time.  The page should auto submit for them.
2. Student starts an attempt, then closes the page and walks away.  The server must auto submit that attempt when the deadline is reached.

The server side auto submit is an Oban job that is scheduled at the due date (plus the grace period).  The job id gets recorded into the resource attempt and is cancelled when the student submits ahead of time, on their own.  

The client side auto submit is driven from a timer that will programmatically click the button and submit the attempt. 

## Testing this out

Testing this out is difficult given that the UX for defining assessment settings isn't available yet.  But to test simply the time limit aspect, you can uncomment lines 556 in `page_delivery_controller.ex`: 

```
# For testing, you can uncomment to introduce a time out
# effective_settings = %{effective_settings | time_limit: 2, late_submit: :disallow}
```

To see server side auto submit in action, you have to replace line 77 in `page_lifecycle.ex` with:

```
effective_settings: %{effective_settings | time_limit: 2, late_submit: :disallow}
```


eliknebel (approved):
Code looks good. Tested and works as expected

PR #3550, opened 2023-05-18
[ENHANCEMENT] Remove engagement and fix objectives display bug [MER-1706]

darrensiegel:
Removes all "Student Engagement" references from UI and code, since we are not moving forward with that analytic

In testing my changes, I caught a bug in how the Learning Objectives tab was populated.  It only showed LO's that were actually attached to pages.  I removed that constraint so that it shows all LOs.

### DIFF
--- a/assets/css/table.css
+++ b/assets/css/table.css
@@ -44,7 +44,7 @@
   @apply bg-red-200 bg-opacity-20;
 }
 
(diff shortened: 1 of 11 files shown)


augusto1024 (changes requested):
Looks good! But there's one comment I left that we might want to address

augusto1024 (approved):

PR #3552, opened 2023-05-18
[ENHANCEMENT] only students have to have posts approved [MER-2041]

darrensiegel:
This fix allows instructors to post without needing to approve their own posts.


nicocirio (approved):
LGTM 🚀

PR #3559, opened 2023-05-19
[BUG FIX] follow relates_to links [MER-2009]

darrensiegel:
Unordered explorations were being culled from sections since they were only accessible via the "relates_to" attribute.  This PR expands the determination of the "reachable set" of pages to take into account relates_to. 

Tested with an unordered exploration and it preserves it in the course section.

### DIFF
--- a/lib/oli/delivery/sections.ex
+++ b/lib/oli/delivery/sections.ex
@@ -911,7 +911,8 @@ defmodule Oli.Delivery.Sections do
 
   # For a given section id and the list of resource ids that exist in its hiearchy,
   # determine and return the list of page resource ids that are not reachable from that
-  # hierarchy and linked pages.
+  # hierarchy, taking into account links from pages to other pages and the 'relates_to'
+  # relationship between pages.
   defp determine_unreachable_pages(publication_ids, hierarchy_ids) do
     # Start with all pages
     unreachable =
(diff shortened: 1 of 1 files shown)


lib/oli/delivery/sections.ex
```diff
@@ -1028,6 +1034,36 @@ defmodule Oli.Delivery.Sections do
...
+    Enum.reduce(results, MapSet.new(), fn [source_id, relates_to], links ->
+
+      # The relates_to field is an array of resource ids, to be future proof
+      # to how relates_to is used, we will follow these 'links' in both directions
+      links = Enum.reduce(relates_to, links, fn target_id, links ->
+        MapSet.put(links, {source_id, target_id})
+      end)
+      Enum.reduce(relates_to, links, fn target_id, links ->
+        MapSet.put(links, {target_id, source_id})
+      end)
```
augusto1024, line 1063: Instead of going through the `relates_to` list twice, WDYT about doing this?
```suggestion
      Enum.reduce(relates_to, links, fn target_id, links ->
        MapSet.put(links, {source_id, target_id}) |> MapSet.put({target_id, source_id})
      end)
```

lib/oli/delivery/sections.ex
```diff
@@ -1028,6 +1034,36 @@ defmodule Oli.Delivery.Sections do
...
+    page_type_id = Oli.Resources.ResourceType.get_id_by_type("page")
+
+    sql = """
+    select
+      rev.resource_id, rev.relates_to
+    from published_resources as mapping
+    join revisions as rev
+    on mapping.revision_id = rev.id
+    where rev.resource_type_id = #{page_type_id} and array_length(rev.relates_to, 1) > 0 and mapping.publication_id IN (#{joined_publication_ids})
+    """
```
augusto1024, line 1050: Is there a reason why we're using raw SQL instead of Ecto queries?

darrensiegel: I couldn't do this in an Ecto query due to the usage of `array_length`.  I guess I could have used just a fragment for that one part.

PR #3569, opened 2023-05-23
MER-2010

dtiwarATS:


assets/src/apps/delivery/store/features/page/actions/loadInitialPageState.ts
```diff
@@ -60,8 +60,12 @@ export const loadInitialPageState = createAsyncThunk(
       sessionState['app.active'] = 'none';
       // read all user state for the assigned everapps into the session state
       /* console.log('INIT PAGE', params); */
-      if (params.content.custom?.everApps) {
```
dtiwarATS, line 63: Previously, the `app.something.value` variables are processed only when the lesson has a Ever App in it. 

This has taken away the ability to use the `app.something.variablename` as a global variable that can be used across multiple lesson. This PR fixes this issue.

rmathis (approved):
This looks good

darrensiegel (approved):

PR #3570, opened 2023-05-23
[FEATURE] Container specific progress CSV download

darrensiegel:
This satisfies https://eliterate.atlassian.net/browse/MER-2042, though it won't initially be available as a clickable link in the UX until we implement: 

https://eliterate.atlassian.net/browse/MER-2059


lib/oli_web/controllers/metrics_controller.ex
```diff
@@ -0,0 +1,26 @@
...
+      nil ->
+        OliWeb.StaticPageController.not_found(conn, params)
+
+      section ->
+        contents = Oli.Delivery.Metrics.progress_datatable_for(section.id, container_id)
+        |> Oli.Analytics.DataTables.DataTable.to_csv_content()
+
+        conn
+        |> send_download({:binary, contents},
+          filename: "progress.csv"
```
augusto1024, line 20: NIT: Perhaps we might want to append the section slug to the file name so it's easier to find?

lib/oli_web/router.ex
```diff
@@ -835,6 +835,12 @@ defmodule OliWeb.Router do
         :preview
       )
     end
+
+    get(
+      "/downloads/progress/:container_id",
+      MetricsController,
+      :download_container_progress
+    )
```
augusto1024, line 843: Since this is downloading a file, I don't think this is the best place to put this route. Maybe we can have it under the `/api` scope like some other endpoints we have?

```suggestion
    scope "/api/v1/sections/:section_slug"
      pipe_through([
        :browser,
        :delivery_and_admin,
        :maybe_gated_resource,
        :pow_email_layout
      ])
      
      get(
        "/:container_id/progress",
        MetricsController,
        :download_container_progress
      )
  end
```

augusto1024 (commented):
Looks good! Just left two minor comments

lib/oli_web/router.ex
darrensiegel: I don't think `/api` is the right place for downloading file routes, only since these routes all originate from direct user interaction (e.g. a button or a link in the UX is clicked to trigger it) as opposed to a piece of a client side typescript code invoking the endpoint.

augusto1024: Yes, you're right that we're invoking `/api` endpoints just from the client side. I still think it's a little bit weird that it's under the `/sections/:section_slug/instructor_dashboard` scope but maybe if we use it somewhere else in the future, we can move it then

augusto1024 (approved):

PR #3587, opened 2023-05-25
[MER-1944] Instructor level assessments settings

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-1944) to the ticket
[Link ](https://docs.google.com/spreadsheets/d/1uEPyPyWxMAanRPB_9fl71nGvHezN9JgU1vC5Y3RlQvg/edit#gid=0) to the UX design

## TODOs
- [x] Add sort option for all columns
- [x] Improve assessment naming
- [x] Improve styling (for small devices)
- [x] Add tests

### DIFF
--- a/assets/css/table.css
+++ b/assets/css/table.css
@@ -59,3 +59,7 @@
 .instructor_dashboard_table tr:has(a[data-score-check='false']) {
   @apply bg-red-200 bg-opacity-20;
 }
+
+.instructor_dashboard_table td:has(div.highlight-exception) {
+  @apply bg-blue-100 rounded-md bg-opacity-50;
+}
--- a/assets/css/table.css
+++ b/assets/css/table.css
@@ -63,3 +63,7 @@
 .instructor_dashboard_table td:has(div.highlight-exception) {
   @apply bg-blue-100 rounded-md bg-opacity-50;
 }
+
+div:has(.instructor_dashboard_table) #header_paging {
+  @apply sticky left-0;
+}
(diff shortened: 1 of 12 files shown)


darrensiegel:
I did usage based testing, and just a brief cursory code review.  This is looking really good!

Some things I saw in using it:
1. When trying to access it via the link from Manage Section, I got an error `assign @title not available in template.`.  I worked around this by adding a `:title` assign in `settings_live.ex`
2. I really the like approach of the "frozen" first column and the horizontal scrollbar for the rest of the columns - as opposed to a more rigid "paging" implementation.  But I noticed that when i made my window very wide, the UI stopped increasing its width at a certain point and I still had to scroll to see everthing.  Maybe it would be best just to let this view go 100% width?
3. The sorting of columns does not seem to work

augusto1024 (commented):
Looks good to me! There's a small bug that we discussed offline, but otherwise great work 👏👏👏

darrensiegel (changes requested):
This looks really good.  

See the two fixes that I noted inline.

Also, when I try to change the "View Feedback" setting to "Scheduled", the modal dialog appears and I can pick a date and time and click "Save".  But, I noticed that then the setting for View Feedback back in the table doesn't display "Scheduled".  It remains as the previous value.  Only after clicking "Scheduled" again, then clicking "save" in the dialog does the value in the table display as "Scheduled"

darrensiegel (approved):

PR #3599, opened 2023-06-01
[BUG FIX] Handle activity to activity references on ingest [MER-2093]

darrensiegel:
The flowchart based adaptive page introduces activity to activity references via resource ids.  These have to be updated to their new resource ids after one exports a project with a flowchart page, then re-ingests. 

How to test this end to end:

1. Create a three screen, flowchart based adaptive page.
2. Export that project.
3. Using V2 ingest, ingest that zip file.
4. Verify that the links still exist between the three screens.


eliknebel:
Fixed by [#3619](https://github.com/Simon-Initiative/oli-torus/pull/3619)

PR #3608, opened 2023-06-01
[Feature] [MER-2078] adaptive pages inside torus navigation

marc-hughes:
Adds the ability to display adaptive lessons within the torus navigation UI.

To enable this, see the "Display Torus Navigation" option on the Lesson panel of the adaptive authoring environment.

image

Demo:

https://github.com/Simon-Initiative/oli-torus/assets/333265/c9231154-4635-41dc-a23f-4c9329216453

### DIFF
--- a/lib/oli/delivery/adaptive_iframe.ex
+++ b/lib/oli/delivery/adaptive_iframe.ex
@@ -5,28 +5,28 @@ defmodule Oli.Delivery.AdaptiveIFrame do
   @chrome_height 175
   @chrome_width 150
 
-  defp get_size(revision) do
-    content_height = Map.get(revision.content["custom"], "defaultScreenHeight", @default_height)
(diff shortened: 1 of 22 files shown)


lib/oli/delivery/adaptive_iframe.ex
```diff
@@ -0,0 +1,41 @@
+defmodule Oli.Delivery.AdaptiveIFrame do
```
darrensiegel, line 1: This module probably should be someplace under `OliWeb`, instead of `Oli` so that we do not introduce a dependency on `OliWeb.Router.Helpers` into the core `Oli` space.

darrensiegel (changes requested):
This is great.  Just one change to avoid a dependency ordering problem

darrensiegel (approved):

Issue #3506, opened 2023-05-09
Tokamak: Drag and Drop doesn't always register feedback


halturner:
When dropping initiators into drag and drop zones, the feedback is not always triggered. See the screen recording, where an initiator was dropped into multiple zones that initially didn't result in feedback, but later did.

https://tokamak.oli.cmu.edu/sections/kth_onelearns_4_may_2023/page/74yy1_knowledge_37jhg
LBD with text "By reading the material above you should be able to infer the correct answers to these questions. Each of the statements below represent a student acquiring a piece of knowledge/a type of knowledge. Classify the knowledge described as Declarative, Procedural, or Concepts & Principles."

https://github.com/Simon-Initiative/oli-torus/assets/24977416/a7828c02-0b80-49b3-a81a-e87ea31ef47e

darrensiegel:
https://eliterate.atlassian.net/browse/MER-1974

eliknebel:
fixed

PR #3636, opened 2023-06-06
[MER-1974] custom dnd race condition

marc-hughes:
This fixes an issue described in MER-1974 / #3506 where, in delivery mode, moving the dropped tile around after the initial drop would sometimes lead to feedback not being triggered.

The root problem was that the custom dnd activity was resetting the part attempt, and then calling resetAndSubmitPart without waiting for the part attempt reset to finish. Most of the time, that would work out just fine. Sometimes, the second call finished before the first. In those cases, the order of the part attempts in the DB would be backwards from what was expected, and the server call to submit the next attempt would fail since it only looks for the latest attempt record.

Preview mode continues to only work for the first drop of each tile. That was the behavior before this fix and I left that as-is until higher priority items can be finished.

### DIFF
--- a/assets/src/components/activities/check_all_that_apply/CheckAllThatApplyDelivery.tsx
+++ b/assets/src/components/activities/check_all_that_apply/CheckAllThatApplyDelivery.tsx
@@ -105,7 +105,7 @@ export const CheckAllThatApplyComponent: React.FC = () => {
 // Defines the web component, a simple wrapper over our React component above
 export class CheckAllThatApplyDelivery extends DeliveryElement<CATASchema> {
   render(mountPoint: HTMLDivElement, props: DeliveryElementProps<CATASchema>) {
-    const store = configureStore({}, activityDeliverySlice.reducer);
+    const store = configureStore({}, activityDeliverySlice.reducer, { name: 'CATADelivery' });
     ReactDOM.render(
       <Provider store={store}>
         <DeliveryElementProvider {...props}>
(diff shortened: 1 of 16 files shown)


assets/src/components/activities/check_all_that_apply/CheckAllThatApplyDelivery.tsx
```diff
@@ -105,7 +105,7 @@ export const CheckAllThatApplyComponent: React.FC = () => {
 // Defines the web component, a simple wrapper over our React component above
 export class CheckAllThatApplyDelivery extends DeliveryElement<CATASchema> {
   render(mountPoint: HTMLDivElement, props: DeliveryElementProps<CATASchema>) {
-    const store = configureStore({}, activityDeliverySlice.reducer);
+    const store = configureStore({}, activityDeliverySlice.reducer, { name: 'CATADelivery' });
```
marc-hughes, line 108: At one point, I had a theory about multiple redux stores being the cause of this bug, and the redux-dev tools were very difficult to use because they weren't labeled. It didn't end up being the case, but I did label these to make it easier for future debugging.

assets/src/components/activities/custom_dnd/DragCanvas.tsx
```diff
@@ -118,35 +118,43 @@ function focusDraggable(shadowRoot: any, draggableId: string | null, props: Drag
...
+        const newlyDropped = (ev.currentTarget as any).children[0];
+        newlyDropped.style.left = '0px';
+        newlyDropped.style.top = '0px';
+        newlyDropped.style.position = 'relative';
+
+        const targetId = (ev.currentTarget as any).getAttribute('input_ref');
+        focusTarget(shadowRoot, targetId, props);
+
+        if (prevTargetId) {
+          await props.onDetach(prevTargetId, inputVal);
```
marc-hughes, line 151: Moving this here and throwing in an await (as well as some upstream changes to make it async) was the actual fix.

eliknebel (approved):

PR #3656, opened 2023-06-07
[BUGFIX][MER-2097] Submit survey button not rendered

augusto1024:
[Link to the story](https://eliterate.atlassian.net/browse/MER-2097)

image

### DIFF
--- a/lib/oli/rendering/activity/html.ex
+++ b/lib/oli/rendering/activity/html.ex
@@ -84,7 +84,12 @@ defmodule Oli.Rendering.Activity.Html do
           render_single_activity_html(tag, context, summary, bib_params, model_json)
         else
           [
-            render_historical_attempts(summary.id, context.historical_attempts, section_slug, context.liveview),
+            render_historical_attempts(
+              summary.id,
+              context.historical_attempts,
+              section_slug,
+              context
+            ),
             render_single_activity_html(tag, context, summary, bib_params, model_json)
           ]
         end
@@ -94,7 +99,7 @@ defmodule Oli.Rendering.Activity.Html do
     end
   end
 
-  defp render_historical_attempts(activity_id, historical_attempts, section_slug, liveview) do
+  defp render_historical_attempts(activity_id, historical_attempts, section_slug, context) do
     case historical_attempts do
       nil ->
         []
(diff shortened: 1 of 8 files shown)


eliknebel:
I think the Jira link might not be correct here

lib/oli/rendering/activity/html.ex
```diff
@@ -109,23 +109,45 @@ defmodule Oli.Rendering.Activity.Html do
...
-                          a.updated_at,
-                          "{Mfull} {D}, {YYYY} at {h12}:{m} {AM} {Zabbr}"
-                        )
-                    }
-                  end),
-                sectionSlug: section_slug
-              })
+              case liveview do
+                true ->
+                  PhoenixLiveReact.live_react_component("Components.AttemptSelector", %{
```
eliknebel, line 114: In an effort to eliminate duplicated ReactPhoenix/ReactLivePhoenix react component definitions like these, I created a module `OliWeb.Common.React` `lib/oli_web/common/react.ex` to handle both cases. Though currently, it only accepts a SessionContext struct I think we could expand it to just accept an `is_liveview` param as well, then we could use this single common component here instead of defining twice? What do you think?

```
defmodule OliWeb.Common.React do
  # use Phoenix.Component

  import PhoenixLiveReact

  alias OliWeb.Common.SessionContext

  def component(_ctx_or_is_liveview?, name, props, opts \\ %{})

  def component(%SessionContext{is_liveview: true}, name, props, opts),
    do: live_react_component(name, props, opts)

  def component(%SessionContext{is_liveview: false}, name, props, opts),
    do: ReactPhoenix.ClientSide.react_component(name, props, opts)

  def component(true = _is_liveview, name, props, opts),
    do: live_react_component(name, props, opts)

  def component(false = _is_liveview, name, props, opts),
    do: ReactPhoenix.ClientSide.react_component(name, props, opts)
end

```

lib/oli/rendering/alternatives/html.ex
```diff
@@ -57,19 +57,37 @@ defmodule Oli.Rendering.Alternatives.Html do
...
-            mode,
-            user,
-            section_slug,
-            alternatives_id,
-            extrinsic_read_section_fn
-          )
-      })
+      case Map.get(context, :liveview) do
+        true ->
+          PhoenixLiveReact.live_react_component("Components.AlternativesPreferenceSelector", %{
```
eliknebel, line 62: Same comment as above

eliknebel (changes requested):

lib/oli/rendering/activity/html.ex
augusto1024: Yup, that sounds great! Maybe we can change it so that it accepts a map instead of a `SessionContext`? WDYT?

eliknebel (approved):

PR #3658, opened 2023-06-07
[BUGFIX][MER-2112] Fix manage section links for instructor

augusto1024:
[Link to the story](https://eliterate.atlassian.net/browse/MER-2112)

### DIFF
--- a/lib/oli_web/live/sections/overview_view.ex
+++ b/lib/oli_web/live/sections/overview_view.ex
@@ -132,22 +132,15 @@ defmodule OliWeb.Sections.OverviewView do
       </Group>
       <Group label="Curriculum" description="Manage content delivered to students">
         <ul class="link-list">
-        <li>
-          <a target="_blank" href={Routes.instructor_dashboard_path(OliWeb.Endpoint, :preview, @section.slug, :overview)} class={"btn btn-link"}><span>Preview Course as Instructor</span> <i class="fas fa-external-link-alt self-center ml-1"></i></a>
-        </li>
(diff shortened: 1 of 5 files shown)


lib/oli_web/live/sections/overview_view.ex
```diff
@@ -132,22 +132,15 @@ defmodule OliWeb.Sections.OverviewView do
...
             <a
               target="_blank"
-              href={Routes.instructor_dashboard_path(OliWeb.Endpoint, :preview, @section.slug, :content)}
+              href={Routes.instructor_dashboard_path(OliWeb.Endpoint, :preview, @section.slug, :overview)}
               class="btn btn-link"
             ><span>Preview Course as Instructor</span> <i class="fas fa-external-link-alt self-center ml-1" /></a>
           </li>
           <li><a
-              href={Routes.page_delivery_path(OliWeb.Endpoint, :index, @section.slug)}
+              href={Routes.page_delivery_path(OliWeb.Endpoint, :index_preview, @section.slug)}
```
eliknebel, line 143: I think this should be `Routes.page_delivery_path(OliWeb.Endpoint, :index, @section.slug)` same as the URL listed in the details above. This link should be the same link that can be shared with students to access the course, or an instructor can click to enter as a student, not preview

eliknebel (changes requested):

lib/oli_web/live/sections/overview_view.ex
eliknebel: After realizing this what what this link used to be, and investigating further, it seems there is something strange about how these URLs are being redirected. This might require further investigation that we can follow up with. For now, this resolves the core of the issue here so I will approve and we can follow up with a targeted fix

eliknebel (approved):

PR #3663, opened 2023-06-08
[MER-2075] Links and Iframes from media library

marc-hughes:
Now, both links and webpages (iframes) in core authoring allow you to select a destination from the media library (including uploading files to use)

I modified the media library S3 upload to add a content-type to make viewing uploaded files in iframes work correctly, otherwise it just wants to download them. 

There is a bit of tension between the desire to have a link that downloads a file, vs. a link that opens in a new tab. Here's where I landed:

1. Links to media library uploads from the past will download (undesirable, but reasonable)
2. IFrames to media library uploads from the past will download (undesirable and useless)
3. Links to new media library items will do the browser-default behavior. Generally:
   - Images, Videos, Audio, PDF will open in a new browser tab 
   - HTML files will open in a new tab
   - Most other file types will download
4. IFrames to new media library items will do the browser-default behavior. Generally:
   - Images, Videos, Audio, PDF will display inside the iframe
   - HTML files will display in the iframe
   - Most other file types will initiate a download with nothing visible in the iframe

https://github.com/Simon-Initiative/oli-torus/assets/333265/1604e2ec-e299-4dab-bbb8-6eeb682a647a

### DIFF
--- a/assets/src/components/editing/elements/webpage/webpageActions.tsx
+++ b/assets/src/components/editing/elements/webpage/webpageActions.tsx
@@ -1,11 +1,10 @@
 import * as React from 'react';
-import { useState } from 'react';
 import { Transforms } from 'slate';
 import { createButtonCommandDesc } from 'components/editing/elements/commands/commandFactories';
-import * as Settings from 'components/editing/elements/common/settings/Settings';
-import { Modal } from 'components/modal/Modal';
 import { modalActions } from 'actions/modal';
 import { Model } from 'data/content/model/elements/factories';
+import { Webpage } from 'data/content/model/elements/types';
+import { WebpageModal } from './WebpageModal';
 
 const dismiss = () => window.oliDispatch(modalActions.dismiss());
 const display = (c: any) => window.oliDispatch(modalActions.display(c));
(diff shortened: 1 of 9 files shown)


lib/oli/authoring/media_library.ex
```diff
@@ -212,7 +212,9 @@ defmodule Oli.Authoring.MediaLibrary do
   end
 
   defp upload_file(bucket, file_name, contents) do
-    S3.put_object(bucket, file_name, contents, [{:acl, :public_read}]) |> ExAws.request()
+    mime_type = MIME.from_path(file_name)
```
marc-hughes, line 215: This defaults to application/octet-stream for unknown types, which is what we used to always use.

assets/src/components/editing/elements/webpage/webpageActions.tsx
```diff
@@ -14,75 +13,50 @@ export type WebpageCreationProps = {
...
+      });
+      dismiss();
+    };
+
+    const onCancel = () => {
+      resolve(null);
+      dismiss();
+    };
+
+    display(
```
marc-hughes, line 34: inserting an iframe behaves a bit different now.

Before it would ask for a url, now it uses the entire iframe settings modal so you can choose from media library and also set the alt text on creation.

eliknebel (approved):

PR #3665, opened 2023-06-08
[BUGFIX][MER-2099] Fix learning objectives table 

simonchoxx:
[MER-2099](https://eliterate.atlassian.net/browse/MER-2099)

This PR refactors the query used to obtain the objectives rendered in the 'learning objectives' tab for the instructor dashboard.
With the changes made, the objectives and sub-objectives will always be rendered in the appropriate columns.
If a sub-objective is added to a page in the course section, its corresponding objective will be searched for in order to display it correctly in the table.
Also a bug that was occurring in the search field when receiving a null value is corrected.

Captura de pantalla 2023-06-08 a la(s) 19 12 56


lib/oli/delivery/sections.ex
```diff
@@ -2666,21 +2669,84 @@ defmodule Oli.Delivery.Sections do
...
+    )
+    |> Map.put(
+      :student_mastery_subobj,
+      Map.get(
+        proficiency_per_learning_objective,
+        obj.subobjective_resource_id,
+        "Not enough data"
       )
-    end)
+    )
```
augusto1024, line 2749: I think this way of building the objectives/subobjectives list is too complex. We're iterating through many list many times and I think we can avoid it. I wrote a piece of codes which I think does the same but with less iterations. Let me know what you think
```elixir
    Enum.reduce(objectives, [], fn objective, acc ->
      # Only consider objectives that belong to a page
      if (Map.get(objectives_pages_map, objective.objective_resource_id)) do
        # Get all the parent objectives of the current objective
        parent_objectives = Enum.filter(objectives, fn obj ->
          obj.subobjective_resource_id == objective.objective_resource_id
        end)

        if (length(parent_objectives) == 0) do
          # If the current objective doesn't have a parent, just render it
          [objective | acc]
        else
          # If the current objective has one or more parents, render their parents
          parent_objectives ++ acc
        end
      else
        acc
      end
    end)
    |> Enum.uniq_by(&({&1.objective_resource_id, &1.subobjective_resource_id}))
    |> Enum.map(fn obj -> add_necessary_fields_to_objectives(obj, objectives_pages_map, proficiency_per_learning_objective) end)
```

augusto1024 (commented):
Tried it and seems to work fine! I left a comment with a suggestion on how we could improve the code a little bit

lib/oli/delivery/sections.ex
simonchoxx: I really like your solution. I'll take it as I think it's more performant to do it this way!

augusto1024 (approved):

lib/oli/delivery/sections.ex
```diff
@@ -2666,23 +2669,61 @@ defmodule Oli.Delivery.Sections do
...
+    Enum.reduce(objectives, [], fn objective, acc ->
+      # Only consider objectives that belong to a page
+      if Map.get(objectives_pages_map, objective.objective_resource_id) do
+        # Get all the parent objectives of the current objective
+        parent_objectives =
+          Enum.filter(objectives, fn obj ->
+            obj.subobjective_resource_id == objective.objective_resource_id
+          end)
+
+        if length(parent_objectives) == 0 do
```
eliknebel, line 2681: According to Credo:
length(parent_objectives) == 0 is expensive. Prefer Enum.empty?/1 or list == [] (warning:Credo.Check.Warning.ExpensiveEmptyEnumCheck)

Though parent_objectives should never be very long, I would prefer we use `Enum.empty?/1` here instead since it is inside a reduce fn

eliknebel (changes requested):

lib/oli/delivery/sections.ex
simonchoxx: Sure, I'll change it

PR #3696, opened 2023-06-13
[BUG FIX] [MER-2138] Fix darkmode issues

augusto1024:
[Link to the ticket](https://eliterate.atlassian.net/browse/MER-2138)

### NOTES
I logged in as admin, instructor, and student, did a pass-through with each of those roles, and checked that everything I saw was at least usable (not pretty). We probably will need to do some QA to check if everything is usable or if we're missing something.

Also, we should consider fixing the styles in the future because it's a little messy.

### DIFF
--- a/assets/styles/common/purpose.scss
+++ b/assets/styles/common/purpose.scss
@@ -201,9 +201,16 @@
     text-decoration: none;
 
     .content-purpose-content {
+
       .title {
         font-size: 1.3em;
-        color: var(--color-body-color);
+        @media (prefers-color-scheme: light) {
+          color: var(--color-body-color);
+        }
+
+        @media (prefers-color-scheme: dark) {
+          color: var(white);
+        }
       }
 
       i {
--- a/assets/styles/common/purpose.scss
+++ b/assets/styles/common/purpose.scss
@@ -204,13 +204,7 @@
 
       .title {
         font-size: 1.3em;
-        @media (prefers-color-scheme: light) {
-          color: var(--color-body-color);
-        }
-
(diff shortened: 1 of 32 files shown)


eliknebel:
Screenshot 2023-06-13 at 4 58 44 PM

I noticed drag and drop styles are not compatible with dark mode, but this is likely because the styles are embedded within the drag and drops themselves and wont be fixed by this PR

eliknebel (changes requested):
There are two areas in delivery I noticed there are still some hard to find issues:

Page links:
Screenshot 2023-06-13 at 5 00 08 PM

Popup Content:
Screenshot 2023-06-13 at 5 01 54 PM

assets/styles/common/purpose.scss
```diff
@@ -201,9 +201,16 @@
...
     .content-purpose-content {
+
       .title {
         font-size: 1.3em;
-        color: var(--color-body-color);
+        @media (prefers-color-scheme: light) {
+          color: var(--color-body-color);
+        }
+
+        @media (prefers-color-scheme: dark) {
```
eliknebel, line 211: Instead of using @media selector, can we add this to the `html.dark` rules below? This will cover both cases where dark mode is system driven and JS preference driven (both add `dark` class to html element) We can also utilize the `--color-body-color-dark` css variable that get generated by tailwind. e.g.
```
html.dark {
  ...

  .content-purpose-content {
    .title {
      color: var(--color-body-color-dark)
    }
  }

```

eliknebel (approved):

Issue #3536, opened 2023-05-17
XREFS in captions don't migrate properly


halturner:
KTH identified a cross-reference (xref) in a caption that didn't migrate properly from Echo to Tokamak. Dev team noted that the migration tool doesn't account for xrefs in captions (and possibly other locations?).

PR #3705, opened 2023-06-15
[MER-2072] Migrate xref links correctly

marc-hughes:
Legacy xref links are converted into regular 'A' links with an idref set to a page reference by the course-digest tool. Torus had two problems with the ingest of those links. 

1. The ingest process rewrites the idref attributes to a url to an ingested page. It recursively does this through the content, but was only recursively following the 'children' attribute. There can also be links in 'caption', 'pronunciation', and 'translations' of our content model.
2. anchor (new) and target (previously existed) attributes of links with an idref were dropped on ingest.

To test this, you must use the course digest with these changes in it: https://github.com/Simon-Initiative/course-digest/pull/168/files

This has been tested:
 * kth - on learns : there is a single xref in this project which is now imported correctly
 * gsu-statistics: there are many xref's here with a few different observations:
    * They all import and can be viewed in authoring correctly
    * I couldn't find some pages that exist in delivery mode. - ex. Distribution of Sample Means (1 of 4)
    * Some pages are broken in delivery mode, both with this change, and in the master branch using a master branch course-digest tool - ex. Estimating a Population Mean (2 of 3)

Fixes https://github.com/Simon-Initiative/oli-torus/issues/3536

### DIFF
--- a/assets/src/data/content/model/elements/types.ts
+++ b/assets/src/data/content/model/elements/types.ts
@@ -387,7 +387,8 @@ export type HyperlinkType = 'page' | 'url' | 'media_library';
 export interface Hyperlink extends SlateElement<Text[]> {
   type: 'a';
   href: string;
-  target: string;
+  target?: string;
+  anchor?: string;
   linkType?: HyperlinkType;
 }
 
--- a/lib/oli/interop/rewire_links.ex
+++ b/lib/oli/interop/rewire_links.ex
@@ -45,11 +45,25 @@ defmodule Oli.Ingest.RewireLinks do
   end
 
   def rewire(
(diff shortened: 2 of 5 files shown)


assets/src/data/content/model/elements/types.ts
```diff
@@ -387,7 +387,8 @@ export type HyperlinkType = 'page' | 'url' | 'media_library';
 export interface Hyperlink extends SlateElement<Text[]> {
   type: 'a';
   href: string;
-  target: string;
+  target?: string;
+  anchor?: string;
```
marc-hughes, line 391: New anchor field for in-page anchors, but it's not actually used right now.

eliknebel (approved):

PR #3706, opened 2023-06-15
[MER-2060] Add summative assessment item analytics

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-2060) to the ticket

## TODO

- [x] Add tests
- [x] Dark mode

## Demo

https://github.com/Simon-Initiative/oli-torus/assets/74839302/44eb773f-feda-46f0-aa00-a212a33ca2fe

### DIFF
--- a/lib/oli/delivery/metrics.ex
+++ b/lib/oli/delivery/metrics.ex
@@ -411,9 +411,6 @@ defmodule Oli.Delivery.Metrics do
         where:
           rev.resource_id in ^pages_ids and r_acc.section_id == ^section_id and
             not is_nil(ra.date_evaluated),
-        where:
-          rev.resource_id in ^pages_ids and
-            not is_nil(ra.date_evaluated),
         group_by: rev.resource_id,
         select: {
           rev.resource_id,
--- a/lib/oli/delivery/metrics.ex
+++ b/lib/oli/delivery/metrics.ex
@@ -1,6 +1,9 @@
 defmodule Oli.Delivery.Metrics do
   import Ecto.Query, warn: false
 
+  alias Oli.Delivery.Attempts.Core.ResourceAttempt
+  alias alias Oli.Resources.Revision
+  alias Oli.Delivery.Attempts.Core.ResourceAccess
   alias Oli.Repo
   alias Oli.Analytics.DataTables.DataTable
   alias Oli.Delivery.Attempts.Core.{ResourceAccess, ActivityAttempt}
@@ -21,31 +24,34 @@ defmodule Oli.Delivery.Metrics do
   def progress_datatable_for(section_id, container_id) do
     learner_id = ContextRoles.get_role(:context_learner).id
 
-    users = from(e in Enrollment,
-      join: ecr in assoc(e, :context_roles),
-      join: u in assoc(e, :user),
-      where: e.section_id == ^section_id,
-      where: ecr.id == ^learner_id,
-      select: u,
-      distinct: u
-    )
-    |> Repo.all()
-    |> Enum.reduce(%{}, fn user, acc -> Map.put(acc, user.id, user) end)
+    users =
+      from(e in Enrollment,
+        join: ecr in assoc(e, :context_roles),
+        join: u in assoc(e, :user),
+        where: e.section_id == ^section_id,
+        where: ecr.id == ^learner_id,
+        select: u,
+        distinct: u
+      )
+      |> Repo.all()
+      |> Enum.reduce(%{}, fn user, acc -> Map.put(acc, user.id, user) end)
 
     user_ids = Map.keys(users)
 
     progress_for(section_id, user_ids, container_id)
     |> Enum.reduce([], fn {user_id, progress}, acc ->
-      [%{
-        id: user_id,
-        name: users[user_id].name,
-        email: users[user_id].email,
-        progress: progress
-      } | acc]
+      [
+        %{
+          id: user_id,
+          name: users[user_id].name,
+          email: users[user_id].email,
+          progress: progress
+        }
+        | acc
+      ]
     end)
     |> DataTable.new()
     |> DataTable.headers([:id, :name, :email, :progress])
-
   end
 
   @doc """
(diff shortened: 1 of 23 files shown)


lib/oli/delivery/sections.ex
```diff
@@ -2699,6 +2700,60 @@ defmodule Oli.Delivery.Sections do
...
+  For example:
+
+  %{1: "Unit 1: Basics", 15: nil, 45: "Module 3: Enumerables"}
+  """
+  def map_resources_with_container_labels(section_slug, resource_ids) do
+    containers =
+      from([sr, s, spp, _pr, rev] in DeliveryResolver.section_resource_revisions(section_slug),
+        join: p in Project,
+        on: p.id == spp.project_id,
+        where: s.slug == ^section_slug and rev.resource_type_id == 2,
```
augusto1024, line 2716: WDYT about using Oli.Resources.ResourceType.get_type_by_id/1 instead of the hardcoded id?

lib/oli_web/components/delivery/scored_activities/activities_table_model.ex
```diff
@@ -0,0 +1,136 @@
...
+  defp best_effort_stem_extract(_), do: "[Empty]"
+
+  defp extract(%{"type" => "p", "children" => children}) do
+    value =
+      Enum.reduce(children, "", fn c, s ->
+        s <> Map.get(c, "text", "")
+      end)
+
+    cond do
+      String.length(value) > 75 -> String.slice(value, 0..75) <> "..."
```
augusto1024, line 128: Instead of doing this slice, WDYT about using [`text-ellipsis`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow)?

lib/oli_web/components/delivery/scored_activities/scored_activities.ex
```diff
@@ -0,0 +1,624 @@
+defmodule OliWeb.Components.Delivery.ScoredActivities do
+  use Surface.LiveComponent
```
augusto1024, line 2: Can't we make this a `Phoenix.LiveComponent` instead?

lib/oli_web/components/delivery/scored_activities/scored_activities.ex
```diff
@@ -0,0 +1,624 @@
...
+              text={@params.text_search}
+            />
+          </form>
+          {#if @current_assessment != nil}
+            <div id="student_attempts_summary" class="flex flex-row mt-auto">
+              <span class="text-xs">
+                {#if @students_with_attempts_count == 0}
+                  No student has completed any attempts.
+                {#else}
+                  {@students_with_attempts_count} {Gettext.ngettext(OliWeb.Gettext, "student has", "students have", @students_with_attempts_count)} completed {@total_attempts_count} {Gettext.ngettext(OliWeb.Gettext, "attempt", "attempts", @total_attempts_count)}.
```
augusto1024, line 189: ❤️

lib/oli_web/components/delivery/scored_activities/scored_activities.ex
```diff
@@ -0,0 +1,624 @@
...
+                    readonly
+                  />
+                  <button
+                    id="copy_emails_button"
+                    class="text-xs text-primary underline ml-auto mb-6"
+                    phx-hook="CopyListener"
+                    data-clipboard-target="#email_inputs"
+                  ><i class="fa-solid fa-copy mr-2" />{Gettext.ngettext(
+                      OliWeb.Gettext,
+                      "Copy his email address",
```
augusto1024, line 216: NIT: Could be a "she" and not a "he". Perhaps we can change it to just "Copy email address" and "Copy email addresses"?

augusto1024 (commented):
Wow, that was a long PR 😰. Amazing work! I already tested it the other day and looks amazing. Just left a couple comments.

lib/oli/delivery/metrics.ex
```diff
@@ -358,6 +361,70 @@ defmodule Oli.Delivery.Metrics do
...
+        join: rev in Revision,
+        on: ra.revision_id == rev.id,
+        join: r_acc in ResourceAccess,
+        on: ra.resource_access_id == r_acc.id,
+        where:
+          rev.resource_id in ^pages_ids and r_acc.section_id == ^section_id and
+            not is_nil(ra.date_evaluated),
+        where:
+          rev.resource_id in ^pages_ids and
+            not is_nil(ra.date_evaluated),
```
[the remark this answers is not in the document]
nicocirio: let's blame copilot for that 😅 🤣

lib/oli/delivery/metrics.ex
```diff
@@ -358,6 +361,70 @@ defmodule Oli.Delivery.Metrics do
...
 
+  @doc """
+  Calculates the average score for all students in a collection of pages
+  (only considering finished attempts).
+
+  The last parameter gives flexibility into excluding specific users
+  from the calculation. This exists primarily to exclude instructors.
+  `user_ids_to_ignore` can be an empty list.
+  """
+  def avg_score_across_for_pages(section_id, pages_ids, user_ids_to_ignore) do
```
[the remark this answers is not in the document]
nicocirio: Not yet actually. But my goal was to keep the same functionality/flexibility as `progress_across/4`

lib/oli_web/components/delivery/scored_activities/scored_activities.ex
nicocirio: We can, sure! I decided to go with this approach because within this component I need to render the `PagedTable` component, which is a `Surface.Component`. 
Although we could render it, I do not like having to pass all the component props to make it work inside a Phoenix.LiveComponent. I find it more readable/clean this way, WDYT?

augusto1024: Although I find it cleaner this way, if we want to stop using Surface I think it would be better to make it a LiveView component. We can leave it as is and change it later because it shouldn't be difficult to change it. I'll just leave it up to you to decide 😅

eliknebel (approved):

lib/oli_web/components/delivery/scored_activities/scored_activities.ex
nicocirio: makes sense, I will refactor it

PR #3722, opened 2023-06-16
[BUGFIX][MER-2158] Show always hint options

simonchoxx:
[MER-2158](https://eliterate.atlassian.net/browse/MER-2158)

This PR fixes the error that happened when rendering the hints when changing the chosen option in a MC. 

Now, the hints are always visible even if the MC option is changed.

https://github.com/Simon-Initiative/oli-torus/assets/16328384/d18651d7-ae52-4182-9c20-eced11c62872

[MER-2158]: https://eliterate.atlassian.net/browse/MER-2158?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/assets/src/components/activities/common/hints/delivery/HintsDeliveryConnected.tsx
+++ b/assets/src/components/activities/common/hints/delivery/HintsDeliveryConnected.tsx
@@ -28,7 +28,7 @@ const shouldShow = (
 ) => {
   if (graded) return false;
   if (surveyId !== null) return false;
-  if (!correct) return true;
+  if (!correct || correct) return true;
 
   return (
     (typeof shouldShow === 'undefined' || shouldShow) &&
@@ -43,9 +43,8 @@ const isRequestHintDisabled = (
   correct: boolean,
   graded: boolean,
 ) => {
-  if (!hasMoreHints) return false;
+  if (!hasMoreHints || correct || !correct) return false;
   if (isEvaluated(uiState) && graded) return true;
-  if (!correct) return false;
 
   return isEvaluated(uiState) || isSubmitted(uiState);
 };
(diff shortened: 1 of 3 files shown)


assets/src/components/activities/common/hints/delivery/HintsDeliveryConnected.tsx
```diff
@@ -28,7 +28,7 @@ const shouldShow = (
 ) => {
   if (graded) return false;
   if (surveyId !== null) return false;
-  if (!correct) return true;
+  if (!correct || correct) return true;
```
augusto1024, line 31: This always evaluates to true because `(!true || true) == (false || true) == true` so we can just replace it with `return true` and remove the other return below this one, right?

assets/src/components/activities/common/hints/delivery/HintsDeliveryConnected.tsx
```diff
@@ -43,9 +43,8 @@ const isRequestHintDisabled = (
   correct: boolean,
   graded: boolean,
 ) => {
-  if (!hasMoreHints) return false;
+  if (!hasMoreHints || correct || !correct) return false;
```
augusto1024, line 46: Similar to the comment above, this is equivalent to  `!hasMoreHints || true` which will always return true

assets/src/components/activities/common/hints/delivery/HintsDeliveryConnected.tsx
simonchoxx: You're right! Thanks!

simonchoxx:
I have made some changes so that in the correct answers the hints are still shown but no new hints are allowed to be requested.

augusto1024 (commented):
Looks good! Just one more small change I think we should do

eliknebel (approved):

PR #3726, opened 2023-06-19
[Bugfix] Fixes duplicate page on ingested projects

### LINKED ISSUE #3592: Duplicate page function not working
From support_torus channel, Guatum

**Describe the bug**
Duplicate page is not working. The issue might be intermittent or project specific.

**To Reproduce**
https://github.com/Simon-Initiative/oli-torus/assets/6248894/f6e21ca7-d720-49c9-8854-9e2cd61e52e9

**AppSignal**
https://appsignal.com/open-learning-initiative/sites/642b4d70d2a5e4452a0e5721/exceptions/incidents/127

marc-hughes:
Steps to reproduce:

1. Ingest any project
2. Go to the all-pages of that project
3. Click the (...) menu on a page and select duplicate

Expected: The page duplicates
Actual: No visual change, error in elixir console

Cause of the bug:

Oli.Authoring.Course.create_and_attach_resource eventually calls Resources.create_revision which passes the attributes passed in through a changeset function. If any of the children of those attributes are a struct and not a map, you fail with an ecto error.

This change makes sure the legacy property passed in is a map.

Fixes #3592

### DIFF
--- a/lib/oli/authoring/editing/container_editor.ex
+++ b/lib/oli/authoring/editing/container_editor.ex
@@ -365,8 +365,15 @@ defmodule Oli.Authoring.Editing.ContainerEditor do
     new_page_attrs =
       original_page
       |> Map.drop([:slug, :inserted_at, :updated_at, :resource_id, :resource])
-      |> Map.put(:title, "Copy of #{original_page.title}")
+      |> Map.put(:title, "#{original_page.title} (copy)")
       |> Map.put(:content, nil)
+      |> then(fn map ->
+        if is_nil(map.legacy) do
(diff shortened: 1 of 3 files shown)


eliknebel (approved):
Nice solution. I made some minor tweaks:
- I noticed this still wasn't working in container live view so I ported the solution over there as well
- I changed the iife expression to then (elixir recommended)
- I noticed when sorting by title, "Copy of" is prepended so the duplicate page is added but isn't obvious. I switched this so that instead (copy) is appended and page appears next to original when sorted by title

PR #3735, opened 2023-06-20
[FEATURE][MER-2160] Add index based navigation for course pages

augusto1024:
[Link to the ticket](https://eliterate.atlassian.net/browse/MER-2160)

### NOTES
I had to use a bunch of JS. We should move this view to a Phoenix LiveView in the future.

https://github.com/Simon-Initiative/oli-torus/assets/53446634/e9df406b-f702-4ba4-869a-e20fcb282b6d

### DIFF
--- a/lib/oli_web/components/delivery/page_navigator.ex
+++ b/lib/oli_web/components/delivery/page_navigator.ex
@@ -1,0 +1,91 @@
+defmodule OliWeb.Components.Delivery.PageNavigator do
+  use Phoenix.LiveComponent
+
+  alias OliWeb.PageDeliveryView
+  alias OliWeb.Router.Helpers, as: Routes
+
+  attr :page_number, :integer, required: true
+  attr :next_page, :map
+  attr :previous_page, :map
+  attr :preview_mode, :boolean, default: false
+  attr :section_slug, :string, required: true
+  attr :numbered_revisions, :list, required: true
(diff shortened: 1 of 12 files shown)


lib/oli_web/components/delivery/page_navigator.ex
```diff
@@ -0,0 +1,91 @@
...
+  attr :section_slug, :string, required: true
+  attr :numbered_revisions, :list, required: true
+  attr :show_navigation_arrows, :boolean, default: true
+  attr :id, :string, required: true
+
+  def render(assigns) do
+    assigns = assign(assigns, %{min_page: 1, max_page: assigns.numbered_revisions |> List.last() |> Map.get(:numbering_index, 1)})
+    ~H"""
+      <script>
+        function handleInputFocus_<%= @id %>(focus = true) {
```
eliknebel, line 20: can we refactor this javascript to use Phoenix Hooks in a typescript module instead of rendering directly in the page?

eliknebel (changes requested):

lib/oli_web/components/delivery/page_navigator.ex
eliknebel: Never mind, Phoenix Hooks appear to only be supported within a LiveView

eliknebel (approved):

PR #3740, opened 2023-06-20
[BUGFIX][MER-2175] Fix "Cancel" button in student onboarding wizard

augusto1024:
[Link to the story](https://eliterate.atlassian.net/browse/MER-2175)

### DIFF
--- a/lib/oli_web/live/delivery/student_onboarding/wizard.ex
+++ b/lib/oli_web/live/delivery/student_onboarding/wizard.ex
@@ -91,7 +91,7 @@ defmodule OliWeb.Delivery.StudentOnboarding.Wizard do
        current_step_name: @intro_step,
        current_step_index: 0,
        datashop_session_id: datashop_session_id,
-       is_lti: !is_nil(session["lti_params_id"])
+       is_lti: section.open_and_free == false
      )}
   end
 
--- a/lib/oli_web/live/delivery/student_onboarding/wizard.ex
+++ b/lib/oli_web/live/delivery/student_onboarding/wizard.ex
@@ -107,6 +107,7 @@ defmodule OliWeb.Delivery.StudentOnboarding.Wizard do
           module={Stepper}
           steps={@steps}
           current_step={@current_step_index}
+          on_cancel={JS.navigate(Routes.delivery_path(OliWeb.Endpoint, :open_and_free_index))}
           data={get_step_data(assigns)}
         />
       </div>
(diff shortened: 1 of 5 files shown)


lib/oli_web/live/delivery/student_onboarding/wizard.ex
```diff
@@ -107,6 +107,7 @@ defmodule OliWeb.Delivery.StudentOnboarding.Wizard do
           module={Stepper}
           steps={@steps}
           current_step={@current_step_index}
+          on_cancel={JS.navigate(Routes.delivery_path(OliWeb.Endpoint, :open_and_free_index))}
```
eliknebel, line 110: I think this makes sense for open and free/independent sections but what about LTI sections? Maybe we shouldn't show this button in that case?

eliknebel (changes requested):

lib/oli_web/live/delivery/student_onboarding/wizard.ex
augusto1024: You're right. Good catch!

lib/oli_web/live/delivery/student_onboarding/wizard.ex
```diff
@@ -90,7 +90,8 @@ defmodule OliWeb.Delivery.StudentOnboarding.Wizard do
        steps: steps,
        current_step_name: @intro_step,
        current_step_index: 0,
-       datashop_session_id: datashop_session_id
+       datashop_session_id: datashop_session_id,
+       is_lti: !is_nil(session["lti_params_id"])
```
eliknebel, line 94: `lti_params_id` is really intended to be used to facilitate the handshake and user creation after launch under the hood and may not always reliable to use.

I think a better method to determine if this is an LTI section is using the `section.open_and_free == false` (admittedly, this is a poor name for this flag since the introduction of independent sections that are not open and free) or `user.independent_learner == false` (see `lib/oli_web/components/delivery/utils.ex:72` is_independent_learner?).

I think `section.open_and_free == false` is probably best. We should eventually rename this to `section.independent`, but maybe at a later time.

eliknebel (changes requested):

eliknebel (changes requested):
Tested and works great for open and free. I wasn't able to find this screen through LTI connection, but I did stumble across a couple instructor facing views that we also need to fix.

Screenshot 2023-06-22 at 11 13 00 AM
Screenshot 2023-06-22 at 11 12 34 AM
Screenshot 2023-06-22 at 10 59 33 AM

lib/oli_web/live/delivery/student_onboarding/wizard.ex
eliknebel: On further investigation, it appears we are using this to load and use `lti_params` in various other places, and it appears to be working fine so we can just continue with that approach.

eliknebel (approved):

PR #3754, opened 2023-06-21
[MER-2060v2] Use Phoenix.LiveComponent :smile::heart:

nicocirio:
This PR replaces `Surface.LiveComponent` with `Phoenix.LiveComponent`, as discussed in [this PR](https://github.com/Simon-Initiative/oli-torus/pull/3706#discussion_r1236931051)

### DIFF
--- a/lib/oli_web/components/delivery/scored_activities/scored_activities.ex
+++ b/lib/oli_web/components/delivery/scored_activities/scored_activities.ex
@@ -1,5 +1,5 @@
 defmodule OliWeb.Components.Delivery.ScoredActivities do
-  use Surface.LiveComponent
+  use Phoenix.LiveComponent
 
   import Ecto.Query
   alias Oli.Repo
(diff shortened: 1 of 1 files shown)


lib/oli_web/components/delivery/scored_activities/scored_activities.ex
```diff
@@ -28,24 +28,6 @@ defmodule OliWeb.Components.Delivery.ScoredActivities do
...
-
-  data(table_model, :map)
-  data(total_count, :integer)
-  data(student_emails_without_attempts, :list)
-  data(students_with_attempts_count, :integer)
-  data(total_attempts_count, :integer)
-  data(current_assessment, :map)
-  data(activities, :list)
-  data(preview_rendered, :map)
-
```
augusto1024, line 48: NIT: Could we use `attr` and not remove this props? 😅

nicocirio: Yeah, that would be great. Nowadays the [attr](https://hexdocs.pm/phoenix_live_view/Phoenix.Component.html#attr/3) is supported for function components, but not yet for live components. There is [an open issue](https://github.com/phoenixframework/phoenix_live_view/issues/2407) that suggests that implementation.
A workaround could be wrapping the live_component with a function component with attrs, but [it is not recommended](https://github.com/phoenixframework/phoenix_live_dashboard/pull/394#issuecomment-1399553639) to do so

augusto1024: Oh, I didn't know that! Thanks for the info 😁

augusto1024 (approved):
🙌 🙌 🙌

eliknebel (approved):

PR #3757, opened 2023-06-22
[MER-1995] Fix explanation for multi inputs

simonchoxx:
[MER-1995](https://eliterate.atlassian.net/browse/MER-1995)

This PR fixes the issue that was happening with the explanation strategy in multi inputs that have more than one part and the "submit per part" option enabled. 
In this case, the explanation strategy will be tied to the part attempt number and not to the activity attempt number.
Therefore, the explanation will be displayed for each input separately, in case the answer is incorrect.

https://github.com/Simon-Initiative/oli-torus/assets/16328384/abae10b2-0e9e-487a-b6a0-7687c8e0101e

[MER-1995]: https://eliterate.atlassian.net/browse/MER-1995?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli/delivery/evaluation/explanation.ex
+++ b/lib/oli/delivery/evaluation/explanation.ex
@@ -4,7 +4,6 @@ defmodule Oli.Delivery.Evaluation.Explanation do
   alias Oli.Delivery.Evaluation.ExplanationContext
   alias Oli.Delivery.Attempts.Core.{ActivityAttempt, ResourceAttempt}
   alias Oli.Delivery.Evaluation.Actions.FeedbackAction
-  alias Oli.Repo
 
   @doc """
   Determines whether an explanation should be shown using the given explanation context.
@@ -13,15 +12,9 @@ defmodule Oli.Delivery.Evaluation.Explanation do
   """
   def get_explanation(
         %ExplanationContext{
-          part: part,
-          activity_attempt: %ActivityAttempt{} = activity_attempt
+          part: part
         } = context
       ) do
-    context = %{
-      context
-      | activity_attempt: activity_attempt |> Repo.preload(:part_attempts)
-    }
-
     case check_explanation_condition(context) do
       {_strategy, true} ->
         part.explanation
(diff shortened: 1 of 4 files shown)


lib/oli/delivery/evaluation/explanation.ex
```diff
@@ -12,9 +13,15 @@ defmodule Oli.Delivery.Evaluation.Explanation do
...
   def get_explanation(
         %ExplanationContext{
-          part: part
+          part: part,
+          activity_attempt: %ActivityAttempt{} = activity_attempt
         } = context
       ) do
+    context = %{
+      context
+      | activity_attempt: activity_attempt |> Repo.preload(:part_attempts)
```
eliknebel, line 22: We need to be careful about deeply nesting database preloads like this. Since this function can be called from a higher up Enum iteration, the number of sequential database calls can be `n` here, possibly hammering the database.

At a first glance, it looks like the two areas where these should be preloaded are:
1. `lib/oli/delivery/attempts/activity_lifecycle.ex:134`
2. `lib/oli/delivery/attempts/activity_lifecycle.ex:256`
3. `lib/oli/delivery/attempts/activity_lifecycle/evaluate.ex:716`
4. ...

Changing in `lib/oli/delivery/attempts/core.ex:734` could handle a few of these..

lib/oli/delivery/evaluation/explanation.ex
```diff
@@ -33,9 +40,15 @@ defmodule Oli.Delivery.Evaluation.Explanation do
...
         {:ok, %FeedbackAction{} = feedback_action},
         %ExplanationContext{
-          part: part
+          part: part,
+          activity_attempt: %ActivityAttempt{} = activity_attempt
         } = context
       ) do
+    context = %{
+      context
+      | activity_attempt: activity_attempt |> Repo.preload(:part_attempts)
```
eliknebel, line 49: Same issue above

eliknebel (changes requested):

eliknebel (approved):

PR #3762, opened 2023-06-23
[MER-1956] View course sections by institution

nicocirio:
[Link ](https://eliterate.atlassian.net/browse/MER-1956?filter=10069)to the ticket

### DIFF
--- a/lib/oli/delivery/sections/browse.ex
+++ b/lib/oli/delivery/sections/browse.ex
@@ -84,7 +84,9 @@ defmodule Oli.Delivery.Sections.Browse do
         join: e in Enrollment,
         on: e.user_id == u.id,
         join: ecr in "enrollments_context_roles",
-        on: ecr.enrollment_id == e.id and ecr.context_role_id == ^instructor_role_id and e.status == :enrolled,
+        on:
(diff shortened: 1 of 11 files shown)


eliknebel (commented):
This looks great, however I couldnt figure out how to set the institution association on an independent section. This should be required for new section creation. LTI sections should automatically set this based on their LTI registration

eliknebel:
We don’t want people entering/selecting erroneous institutions just because they have to. Having bad/incorrect data in the system is worse than none imo. Also we need to consider if we want to present a list of torus known institutions to instructors which they must select from (leaking potentially sensitive information) or an open text box. Since the entire purpose of this is to associate with a known registered institution, only the selection dropdown makes sense.

For now, we should just proceed with adding an admin facing capability to associate an independent section with an known institution_id from the “Edit Section Details” link. i.e. Have a drop down of all institutions that one can select from with one of the options being nil to support existing sections and be able to disassociate (This option can simply have placeholder value “Select an institution”). For LTI sections, this should be read-only since that will always be driven by the deployment a section is associated with

darrensiegel (approved):
Tested, works great.

PR #3774, opened 2023-06-26
[MER-2167] Summative Assessment Analytics - Part 2

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-2167?filter=10069) to the ticket

### Multiple choice
image

### Short Answer
image


lib/oli_web/components/delivery/scored_activities/scored_activities.ex
```diff
@@ -605,7 +629,93 @@ defmodule OliWeb.Components.Delivery.ScoredActivities do
...
+            Enum.reduce(activity_attempt.part_attempts, [], fn part_attempt, pa_acc ->
+              responses =
+                part_attempt.response["input"]
+                |> String.split(" ", trim: true)
+
+              responses ++ pa_acc
+            end) ++
+              acc
+          end)
+          |> Enum.frequencies()
```
augusto1024, line 650: I didn't know about this function 🙌

augusto1024 (approved):
LGTM 🚀

lib/oli_web/components/delivery/scored_activities/scored_activities.ex
```diff
@@ -605,7 +629,93 @@ defmodule OliWeb.Components.Delivery.ScoredActivities do
...
+
+        activity_attempt = hd(activity_attempts)
+
+        choices =
+          activity_attempt.transformed_model["choices"]
+          |> Enum.map(fn choice ->
+            Map.merge(choice, %{"frequency" => Map.get(frequency_per_choice_id, choice["id"])})
```
darrensiegel, line 657: This is not robust against the possibility that choices can be removed in later revisions of an activity.

darrensiegel (changes requested):
This impl can't be taken as-is as it isn't robust against the fact that activities can structurally change from revision to revision.  Consider the following scenario:

1. An MCQ is published which contains 3 choices.
2. A student takes a quiz with that MCQ and picks choice A.
3. Author revises the activity to remove choice A, republishes course.
4. Instructor goes to view the Scored Activities for that page.

The line I noted in the review breaks with:

```
key :id not found in: nil. If you are using the dot syntax, such as map.field, make sure the left-hand side of the dot is a map
```

darrensiegel:
I am retargeting this work to version 25. 

I'm concerned about the performance impact (particularly memory footprint) of fetching all activity and part attempts in the manner that this is doing - and especially landing this so close to the end of the development cycle. 

Also, I've filed: https://eliterate.atlassian.net/browse/MER-2228, which needs to be done to fix performance issues for this implementation, which exist on `master`

PR #3775, opened 2023-06-27
[BUG FIX] [MER-2200] Can't restart adaptive page when torus navigation is present

dtiwarATS:
Fixes an issue with adaptive pages, when the Display Torus Navigation check box is checked, the user cannot restart the lesson.

https://eliterate.atlassian.net/browse/MER-2200

### DIFF
--- a/assets/src/apps/delivery/layouts/deck/RestartLessonDialog.tsx
+++ b/assets/src/apps/delivery/layouts/deck/RestartLessonDialog.tsx
@@ -64,8 +64,10 @@ const RestartLessonDialog: React.FC<RestartLessonDialogProps> = ({ onRestart }) 
     }
     if (isPreviewMode) {
       window.location.reload();
-    } else {
+    } else if (window.top) {
       window.top.location.href = redirectTo;
+    } else {
+      window.location.href = redirectTo;
     }
   };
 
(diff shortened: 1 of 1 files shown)


assets/src/apps/delivery/layouts/deck/RestartLessonDialog.tsx
```diff
@@ -62,7 +62,7 @@ const RestartLessonDialog: React.FC<RestartLessonDialogProps> = ({ onRestart })
       const newAttemptUrl = `/sections/${sectionSlug}/page/${revisionSlug}/attempt`;
       redirectTo = newAttemptUrl;
     }
-    if (!graded || isPreviewMode) {
+    if (isPreviewMode) {
```
dtiwarATS, line 65: Hey @rmathis, please review.

If the lesson is not graded then just reloading the windows does not restart the lesson. the correct url logic is already in place (i.e. `newAttemptUrl` ).

assets/src/apps/delivery/layouts/deck/RestartLessonDialog.tsx
```diff
@@ -62,7 +62,7 @@ const RestartLessonDialog: React.FC<RestartLessonDialogProps> = ({ onRestart })
...
       window.location.reload();
     } else {
       window.location.href = redirectTo;
```
marc-hughes, line 68:
```suggestion
      window.top.location.href = redirectTo;
```

Let's change this to navigate the top-level window instead of the iframe to prevent that bug we talked about.

dtiwarATS: Done

rmathis (approved):

eliknebel (approved):

PR #3781, opened 2023-06-27
[MER-2118] [Bugfix] Center block formulas

marc-hughes:
Block level formulas were not centered, adjusted styles so they were again. The other piece mentioned in the bug about top/bottom margins had previously been fixed somewhere else.

image

### DIFF
--- a/assets/styles/common/elements.scss
+++ b/assets/styles/common/elements.scss
@@ -344,6 +344,13 @@ $element-margin-bottom: 1.5em;
     }
   }
 
+  .formula {
+    /* MER-2118 - Center block level formulas */
+    width: 100%;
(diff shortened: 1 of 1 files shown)


assets/styles/common/elements.scss
```diff
@@ -344,6 +344,13 @@ $element-margin-bottom: 1.5em;
     }
   }
 
+  .formula {
```
marc-hughes, line 347: I didn't touch the shared style with callouts, should they be centered too?

marc-hughes:
Just saw [MER-2121](https://eliterate.atlassian.net/browse/MER-2121) which seems similar & related, looking into that before we merge this.

PR #3795, opened 2023-06-28
[ENHANCEMENT] Revert default creds cache ttl

eliknebel:
After landing #3773, we started seeing a bunch of warnings in the build advising against this change for security reasons. This PR reverts that change. I suspect a better solution is to simply use the defaults as advised and pin down any adaptive author/player authorization issues as we find them in the test cycle.

We have adjusted several plug pipelines in the past, I suspect the timeout is no longer an issue for active sessions, but for sessions where someone is inactive for more than 30 mins, they start seeing authorization redirects returned for API calls from the adaptive player/author. We should properly handle these 403 statuses in the API client (adaptive author/player) and handle them there accordingly.

This PR also adds a system level configuration variable `ALWAYS_USE_PERSISTENT_LOGIN_SESSIONS` which when set removes the "Remember me" checkbox from login forms, effectively making all logins persistent (expire after 30 days of inactivity) by default. This is currently how LTI logins work. Checking the "Remember me" checkbox seems to be a constant point of confusion for users. Other common web services have adopted this sort of login policy as well so this option will allow the platform vendor to decide how they want user sessions to work.


lib/oli_web/pow/pow_helpers.ex
Charles-Jones-Unicon, line 1: Is credentials_cache_store supposed to be commented out?

eliknebel: Yes, I left these here for future reference but commented out

darrensiegel (approved):

PR #3796, opened 2023-06-28
[ENHANCEMENT] Add some marketing to torus homepage

eliknebel:
This PR adds some marketing materials to the torus homepage. I used ChatGPT to help create some of this content.

Screenshot 2023-06-28 at 12 44 12 PM


lib/oli_web/templates/static_page/index.html.eex
```diff
@@ -1,22 +1,85 @@
...
       <% end %>
-    </p>
-  <% end %>
 
+    </section>
+
+</div>
+<div class="bg-gray-50 dark:bg-gray-800 p-12 mt-12">
+    <section class="container mx-auto">
+      OLI Torus brings together an intuitive authoring experience, data-driven improvement, and
```
darrensiegel, line 47: I don't think we want static references to the product here.  This needs to use the vendor properties to allow folks to replace the product name.

darrensiegel (changes requested):
I don't think we want to take this for v24 without getting feedback from ASU and CMU LEs on the text here.  Can we retarget this for post v24?

lib/oli_web/templates/static_page/index.html.eex
eliknebel: Certainly. I will convert to draft for now

PR #3807, opened 2023-06-29
[BUGFIX] set width when present in model for client rendered block images [MER-2125]

darrensiegel:
This fixes a problem where block images in activities that had a specific width sets in the XML were not being rendered correctly. 

The fix was to simply set the image `width` attr when present.  Tested locally and it fixes the issue.


simonchoxx (approved):
Great!

PR #3816, opened 2023-06-30
[MER-2227] Fix activity bank in scored activities

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-2227) to the ticket

https://github.com/Simon-Initiative/oli-torus/assets/74839302/b2e4d41b-da10-4d71-95e2-39b55e007e50

This PR also addresses [this other ticket](https://eliterate.atlassian.net/browse/MER-2228)

### DIFF
--- a/lib/oli_web/components/delivery/scored_activities/scored_activities.ex
+++ b/lib/oli_web/components/delivery/scored_activities/scored_activities.ex
@@ -69,10 +69,13 @@ defmodule OliWeb.Components.Delivery.ScoredActivities do
               a.id == assessment_id
             end)
 
-          activities = get_activities(current_assessment, assigns.section.slug)
+          activities = get_activities(current_assessment, assigns.section)
 
           students_with_attempts =
-            DeliveryResolver.students_with_attempts_for_page(current_assessment.id)
+            DeliveryResolver.students_with_attempts_for_page(
+              current_assessment.id,
+              assigns.section.id
+            )
 
           student_emails_without_attempts =
             Enum.reduce(assigns.students, [], fn s, acc ->
@@ -103,7 +106,8 @@ defmodule OliWeb.Components.Delivery.ScoredActivities do
             students_with_attempts_count: Enum.count(students_with_attempts),
             student_emails_without_attempts: student_emails_without_attempts,
             total_attempts_count:
-              Enum.reduce(activities, 0, fn a, acc -> a.total_attempts + acc end)
+              Enum.reduce(activities, 0, fn a, acc -> a.total_attempts + acc end),
+            rendered_activity_id: UUID.uuid4()
           )
       end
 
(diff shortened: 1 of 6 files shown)


lib/oli_web/components/delivery/scored_activities/scored_activities.ex
```diff
@@ -520,51 +524,50 @@ defmodule OliWeb.Components.Delivery.ScoredActivities do
...
+        join: res_access in ResourceAccess,
+        on: res_attempt.resource_access_id == res_access.id,
+        where: res_access.section_id == ^section.id,
+        join: rev in Revision,
+        on: aa.revision_id == rev.id,
+        group_by: rev.id,
+        select: {rev, count(aa.id), sum(aa.score) / sum(aa.out_of)}
+      )
+      |> Repo.all()
+      |> Enum.map(fn {rev, total_attempts, avg_score} ->
```
darrensiegel, line 542: This query joins from the ActivityAttempt, through to the Revision based on the revision ID from that attempt.  It then merges in `total_attempts` and `avg_score` into that Revision.  The revisions are then used as the basis for the table model.  

This is problematic in that it incorrectly handles the case when there are more than one revision present for a single resource_id.  To replicate that situation:

1. Create an assessment with one question.  Publish the course and have one student take an attempt on that page.
2. Then, as an author, make a change to the activity (add a new choice, for instance).  Republish and have the student take a second attempt.

This UI will then incorrectly show two rows for the same activity, since this is driven from revision, and not from resource id uniqueness.

darrensiegel (changes requested):

lib/oli_web/components/delivery/scored_activities/scored_activities.ex
nicocirio: I could not reproduce the bug you mentioned, since the current_assessment already corresponds to the latest revision (assigned in the instructor_dashboard_live [here](https://github.com/Simon-Initiative/oli-torus/blob/eff16f50cf0c711c3366166497c09a004e23c3c6/lib/oli_web/live/delivery/instructor_dashboard/instructor_dashboard_live.ex#L113)).

https://github.com/Simon-Initiative/oli-torus/assets/74839302/8812b036-4e13-48f7-b00b-a72186dd77c1

nicocirio: I misunderstood your comment and finally could replicate the error. I will let you know as soon as I fix it.

lib/oli_web/components/delivery/scored_activities/scored_activities.ex
```diff
@@ -535,6 +535,9 @@ defmodule OliWeb.Components.Delivery.ScoredActivities do
         where: res_access.section_id == ^section.id,
         join: rev in Revision,
         on: aa.revision_id == rev.id,
+        left_join: rev2 in Revision,
+        on: rev.resource_id == rev2.resource_id and rev2.id > rev.id,
```
darrensiegel, line 539: This isn't correct - this will join to retrieve the latest version of this resource, which is not guaranteed to be the correct revision.  I think this query needs to group by `rev.resource_id`.  That will prevent duplicates due to differing revisions.  Upon selection of a row (i.e., an activity) this code should resolve the current version of that activity resource for that activity (regardless of how many revisions may have existed during the collection of the attempts)

darrensiegel (changes requested):

lib/oli_web/components/delivery/scored_activities/scored_activities.ex
```diff
@@ -520,51 +526,55 @@ defmodule OliWeb.Components.Delivery.ScoredActivities do
...
+        on: res_attempt.resource_access_id == res_access.id,
+        where: res_access.section_id == ^section.id,
+        join: rev in Revision,
+        on: aa.revision_id == rev.id,
+        join: pr in PublishedResource,
+        on: rev.id == pr.revision_id,
+        join: spp in SectionsProjectsPublications,
+        on: pr.publication_id == spp.publication_id,
+        where: spp.section_id == ^section.id,
+        group_by: rev.id,
```
darrensiegel, line 545: This should be changed to group by `rev.resource_id`.

darrensiegel (changes requested):
The latest changes update the query to fetch the latest published version of the activity revision, but that doesn't address the original problem when there are two attempts of the same activity across two different revisions of the activity.   The key issue here is that the query is grouping by the revision id.  See the note inline about the exact change that should be made.

darrensiegel (approved):

PR #3821, opened 2023-07-03
Add leading zero when necessary

augusto1024:
[Link to the ticket](https://eliterate.atlassian.net/browse/MER-2216)

It seems that [Float.parse/1](https://hexdocs.pm/elixir/1.12.3/Float.html#parse/1) doesn't work with values without leading zeros.

https://github.com/Simon-Initiative/oli-torus/assets/53446634/f216f106-9e34-4a78-9aab-096d531eb70f


darrensiegel (changes requested):
Can you add a unit test that demonstrates that this handles this case?

darrensiegel (approved):

PR #3822, opened 2023-07-04
[MER-2211] Unordered pages in assignment tab

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-2211) to the ticket

### DIFF
--- a/lib/oli/delivery/sections.ex
+++ b/lib/oli/delivery/sections.ex
@@ -2437,13 +2437,7 @@ defmodule Oli.Delivery.Sections do
     Repo.all(query)
   end
 
-  defp get_student_pages(section_slug, user_id, opts \\ [graded: false]) do
-    graded_filter =
-      case opts[:graded] do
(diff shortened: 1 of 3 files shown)


darrensiegel (changes requested):
This is not sufficient to fully address the problem - as it does not always render unordered pages correctly.  Consider a simply course hierarchy:

Unit 1
--Ungraded Page 1
--Quiz 1

Then, consider an unordered page "Quiz 2", which is referenced from Ungraded Page 1 via a "page_link" construct.

This implementation will end up rendering Quiz 1, and Quiz 2 - instead of the correct ordering of Quiz 2, Quiz 1.  

The missing part of this implementation is the querying and then traversal of all page to page hyperlinks and page to page "page_link" references in order to property assign the `numbering_index` for each unordered page to be the same value as the `numbering_index` of the ordered page that is referencing it.   This code has to be in both the code lines that execute when creating a course section *and* when performing a remix - so that when the content changes and when the ordering changes that these indices get updated.

darrensiegel (approved):

PR #3830, opened 2023-07-05
[BUG FIX] [MER-2204] [MER-2207] [MER-2243] Adaptive player styling fixes

marc-hughes:
When we added Tailwind, it applied the [tailwind preflight](https://tailwindcss.com/docs/preflight) global styles which broke adaptive player lessons in several places. This PR adds some global resets to the adaptive player to undo those changes and make the adaptive player look like it did before.

For all examples I looked at, the content is now matches pixel-perfect with the previous adaptive player content (some navigation UI has changed)

This fixes all reported issues except BioBeyond Genetic Blueprint, the double dropdown arrows will still appear and I'd suggest fixing this in the lesson itself.

https://eliterate.atlassian.net/browse/MER-2243
https://eliterate.atlassian.net/browse/MER-2207
https://eliterate.atlassian.net/browse/MER-2205

### DIFF
--- a/assets/css/app.css
+++ b/assets/css/app.css
@@ -8,6 +8,8 @@
 @import 'button';
 @import 'table';
 
+@import 'adaptive-reset.scss';
+
 /**
  * Automatically style all links with blue text and underline on hover.
  * External links will automatically get an arrow icon appended.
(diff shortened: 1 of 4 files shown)


eliknebel (approved):
Tested and fixes all issues identified in the original tickets. I suspect it also covers other issues which may not have been reported

PR #3841, opened 2023-07-07
[BUG FIX] Correct progress tracking in surveys [MER-2236]

darrensiegel:
Pages that only contain survey activities were not updating their progress.  Since activities within surveys are considered "unscoreable", the system needs to detect when a page *only* has unscoreable activities and mark the page progress as 100% upon visit.

In fixing this, I caught a regression where the "Enter course as a student" link had changed.  I corrected that.

### DIFF
--- a/lib/oli_web/live/sections/overview_view.ex
+++ b/lib/oli_web/live/sections/overview_view.ex
@@ -160,7 +160,7 @@ defmodule OliWeb.Sections.OverviewView do
             ><span>Preview Course as Instructor</span> <i class="fas fa-external-link-alt self-center ml-1" /></a>
           </li>
           <li><a
-              href={Routes.page_delivery_path(OliWeb.Endpoint, :index_preview, @section.slug)}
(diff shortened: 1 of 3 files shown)


lib/oli_web/live/sections/overview_view.ex
```diff
@@ -160,7 +160,7 @@ defmodule OliWeb.Sections.OverviewView do
             ><span>Preview Course as Instructor</span> <i class="fas fa-external-link-alt self-center ml-1" /></a>
           </li>
           <li><a
-              href={Routes.page_delivery_path(OliWeb.Endpoint, :index_preview, @section.slug)}
+              href={Routes.page_delivery_path(OliWeb.Endpoint, :index, @section.slug)}
```
nicocirio, line 163: This change reverts the fix made on [MER-1973](https://eliterate.atlassian.net/browse/MER-1973)

darrensiegel: Yes, we need this link to be the link that routes to actual student view, not a preview.

eliknebel (approved):

PR #3859, opened 2023-07-07
[MER-2238] Fix student proficiency in content tab

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-2238) to the ticket

### Before
image

### After
image


darrensiegel (approved):
Excellent!  This solution is great as it only does the page specific proficiency queries when the course only has pages

PR #3870, opened 2023-07-09
[ENHANCEMENT] Improve course creation [MER-2095]

darrensiegel:
This PR addresses the core issue pointed in [MER-2095], which is that the "Next" button can be scrolled off of the page, making it confusing for the user to know how to proceed to Step 2.  This PR makes a change to automatically move to Step 2 after the user 'selects' a source. 

This PR also makes a few UI adjustments to simplify some language and reduce the size of the "Select" buttons.

[MER-2095]: https://eliterate.atlassian.net/browse/MER-2095?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli_web/live/new_course/new_course.ex
+++ b/lib/oli_web/live/new_course/new_course.ex
@@ -22,9 +22,9 @@ defmodule OliWeb.Delivery.NewCourse do
 
     steps = [
       %Step{
-        title: "Select your base course project or course product",
+        title: "Select your source materials",
         description:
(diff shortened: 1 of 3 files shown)


augusto1024 (approved):
LGTM! Just pushed some changes to fix a bug when trying to go back to step 1 from step 2

PR #3874, opened 2023-07-10
[FEATURE] [MER-2258] | Adaptive Player API to share Torus section ID with a SIM

dtiwarATS:


dtiwarATS:
I am not sure on which branch this should go to hence closing. I will ask @Tess about it.

PR #3883, opened 2023-07-10
[BUG FIX] Fix counts [MER-2227]

darrensiegel:
This fixes the issues with the counts documented in https://eliterate.atlassian.net/browse/MER-2227

The counts of attempts never should have actually been the count of activity attempts, it should have simply been the count of the page (aka 'assessment') attempts. Tested this with an assessment that contains bank selections, works correctly. 

This also fixes a subtle error where the `get_activities` joined through to the ResourceAttempt on the revision id of the latest published revision.  This would lead to the dropping of results whenever this assessment gets revised, and has attempts tied to two (or more) different revisions.

### DIFF
--- a/lib/oli/delivery/metrics.ex
+++ b/lib/oli/delivery/metrics.ex
@@ -2,7 +2,6 @@ defmodule Oli.Delivery.Metrics do
   import Ecto.Query, warn: false
 
   alias Oli.Delivery.Attempts.Core.ResourceAttempt
-  alias Oli.Resources.Revision
   alias Oli.Delivery.Attempts.Core.ResourceAccess
   alias Oli.Repo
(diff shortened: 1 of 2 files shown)


nicocirio:
LGTM! 
I added a commit to match the attempts count (and other metrics) between the assessment table and the activities table (after the subtle change of criteria)

### Before

https://github.com/Simon-Initiative/oli-torus/assets/74839302/8a8662c7-6317-4062-a037-e63620dc76e9



### After

https://github.com/Simon-Initiative/oli-torus/assets/74839302/46766190-dce9-4f3b-a3d6-3abff27df2ad

nicocirio (approved):

PR #3885, opened 2023-07-10
[BUGFIX] Fix setting objectives on multi-part activities. [MER-2270]

marc-hughes:
Setting objectives on any multi-part activity was broken

To reproduce: 

1. From course project overview, enable the “Likert” activity
2. Go into the basic page page editor and create a Likert activity
3. Add a second “item” (by clicking the second “Add choice” button link, this really should read “Add item”)
4. From the Learning Objectives attachment dialog for that activity, attempt to attach LOs to each of the different parts
5. Refresh and reload the page

You will see that after reloading, both objectives are attached to the first part.

This was caused by logic meant for only saving adaptive lesson activities being applied to saving all activities.

### DIFF
--- a/assets/src/components/resource/objectives/ActivityLOs.tsx
+++ b/assets/src/components/resource/objectives/ActivityLOs.tsx
@@ -45,9 +45,9 @@ const MultiPartSelection = (props: Props & { id: string; index: number }) => {
         {...props}
         selected={props.objectives[props.id] || []}
         objectives={props.allObjectives}
-        onEdit={(objectives) =>
-          props.onEdit({ ...props.objectives, ...{ [props.id]: objectives } })
-        }
+        onEdit={(objectives) => {
+          return props.onEdit({ ...props.objectives, [props.id]: objectives });
+        }}
       />
     </div>
   );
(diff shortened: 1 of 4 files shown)


assets/src/data/persistence/activity.ts
```diff
@@ -246,7 +245,6 @@ export function edit(
       indexBibrefs(update.authoring, citationRefs);
     }
     update.content.bibrefs = citationRefs;
-    update = fixObjectiveParts(update) as ActivityUpdate & { releaseLock: boolean };
```
marc-hughes, line 249: This line was the cause of the bug, it should have been over in adaptive related saving routine instead.

assets/src/components/resource/objectives/ActivityLOs.tsx
```diff
@@ -45,9 +45,9 @@ const MultiPartSelection = (props: Props & { id: string; index: number }) => {
         {...props}
         selected={props.objectives[props.id] || []}
         objectives={props.allObjectives}
-        onEdit={(objectives) =>
-          props.onEdit({ ...props.objectives, ...{ [props.id]: objectives } })
-        }
+        onEdit={(objectives) => {
+          return props.onEdit({ ...props.objectives, [props.id]: objectives });
```
marc-hughes, line 49: Stumbled on this unrelated unnecessary object / destructure while debugging.

darrensiegel (approved):

PR #3889, opened 2023-07-11
[BUG FIX] Fix top level objectives display in dashboard reports [MER-2267]

darrensiegel:
This PR reworks the arrangement of objectives within the rows of the "Learning Objectives" reports in the instructor dashboard.  

Consider an authored project with the following objectives, noting how `Sub2` is a sub objective that appears under both of the top-level objectives:
```
TOP
 --Sub1
 --Sub2
TOP2
 --Sub2
 --Sub3
```

This PR now correctly renders this structure:
Screenshot 2023-07-11 at 9 17 08 AM

Moreover, for the rows where only a top-level objective is shown, the proficiency results are for only when that top-level objective is directly attached to activities.  For the other rows where a sub objective is displayed, the values shown for the "top level" objective are the aggregated results across all sub objectives that appear under that top-level objective.

### DIFF
--- a/lib/oli/delivery/sections.ex
+++ b/lib/oli/delivery/sections.ex
@@ -2649,7 +2649,6 @@ defmodule Oli.Delivery.Sections do
   directly attached to them.
   """
   def get_objectives_and_subobjectives(section_slug, student_id \\ nil) do
-
     calc = fn count, total ->
       case total do
         0 -> nil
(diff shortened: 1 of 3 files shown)


lib/oli/delivery/sections.ex
```diff
@@ -2631,57 +2631,123 @@ defmodule Oli.Delivery.Sections do
...
-          Metrics.proficiency_per_learning_objective(section_slug)
+          Metrics.raw_proficiency_per_learning_objective(section_slug)
 
         student_id ->
-          Metrics.proficiency_for_student_per_learning_objective(section_slug, student_id)
+          Metrics.raw_proficiency_for_student_per_learning_objective(section_slug, student_id)
       end
+      |> Enum.reduce(%{}, fn {lo_id, correct, total}, acc ->
+        Map.put(acc, lo_id, {correct, total})
+      end)
```
nicocirio, line 2670: `Metrics.raw...` could directly return the desired map

nicocirio (approved):
I added a commit to:
* directly get a map from both `Metrics.raw_proficiency...` functions
* remove unused `Metrics.proficiency...` functions
* Adapt the previous test to match the new `raw_proficiency` functions

nicocirio:
I think having different `student proficiency obj` values for the same "TOP" learning objective may be confusing for the user at first glance. Maybe we could add some explanation for that metric in the UI, wdyt? 

The explanation you gave in the PR made me finally understand the idea and get the correct interpretation of the values

```
Moreover, for the rows where only a top-level objective is shown, the proficiency results are for only when that top-level objective is directly attached to activities. For the other rows where a sub objective is displayed, the values shown for the "top level" objective are the aggregated results across all sub objectives that appear under that top-level objective.
```

image

PR #3890, opened 2023-07-11
[BUGFIX]  Image coding on graded page not evaluated [MER-2259]

andersweinstein:
This provides support for image coding on a graded page. The activity's submit button is shown; it must be clicked before page submission to finalize the image coding answer (this kicks off the needed client-side evaluation). Also adjusts conditional rendering of hints, evaluation feedback, and graded points to be appropriate for graded delivery, graded review mode, and practice page contexts, respecting new showFeedback flag. Returning to an evaluated page (as in review mode) also attempts to regenerate output from student code, since this output is not part of persistent activity state. 

There are still issues in graded context: if activity-local Submit is never used before page-wide Submit Answers, question will be marked incorrect even if code is correct and there is no indication this is the reason in the review display. And if code is never edited at all (comparable to leaving text input question blank) before page-wide Submit Answers, code input box will show up as empty on review (which at least indicates what happened). Better treatment probably requires more support from the torus framework for client-evaluated activities on graded pages. But as long as code is finalized by Submit button, image coding questions should behave reasonably.

This PR also explicitly styles the code box with a border which had disappeared at some point for clearer display.

### DIFF
--- a/assets/src/components/activities/image_coding/ImageCodingAuthoring.tsx
+++ b/assets/src/components/activities/image_coding/ImageCodingAuthoring.tsx
@@ -189,9 +189,17 @@ const ImageCoding = (props: AuthoringElementProps<ImageCodingModelSchema>) => {
 
       {!model.isExample && (
         <div>
-          {solutionParameters()}
-
           <TabbedNavigation.Tabs>
+            <TabbedNavigation.Tab label="Answer Key">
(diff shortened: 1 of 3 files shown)


andersweinstein:
This PR requests merge into master. Closed and superceded by new PR into prerelease-v.0.24

PR #3902, opened 2023-07-12
[BUG FIX] Fix auto submit [MER-2253]

darrensiegel:
This PR appears to fix a very strange bug, where during the execution of an `Oban` auto submit job, other `Oban` jobs cannot be created.  This PR also fixes two other bugs that I discovered during this work:

1. Auto submission via the delayed Oban auto submit job will still mark the attempt as being "LATE".  Auto submit is executed because the instructor has decided for that assessment to not allow late submissions, therefore, no submissions including auto submitted ones should be marked as late.  The fix here was to take into account `late_submit: :disallow` in `Settings.was_late?`. 
2. `page_delivery_controller.ex` code was still using the max number of attempts from the page revision, instead of the effective settings to determine whether or not to allow a user to start a new attempt.  This resulted in situations where the instructor could increase the number of attempts up to 20, but the student would not be able to start an attempt beyond the value found in the page revision (typically 5)

The main issue:

The original report was that grade passback was failing during an auto submission. While testing and debugging this I triggered an auto submit (by creating an assessment with a 1 minute time limit, starting an attempt and then closing the browser).  A couple of minutes later I could see the auto submit job run and logging statements from the DB calls to finalize the attempt.  BUT, I noticed nothing appeared at all in the logs related to the last two steps of this process: snapshot creation AND grade passback.  Even logging messages that I placed after these two steps didn't show up in the logs. Wrapping code with `try catch` yielded nothing. Sometimes I would also see a message in the console that a DBConnection from a particular PID had been closed/dropped. 

After banging my head against this for several hours, I hypothesized that something might be breaking when we were trying to schedule the snapshot creation Oban job (and the grade update worker job) FROM the current auto submit Oban job.  I reworked the auto submit job to synchronously create the snapshots and perform the passback and that immediately allowed things to start working again.

### DIFF
--- a/lib/oli/delivery/attempts/auto_submit/worker.ex
+++ b/lib/oli/delivery/attempts/auto_submit/worker.ex
@@ -2,11 +2,13 @@ defmodule Oli.Delivery.Attempts.AutoSubmit.Worker do
   use Oban.Worker, queue: :auto_submit, max_attempts: 5
 
   alias Oli.Delivery.Attempts.AutoSubmit.Worker
-  alias Oli.Delivery.Attempts.PageLifecycle
-  alias Oli.Delivery.Attempts.Core.ResourceAttempt
-  alias Oli.Delivery.Attempts.PageLifecycle.FinalizationSummary
+  alias Oli.Delivery.Attempts.Core.{ResourceAttempt, ResourceAccess}
+  alias Oli.Delivery.Attempts.PageLifecycle.{FinalizationSummary, FinalizationContext}
   alias Oli.Delivery.Settings
   alias Oli.Delivery.Sections
+  alias Oli.Delivery.Attempts.PageLifecycle.Graded
+
+  require Logger
 
   @moduledoc """
   An Oban worker driven page attempts auto submission creator.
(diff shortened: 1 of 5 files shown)


lib/oli/delivery/attempts/auto_submit/worker.ex
```diff
@@ -2,35 +2,119 @@ defmodule Oli.Delivery.Attempts.AutoSubmit.Worker do
...
+                )
+
+                Logger.info("Auto submit: gradepassback FINISHED for guid #{resource_attempt_guid} and section slug #{section_slug}")
+
+                result
+              else
+                Logger.info("Auto submit: gradepassback NOT ENABLED for guid #{resource_attempt_guid} and section slug #{section_slug}")
+
+                :ok
+              end
```
augusto1024, line 104: Should we still send the score to the lms in this case?

darrensiegel: Yes, this was intentional.  Imagine the situation where auto submit finalizes the attempt BUT then the grade passback fails.  This will fail the entire job, which will be retried.  In the second attempt we will get the `{:error, :already_submitted}` when we call `finalize`, which we handle, and then take another crack at grade sync.

augusto1024 (approved):
I spent some time looking at the previous code and debugging it but I couldn't figure out what was happening either. Doing everything in this same job makes sense to me

PR #3904, opened 2023-07-12
[BUGFIX][MER-2209] Reingest project does not retain objectives hierarchy

simonchoxx:
[MER-2209](https://eliterate.atlassian.net/browse/MER-2209)

This PR fixes the issue that happened when ingest a project and build the objectives hierarchy. 

With this fix, the objectives and their sub-objectives added to the ingested project are rendered correctly as their hierarchy specifies.

- **Ingest Project**

https://github.com/Simon-Initiative/oli-torus/assets/16328384/ed0c0a44-f882-4574-b5c5-bffbdb5f943a

- **V2 Ingest Project**

https://github.com/Simon-Initiative/oli-torus/assets/16328384/5b9b2a0a-90a0-432a-bc79-43bfeeb75c4a

[MER-2209]: https://eliterate.atlassian.net/browse/MER-2209?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli/interop/export.ex
+++ b/lib/oli/interop/export.ex
@@ -61,8 +61,8 @@ defmodule Oli.Interop.Export do
         tags: transform_tags(r),
         unresolvedReferences: [],
         content: %{},
-        objectives: [],
-        children: Enum.map(r.children, fn id -> "#{id}" end)
+        objectives: Enum.map(r.children, fn id -> "#{id}" end),
+        children: []
       }
       |> entry("#{r.resource_id}.json")
     end)
--- a/test/oli/interop/digest/22.json
+++ b/test/oli/interop/digest/22.json
@@ -1,11 +1,8 @@
 {
-  "children": [
-    "26",
-    "25"
-  ],
+  "children": [],
   "content": {},
   "id": "22",
-  "objectives": [],
+  "objectives": ["26", "25"],
   "originalFile": "",
   "tags": [],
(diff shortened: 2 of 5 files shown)


augusto1024 (approved):
🚀 🚀 🚀

darrensiegel (changes requested):
So it looks like this PR changes one of the "ingest" processors to read "children" instead of "objectives" so that it matches how the export step works. 

This would unfortunately break Torus course digests that are produced by the course digest tool (https://github.com/Simon-Initiative/course-digest).  That tool populates the `objectives` attribute of a top-level objective with the ids of its sub objectives. 

I think then the change that we need is to change the export.ex so that it outputs sub objectives to "objectives".  This is what that code looks like right now:

```
 |> Enum.map(fn r ->
      %{
        type: "Objective",
        id: Integer.to_string(r.resource_id, 10),
        originalFile: "",
        title: r.title,
        tags: transform_tags(r),
        unresolvedReferences: [],
        content: %{},
        objectives: [],
        children: Enum.map(r.children, fn id -> "#{id}" end)
      }
      |> entry("#{r.resource_id}.json")
    end)
``` 

Then we just need to ensure that both Ingest versions read from "objectives"

darrensiegel (approved):

PR #3905, opened 2023-07-12
[BUG FIX] Contained pages rebuild migration [MER-2305]

darrensiegel:
This PR adds a DB migration to rebuild the contained_pages relation for all course sections. 

I optimized its impl so that it is pulling far smaller payloads with its internal queries.

### DIFF
--- a/priv/repo/migrations/20230712163553_rebuild_contained_pages.exs
+++ b/priv/repo/migrations/20230712163553_rebuild_contained_pages.exs
@@ -1,0 +1,17 @@
+defmodule Oli.Repo.Migrations.RebuildContainedPages do
+  use Ecto.Migration
+  import Ecto.Query, warn: false
+
+  alias Oli.Repo
+
+  def change do
+      from(s in "sections",
+      select: %{id: s.id, slug: s.slug, root_section_resource_id: s.root_section_resource_id}
+    )
+    |> Repo.all()
+    |> Enum.each(fn s ->
+      Oli.Delivery.Sections.rebuild_contained_pages(s)
+    end)
+
+  end
+end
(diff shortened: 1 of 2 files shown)


nicocirio:
I was wondering if we need to run the migration for every section,  or if doing so for those sections that are not present in the conained_pages table is enough

darrensiegel:
> I was wondering if we need to run the migration for every section, or if doing so for those sections that are not present in the conained_pages table is enough

For Argos Prod, this will be every section (since none of the ContainedPage infrastructure exists there, whatsoever), so this would only optimize for some OLI servers, which do not have a lot of sections to begin with.

priv/repo/migrations/20230712163553_rebuild_contained_pages.exs
```diff
@@ -0,0 +1,17 @@
...
+
+  alias Oli.Repo
+
+  def change do
+      from(s in "sections",
+      select: %{id: s.id, slug: s.slug, root_section_resource_id: s.root_section_resource_id}
+    )
+    |> Repo.all()
+    |> Enum.each(fn s ->
+      Oli.Delivery.Sections.rebuild_contained_pages(s)
```
eliknebel, line 13: I wonder if we should copy+paste this `rebuild_contained_pages` implementation directly into this migration so that if we have to ever change this function in the future it doesn't affect this migration?

darrensiegel: I think in this case we will always want this migration to run whatever is the latest `rebuild_contained_pages`, regardless of when the migration is run.

eliknebel (approved):

PR #3907, opened 2023-07-12
[BUG FIX] [MER-2312] Enter course as student doesn't work

eliknebel:
There were a couple issues related to manage links to Preview as instructor and Enter course as student around the different roles in the system. This PR attempts to fix those issues by:

1. "Preview as Instructor" always takes Instructors and System Admins to the instructor preview overview
2. Removes "Enter course as student" link, as this was not useful to instructors or system admins
3. Fixes a couple other issues identified through testing:
  - Sign out was broken for system admins in instructor dashboard
  - Various buttons were siblings with block elements, making margins not render correctly 

Screenshot 2023-07-12 at 1 06 03 PM

As a side note: the copy button does not work inside LTI iframes, as it would require a clipboard permission that, at least when tested in canvas, doesn't have.

### DIFF
--- a/lib/oli_web/components/delivery/user_account_menu.ex
+++ b/lib/oli_web/components/delivery/user_account_menu.ex
@@ -3,6 +3,7 @@ defmodule OliWeb.Components.Delivery.UserAccountMenu do
 
   import OliWeb.Components.Delivery.Utils
 
+  alias Oli.Accounts.{Author, SystemRole}
   alias OliWeb.Router.Helpers, as: Routes
   alias Oli.Delivery.Sections
   alias OliWeb.Common.SessionContext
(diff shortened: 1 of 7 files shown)


lib/oli_web/components/delivery/user_account_menu.ex
```diff
@@ -106,4 +107,16 @@ defmodule OliWeb.Components.Delivery.UserAccountMenu do
       </div>
     """
   end
+
+  defp signout_path(%SessionContext{user: user_or_admin}) do
```
eliknebel, line 111: Fixes sign out for system admins in instructor dashboard

lib/oli_web/live/common/properties/read_only.ex
```diff
@@ -23,6 +24,21 @@ defmodule OliWeb.Common.Properties.ReadOnly do
     """
   end
 
+  defp render_property(%{show_copy_btn: true} = assigns) do
+    copy_id = "copy-#{UUID.uuid4(:hex)}"
+
+    ~F"""
+    <div class="relative">
+      <input id={copy_id} class="form-control" type={@type} disabled value={@value}/>
```
eliknebel, line 32: Adds Copy button to readonly URL field to making it easier to open this link

lib/oli_web/router.ex
```diff
@@ -997,7 +997,7 @@ defmodule OliWeb.Router do
       :pow_email_layout
     ])
 
-    live("/:section_slug", Sections.OverviewView)
+    get("/:section_slug", PageDeliveryController, :index)
```
eliknebel, line 1000: There is logic to handle instructors vs students in PageDeliveryController index, we should be routing all base section/<slug> requests through that so there is only a single codepath for all users entering a section.

Previously, this would show the Manage LiveView, which wasn't being wrapped in the dashboard layout and isn't applicable to students

lib/oli_web/live/common/properties/read_only.ex
darrensiegel: Very nice addition!

darrensiegel (approved):

PR #3910, opened 2023-07-13
[MER-2248][IMPROVEMENT] Add enrollment button to student details

augusto1024:
[Link to the story](https://eliterate.atlassian.net/browse/MER-2248)

### DIFF
--- a/lib/oli_web/components/delivery/actions/actions.ex
+++ b/lib/oli_web/components/delivery/actions/actions.ex
@@ -8,6 +8,7 @@ defmodule OliWeb.Components.Delivery.Actions do
   alias Phoenix.LiveView.JS
   alias Oli.Delivery.Paywall
   alias Oli.Delivery.Paywall.Payment
+  alias OliWeb.Router.Helpers, as: Routes
 
   prop(enrollment_info, :map, required: true)
   prop(section, :map, required: true)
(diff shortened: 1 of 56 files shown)


lib/oli_web/components/delivery/actions/actions.ex
```diff
@@ -92,10 +105,35 @@ defmodule OliWeb.Components.Delivery.Actions do
               </button>
           </div>
         {/if}
+
+        <div class="ml-auto">
+          <button phx-click={JS.push("open", target: "#unenroll_user_modal")} class="btn btn-danger">Unenroll</button>
+        </div>
```
simonchoxx, line 111: Looks good! But... What do you think about adding some text on the left side of the button so that it is not loose on the right side? 
I'm thinking of some text like the one that has the functionality to change the role of the user above the unenroll button.

test/oli_web/live/delivery/student_dashboard/components/actions_tab_test.exs
```diff
@@ -109,6 +111,59 @@ defmodule OliWeb.Delivery.StudentDashboard.Components.ActionsTabTest do
...
+          section.slug,
+          :manage
+        )
+      )
+
+      assert Enrollment
+             |> where([e], e.user_id == ^student.id and e.section_id == ^section.id)
+             |> limit(1)
+             |> Oli.Repo.one()
+             |> Map.get(:status) == :suspended
```
simonchoxx, line 165: I'm not sure, but maybe this piece of test could be moved to the `sections_test.exs` module where there are some tests referring to enroll and unenroll actions (I don't know if this has been tested yet).

simonchoxx (commented):
Nice work 👏👏!! Just left you some small comments

simonchoxx:
@augusto1024 Just to let you know, I found an issue in the student's actions tab liveview when the following behaviour happened:

1 - The instructor unenroll a student
2 - The instructor filters the list of students by `suspended` and re-enters the profile of the previously unenrolled student.
3 - The instructor goes to the actions tab and the issue is happening right now.

test/oli_web/live/delivery/student_dashboard/components/actions_tab_test.exs
augusto1024: As discussed offline, I'm only querying the database to see if the user interaction actually made the user unenrolled, so it's ok for this to be here 😁

darrensiegel (approved):

PR #3913, opened 2023-07-13
[ENHANCEMENT] Adjust content browser for when only pages are present [MER-2319]

darrensiegel:
The content browser (both student and instructor facing) does not handle well courses that *only* contain pages.  These pages appear almost as "units" that the user can navigate between via the left and right arrow, but the user cannot click to access, nor see any page pertinent information such as due dates.  

The PR corrects that by detecting this case and just listing the pages, as pages.

For example:
Screenshot 2023-07-13 at 2 24 57 PM

There is also a change here to handle the root path for the dashboard explicitly and redirect to the "overview" route.  This fixes a bug where an Internal Server Error was being thrown when trying to hit this route directly.

### DIFF
--- a/lib/oli_web/components/delivery/course_content.ex
+++ b/lib/oli_web/components/delivery/course_content.ex
@@ -17,6 +17,30 @@ defmodule OliWeb.Components.Delivery.CourseContent do
   attr(:is_instructor, :boolean, default: false)
   attr(:preview_mode, :boolean, default: false)
 
+  def adjust_hierarchy_for_only_pages(hierarchy) do
+    case Enum.all?(hierarchy["children"], fn child -> child["type"] == "page" end) do
+      true ->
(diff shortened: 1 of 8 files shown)


lib/oli_web/components/delivery/course_content.ex
```diff
@@ -17,6 +17,30 @@ defmodule OliWeb.Components.Delivery.CourseContent do
...
+              "slug" => nil,
+              "title" => "Curriculum",
+              "type" => "container",
+              "children" => hierarchy["children"]
+            }
+          ]
+        }
+
+      _ -> hierarchy
+    end
```
augusto1024, line 41: Shouldn't we also consider the case where the course has units/modules AND pages?

eliknebel (approved):

PR #3945, opened 2023-07-18
[ENHANCEMENT] track dirty state of input for onBlur submission [MER-2344]

darrensiegel:
This PR adds "isDirty" tracking to the "submitPerPart" mode of the MultiInput activity.  This changes the behavior so that losing focus only submits the part when there had been a content change made to the input.


assets/src/components/activities/multi_input/MultiInputDelivery.tsx
```diff
@@ -231,6 +239,10 @@ export const MultiInputComponent: React.FC = () => {
           deferredSaves.current[id].save(fn);
         }
       }
+
+      if (!isInputDirty[id]) {
```
marc-hughes, line 243: There is a case on line 234 that saves the part, this looks like it will set it to dirty anyways.

Probably not a big deal, but if we care, could move this up to the top and have handlePerPartSubmission clear the flag on save instead of it being after each call.

PR #3969, opened 2023-07-19
[BUGFIX][MER-2298] Disable sort by published date in add materials modal

augusto1024:
[Link to the ticket](https://eliterate.atlassian.net/browse/MER-2298)

### NOTES
The issue is not because of the double click. It's because the sorting by published date wasn't being handled. I removed the ability to sort by this attribute because all revisions for a given publication are going to have the same publication date so it doesn't make sense to sort them by it.


simonchoxx (approved):
LGTM 🚀🚀

PR #4006, opened 2023-07-24
[IMPROVEMENT][MER-2332] Improve product and section creation

augusto1024:
[Link to the story](https://eliterate.atlassian.net/browse/MER-2332)

### DIFF
--- a/lib/oli/delivery/sections.ex
+++ b/lib/oli/delivery/sections.ex
@@ -831,6 +831,20 @@ defmodule Oli.Delivery.Sections do
   end
 
   @doc """
+  Creates multiple section resources.
+  ## Examples
+      iex> bulk_create_section_resource([%{slug: slug_value_1, ...}, %{slug: slug_value_2, ...}])
+      {2, [%SectionResource{}, %SectionResource{}]}
+  """
+  def bulk_create_section_resource(_section_resource_rows, _opts \\ [])
+
+  def bulk_create_section_resource([], _opts), do: {0, []}
+
+  def bulk_create_section_resource(section_resource_rows, opts) do
+    Repo.insert_all(SectionResource, section_resource_rows, returning: opts[:returning] || true)
+  end
+
+  @doc """
   Updates a section resource.
   ## Examples
       iex> update_section_resource(section, %{field: new_value})
@@ -842,6 +856,24 @@ defmodule Oli.Delivery.Sections do
     section
     |> SectionResource.changeset(attrs)
     |> Repo.update()
+  end
+
+  @doc """
+  Updates multiple section resources.
+  ## Examples
+      iex> bulk_update_section_resource(section, %{field: new_value})
+      {2, [%SectionResource{}, %SectionResource{}]}
+  """
+  def bulk_update_section_resource(_section_resource_rows, _opts \\ [])
+
+  def bulk_update_section_resource([], _), do: {0, []}
+
+  def bulk_update_section_resource(section_resource_rows, opts) do
+    Repo.insert_all(SectionResource, section_resource_rows,
+      returning: opts[:returning] || true,
+      on_conflict: {:replace, [:children]},
+      conflict_target: [:id]
+    )
   end
 
   @doc """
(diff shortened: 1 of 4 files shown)


lib/oli/delivery/sections.ex
```diff
@@ -1205,69 +1296,78 @@ defmodule Oli.Delivery.Sections do
...
-      end)
 
-    %SectionResource{id: section_resource_id} =
-      Oli.Repo.insert!(%SectionResource{
+    # Return the section resource for the revision along with all the section resources
+    # for the revision's children
+    section_resources = [
+      # The below is necessary because Repo.insert_all/3 receives a map, not a struct
+      # The below is necessary because Repo.insert_all/3 doesn't autogenerate values
+      # TODO: CAMBIAR ESTO
```
darrensiegel, line 1349: TODO?

darrensiegel (changes requested):
There is a problem here with how this PR attempts to bulk allocate slugs for the SectionResources.  Lines 1204-1211 in `sections.ex` passes in a list of all of the titles of all of the revisions to `Slug.generate`.   This code assumes that titles are unique.  In many course projects, they will not be unique (e.g. imagine a graded page in Unit 1 titled "End of Unit Quiz", and the same title used for a graded page in Unit 2, Unit 3, etc).  

`Oli.Utils.Slug.unique_slugs/5` zips the titles and their slugs into a map, keyed on title. This means that we will only get one slug per unique title, which is incorrect.  We need a slug per each unique resource.  So fundamentally this function is broken and should not be used as-is. 

That being said, I don't believe that we are even using the `:slug` field from SectionResource.  Even with this slug problem present, I can still create a course section just fine where I have two resources with the same title.  

So I think what we should do here is **remove** the slug field from SectionResource schema and from and write a migration to remove that field. Then we can remove the "Bulk Slug creation code"

augusto1024:
@darrensiegel I tried to remove the `slug` field from the section resources but when I ran the tests I had 700+ failed tests. It seems we're using it quite a bit and might be risky to remove them. Are you ok if I just modify the function for querying multiple slugs so that there are no collisions?
image

darrensiegel:
Ok, let's leave the `slug` field in.  

One easy way to fix this would be use `slug_with_prefix` instead of `slugify` inside of the call to `generate`.  The randomness will help ensure that the slugs generated are unique, even when duplicate titles are present. 

You'll also want to change `unique_slugs` to return the list of `candidates` instead of putting it into a map

augusto1024:
> You'll also want to change `unique_slugs` to return the list of `candidates` instead of putting it into a map

But in that case, how would you know which slug corresponds to which resource?

darrensiegel:
The changes present in this branch to `Oli.Utils.Slug` are not correct in that they do not guarantee unique slugs.  For instance, if we use this call with duplicated titles, we get duplicated slugs:

```
Slug.generate("section_resources", ["title one", "title one"]) |> IO.inspect()
```

The above outputs:

```
%{"title one" => ["title_one", "title_one"]}
```

To fix this, and to simplify the interface between `Oli.Utils.Slug` and client code, I would suggest doing this:

1.  Instead of a list of string titles as a parameter, change the `when is_list` version of `Oli.Utils.Slug.generate` to expect a list of maps of the form `%{id: <unique integer identifier>, title: <string title>}`.   Change the return value to be a simple list, where the ordinal position of the list element in the return list corresponds to the same ordinal position of the input list. 
2. Then, within `generate`, map those string titles to slugs, by prepending a string representation of the id as well as a  string representation of a random 5 alphanumeric characters to the title and then calling `slugify` on each.  This will look like:

```
<string id>_<random 5 characters>_<the title>
```

Calling `slugify` on the above absolutely guarantees that we get uniqueness "within the set of candidates".  Including the random 5 characters gives fairly good chance that we have no collisions between any of our candidates and *existing* slugs in the target table.  

If the check against the database fails, try again, but with a new set of candidate slugs (where we get all new random characters).  If we exhaust the number of attempts we must return `{:error, :failed_to_generate_slugs}` otherwise we return back to the caller `{:ok, list_of_slugs}`.  

The calling client code just has to pick a unique ID to use to pass in as the `id` field.  Ideally, this would be the SectionResource ID, but if we haven't created the SR yet we won't have it available.  Instead use the Revision ID

augusto1024:
@darrensiegel I took some of your suggestions and went in a slightly different path. It felt weird to have to pass a map to the `Oli.Utils.Slug.generate` function, so I'm still passing a list of titles but now I'm making sure that there are no repeated candidates in the "candidates list". I'm also returning a list as you said and I'm doing the mapping outside of the `generate` function so I know which slug corresponds to which revision.

The `generate` function behaves like this now:
```elixir
iex > Oli.Utils.Slug.generate(:section_resources, ["title_1", "title_2",  "title_1"])
["title_1", "title_2", "title_1_cj461"]
```

Let me know what you think

lib/oli/utils/slug.ex
```diff
@@ -115,12 +115,38 @@ defmodule Oli.Utils.Slug do
...
-  Generates a unique slug for the table using the title provided
+  Generates a unique slug or slugs for the table using the title or titles provided
   """
+  def generate(table, titles) when is_list(titles) do
+    # Ensure no repeated titles
+    {titles, _} =
+      Enum.reduce(titles, {[], 1}, fn title, {titles, index} ->
+        slugified_title = slugify(title)
+
+        case Enum.member?(titles, slugified_title) do
```
darrensiegel, line 126: There is some `n^2` behavior that we should remove here.  We need to rework this `Enum. reduce` so that inside of the loop we do not make a call to `Enum.member`.  Instead, have the initial value of the reduce be `{[], MapSet.new(), 1}`, and add each unique title to the mapset as we go.

darrensiegel (changes requested):
Ok, this approach looks good, let's just remove the n^2 behavior noted here.

lib/oli/utils/slug.ex
augusto1024: I thought about using a MapSet but when I transform that MapSet into a list, the order is not the same as the input list. I'm going to think of a way to make it o(n) and guarantee that the order of the output list is the same one as the input one

augusto1024:
@darrensiegel @eliknebel I'm running the tests from this branch locally and they're passing. The tests that are failing are from `test/oli_web/pow/pow_test.exs` (maybe an issue with the new version of LiveView?

augusto1024:
@eliknebel I don't remember but I'll try branching off of master, cherrypick the commits from this branch and see if it fixes the issue. Thanks for the tip 🙌

PR #4024, opened 2023-07-26
[MER-2167] Summative assessment analytics p2

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-2167?filter=10069) to the ticket

### Multiple choice
image

### Multiple choice (with blank attempts)
image

image

### Short Answer
image

### DIFF
--- a/assets/src/components/activities/common/choices/authoring/ChoicesAuthoring.tsx
+++ b/assets/src/components/activities/common/choices/authoring/ChoicesAuthoring.tsx
@@ -40,49 +40,71 @@ export const Choices: React.FC<Props> = ({
 }) => {
   return (
     <>
-      <Draggable.Column items={choices} setItems={setAll}>
-        {choices.map((choice) => (
-          <Draggable.Item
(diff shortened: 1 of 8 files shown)


darrensiegel:
For v25 we are going to need a more complete solution that allows aggregate response reporting across many different activity types, in both a performant and revision-aware implementation.  I've started some design work in that space, but it isn't ready yet.  So let's close this PR and wait until we build the new impl.

nicocirio (commented):
This looks great! I think the error messages beside the corresponding input are much more clear than the error flash message.
I've been playing around with the form and I *might* have found a bug (not 100% sure). Should an admin be allowed to change a student email to an already existing email (used by another independent learner) if the current student is not an independent learner?
In the following example `student3@test.com` has already been taken by another independent learner, and I want to use the same email but for a non independent learner

https://github.com/Simon-Initiative/oli-torus/assets/74839302/5d5d4f39-2b3c-42ab-91f9-08990ba2ef68

PR #4051, opened 2023-07-28
[Bugfix] ctrl-z undo fixes [MER-2282, MER-2284]

marc-hughes:
This fixes a few instances of performing an action, and then immediately pressing ctrl-z to undo that action. Unfortunately, this is a tricky bug to solve that doesn't have a single global answer. The complexity stems from:

* There are multiple slate editors on a page
* Some elements in those editors have embedded slate editors
* Each of those editors maintains it's own undo history

Because of this, it matters where user focus is for what will happen when you attempt to undo. For some things, like deleting a speaker in a dialog, or removing a row in a table, the element with focus is removed, and a default focus to the page body is set. 

This fix manually sets focus on a relevant element after certain operations are performed. This lets you immediately ctrl-z and get the desired behavior. However, if you move your cursor to a different editor before trying to undo, you'll end up undoing that editor's last action, and not the last thing you did.

These actions should now be undoable:

Dialog:
Adding, removing, switching speakers.
Adding or removing a dialog line

Conjugations:
Adding or removing columns or rows.
Setting which audio clip plays in a pronunciation.
Changing table formatting, such as alternating row color or border style.


darrensiegel (approved):

eliknebel:
Due to CI issues, I had to squash and merge in https://github.com/Simon-Initiative/oli-torus/pull/4067

PR #4055, opened 2023-07-31
[MER-2415] Fix due date on assessment settings

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-2415) to the ticket


darrensiegel (approved):
Works well, also in the case of editing an assessment that does not have a due date to make it have a due date.

PR #4060, opened 2023-07-31
[MER-2425] Phoenix storybook

eliknebel:
Adds phoenix storybook for server side component library future development.

Storybook is accessible in a dev environment via [http://localhost/dev/storybook](http://localhost/dev/storybook)

Torus storybook components and docs can be created [according to the docs](https://hexdocs.pm/phoenix_storybook/PhoenixStorybook.html). Another great example is [https://phx-live-storybook-sample.fly.dev/storybook](https://phx-live-storybook-sample.fly.dev/storybook), GitHub: https://github.com/phenixdigital/phoenix_storybook_sample


darrensiegel (changes requested):
Everything looks good, but I see to access this I hit `/storybook`.  

But it seems like we instead should route it through `/dev/storybook`. And also ensure that it is a dev available route only? (like what we do with `/dev/sent_emails`, for instance)

eliknebel:
Ah, good idea. Changed the route to be dev only

darrensiegel (approved):

PR #4064, opened 2023-08-01
[BUGFIX][MER-2429] Fix section creation from product

augusto1024:
[Link to the ticket](https://eliterate.atlassian.net/browse/MER-2428)


eliknebel (dismissed):
Tested and appears to work great, except I am seeing an issue where a section created from a product doesn't have the same defaults as a section created from project. Here the check boxes "Registration Open" and "Requires Enrollment" should be checked. I tested this on stellarator and it appears to do the correct thing there so I believe it is related to these changes.

Screenshot 2023-08-02 at 11 53 50 AM

darrensiegel (approved):

darrensiegel (approved):

PR #4065, opened 2023-08-01
[BUGFIX]  fix deletion of hints on multipart activities [MER-2432]

andersweinstein:
Utility function Hints.removeHint used by hint editing components was not coded to handle multi-part items, causing bad behavior on author deletion of a hint from any multi-part activity. This recodes it to take partId parameter and use appropriate JSON path operations to operate on the specified part's hint list only.


eliknebel (changes requested):
I tested this locally and works great for the case of new multi inputs, however it doesn't seem to handle cases where an existing multi input that has already been modified given the previous bug and tries to delete one of the duplicated hints.

I think we also need to consider this case where someone has tried to remove and hint with the old version, putting the multi input into an erronous state, this fix should re-stablize the content when a hint is removed in those cases so that the activity can return to a normal functioning state

eliknebel:
The issue turned out to be a React rendering issue related to duplicate keys. There is no low hanging easy fix for this, so it makes sense to take this as is and manually fix existing problematic activities or implement a different fix in the future for those

eliknebel (approved):

PR #4068, opened 2023-08-02
[FEATURE][MER-1912] Add ability to enroll users

augusto1024:
[Link to the ticket](https://eliterate.atlassian.net/browse/MER-1912)

https://github.com/Simon-Initiative/oli-torus/assets/53446634/9ed80ef9-154e-4cc5-8bab-ed5bda893ba2

### DIFF
--- a/lib/oli_web/live/sections/enrollments/enrollments_view.ex
+++ b/lib/oli_web/live/sections/enrollments/enrollments_view.ex
@@ -143,7 +143,7 @@ defmodule OliWeb.Sections.EnrollmentsViewLive do
             case @add_enrollments_step do
               :step_1 -> JS.push("add_enrollments_go_to_step_2")
               :step_2 -> JS.push("add_enrollments_go_to_step_3")
-              :step_3 -> nil # Add enrollments and send invitations
+              :step_3 -> JS.dispatch("click", to: "#add_enrollments_form button")
             end
           }
           on_confirm_label={if @add_enrollments_step == :step_3, do: "Confirm", else: "Next"}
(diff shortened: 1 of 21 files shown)


lib/oli_web/templates/email/enrollment_invitation.text.eex
nicocirio, line 1: In case the receiver does not support HTML emails, this text email will be the one to be shown, so I think we should replicate the HTML text content here. 
When we see emails in dev (https://localhost/dev/sent_emails), we can see both HTML and text examples rendered

image

image

nicocirio (commented):
Really loved the UI for adding emails!! 😍 👏 👏

lib/oli_web/templates/email/enrollment_invitation.text.eex
augusto1024: I just tested this and it's also working for the enrollments emails so I think we're good 😁 

image

lib/oli_web/live/sections/enrollments/enrollments_view.ex
```diff
@@ -121,9 +133,36 @@ defmodule OliWeb.Sections.EnrollmentsViewLive do
 
   def render(assigns) do
     ~F"""
-    <div class="container mx-auto">
-
-      <div class="d-flex justify-content-between">
+    <div id="enrollments_view" class="container mx-auto">
+      {#if @is_admin}
```
darrensiegel, line 137: In addition to only showing this when the user is an Admin, we only want to show this feature for Direct Delivery course sections.  LTI course sections cannot support this email based enrollment.

darrensiegel (changes requested):
This is some really impressive work, from the slick working UX to the impl itself. I had one inline comment and one request: Can we change the default of "Add as role" from "instructor" to "student"?  This will help prevent folks from accidentally adding users as instructors.

lib/oli_web/live/sections/enrollments/enrollments_view.ex
augusto1024: Done!

darrensiegel (approved):

PR #4069, opened 2023-08-02
[FEATURE] [MER-1908] Transfer student data to another section

simonchoxx:
[MER-1908](https://eliterate.atlassian.net/browse/MER-1908)

This PR adds the possibility for an admin user to transfer the enrollment of a student from one course section to another course section.

This functionality can be found in the `actions` tab within the `student dashboard`.

https://github.com/Simon-Initiative/oli-torus/assets/16328384/aeed67db-71e4-4bc1-8fdb-2575e8a631fa

[MER-1908]: https://eliterate.atlassian.net/browse/MER-1908?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli/delivery/attempts/core.ex
+++ b/lib/oli/delivery/attempts/core.ex
@@ -25,11 +25,12 @@ defmodule Oli.Delivery.Attempts.Core do
 
   def get_user_from_attempt(%ResourceAttempt{resource_access_id: resource_access_id}) do
     query =
-      from access in ResourceAccess,
+      from(access in ResourceAccess,
         join: user in User,
         on: access.user_id == user.id,
         select: user,
         where: access.id == ^resource_access_id
+      )
 
     Repo.one(query)
   end
@@ -40,13 +41,14 @@ defmodule Oli.Delivery.Attempts.Core do
   """
   def has_any_attempts?(%User{id: user_id}, %Section{id: section_id}, resource_id) do
     query =
-      from access in ResourceAccess,
+      from(access in ResourceAccess,
         join: attempt in ResourceAttempt,
         on: access.id == attempt.resource_access_id,
         select: count(attempt.id),
         where:
           access.user_id == ^user_id and access.section_id == ^section_id and
             access.resource_id == ^resource_id
+      )
 
     Repo.one(query) > 0
   end
@@ -419,7 +421,7 @@ defmodule Oli.Delivery.Attempts.Core do
   def get_part_attempts_and_users(project_id) do
     # This is our base, reusable query designed to get the part attempts
     core =
-      from snapshot in Snapshot,
+      from(snapshot in Snapshot,
         as: :snapshot,
         join: part_attempt in PartAttempt,
         as: :part_attempt,
@@ -427,6 +429,7 @@ defmodule Oli.Delivery.Attempts.Core do
         where:
           snapshot.project_id == ^project_id and
             part_attempt.lifecycle_state == :evaluated
+      )
 
     # Now get the resource attempt revision for those part attempts, distinctly, and
     # create a map of their ids to the attempts
@@ -589,16 +592,18 @@ defmodule Oli.Delivery.Attempts.Core do
   end
 
   def get_latest_non_active_activity_attempts(resource_attempt_id) do
-
-    query = from(
-      aa in ActivityAttempt,
-      where: (aa.lifecycle_state == :evaluated or aa.lifecycle_state == :submitted) and aa.resource_attempt_id ==  ^resource_attempt_id,
-      group_by: aa.resource_id,
-      select: max(aa.id)
-    )
+    query =
+      from(
+        aa in ActivityAttempt,
+        where:
+          (aa.lifecycle_state == :evaluated or aa.lifecycle_state == :submitted) and
+            aa.resource_attempt_id == ^resource_attempt_id,
+        group_by: aa.resource_id,
+        select: max(aa.id)
+      )
 
     Repo.all(
-      from(aa in ActivityAttempt ,
+      from(aa in ActivityAttempt,
         where: aa.id in subquery(query),
         select: aa
       )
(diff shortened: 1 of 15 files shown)


lib/oli_web/components/delivery/actions/actions.ex
```diff
@@ -194,6 +213,110 @@ defmodule OliWeb.Components.Delivery.Actions do
...
+
+    %{
+      section: section,
+      user: user,
+      enrollment: enrollment
+    } = socket.assigns
+
+    result =
+      Repo.transaction(fn ->
+        with {:ok, %Enrollment{} = new_enrollment} <-
```
darrensiegel, line 258: All of the code here inside (and including) the `Repo.transaction` should be moved out of `OliWeb` and into `Oli`, since it represents core application functionality, rather than UI specific functionality.  As a general rule we should be placing as little code as possible in the UX layer, allowing greater code reuse and easier testing.

lib/oli_web/components/delivery/actions/actions.ex
```diff
@@ -194,6 +213,110 @@ defmodule OliWeb.Components.Delivery.Actions do
...
+                 section_id_to_transfer
+               ),
+             ## Delete current enrollment
+             {:ok, _enrollment} <- Sections.delete_enrollment(enrollment),
+
+             ## Delete attempts and resource accesses for section to transfer
+             {:ok, _} <- Core.delete_attempts_and_resource_accesses(section_to_transfer, user.id),
+             ## Delete snapshots for section to transfer
+             {changes_count, nil} when is_integer(changes_count) <-
+               Snapshots.delete_snapshots_by_section(section_id_to_transfer),
```
darrensiegel, line 277: This will have dire, unintended side effects.  `Snapshots.delete_snapshots_by_section` as currently written will delete **ALL** snapshot records for the section when the intent is to delete only the snapshot records for that user in that section. `Snapshots.update_snapshots_by_section` has the same problem.

lib/oli/delivery/attempts/core.ex
```diff
@@ -868,6 +881,18 @@ defmodule Oli.Delivery.Attempts.Core do
     |> Repo.update()
   end
 
+  @doc """
+  Updates all resource accesses for a given section.
+  """
+
+  def update_resource_accesses_by_section(section_id, new_section_id) do
```
darrensiegel, line 888: I don't think this is being used, and it would be dangerous to use it as-is since it is not user specific. We should remove it.

lib/oli_web/components/delivery/actions/actions.ex
```diff
@@ -194,6 +213,110 @@ defmodule OliWeb.Components.Delivery.Actions do
...
+
+             ## Update payments for current enrollment with new enrollment id
+             {changes_count, nil} when is_integer(changes_count) <-
+               Paywall.update_payments_for_enrollment(
+                 enrollment,
+                 new_enrollment.id,
+                 section_id_to_transfer
+               ),
+             ## Delete current enrollment
+             {:ok, _enrollment} <- Sections.delete_enrollment(enrollment),
```
darrensiegel, line 271: Creating a new enrollment and deleting the existed fails to carry forward "Course section specific user state".  This is found in the `state` field of the `enrollments` schema.

darrensiegel (changes requested):
There is some rework needed here, partly because I have also just realized that we need to change how this transfer process works so that we can actually target a key use case where a student needs to be transferred from a Direct Delivery section into a LTI section (and vice-versa).  In this key use case, the user will be different. 

This is what I am recommending:
A. Create a new context for all of this "section transfer" code.  Perhaps `Oli.Delivery.Sections.Transfer` or even `Oli.Delivery.Transfer`.  Move `Sections.get_sections_to_transfer_data` into there as well as the core implementation of this that is currently found in `actions.ex`. 
B. Change this implementation - and entire feature - to assume that the student must already be enrolled in the target course section.   core implementation will then look like this:
    
1. From a list of eligible course sections **that includes the section title, start and end dates, and instructor name**, user selects a target course section. 
2. From a list of all enrolled students (showing student name and enrollment date and time) in the selected section, user selects an enrolled student.
3. System copies the `state` field from the source enrollment to the target enrollment
4. If the source enrollment *has a Payment record that is `paid`* **AND** the target enrollment *does not have a Payment record that is already `paid`*, the system updates the target enrollment to carry forward the source enrollment Payment.
5. The system deletes all snapshot, resource access, resource_attempts, activity_attempts and part_attempts for the **selected user** in the **target course section**
6. The system updates the resource access records for the source student in the source section, changing the `section_id` and `user_id` to point to the target section and target user. 
7. The system updates the snapshot records from the source section for the source user, changing both the section_id and the user_id to point to the target section_id and the target user_id. 

Note the UX changes described above (a two step UX, selecting first the target section and then an enrolled user)

simonchoxx:
The UI/UX is modified to be able to select a target section and then a target student to transfer the enrollment to. 


https://github.com/Simon-Initiative/oli-torus/assets/16328384/0b75f068-b07e-4800-861b-ccd7b8c0f4ed

darrensiegel (approved):
This all looks good.  

In testing this locally I ran into two problems, which I have pushed commits to fix:
1. If I start Torus in my local dev server, when I already have open the page editor in a tab, and simply reload the page, and then try to insert an MCQ I see a failure, which was triggered by parsing of the MCQ part, and a call to `String.to_existing_atom` where the string passed to it was `"automatic"`.  I imagine that the atom `:automatic` hadn't yet been created in the BEAM and so this code fails.  This is likely not the fault of this PR, but I handled yet via statically defining both `:manual` and `:automatic` here. 

2. In trying to transfer one student enrollment to another section, I saw a server side failure where `String.downcase` was being called on a `nil`.  Root cause was that any section with guest users will have users whose `:name` is `nil`.  I handled this here and it works fine now.

simonchoxx:
Great! Thanks @darrensiegel for testing and identifying those errors and pushed those changes!

PR #4080, opened 2023-08-03
[BUG FIX] [MER-2450] multi input text inputs on graded pages restoring only first word

eliknebel:
Fixes an issue with multi input questions in review mode of an assessment where only the first word in a text input was being rendered.

### DIFF
--- a/assets/src/components/activities/multi_input/MultiInputDelivery.tsx
+++ b/assets/src/components/activities/multi_input/MultiInputDelivery.tsx
@@ -121,7 +121,7 @@ export const MultiInputComponent: React.FC = () => {
   );
 
   const inputValue = (input: MultiInput) => {
-    const studentInput = uiState.partState[input.partId].studentInput;
+    const studentInput = uiState.partState[input.partId]?.studentInput;
 
     // safeguard against null/undefined studentInput
     if (!studentInput) return '';
--- a/assets/src/components/activities/multi_input/MultiInputDelivery.tsx
+++ b/assets/src/components/activities/multi_input/MultiInputDelivery.tsx
@@ -83,10 +83,12 @@ export const MultiInputComponent: React.FC = () => {
       initializeState(
         activityState,
         safelySelectStringInputs(activityState).caseOf({
-          just: (input) => input,
-          nothing: () => ({
-            [castPartId(activityState.parts[0].partId)]: [''],
-          }),
+          just: (inputs) => inputs,
+          nothing: () =>
+            model.inputs.reduce((acc, input) => {
+              acc[input.partId] = [''];
+              return acc;
+            }, {} as PartInputs),
         }),
         model,
         context,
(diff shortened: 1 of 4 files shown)


assets/src/components/activities/multi_input/MultiInputDelivery.tsx
```diff
@@ -120,6 +120,19 @@ export const MultiInputComponent: React.FC = () => {
...
   );
 
+  const inputValue = (input: MultiInput) => {
+    const studentInput = uiState.partState[input.partId].studentInput
+
+    // safeguard against null/undefined studentInput
+    if (!studentInput) return '';
+
+    if (input.inputType === 'text') {
+      return uiState.partState[input.partId].studentInput.join(' ');
```
darrensiegel, line 130: So it seems like for 'text' inputs, student input was being stored as an array of strings, one element for each word?

If there is even a chance at some point in time "some" inputs were being stored correctly as full strings, but then an implementation changed which caused them to be stored as an array - we should make this impl here more robust and perhaps be looking to see what `typeof  uiState.partState[input.partId].studentInput` returns.  If it is a string, we use it as is, if it is an array, we `join`

darrensiegel (changes requested):

assets/src/components/activities/multi_input/MultiInputDelivery.tsx
```diff
@@ -120,6 +120,22 @@ export const MultiInputComponent: React.FC = () => {
     {},
   );
 
+  const inputValue = (input: MultiInput) => {
+    const studentInput = uiState.partState[input.partId].studentInput;
```
marc-hughes, line 124: The old version has a null guard:
uiState.partState[input.partId]?.studentInput || []

Might that be null or undefined?

andersweinstein:
It looks to me like the proper fix would be to replace `safelySelectInputs` with `safelySelectStringInputs` where it is used in MultiInputDelivery. The latter is a version of the utility function that does almost exactly the same thing, only does not apply "stringToStudentInput" to split the input strings at spaces. 

I suspect use of the wrong version of the utility function was the source of the bug.

assets/src/components/activities/multi_input/MultiInputDelivery.tsx
andersweinstein: I don't think it was ever stored as an array. I believe it was always stored as a string. The bug arised because a client side code uses a utility function safelySelectInputs to fetch the input values from the state. That utility function splits string-valued inputs into arrays before returning. 

It seems use of safelySelectInputs  caused a bug for ShortAnswers two years ago which was fixed by adding a different function safelySelectInputStrings which does not do any splitting. I believe multi-input should use safelySelectinputStrings as well.

eliknebel (commented):
Tested with all input variants in multi input and fixes the issue, works as expected

andersweinstein (approved):

andersweinstein (approved):

andersweinstein (approved):

andersweinstein (approved):

eliknebel:
@darrensiegel can you take another look and approve if this looks good?

PR #4081, opened 2023-08-04
[ENHANCEMENT] author multiple correct answers [MER-2294]

andersweinstein:
This adds the ability for an author to specify multiple correct answers. It works by adding a checkbox labeled "Correct" to the ResponseCard component used to display a response when adding targeted feedback. Value is bound by code to the "score" attribute of the response, with non-zero score being correct, anything else incorrect. Checking the box on a previously incorrect response sets a score of 1.

Note score was preserved through migration tool, so migrated courses may have multiple correct responses, though this fact was invisible on the interface. 

Adds a simple AuthoringCheckbox[Connected] component wrapping a checkbox; "connected" version picks up the editMode from the authoring context. Modeled on AuthoringButton.

### DIFF
--- a/assets/src/components/activities/common/authoring/AuthoringCheckbox.tsx
+++ b/assets/src/components/activities/common/authoring/AuthoringCheckbox.tsx
@@ -17,20 +17,21 @@ interface Props {
 export const AuthoringCheckbox: React.FC<Props> = (props: Props) => {
   return (
     <div>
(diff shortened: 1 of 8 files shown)


eliknebel:
Looking into this more, I just double checked an it seems like I had not clicked the "Correct" checkbox for these, after doing so I verified the change persisted correctly and feedback was now working as expected

eliknebel (approved):

PR #4088, opened 2023-08-07
[Feature] Support html paste [MER-2159]

marc-hughes:
This adds the ability to paste in formatted content to torus core authoring pages. I've gotten some reasonable results from plain html, echo preview, MSWord, and google docs.  Each of these copy sources tended to do things a bit differently, so there's a fairly extensive set of unit tests that cover all the cases I saw while developing.

Example echo paste:
image

Echo paste notes:

1. Echo uses `<em>` with no additional info for bold. Generally, em is italic, and I stuck with that more semantic meaning.
3. An echo table has a table inside a table to support the title & caption.  Torus does not support nested tables. There's not a great way for me to identify when it is or isn't a paste from echo, so whenever we see a nested table, we take the inner table and put it as a sibling as the outer table. It's not the desired formatting in this case, but it has the benefit of not losing any data.

Example plain HTML:
image

Example google doc:
image

Example msword:
image

Some future improvements we could consider:

1. Pasting images (either just the image, or inline in other content)
2. Pasting rendered formulas from echo


darrensiegel (approved):
Tested locally and works nicely.

PR #4095, opened 2023-08-09
[PERFORMANCE] Optimize container based proficiency calculation [MER-2447]

darrensiegel:
The original reported issue (https://eliterate.atlassian.net/browse/MER-2447) noted course sections on Argos Prod that would not load the Instructor Dashboard "Reports" view, with the UI eventually getting a "Gateway Timeout" error.

## Part 1: Query Optimization and Post Processing

I pinned down the problem to be the query used to calculate student proficiency, per container, from `Metrics.proficiency_per_container/1`.  The raw SQL for that query originally was:

```
SELECT c2.container_id, 
  CAST(COUNT(CASE WHEN s0.correct THEN 1 END) as float) / CAST(COUNT(*) as float) 
FROM snapshots AS s0 
INNER JOIN sections AS s1 ON s0.section_id = s1.id 
INNER JOIN contained_pages AS c2 ON c2.page_id = s0.resource_id 
WHERE (((s0.attempt_number = 1) AND (s0.part_attempt_number = 1)) 
  AND (s1.slug = 'bio_100_on_demand__self_paced_')) 
GROUP BY c2.container_id
```

This query runs just fine on Torus Prod with its 200K snapshot records.  On Argos Prod there are 109 million snapshot records and this query doesn't finish even after letting it go for close to 5 minutes from within pgAdmin. 

To address the problem I've reworked the overall approach here to no longer attempt to do the container "bucketing" within the query, and instead to do that in code as a post-processing step after the query finishes.  I also removed the JOIN to `sections` that was necessary because the query was slug based. So the new query now looks like:

```
SELECT s0.resource_id, 
  CAST(COUNT(CASE WHEN s0.correct THEN 1 END) as float),
  CAST(COUNT(*) as float) 
 FROM snapshots AS s0 
 WHERE (((s0.attempt_number = 1) AND (s0.part_attempt_number = 1)) 
		AND (s0.section_id = 725)) 
GROUP BY s0.resource_id
```

The above query, on Argos Prod, runs anywhere between 2 seconds to about 25 seconds (due to cache warmth, I believe). The 25 seconds upper bound is concerning as we could still face the issue of this view still timing out if this grows.  There would also be a negative UX by forcing the user to wait this long before the view would display.  So, on to:

## Part 2: Async Processing

This PR also changes the Instructor and Student Details dashboards LiveViews to invoke the potentially expensive proficiency calculation as an async, background task.  When that task finishes it will notify the LV and the assigns are updated.  The UX now will be a quick loading view, with some "Loading..." placeholder text in place for the proficiency values until those calculations finish.

### DIFF
--- a/lib/oli/delivery/metrics.ex
+++ b/lib/oli/delivery/metrics.ex
@@ -885,10 +885,14 @@ defmodule Oli.Delivery.Metrics do
 
     Enum.reduce(page_totals, container_totals, fn {page_id, correct, total}, map ->
       container_ids = Map.get(inverted_cp_index, page_id)
-      Enum.reduce(container_ids, map, fn container_id, map ->
-        {current_correct, current_total} = Map.get(map, container_id)
-        Map.put(map, container_id, {current_correct + correct, current_total + total})
-      end)
+      case container_ids do
+        nil -> map
+        _ ->
+          Enum.reduce(container_ids, map, fn container_id, map ->
+            {current_correct, current_total} = Map.get(map, container_id)
+            Map.put(map, container_id, {current_correct + correct, current_total + total})
+          end)
+      end
     end)
     |> Enum.into(%{}, fn {container_id, {correct, total}} ->
       proficiency = case total do
--- a/lib/oli/delivery/metrics.ex
+++ b/lib/oli/delivery/metrics.ex
@@ -868,9 +868,7 @@ defmodule Oli.Delivery.Metrics do
   # Given a list of ContainedPage records, return a map of container ids to a tuple of
   # correct and total values, initialized to {0.0, 0.0}
   defp init_container_totals(contained_pages) do
-    Enum.map(contained_pages, fn %ContainedPage{container_id: container_id} -> container_id end)
-    |> Enum.dedup()
-    |> Enum.reduce(%{}, fn container_id, map ->
+     Enum.reduce(contained_pages, %{}, fn %ContainedPage{container_id: container_id}, map ->
       Map.put(map, container_id, {0.0, 0.0})
     end)
   end
@@ -885,12 +883,14 @@ defmodule Oli.Delivery.Metrics do
 
     Enum.reduce(page_totals, container_totals, fn {page_id, correct, total}, map ->
       container_ids = Map.get(inverted_cp_index, page_id)
+
       case container_ids do
         nil -> map
         _ ->
           Enum.reduce(container_ids, map, fn container_id, map ->
-            {current_correct, current_total} = Map.get(map, container_id)
-            Map.put(map, container_id, {current_correct + correct, current_total + total})
+            update_in(map, [container_id], fn {current_correct, current_total} ->
+              {current_correct + correct, current_total + total}
+            end)
           end)
       end
     end)
--- a/lib/oli/delivery/metrics.ex
+++ b/lib/oli/delivery/metrics.ex
@@ -843,6 +843,18 @@ defmodule Oli.Delivery.Metrics do
     Repo.one(query)
   end
 
+  # Given a list of ContainedPage records, return a map of
+  # page ids to a list of container ids (their ancestor containers)
+  # that contain that page
+  #
+  # For instance, given the following ContainedPage records:
+  #
+  #  %ContainedPage{container_id: 1, page_id: 10}
+  #  %ContainedPage{container_id: 1, page_id: 11}
+  #  %ContainedPage{container_id: 2, page_id: 10}
+  #
+  # This function will return:
+  # %{10 => [1, 2], 11 => [1]}
   defp page_to_parent_containers_map(contained_pages) do
     Enum.reduce(contained_pages, %{},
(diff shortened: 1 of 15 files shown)


nicocirio (approved):
LGTM! 
Loved the async implementation :smile:

I tested it locally adding a `:timer.sleep` in the async function to see the behavior when receiving the message after navigating to another table page and worked fine!

I left some minor NITs that may improve a little bit the performance by reducing the amounts of nested `Enum`

lib/oli/delivery/metrics.ex
```diff
@@ -857,4 +842,59 @@ defmodule Oli.Delivery.Metrics do
...
+  # ContainedPage records, return a map of container ids to a tuple of correct and total values,
+  # where the container totals are the sum of the page totals for all pages contained in that container.
+  defp bucket_into_container_totals(page_totals, contained_pages) do
+
+    inverted_cp_index = page_to_parent_containers_map(contained_pages)
+    container_totals = init_container_totals(contained_pages)
+
+    Enum.reduce(page_totals, container_totals, fn {page_id, correct, total}, map ->
+      container_ids = Map.get(inverted_cp_index, page_id)
+      Enum.reduce(container_ids, map, fn container_id, map ->
```
eliknebel, line 886: I'm running into an error here when testing with the chemistry course, I believe it is related to `container_ids` being null, possibly from detatched pages?

```[error] Task #PID<0.30102.0> started from #PID<0.30094.0> terminating
** (Protocol.UndefinedError) protocol Enumerable not implemented for nil of type Atom
    (elixir 1.14.0) lib/enum.ex:1: Enumerable.impl_for!/1
    (elixir 1.14.0) lib/enum.ex:166: Enumerable.reduce/3
    (elixir 1.14.0) lib/enum.ex:4307: Enum.reduce/3
    (elixir 1.14.0) lib/enum.ex:2468: Enum."-reduce/3-lists^foldl/2-0-"/3
    (oli 0.24.3) lib/oli/delivery/metrics.ex:886: Oli.Delivery.Metrics.bucket_into_container_totals/2
    (oli 0.24.3) lib/oli_web/live/delivery/instructor_dashboard/instructor_dashboard_live.ex:599: anonymous fn/2 in OliWeb.Delivery.InstructorDashboard.InstructorDashboardLive.async_calculate_proficiency/1
    (elixir 1.14.0) lib/task/supervised.ex:89: Task.Supervised.invoke_mfa/2
    (elixir 1.14.0) lib/task/supervised.ex:34: Task.Supervised.reply/4
    (stdlib 3.15) proc_lib.erl:226: :proc_lib.init_p_do_apply/3
Function: #Function<0.102994098/0 in OliWeb.Delivery.InstructorDashboard.InstructorDashboardLive.async_calculate_proficiency/1>
    Args: []
[error] an exception was raised:
    ** (Protocol.UndefinedError) protocol Enumerable not implemented for nil of type Atom
        (elixir 1.14.0) lib/enum.ex:1: Enumerable.impl_for!/1
        (elixir 1.14.0) lib/enum.ex:166: Enumerable.reduce/3
        (elixir 1.14.0) lib/enum.ex:4307: Enum.reduce/3
        (elixir 1.14.0) lib/enum.ex:2468: Enum."-reduce/3-lists^foldl/2-0-"/3
        (oli 0.24.3) lib/oli/delivery/metrics.ex:886: Oli.Delivery.Metrics.bucket_into_container_totals/2
        (oli 0.24.3) lib/oli_web/live/delivery/instructor_dashboard/instructor_dashboard_live.ex:599: anonymous fn/2 in OliWeb.Delivery.InstructorDashboard.InstructorDashboardLive.async_calculate_proficiency/1
        (elixir 1.14.0) lib/task/supervised.ex:89: Task.Supervised.invoke_mfa/2
        (elixir 1.14.0) lib/task/supervised.ex:34: Task.Supervised.reply/4
        (stdlib 3.15) proc_lib.erl:226: :proc_lib.init_p_do_apply/3```

**Review by eliknebel — CHANGES_REQUESTED** [review]

File: lib/oli/delivery/metrics.ex (hunk 1 above)
Comment (eliknebel) [reply lib/oli/delivery/metrics.ex:886]: Indeed adding a case statement around this code fixes the issue, however I am not certain this is what we want here or not
```
      case container_ids do
        nil -> map
        _ ->
          Enum.reduce(container_ids, map, fn container_id, map ->
            {current_correct, current_total} = Map.get(map, container_id)
            Map.put(map, container_id, {current_correct + correct, current_total + total})
          end)
      end
```

darrensiegel: Ah, yes, great catch.  That is exactly what is happening here.  You're solution makes sense since an unordered page has no ancestor containers - we simply have to ignore it.

eliknebel (approved):

nicocirio:
@darrensiegel It seems liveview is about to support async assigns natively, so we should keep an eye on [this PR](https://github.com/phoenixframework/phoenix_live_view/pull/2763)

PR #4117, opened 2023-08-17
[BUGFIX][MER-2339] Fix error quiz score tab in suspended student

simonchoxx:
[MER-2339](https://eliterate.atlassian.net/browse/MER-2339)

This PR fixes a bug that happened when accessing the quiz scores tab within a student's dashboard and the student was in suspended status in that section. 

Now, when accessing the quiz scores tab for a specific student who is suspended in that section, a text is rendered indicating that there are no quiz scores to show.

[MER-2339]: https://eliterate.atlassian.net/browse/MER-2339?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli/grading.ex
+++ b/lib/oli/grading.ex
@@ -8,9 +8,9 @@ defmodule Oli.Grading do
   import Ecto.Query, warn: false
   import Oli.Utils, only: [log_error: 2]
 
-  alias Oli.Publishing.DeliveryResolver
+  alias Oli.Publishing.{DeliveryResolver, PublishedResource}
   alias Oli.Delivery.Sections
-  alias Oli.Delivery.Sections.{Section}
+  alias Oli.Delivery.Sections.{Section, SectionResource, SectionsProjectsPublications}
   alias Oli.Delivery.Attempts.Core, as: Attempts
   alias Oli.Delivery.Attempts.Core.ResourceAccess
   alias Oli.Grading.GradebookRow
@@ -20,6 +20,7 @@ defmodule Oli.Grading do
   alias Lti_1p3.Tool.Services.AGS.Score
   alias Oli.Resources.Revision
   alias OliWeb.Common.Utils
+  alias Oli.Repo
 
   @doc """
   If grade passback services 2.0 is enabled, sends the current state of a ResourceAccess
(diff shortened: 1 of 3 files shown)


darrensiegel (changes requested):
These changes look good, but there is a performance optimization here that we should take care of in this PR:

`get_scores` is intending to return assessment results for a specific student, but it does so by fetching the assessment results **for all students** (via the call to `Oli.Grading.generate_gradebook_for_section), and then filtering for a specific user.  Instead, we should have a single, custom query written that fetches this information for the current user.

darrensiegel:
From the "Student Details / Quiz Scores" we should show all scores, regardless of whether the student is enrolled or suspended.  This will allow an instructor or admin to go back and see what a student has done, even if the student eventually was suspended.

lib/oli/grading.ex
```diff
@@ -211,6 +211,44 @@ defmodule Oli.Grading do
...
 
+  @doc """
+  Returns a list of GradebookScore for enrolled user in the provided section
+
+  `[%GradebookScore{}, GradebookScore{}, ...]`
+  """
+  def get_scores_for_section_and_user(%Section{} = section, student_id) do
+    # get publication page resources, filtered by graded: true
+    graded_pages = Sections.fetch_scored_pages(section.slug)
+
```
darrensiegel, line 222: This should be optimized one step further.  

Here is what is taking place with this latest change:
1. Query for all scored_pages
2. Query for all resource accesses (for all students)
3. Post process to filter down to only those resource accesses for this specific student (and convert to GradebookScore)

Instead, the two queries can be merged into one in a way that eliminates the need for any post processing:

Something like: 

```
SELECT page.resource_id, page.title, resource_access.score, resource_access.out_of, resource_access.was_late
FROM subquery_for_graded_page_titles_and_resource_id_only(section.id) as page
LEFT JOIN resource_accesses resource_access ON resource_access.id = page.resource_access_id
WHERE resource_access.section_id = ^section.id and resource_access.user_id = ^user.id
```

The above isn't tested and likely not valid SQL, but it conveys the intent of what we want here - a single query that gets just the information that we need, which can simply then be mapped into the `GradebookScore` structs.  Also, this SQL approach will right-pad with `nil` resource access records that are missing.  A "missing" resource_access record means that the student has never visited that particular page yet.  In that case, we still want to list the quiz in the table, with some indication that the student has no score yet.  The existing approach I believe will omit graded pages that student has not visited (given how it reduces only non `nil` values into the final list)

darrensiegel (changes requested):

darrensiegel (approved):

PR #4118, opened 2023-08-17
[BUGFIX][MER-2434] Only return pages for upcoming activities

augusto1024:
[Link to the ticket](https://eliterate.atlassian.net/browse/MER-2434)

### DIFF
--- a/test/oli/sections_test.exs
+++ b/test/oli/sections_test.exs
@@ -1444,7 +1444,10 @@ defmodule Oli.SectionsTest do
       assert target_section_2.title == section_2.title
     end
 
-    test "deletes attempts and resource accesses by section and user", %{section_1: section_1, user_1: user_1} do
+    test "deletes attempts and resource accesses by section and user", %{
+      section_1: section_1,
+      user_1: user_1
+    } do
       # gets resource accesses, resource attempts, activity attempts and part attempts from target section
       resource_accesses =
         Core.get_resource_accesses(section_1.slug, user_1.id) |> Enum.map(& &1.id)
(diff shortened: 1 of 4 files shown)


test/oli/sections_test.exs
```diff
@@ -1530,19 +1536,76 @@ defmodule Oli.SectionsTest do
...
+        is_liveview: false
+      }
+
+      enroll_user_to_section(student, section, :context_learner)
+
+      next_activities =
+        Sections.get_next_activities_for_student(section.slug, student.id, session_context)
+
+      assert length(next_activities) == 1
+      assert Enum.at(next_activities, 0).title == "Upcoming assessment"
```
simonchoxx, line 1608: I like the use of `Enum.at` here (I usually use List.first)!!!

simonchoxx (approved):
Great work 🚀🚀 🚀

I would only add that there are some lines that were modified (maybe because of the formatting) and I understand that they do not correspond to the changes needed for the ticket!

test/oli/sections_test.exs
augusto1024: I think I actually like `List.first` better 😅. I don't know why I used `Enum.at`

augusto1024:
@simonchoxx yeah, everytime I make changes in a file, I run the formatter for it. We should open a PR for running it across the whole project

test/oli/sections_test.exs
simonchoxx: We get the same thing with both functions, but I was glad to see it because I don't think I've used it!

darrensiegel (approved):

PR #4137, opened 2023-08-24
[FEATURE][MER-2391] Include products related project during project export

simonchoxx:
[MER-2391](https://eliterate.atlassian.net/browse/MER-2391)

This PR includes the products contained in a project when exporting this project. 

It excludes those products that have remixed materials from other projects in their curriculum.

[MER-2391]: https://eliterate.atlassian.net/browse/MER-2391?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli/interop/export.ex
+++ b/lib/oli/interop/export.ex
@@ -5,6 +5,8 @@ defmodule Oli.Interop.Export do
   alias Oli.Authoring.MediaLibrary
   alias Oli.Authoring.MediaLibrary.ItemOptions
   alias Oli.Utils
+  alias Oli.Delivery.Sections.Blueprint
+  alias Oli.Delivery.Sections
 
   @doc """
   Generates a course digest for an existing course project.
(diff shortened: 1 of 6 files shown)


lib/oli/interop/export.ex
```diff
@@ -351,6 +371,21 @@ defmodule Oli.Interop.Export do
     |> entry("_hierarchy.json")
   end
 
+  # create the singular hierarchy file
+  defp create_product_file(resources, product, publication) do
```
darrensiegel, line 375: This implementation ends up duplicating the project hierarchy, and not exporting the actual product hierarchy.  To see an example of this, follow these steps:

1. Create a course project that contains three top level Units.  In each unit add one practice page.
2. Publish that course project.
3. Create a product from that course project.
4. For the Product, using course remix and remove the first two units, then click "Save"
5. Now Export the Project.
6. Import the ZIP file as a new project.
7. In the new, imported project, visit the Product that was imported. 

You will see that it contains all three units, instead of only one.

The reason here is that the logic for outputting the Product hierarchy is driven from the `root_resource_id` of the publication from the project (which of course still points to all three units).  This code should instead construct the hierarchy from the structure of the `SectionResource` records (and specifically from their `children` attributes) rooted from the `root_section_resource` from the Product itself, not from the Project.

darrensiegel (changes requested):

darrensiegel (approved):

PR #4144, opened 2023-08-28
[MER-2316] Fix issues with discussions

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-2316) to the ticket

## Before 
image

## After
image


lib/oli_web/live/collaboration_live/posts/list.ex
```diff
@@ -70,4 +70,15 @@ defmodule OliWeb.CollaborationLive.Posts.List do
...
+  defp reply_margin(level) do
+    case level do
+      0 -> nil
+      1 -> "ml-[12px]"
+      2 -> "ml-[24px]"
+      3 -> "ml-[36px]"
+      4 -> "ml-[48px]"
+      _ -> "ml-[60px]"
+    end
+  end
```
augusto1024, line 83: Personally, I don't like the padding after the second level. I think it's more aesthetic to have only one level of padding just for the replies of the root post. But that's just my opinion 😁

lib/oli_web/live/collaboration_live/collab_space_view.ex
```diff
@@ -683,27 +683,36 @@ defmodule OliWeb.CollaborationLive.CollabSpaceView do
...
+  defp build_replies([], _mapper, _post_level, acum_replies), do: List.flatten(acum_replies)
+
+  defp build_replies([reply | replies], mapper, post_level, acum_replies) do
+    post_full_replies =
+      Enum.map(Map.get(mapper, reply.id, []), fn r ->
+        [Map.put(r, :post_level, post_level) | build_replies([r], mapper, post_level + 1, [])]
+      end)
+
+    build_replies(replies, mapper, post_level, [post_full_replies | acum_replies])
+  end
```
augusto1024, line 714: Wouldn't it be more efficient if we concatenate the lists like this?
```suggestion
  defp build_replies([], _mapper, _post_level, acum_replies), do: []

  defp build_replies([reply | replies], mapper, post_level, acum_replies) do
    post_full_replies =
      Enum.map(Map.get(mapper, reply.id, []), fn r ->
        Map.put(r, :post_level, post_level) ++ build_replies([r], mapper, post_level + 1, [])
      end)

    build_replies(replies, mapper, post_level, [post_full_replies | acum_replies])
  end
```

This way we don't need to recursively transverse the list to make the flatten

augusto1024 (approved):
Looks good! I just left a small suggestion and a very personal opinion 😅

PR #4160, opened 2023-08-30
[IMPROVEMENT][MER-2389] Preserve student survey in project export

augusto1024:
[Link to the story](https://eliterate.atlassian.net/browse/MER-2389)


lib/oli/interop/export.ex
```diff
@@ -306,7 +322,8 @@ defmodule Oli.Interop.Export do
     %{
       title: project.title,
       description: project.description,
-      type: "Manifest"
+      type: "Manifest",
+      required_student_survey: project.required_survey_resource_id
```
darrensiegel, line 326: This needs to convert the `required_survey_resource_id` to a string.  All ids in export archives must be strings to accomodate that in the OLI legacy system these strong ids were alphanumeric in nature.

lib/oli/interop/ingest.ex
```diff
@@ -338,7 +338,8 @@ defmodule Oli.Interop.Ingest do
         Oli.Authoring.Course.create_project(title, as_author, %{
           description: Map.get(project_details, "description"),
           legacy_svn_root: Map.get(project_details, "svnRoot"),
-          customizations: custom_labels
+          customizations: custom_labels,
+          required_survey_resource_id: Map.get(project_details, "required_student_survey")
```
darrensiegel, line 342: The value of "required_student_survey" must be read, and if not `nil`, mapped from the string id to the actual Torus database id of that resource.  This ID won't exist into all pages have been ingested and we know what the mapping of string ids to actual Torus Ids are.  So we probably can't set `required_survey_resource_id` here, it will need to be done as an update after all pages are processed.

darrensiegel (changes requested):
See my comments inline, but at the moment this solution does not work (after ingesting an export with a survey the checkbox is checked but the `required_student_survey` will have an invalid value)

darrensiegel (approved):
Tested, works as expected

PR #4164, opened 2023-08-31
[BUGFIX] [MER-2387] Fix institution name suggestion

gastonabella:
[[MER-2387]]

After Bootstrap was removed in favor of Tailwind, the `typeahead` function was implicitly removed too since it was part of the library.
However, it is now available separately and can be imported as a new plugin.

**Solution:**
Import `typeahead` library only for the contexts where it is being used (authoring and lti layouts).

https://github.com/Simon-Initiative/oli-torus/assets/26532202/9f028157-c04a-4500-a0cf-f17482c87618

[MER-2387]: https://eliterate.atlassian.net/browse/MER-2387?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


augusto1024 (approved):
LGTM 🚀

PR #4168, opened 2023-09-01
[Feature] Adds markdown editor support to torus core authoring [MER-2491]

marc-hughes:
This adds the option to toggle between our current rich text / slate based editor and a simple markdown based editor.

There were a few major pieces to this work:

- Serialize & Deserialize between our current data format and markdown
  - see markdown_serializer_test.ts and markdown_deserializer_test.ts for examples on how that works
- An account-level setting to set the default editor option for an author
- A UI for markdown editing
- Ability to switch between editors, with appropriate warnings when content would be lost
- Retrieve & store which editor to use in all the content elements that should be switchable

Risks / Places to test:

- If the serialization/deserialization isn't correct, we could lose authored content. 
- Each location that has an editor needed to be modified, so I may have missed some.
- It'd be good to get some actual user testing on this, to make sure expectations are being met 

Demo Video:

https://github.com/Simon-Initiative/oli-torus/assets/333265/9c1ecafe-a221-438f-96e9-31779513c3eb


assets/css/app.css
```diff
@@ -18,10 +20,12 @@
  */
 a,
 a.btn.btn-link {
+  /* prettier-ignore */
```
marc-hughes, line 23: My prettier format, was putting a space between hover and text-blue, was weird.

assets/src/components/activities/common/authoring/actions/stemActions.ts
```diff
@@ -5,6 +5,13 @@ import { PREVIEW_TEXT_PATH, STEM_PATH } from 'data/activities/model/utils';
 import { Operations } from 'utils/pathOperations';
 
 export const StemActions = {
+  changeEditorMode(mode: 'slate' | 'markdown') {
+    return (model: any) => {
+      model.stem.editor = mode;
```
marc-hughes, line 10: TODO: I need to go back and revisit this, it's not currently handling the stemPath right.

assets/src/components/activities/common/choices/authoring/ChoicesAuthoring.tsx
```diff
@@ -61,15 +67,23 @@ export const Choices: React.FC<Props> = ({
                     onChange={(e) => onEdit(choice.id, makeContent(e.target.value).content)}
                   />
                 ) : (
-                  <RichTextEditorConnected
+                  <SlateOrMarkdownEditor
```
marc-hughes, line 70: This is a common pattern you'll see in a lot of places... Replace the RichText editor with this new editor that can switch back and forth, and set up an appropriate onEditorTypeChange handler.

assets/src/components/activities/types.ts
```diff
@@ -124,10 +125,14 @@ export interface HasContent {
...
  */
-export function makeContent(text: string, id?: string): { id: string; content: RichText } {
+export function makeContent(
+  text: string,
+  id?: string,
+): { id: string; content: RichText; editor: EditorType } {
   return {
     id: id ? id : guid(),
     content: [Model.p(text)],
+    editor: getDefaultEditor(),
```
marc-hughes, line 135: The default editor is set at content-creation time.

So if your default editor is slate, you create something, and then change your default editor... that content item will still be slate.

Not having an editor set defaults to slate.

assets/src/components/editing/markdown_editor/content_markdown_deserializer.ts
```diff
@@ -0,0 +1,264 @@
...
+    if ('children' in node && node.children) {
+      // Unknown nodes with children
+      console.info('Unknown node', node);
+      return contentMarkdownDeserializer(node.children as AllModelElements[]);
+    }
+
+    return null;
+  };
+
+// Note: We are not supporting colspan or rowspan
```
marc-hughes, line 129: TODO: Remove this note, we do support these!

assets/src/components/editing/toolbar/editorToolbar/EditorToolbar.tsx
```diff
@@ -27,6 +29,9 @@ export const EditorToolbar = (props: Props) => {
       <BlockToggle blockInsertOptions={props.insertOptions} />
       <BlockSettings />
       <BlockInsertMenu blockInsertOptions={props.insertOptions} />
+      {props.onSwitchToMarkdown && (
```
marc-hughes, line 32: Note: Availability of the new menu is dependent on an event handler for it being set. No need for two params.

assets/src/components/resource/editors/ContentEditor.tsx
```diff
@@ -57,11 +57,11 @@ interface ContentOutlineItemProps extends OutlineItemProps {
 
 export const ContentOutlineItem = (props: ContentOutlineItemProps) => {
   const { contentItem } = props;
-
+  const title = contentItem.editor === 'markdown' ? 'Markdown' : 'Paragraph';
```
marc-hughes, line 60: The outline in the left will list "Markdown" instead of "Paragraph" now, but maybe even "Paragraph" should be changed to "Rich text" or something since it can represent much more than a paragraph.

assets/src/phoenix/app.ts
```diff
@@ -134,7 +134,13 @@ document.addEventListener('DOMContentLoaded', () => {
     }),
   );
 
-  (window as any).hljs.highlightAll();
+  const hljs = (window as any).hljs;
+
+  hljs.configure({
```
marc-hughes, line 139: HLJS was trying to format our markdown code, so I made it much more targeted to the torus code blocks.

If it was supposed to be highlighting other things, this will introduce a bug, so need to watch for that.

assets/src/components/editing/markdown_editor/MarkdownEditor.tsx
```diff
@@ -0,0 +1,115 @@
...
+    console.info('onblur');
+    saveChanges(value || '');
+  }, [value, saveChanges]);
+
+  return (
+    <MDEditor
+      value={value}
+      className={props.className}
+      onChange={onChange}
+      height={initialHeight || 600}
```
darrensiegel, line 92: Is there any way to allow to have the height of the `MDEditor` component be driven from the content that is actually present?  It seems like each MDEditor is fixed to a certain height, and then internally scrolls the content through that fixed viewport.  It would be strongly preferred if the height of this thing could React to the content that is present, so that the author never has to scroll and never is wasting UI real estate (when they have only a few lines of content).

darrensiegel (changes requested):
I tested this out and it works well, I have just one request regarding the UX. See inline

assets/src/components/editing/markdown_editor/MarkdownEditor.tsx
marc-hughes: I just made this change, it is a big improvement. Also got rid of the trailing \n\n in each editor's content when it first starts, makes that editing slightly easier and worked well with the resize change.

image

image

darrensiegel (approved):

Issue #4123, opened 2023-08-18
Bullet points not working beside default one


gautamyadavs:
**Describe the bug**
a minor bug in torus which wasn't there before but after trying in multiple pages, I cannot seem to make it work so thought I will report, font in bullet works if it's default one but if I change it to numbered fonts than I cannot do any related action like create sub-bullets, indentation, remove bullets, etc.

**Expected behavior**
Bullet options for editing shouldn't go away.

**Screenshots**
https://github.com/Simon-Initiative/oli-torus/assets/25340998/979ba2aa-c4cc-4a8b-b40e-9cc0ca5a9f42

**Environment (please complete the following information):**
 - OS: Mac
 - Browser: Chrome

bitbldr:
Maybe related to https://github.com/Simon-Initiative/oli-torus/issues/4090

PR #4180, opened 2023-09-07
Bugfix: Numeric lists didn't have correct menu [MER-2523]

marc-hughes:
The list menu wasn't always appearing correctly.

Steps to reproduce:

1. Type some text on a single line
2. Convert it into a list.
3. Change it to a numeric ordered list

Expected: List menu is visible.
Actual: List menu is not visible.

timer

Fixes: #4123

### DIFF
--- a/assets/src/components/editing/elements/list/listActions.tsx
+++ b/assets/src/components/editing/elements/list/listActions.tsx
@@ -86,7 +86,7 @@ export const toggleOrderedList: CommandDescription = {
 const listStyleLabels: Record<OrderedListStyle | UnorderedListStyle, string> = {
   none: 'No Bullet',
   decimal: 'Decimal - 1',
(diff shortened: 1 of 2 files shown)


assets/src/components/editing/elements/list/listActions.tsx
```diff
@@ -86,7 +86,7 @@ export const toggleOrderedList: CommandDescription = {
 const listStyleLabels: Record<OrderedListStyle | UnorderedListStyle, string> = {
   none: 'No Bullet',
   decimal: 'Decimal - 1',
-  'decimal-leading-zero': 'Decimal w/ Zero - 01',
```
marc-hughes, line 89: I also noticed this menu item started wrapping to a second line and looking bad, so reworded it.

darrensiegel (approved):

PR #4182, opened 2023-09-07
[MER-2525] Add available date to assessment settings

gastonabella:
[MER-2525](https://eliterate.atlassian.net/browse/MER-2525)

This PR adds the following changes:
- Add available date to both **Assessment Settings** and **Student Exceptions** pages.
    - If both Available and Due dates are already defined and one of them is inconsistently updated (i.e. `due_date < available_date`), both dates are updated to keep the time distance. (See Video 1)
    - If the Available Date is set but the Due Date **is not** and the instructor attempts to set it, the datetime select does not allow dates before or equal to the Available Date. The reason is that we cannot calculate a time distance in this scenario. (See Video 2)
    - If the Due Date is set but the Available Date **is not** and the instructor attempts to set it, the datetime select does not allow dates after or equal to the Due Date. Same reason as above.
- Do not allow students to start new attempts if the assessment is not available yet.
- Do not allow students to start new attempts if the assessment was available when visiting the prologue page and the instructor sets a new available date before the student refreshes the page.

In order to keep PRs small, the Graphical Scheduler changes will be addressed in a separate PR.

https://github.com/Simon-Initiative/oli-torus/assets/26532202/94d69412-d145-46ba-b479-e1f009c45384

https://github.com/Simon-Initiative/oli-torus/assets/26532202/90f69498-0fda-4e2e-ad6e-f4d08b3d26cb

[MER-2525]: https://eliterate.atlassian.net/browse/MER-2525?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli/delivery/settings.ex
+++ b/lib/oli/delivery/settings.ex
@@ -133,11 +133,9 @@ defmodule Oli.Delivery.Settings do
   def check_start_date(%Combined{start_date: nil}), do: {:allowed}
 
   def check_start_date(%Combined{start_date: start_date}) do
-    if DateTime.compare(start_date, DateTime.utc_now()) == :gt do
-      {:before_start_date}
-    else
-      {:allowed}
-    end
+    if DateTime.compare(start_date, DateTime.utc_now()) == :gt,
+      do: {:before_start_date},
+      else: {:allowed}
   end
 
   def check_end_date(%Combined{end_date: nil}), do: {:allowed}
@@ -176,11 +174,9 @@ defmodule Oli.Delivery.Settings do
 
         # both an end date and a time limit, use the earlier of the two
         {end_date, time_limit} ->
-          if end_date < DateTime.add(resource_attempt.inserted_at, time_limit, :minute) do
-            end_date
-          else
-            DateTime.add(resource_attempt.inserted_at, time_limit, :minute)
-          end
+          if end_date < DateTime.add(resource_attempt.inserted_at, time_limit, :minute),
+            do: end_date,
+            else: DateTime.add(resource_attempt.inserted_at, time_limit, :minute)
       end
 
     case deadline do
--- a/lib/oli/delivery/settings.ex
+++ b/lib/oli/delivery/settings.ex
@@ -54,6 +54,7 @@ defmodule Oli.Delivery.Settings do
 
     %Combined{
       resource_id: resolved_revision.resource_id,
+      start_date: combine_field(:start_date, section_resource, student_exception),
       end_date: combine_field(:end_date, section_resource, student_exception),
       max_attempts: max_attempts,
       retake_mode: combine_field(:retake_mode, section_resource, student_exception),
--- a/lib/oli/delivery/settings.ex
+++ b/lib/oli/delivery/settings.ex
@@ -95,8 +95,12 @@ defmodule Oli.Delivery.Settings do
   end
 
   def was_late?(_, %Combined{late_submit: :disallow}, _now), do: false
-  def was_late?(%ResourceAttempt{} = resource_attempt, %Combined{late_submit: :allow} = effective_settings, now) do
 
+  def was_late?(
+        %ResourceAttempt{} = resource_attempt,
+        %Combined{late_submit: :allow} = effective_settings,
+        now
+      ) do
     case determine_effective_deadline(resource_attempt, effective_settings) do
       nil -> false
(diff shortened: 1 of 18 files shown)


lib/oli/delivery/settings.ex
```diff
@@ -126,6 +130,16 @@ defmodule Oli.Delivery.Settings do
...
   end
 
+  def check_start_date(%Combined{start_date: nil}), do: {:allowed}
+
+  def check_start_date(%Combined{start_date: start_date}) do
+    if DateTime.compare(start_date, DateTime.utc_now()) == :gt do
+      {:before_start_date}
+    else
+      {:allowed}
+    end
```
simonchoxx, line 140: Maybe this could be written using the short `if` statement, WDYT?

lib/oli/delivery/settings.ex
```diff
@@ -143,37 +157,36 @@ defmodule Oli.Delivery.Settings do
...
-        end
-    end
+
+        # both an end date and a time limit, use the earlier of the two
+        {end_date, time_limit} ->
+          if end_date < DateTime.add(resource_attempt.inserted_at, time_limit, :minute) do
+            end_date
+          else
+            DateTime.add(resource_attempt.inserted_at, time_limit, :minute)
+          end
```
simonchoxx, line 183: Same as above!!

simonchoxx (commented):
Great work mate 🚀🚀

I left you some annotations that you can omit if you want!

marc-hughes:
Hi @gastonabella !

There's a second graphical timeline based scheduler interface that instructors can use to set the due date, it looks like the screenshot below.

Could you add support for the new available date to the API's it uses, and I'll update that interface to show/edit this new setting?

It uses GET/PUT to a route like: /api/v1/scheduling/chemistry_20230627


image

lib/oli/delivery/settings.ex
gastonabella: I like short `if`s :)

lib/oli/delivery/settings.ex
gastonabella: Done!

gastonabella:
> Hi @gastonabella !
> 
> There's a second graphical timeline based scheduler interface that instructors can use to set the due date, it looks like the screenshot below.
> 
> Could you add support for the new available date to the API's it uses, and I'll update that interface to show/edit this new setting?
> 
> It uses GET/PUT to a route like: /api/v1/scheduling/chemistry_20230627
> 
> This should be a new field, not a new scheduling_type value since we should be able to set the availability AND the due or suggested-by date at the same time.
> 
> image

Hi @marc-hughes !

I was aware of that page and created a ticket for making the updates ([MER-2536](https://eliterate.atlassian.net/browse/MER-2536)), but I agree it makes more sense that you work on the React side and I can do the Elixir part. I've just assigned you to that ticket and created a new one for the API changes ([MER-2540](https://eliterate.atlassian.net/browse/MER-2540)). Does it make sense?

lib/oli_web/live/sections/assessment_settings/settings_table_model.ex
```diff
@@ -141,6 +150,29 @@ defmodule OliWeb.Sections.AssessmentSettings.SettingsTableModel do
...
+
+    ~H"""
+    <button
+      class="hover:underline whitespace-nowrap"
+      type="button"
+      phx-click={edit_date_and_show_modal(@on_edit_date, "available_date")}
+      phx-value-assessment_id={@id}
+    >
+      <%= if is_nil(@available_date) do %>
+        No available date
```
darrensiegel: I think we should change "No available date" to "Always available" as it better conveys to the instructor the state that an assessment is in when they have not set an available date.

darrensiegel: I pushed that change.

darrensiegel (approved):
One minor change request (wording of "No available date") which I pushed already.  

Great work!

lib/oli_web/live/sections/assessment_settings/settings_table_model.ex
gastonabella: Sounds better! Thanks

darrensiegel (approved):

PR #4186, opened 2023-09-08
[MER-2528] Extend revision history tool

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-2528) to the ticket

## Suggestion

For "Curriculum" and "Objectives" liveviews, links to view revision history are shown for admin users that add the "#show_links" to the URL (as the ticket says).
I suggest showing the revision history links for admins without them having to add the "#show_links" to the URL, to speed up their experience.
Please let me know if you want me to implement this suggested behavior.

### Navigate through resources and go back to the root resource (curriculum)

https://github.com/Simon-Initiative/oli-torus/assets/74839302/cf2d65cb-be13-462e-ba05-9ace9b75e933

### Accessing through different pages (objectives, curriculum and activity bank)

https://github.com/Simon-Initiative/oli-torus/assets/74839302/6e592033-53c2-4b2f-9563-f1b67592a401

https://github.com/Simon-Initiative/oli-torus/assets/74839302/98cab7e5-eb23-489c-a45e-29c62cf2bcfe

https://github.com/Simon-Initiative/oli-torus/assets/74839302/3d08609b-455d-4579-871c-a246795c93ab

### Editing fields (of any revision) with flash messages for error and success

https://github.com/Simon-Initiative/oli-torus/assets/74839302/21fe2f78-ed0a-4c5e-a98d-a7125d24da5a


lib/oli_web/live/history/revision_history.ex
```diff
@@ -360,6 +361,64 @@ defmodule OliWeb.RevisionHistory do
     end
   end
 
+  def handle_event("reset-monaco", _params, socket) do
+    {:noreply, assign(socket, details_modal_assigns: nil)}
+  end
+
+  def handle_event("edit-attribute", %{"attr-key" => attr_key}, socket) do
```
gastonabella, line 368: Is there a convention for why we define some events with dashes and some with underscores?

lib/oli_web/live/history/revision_history.ex
```diff
@@ -360,6 +361,64 @@ defmodule OliWeb.RevisionHistory do
...
+      assign(socket,
+        details_modal_assigns: %{
+          title: attr_key,
+          key: String.to_existing_atom(attr_key)
+        }
+      )
+
+    # we want to ensure monaco editor is reseted correctly
+    # before assigning new values and showing the modal
+    :timer.sleep(200)
```
gastonabella, line 379: The `reset-monaco` event is dispatched before this event so why is this necessary?

lib/oli_web/live/history/details.ex
```diff
@@ -1,76 +1,258 @@
...
+  def render_attribute(
+        %{
+          key: key,
+          keys: keys,
+          value: value,
+          project_slug: project_slug,
+          slug_mapper: slug_mapper
+        } = assigns
+      ) do
+    if key in keys do
```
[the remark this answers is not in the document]
nicocirio: Yes, makes totally sense, thanks! 😄

lib/oli_web/live/history/revision_history.ex
nicocirio: I had the same thought as you and, while developing the feature and testing it locally, sometimes I opened the modal for a new edit and got the values of the previous edited field. 
To be honest, I really do not like having to add a `timer.sleep`. But from the user's perspective, this delay is unnoticeable, the modal seems to open immediately. And having to close the modal and reopen it when you get the wrong values is really annoying.

lib/oli_web/live/history/details.ex
```diff
@@ -1,76 +1,258 @@
...
+    :children,
+    :objectives,
+    :relates_to
+  ]
+
+  def update(assigns, socket) do
+    slug_mapper =
+      fetch_resource_ids(@linkable_keys, assigns.revision)
+      |> Resources.map_slugs_from_resources_ids()
+      |> Enum.into(%{})
```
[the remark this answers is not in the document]
nicocirio: You are right about returning duplicated resource_id with different slugs. For example, if I change a page title for the resource_id=54, `Resources.map_slugs_from_resources_ids()` will return two different slugs:
```
[{54, "new_page_95luq"}, {54, "changed_name"}]
```

And the result of the `Enum.into(%{})` will be: 

```
%{54 => "changed_name"}
```

Luckily, the way the mount function fetches the revisions accepts any of the slugs, fetches the first revision and finally fetches all the revisions for that resource. So we can use any of the revision slugs


https://github.com/Simon-Initiative/oli-torus/assets/74839302/667e19ac-57b1-4432-8c3b-e8a7df7dbaf2

BTW, reviewing the code to check this I found I had a hardcoded project_slug that I just fixed! so thanks! 🙌

gastonabella (approved):
LGTM! Thanks for doing the changes 🚀

darrensiegel (approved):
This is great, nice work!

PR #4191, opened 2023-09-11
[FEATURE][MER-2501] archive products

rgachuhi:
This PR adds a feature to allow project authors or admins to manage the list of products listed in the products list view by toggling the view or hiding archived products.


lib/oli_web/live/products/products_view.ex
```diff
@@ -147,6 +163,42 @@ defmodule OliWeb.Products.ProductsView do
...
+
+    socket = assign(socket,
+      include_archived: include_archived,
+      products: products,
+      total_count: total_count,
+      table_model: table_model
+    )
+
+    {:noreply,
+     push_patch(socket,
```
darrensiegel, line 189: It doesn't make sense to call `push_patch` here unless you've implemented also the `handle_params`.  We are going to file another ticket to convert this view over to a fully DB driven table with URL param support, so for this PR, just change this method to return `{:noreply, socket}` after you've updated the socket assigns.

darrensiegel (changes requested):
I had two comments, see inline.  

But I'm realizing now that this `products_view.ex` LiveView does in memory table operations (sorting, filtering).  This is perfectly fine for individual course projects which likely never have more than a dozen products, but it can be problematic for the "All Products" view which this view also powers.  It might make sense to convert this view implementation over to using a DB powered table.  Same thing applies for the "Payments" view.  I'll file another ticket for these conversions.

lib/oli/delivery/sections/blueprint.ex
```diff
@@ -350,6 +339,25 @@ defmodule Oli.Delivery.Sections.Blueprint do
...
 
+  def list(%BlueprintBrowseOptions{} = options) do
+    filter_by_project =
+      if is_nil(options.project_id),
+        do: true,
+        else: dynamic([s], s.base_project_id == ^options.project_id)
+
+    filter_by_status =
+      if options.include_archived,
+        do: dynamic([s], s.status in [:active, :archived]),
```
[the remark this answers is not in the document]
rgachuhi: There are three valid cases :active, :deleted and :archived. We don't want to include the :deleted case hence the reason I am using dynamic

lib/oli_web/live/products/products_view.ex
rgachuhi: `handle_params` is implemented in the `use OliWeb.Common.SortableTable.TableHandlers`, which extends the `product_view`. This implementation is designed to work well (play well) with the other "filtering" conditions. i.e. you can filter and either include or exclude `:archived` products as part of the filter. If not that nuance gets lost

darrensiegel: `handle_params` is not able to handle "include_archived" - so the issue is that, despite the fact that "push_path" is being called on update to "include_archived", it isn't pushed as a URL param and handled.  To see this, do this:

1. Go into products view. 
2. Toggle "Include archived" (see the archived results appear)
3. Directly navigate away from page to another page. 
4. Use browser back button to return. 

Expected: Include Archived (since it was a path param) is preserved
Actual: it is not

It probably is okay that it is not beign handled, there is another ticket to redo that aspect of this view.  But since we aren't handling it with "handle_params" - we shouldn't push it as a path param.

rgachuhi: I have tried these same steps on several browsers. Navigating back via the back button works fine no matter where else I navigate. `include_archived` state is correctly preserved and processed. The reason for this is that it is maintained on the mount calls as well.

darrensiegel (approved):

PR #4192, opened 2023-09-11
[MER-2541] [ENHANCEMENT] Allow configurable max header value length

bitbldr:
Adds a configuration env variable `HTTP_MAX_VALUE_HEADER_LENGTH` for Cowboy http server max header value length. Configuring to a larger value should resolve some of the HTTP 431 errors being returned for users.

If no value is set, then a value of 4096 will be used, which is the default value listed in the Cowboy docs https://ninenines.eu/docs/en/cowboy/2.5/manual/cowboy_http/

### DIFF
--- a/assets/webpack.config.js
+++ b/assets/webpack.config.js
@@ -63,7 +63,8 @@ const populateEntries = () => {
   const styleSheets = [
     { adaptive: './styles/adaptive/adaptive-reset.scss' },
     { styles: './styles/index.scss' },
-    { preview: './styles/preview.scss' }];
+    { preview: './styles/preview.scss' },
+  ];
 
   // Merge the attributes of all found activities and the initialEntries
   // into one single object.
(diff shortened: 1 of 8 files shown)


darrensiegel (changes requested):
We probably should expose these other two configurable settings: 

```
 protocol_options: [
      max_header_name_length: 64,
      max_header_value_length: 4096,
      max_headers: 100
    ]
```

`max_headers` and `max_header_name_length`

assets/webpack.config.js
```diff
@@ -319,7 +320,7 @@ module.exports = (env, options) => ({
       outputFilename: '../licenses.json',
       licenseOverrides: {
         'janus-script@1.9.2': 'MIT',
-        'phoenix_html@3.2.0': 'MIT',
+        'phoenix_html@3.3.1': 'MIT',
```
bitbldr, line 323: I noticed a warning related to this license version during assets build, this resolves it

darrensiegel (approved):

PR #4205, opened 2023-09-13
[Chore] Build issue [MER-2543] 

marc-hughes:
There was a build issue causing resulting builds to be unusable. That issue was silently ignored by our current build process. This PR fixes both of those issues.

**Note:** The final step in testing this PR will be to merge to master and verify the resulting build actually works on tokamak.

### DIFF
--- a/.github/actions/torus-builder/entrypoint.sh
+++ b/.github/actions/torus-builder/entrypoint.sh
@@ -14,9 +14,7 @@ MIX_ENV=prod SHA=$RELEASE_SHA mix compile
 
 yarn --cwd ./assets
 NODE_ENV=production npm run deploy --prefix ./assets
-echo $?
 NODE_ENV=production npm run deploy-node --prefix ./assets
-echo $?
 
 MIX_ENV=prod mix assets.deploy
 MIX_ENV=prod SHA=$RELEASE_SHA mix release
--- a/.github/actions/torus-builder/entrypoint.sh
+++ b/.github/actions/torus-builder/entrypoint.sh
@@ -1,17 +1,22 @@
 #!/bin/sh
 
+set -e
+set -x
+
 RELEASE_SHA=$1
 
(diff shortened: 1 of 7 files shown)


.github/actions/torus-builder/entrypoint.sh
```diff
@@ -1,10 +1,13 @@
 #!/bin/sh
 
+set -e
```
marc-hughes, line 3: The `set -e` here was the fix for the build silently failing. Now, if any step in entrypoint.sh fails, the entire script will fail. Before, only the last command determined the exit status.

Also added -x so you can see what commands are being executed, makes debugging the build a bit easier.

.github/actions/torus-builder/entrypoint.sh
```diff
@@ -1,10 +1,13 @@
...
+set -x
+
 RELEASE_SHA=$1
 
 mix local.hex --force
 mix local.rebar --force
-mix archive.install hex phx_new 1.5.9
+# mix archive.install hex phx_new 1.5.9 --force
```
marc-hughes, line 10: Talking with Eli, this line was no longer needed, and it was making local testing of the docker image difficult.

assets/src/components/editing/markdown_editor/markdown_util.ts
```diff
@@ -64,7 +64,7 @@ export const getMarkdownWarnings = (model: (AllModelElements | FormattedText)[])
     }
   });
 
-  const uniqueWarnings = [...new Set(warnings)];
+  const uniqueWarnings = Array.from(new Set(warnings));
```
marc-hughes, line 67: This, and the change over in factories.ts were the build-time errors that were not previously detected. I did attempt a build with them broken to verify those builds will now fail.

bitbldr (approved):

PR #4207, opened 2023-09-14
[BUGFIX][MER-2549] Change error message existing authoring collaborator

simonchoxx:
[MER-2549](https://eliterate.atlassian.net/browse/MER-2549)

This PR adds a check to know if the collaborator(s) you want to invite to a project already exist in the project.

In this case, an error message is displayed indicating that this person is already a collaborator in the project. 
 
Captura de pantalla 2023-09-14 a la(s) 13 19 50

[MER-2549]: https://eliterate.atlassian.net/browse/MER-2549?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/oli_web/controllers/collaborator_controller.ex
```diff
@@ -26,45 +27,88 @@ defmodule OliWeb.CollaboratorController do
...
-            add_collaborator(conn, email, project_id, failures)
+        authors =
+          authors_emails
+          |> String.split(",")
+          |> Enum.map(&String.trim/1)
+          |> Enum.filter(fn e -> e != "" end)
+
+        is_collaborator? =
+          Enum.all?(emails, fn elem ->
+            Enum.member?(authors, elem)
```
nicocirio, line 38: Since we are not giving detailed feedback about the email/emails that is/are already a collaborator, we could stop the iteration as soon as we get the first email that is contained in the authors list. So, a `reduce_while` could avoid the full iteration in many cases (except the case where the last email element is the one contained in the author's list)

`Enum.reduce_while(emails, false, fn email, acc -> if email not in authors, do: {:cont, acc}, else: {:halt, true} end)`

nicocirio (approved):
LGTM! 🚀 
I left a NIT about another way of checking if an email is already a collaborator. But, to be honest, I think its performance improvement is going to be imperceptible unless we expect to have extensive lists of collaborators and/or extensive lists of emails.

lib/oli_web/controllers/collaborator_controller.ex
simonchoxx: I understand what you are telling me and it seems like a great solution, but in this part of the code I want to identify if all the emails entered to send invitations are included in the existing collaborators. 

In the code, below, there is a case in which some of the emails match with some existing collaborator then another error message is displayed.

darrensiegel (approved):

PR #4211, opened 2023-09-14
[MER-2547] Preferred time for Scheduler

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-2547) to the ticket

#### Setting the preferred time on the course creation wizard + editing its value after the course was created:

https://github.com/Simon-Initiative/oli-torus/assets/74839302/83ec111c-9bef-4db7-a1b9-925d8ad69f20

@marc-hughes , we talked with @darrensiegel and wanted to ask you for some help to address the React side of this ticket: 

```
The value should be passed into the Scheduler, and the Scheduler must 
use that time value when setting new date/times.  
But if an existing page has a time other than the preferred, that existing time should be preserved.  
```

I pass the value from the Elixir side to the Scheduler [here](https://github.com/Simon-Initiative/oli-torus/blob/93aed9cd731d0156c042ed83ebacd4ecb50e69f1/lib/oli_web/live/sections/schedule_view.ex#L38), and left an initial step done where we receive it from the component side [here](https://github.com/Simon-Initiative/oli-torus/blob/93aed9cd731d0156c042ed83ebacd4ecb50e69f1/assets/src/apps/scheduler/ScheduleEditor.tsx#L15) (as a placeholder).

Could you please help us address the changes from the React components side?
Please let me know if you need something else.

Thanks! 😄


marc-hughes:
I can help with that. Probably makes sense to land https://github.com/Simon-Initiative/oli-torus/pull/4208 first, because it introduces the availability-time which should use this preferred time as well.

marc-hughes:
The preferred time should be used by default in the scheduler app now. This mainly works in two cases:

1. A user uses the reset-schedule option
2. A user changes a date from a type without a time to "due date", the previously selected date will appear with the new preferred  time. For example, if an entry is **suggested by** "1/1/2023", changing it to **due-by** will set it to that date at the preferred time ex "1/1/2023 3:00PM". Previously, it would get set to "1/1/2023 23:59:59"

The preferred time will NOT be used if the user edits a blank date/time input. The default behavior of the system date/time input is to use the current time when editing the field with no date set. I believe we would have to replace the input with a custom control to make the preferred time default here. 

preferredtime

darrensiegel (approved):
Tested, works well!

PR #4230, opened 2023-09-19
[BUGFIX][MER-2552] Admin not authorized to view student pages

simonchoxx:
[MER-2552](https://eliterate.atlassian.net/browse/MER-2552)

This PR allows admin users to access student views in preview mode if they access any of the URLs of a student's course.

A plug is added to the delivery sections pipeline within the router that identifies if the current user is an administrator so that they can be redirected to a preview mode view. 

[MER-2552]: https://eliterate.atlassian.net/browse/MER-2552?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/oli_web/plugs/redirect_if_is_admin.ex
```diff
@@ -0,0 +1,18 @@
...
+  end
+
+  def call(conn, _params) do
+    %{current_author: current_author, current_user: current_user, section: section} = conn.assigns
+
+    if not is_nil(current_author) and is_nil(current_user),
+      do:
+        redirect(conn,
+          to: Routes.page_delivery_path(OliWeb.Endpoint, :index_preview, section.slug)
+        ),
```
augusto1024, line 15: I'm getting this error when the redirect happens because it keeps going down the other plugs. I think we should pipe this with a [Plug.Conn.halt/1](https://hexdocs.pm/plug/Plug.Conn.html#halt/1)
image

lib/oli_web/plugs/redirect_if_is_admin.ex
```diff
@@ -0,0 +1,18 @@
+defmodule Oli.Plugs.RedirectIfIsAdmin do
```
augusto1024, line 1: The name of the module isn't very representative of what it does (we know that it redirects the admin but we don't know where). WDYT about changing it to something like `DeliveryPreview` or something like that?

augusto1024 (commented):
Nice! I just left two small comments that I think we should address 😀

augusto1024 (approved):
LGTM! Thanks for the changes

PR #4244, opened 2023-09-25
[BUGFIX][MER-2570] Fix error accessing project preview

simonchoxx:
[MER-2570](https://eliterate.atlassian.net/browse/MER-2570)

Fixed a couple of bugs when rendering the project preview. 

- Error when calling `DeliveryResolver` instead of calling `AuthoringResolver`. This was failing when you have one selection criteria for an activity bank by objective and there is no objective added in the activity banks.

- Error when applying a selection criteria by item type and there is no such activity in the activity bank.

[MER-2570]: https://eliterate.atlassian.net/browse/MER-2570?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli/rendering/content/selection.ex
+++ b/lib/oli/rendering/content/selection.ex
@@ -94,7 +94,15 @@ defmodule Oli.Rendering.Content.Selection do
         %{}
 
       ids ->
-        Oli.Publishing.AuthoringResolver.from_resource_id(section_slug, ids)
+        revisions =
+          if Enum.any?(
+               Oli.Publishing.AuthoringResolver.from_resource_id(section_slug, ids),
+               fn elem -> is_nil(elem) end
+             ),
+             do: Oli.Publishing.DeliveryResolver.from_resource_id(section_slug, ids),
+             else: Oli.Publishing.AuthoringResolver.from_resource_id(section_slug, ids)
+
+        revisions
         |> Enum.reduce(%{}, fn rev, m -> Map.put(m, rev.resource_id, rev.title) end)
     end
   end
@@ -139,6 +147,8 @@ defmodule Oli.Rendering.Content.Selection do
     |> Enum.join(" ")
   end
 
-  defp titles(id, title_map, class),
-    do: "<span class=\"#{class}\">#{Map.get(title_map, id)}</span>"
+  defp titles(id, title_map, class) do
+    unless Map.equal?(title_map, %{}),
+      do: "<span class=\"#{class}\">#{Map.get(title_map, id)}</span>"
+  end
 end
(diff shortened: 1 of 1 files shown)


lib/oli/rendering/content/selection.ex
```diff
@@ -94,7 +94,7 @@ defmodule Oli.Rendering.Content.Selection do
         %{}
 
       ids ->
-        Oli.Publishing.DeliveryResolver.from_resource_id(section_slug, ids)
+        Oli.Publishing.AuthoringResolver.from_resource_id(section_slug, ids)
```
gastonabella, line 97: I'm not sure if I'm understanding this change. The `DeliveryResolver` function expects a section slug while the `AuthoringResolver` one expects a project slug, so why are we switching functions here?

lib/oli/rendering/content/selection.ex
```diff
@@ -129,8 +129,10 @@ defmodule Oli.Rendering.Content.Selection do
     |> Enum.join(" ")
   end
 
-  defp activity_names(id, activity_types_map),
-    do: "<span class=\"activity-name\">#{Map.get(activity_types_map, id).title}</span>"
+  defp activity_names(id, activity_types_map) do
+    if not Map.equal?(activity_types_map, %{}),
```
gastonabella, line 133: Nit but you could use `unless` instead

lib/oli/rendering/content/selection.ex
```diff
@@ -94,7 +94,15 @@ defmodule Oli.Rendering.Content.Selection do
         %{}
 
       ids ->
-        Oli.Publishing.DeliveryResolver.from_resource_id(section_slug, ids)
+        revisions =
+          if Enum.any?(
```
darrensiegel, line 98: This selection renderer is unique - as it only ever gets called during page preview.  In normal delivery, by the time we get to rendering the activity bank selections have been "fulfilled" and have been replaced by actual activities.  For page preview, we do not fulfill bank selections, we instead leave them in the page model and this rendering code provides a summary of them to the author.  So that being said, we never want to use the `DeliveryResolver` in this code, we only want to resolve to the Authoring revisions. 

I suspect the use of DeliveryResolver was the problem and what was yielding the `nil` ids.  I think we should simplify this code to:

```
 Oli.Publishing.AuthoringResolver.from_resource_id(section_slug, ids). # This section_slug really is the project slug
|> Enum.filter(fn r -> !is_nil(r) end)
|> Enum.reduce(%{}, fn rev, m -> Map.put(m, rev.resource_id, rev.title) end)
```

darrensiegel (changes requested):

darrensiegel (approved):

PR #4259, opened 2023-09-28
[BUGFIX] [MER-2612] Scoring in assessment settings defaulting to average

gastonabella:
[MER-2612](https://eliterate.atlassian.net/browse/MER-2612)

This PR fixes the default scoring strategy shown on the Assessment Settings page when an assessment has no scoring strategy set.

Since there was not an option in the select for when the scoring strategy is not set (is `nil`), the select was taking the first option (`Average`) as the default one.

[MER-2612]: https://eliterate.atlassian.net/browse/MER-2612?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli/delivery/sections.ex
+++ b/lib/oli/delivery/sections.ex
@@ -1506,7 +1506,6 @@ defmodule Oli.Delivery.Sections do
     section_resources = [
       # The below is necessary because Repo.insert_all/3 receives a map, not a struct
       # The below is necessary because Repo.insert_all/3 doesn't autogenerate values
-      # TODO: CAMBIAR ESTO
       %SectionResource{
         numbering_index: numbering_index,
         numbering_level: level,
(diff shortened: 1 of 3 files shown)


darrensiegel (changes requested):
There might be a deeper problem here: Every SectionResource, when created, should have all of its assessment settings set to default values.  None of them should be `nil`.  @gastonabella Can you do some testing, add some IO.inspects, to see what these SectionResource attributes for scoring strategy (and others) look like?

gastonabella:
> There might be a deeper problem here: Every SectionResource, when created, should have all of its assessment settings set to default values. None of them should be `nil`. @gastonabella Can you do some testing, add some IO.inspects, to see what these SectionResource attributes for scoring strategy (and others) look like?

@darrensiegel after some research, I was able to reproduce a section with resources having a `nil` scoring strategy. The origin of the issue was product creation (which was later propagated to sections created from products). When bulk-inserting all the section resources, the `SectionResource.to_map/1` function being used in an inner step was missing the `scoring_strategy_id`.
This probably caused many section resources to have `nil` scoring strategies. How do you want to handle that?
I think we could create a migration that adds a null restriction on the field and set `"Best"` as the default value for the broken resources.

darrensiegel (approved):

PR #4274, opened 2023-10-03
[BUG FIX] Remove author specific columns [MER-2623]

darrensiegel:
Due to some very specific logic in the table model, a few columns were not rendering if the user had an active authoring session.  I've removed that logic, allowing "Last Interacted" and all the other columns to always render.


gastonabella (approved):
LGTM!

PR #4280, opened 2023-10-03
[BUG FIX] Empty slugs [MER-2625]

darrensiegel:
If a title of a revision only contains non-alphanumeric characters, the slug generated was the empty string, which was downstream breaking assertions about uniqueness.  This edge case is handled by generating a random slug.

### DIFF
--- a/lib/oli/utils/slug.ex
+++ b/lib/oli/utils/slug.ex
@@ -172,12 +172,23 @@ defmodule Oli.Utils.Slug do
   def slugify(nil), do: ""
 
   def slugify(title) do
-    String.downcase(title, :default)
+    case String.downcase(title, :default)
     |> String.trim()
     |> String.replace(" ", "_")
     |> alpha_numeric_only()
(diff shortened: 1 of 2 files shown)


nicocirio (approved):
LGTM!

lib/oli/utils/slug.ex
```diff
@@ -172,12 +172,23 @@ defmodule Oli.Utils.Slug do
...
 
   def slugify(title) do
-    String.downcase(title, :default)
+    case String.downcase(title, :default)
     |> String.trim()
     |> String.replace(" ", "_")
     |> alpha_numeric_only()
     |> URI.encode_www_form()
-    |> String.slice(0, 30)
+    |> String.slice(0, 30) do
```
gastonabella, line 180: Nit but I think this long case clause could be more readable by piping the result into a new function or even piping it into the case like `|> case do`

gastonabella (approved):
Looks good!

PR #4289, opened 2023-10-06
[Feature] Custom scoring options [MER-2586]

marc-hughes:
Opening this as a draft, we need some backend support for the evaluation portions.

This PR adds new scoring options to the authoring interface to capture the number of points a correct or incorrect answer will award.
image

It also adds the ability to specify the points of targeted feedback.
image

In multi-input questions, you can specify the scoring strategy:
image

### DIFF
--- a/assets/src/components/activities/check_all_that_apply/CheckAllThatApplyAuthoring.tsx
+++ b/assets/src/components/activities/check_all_that_apply/CheckAllThatApplyAuthoring.tsx
@@ -26,6 +26,7 @@ import { VariableEditorOrNot } from '../common/variables/VariableEditorOrNot';
 import { VariableActions } from '../common/variables/variableActions';
 import * as ActivityTypes from '../types';
 import { CATAActions } from './actions';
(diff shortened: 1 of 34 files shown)


darrensiegel (changes requested):
This all looks good impl-wise, but I did catch one bug which we need to fix.  Here are the steps to reproduce:

1. Add a multi input question to an assessment.
2. For the input that gets created, uncheck the "Use Default Scoring". (you will see that the "Scoring" view doesn't actually display anything, we should fix this too).  Check it and uncheck it to see the two inputs for setting the Points for Correct and Points for Incorrect.
3. Do not change anything, leave it as 1 point for correct and 0 for incorrect.
4. Publish and take the quiz.  It will not calcualte the score correctly. 

If you look at the Revision for this activity you will see that for both responses that exist in this part, the `score` is set to `0`.  The "Correct" response in this case is not getting its `score` initialized to `1`.  

Changing the points in UI fixes the issue.

darrensiegel (approved):

PR #4290, opened 2023-10-06
[BUG FIX] [MER-2643] Creating new institution with same issuer and client_id fails

bitbldr:
Fixes an issue where admin attempting to create a new institution when an existing registration with the same issuer and client_id crashes.

I tested simply dropping the index constraint but unfortunately it does not resolve the issue. While it does allow the new institution and registration to be created, it creates a new issue on LTI launch where the system expects only a single registration with issuer and client_id to exist. A code change is required to fix this.

The new approach taken here is for the system to check on institution creation if a registration already exists for a given issuer client_id and if so, continue to create the new institution but reuse the existing registration. The system was already doing this in a previous code path but the enhancements made to allow institution customization were not using the `find_or_create_registration` function.

I also made a couple of easy QOL style enhancements/bug fixes while I was here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### Bug Fixes
 
+- Fix an issue where admin attempting to create a new institution when an existing registration with the same issuer and client_id fails
+
 ## 0.25.0 (2023-10-5)
 
 ### Enhancements
--- a/lib/oli/institutions.ex
+++ b/lib/oli/institutions.ex
@@ -650,7 +650,7 @@ defmodule Oli.Institutions do
                institution_id: institution.id,
                tool_jwk_id: active_jwk.id
              }),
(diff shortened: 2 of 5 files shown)


nicocirio (approved):
LGTM.
Regarding flash messages, I saw the sticky behavior is not looking so good for components and liveviews, so I created a [ticket for v0.26](https://eliterate.atlassian.net/browse/MER-2651?filter=10069&jql=project%20%3D%20MER%20AND%20issuetype%20in%20(Bug,%20Story)%20AND%20status%20in%20(%22In%20Progress%22,%20PR,%20Ready,%20%22To%20Do%22)%20AND%20labels%20%3D%20wyeworks%20ORDER%20BY%20priority%20DESC,%20status%20ASC,%20created%20DESC) (should we address it as a hotfix for v25?)

darrensiegel (approved):

PR #4298, opened 2023-10-10
[MER-2654] Implement where you left off option

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-2645) to the ticket

https://github.com/Simon-Initiative/oli-torus/assets/74839302/72511889-f062-4be2-a647-03e2ae85bb95


darrensiegel (approved):
Looks good. Tested and works well, handles the case of deleted page

PR #4304, opened 2023-10-11
[MER-2644] Allow changing of table page sizes for instructor dashboard tables

gastonabella:
[MER-2644](https://eliterate.atlassian.net/browse/MER-2644)

This PR enables changing a table page size in all the Instructor Dashboard tables. By default, the page size is 20.
When the user is on a specific page and updates the page size, the offset is recalculated so that the user still sees the same table elements.
It also fixes a bug in the pagination component that allows moving to infinite (even negative) pages.

The page size options in the videos below are not the right ones. They were updated for testing purposes, but they should be 10, 20, 50, and 100.

**Tables with new page size dropdown:**

https://github.com/Simon-Initiative/oli-torus/assets/26532202/120998e2-8d6f-4aa9-821a-87abd86292fb

**Pagination before the bugfix:**

https://github.com/Simon-Initiative/oli-torus/assets/26532202/4847e1b7-ddef-487d-8587-2f0b9e74b81d

[MER-2644]: https://eliterate.atlassian.net/browse/MER-2644?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli_web/components/delivery/content/content.ex
+++ b/lib/oli_web/components/delivery/content/content.ex
@@ -10,7 +10,7 @@ defmodule OliWeb.Components.Delivery.Content do
 
   @default_params %{
     offset: 0,
-    limit: 25,
+    limit: 20,
     container_id: nil,
     sort_order: :asc,
     sort_by: :numbering_index,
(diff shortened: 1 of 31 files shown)


darrensiegel (approved):
Tested, looks great and works perfectly.  Nice work!

PR #4315, opened 2023-10-12
[FEATURE] Disable DataShop download [MER-2662]

darrensiegel:


bitbldr:
https://eliterate.atlassian.net/browse/MER-2663

Issue #4158, opened 2023-08-29
Need to Re-enroll Unenrolled Students and to Suspend Students


kwahlenmayer:
Need to re-enroll a previously un-enrolled student, allowing them to continue with their previous progress.
(What happens now if we unenroll a student and then re-invite them to the same section and they sign in with the same account? Is their previous work there or not?)

There are several situations where we re-enroll previously unenrolled students and expect progress to be restored (accident, changed mind, schedule change, payment dispute, a school can ask us for student's progress from forever ago, and more). 

Further, there are several situations where it is best to preserve the UI of a previously enrolled student (i.e. Suspended enrollment status) (payments, history, accident, changed mind, grade disputes at their institution, preserving payment / enrollment history (how many paid), invoicing, and more). This requires a new enrollment status of Suspended.

bitbldr:
The underlying functionality exists, we just need a UX to expose this ability to an admin or instructor

PR #4324, opened 2023-10-17
[MER-2589] Re-enroll students from Student Dashboard

gastonabella:
[MER-2589](https://eliterate.atlassian.net/browse/MER-2589)

Addresses https://github.com/Simon-Initiative/oli-torus/issues/4158

Allow instructors to re-enroll previously suspended students. The Actions tab in the Student Dashboard is no longer hidden for instructors when a student is suspended. Instead, it now has a **"Re-enroll"** button.

https://github.com/Simon-Initiative/oli-torus/assets/26532202/e8ad8698-7f32-4e90-b0b6-c5a0d88d490a

This also fixes an issue on the Modal function component whose css classes were not properly set on Dark mode.

Modal before the fix:

Screenshot 2023-10-19 at 2 24 28 PM

[MER-2589]: https://eliterate.atlassian.net/browse/MER-2589?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli_web/components/delivery/actions/actions.ex
+++ b/lib/oli_web/components/delivery/actions/actions.ex
@@ -23,14 +23,16 @@ defmodule OliWeb.Components.Delivery.Actions do
         } = _assigns,
         socket
       ) do
-    %{enrollment: enrollment, user_role_id: user_role_id, current_user: current_user} =
+    %{
+      enrollment: enrollment,
+      user_role_id: user_role_id,
+      current_user: current_user,
+      is_suspended?: is_suspended?
+    } =
       enrollment_info
 
     has_payment =
-      case Repo.get_by(Payment, enrollment_id: enrollment.id) do
-        nil -> false
-        _ -> true
-      end
+      !is_suspended? and !is_nil(Paywall.get_payment_by(enrollment_id: enrollment.id))
 
     {:ok,
      assign(socket,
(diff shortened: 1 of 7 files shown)


lib/oli_web/components/delivery/actions/actions.ex
```diff
@@ -145,6 +174,27 @@ defmodule OliWeb.Components.Delivery.Actions do
...
+           to:
+             Routes.live_path(
+               OliWeb.Endpoint,
+               OliWeb.Delivery.InstructorDashboard.InstructorDashboardLive,
+               socket.assigns.section.slug,
+               :manage
+             )
+         )}
+
+      _ ->
```
nicocirio, line 193: Should we handle the error case? Perhaps with a flash message?

lib/oli_web/components/delivery/actions/actions.ex
```diff
@@ -55,71 +57,98 @@ defmodule OliWeb.Components.Delivery.Actions do
...
-            <span class="text-xs text-gray-400 dark:text-gray-950">Apply bypass payment</span>
-          </div>
-          <button
-            class="torus-button flex justify-center primary h-9 w-48"
-            disabled={@has_payment}
-            phx-click="display_bypass_modal"
-            phx-target={@myself}
+      <%= if @is_suspended? do %>
+        <div id="unenrolled_student_actions">
+          <.live_component
```
nicocirio, line 62: NIT: instead of using a LiveComponent for the modal we can use the modal in its function component flavour `lib/oli_web/components/modal.ex`

lib/oli_web/components/delivery/actions/actions.ex
gastonabella: Makes sense! Done :)

nicocirio (approved):

darrensiegel (approved):

PR #4327, opened 2023-10-18
[BUGFIX] Sort attempts by date submitted

nicocirio:
This PR fixes the sorting of the submitted attempts by using `DateTime.before?/2` instead of directly comparing the two datetimes.

## Before
Screenshot 2023-10-18 at 11 20 11

## After
Screenshot 2023-10-18 at 11 25 01


gastonabella (approved):
🪄

PR #4332, opened 2023-10-19
[CHORE] fix tests and warnings

darrensiegel:


bitbldr (approved):

config/config.exs
```diff
@@ -198,19 +198,6 @@ config :lti_1p3,
...
   ags_line_item_prefix: "oli-torus-"
 
-
-config :openai,
-  # find it at https://platform.openai.com/account/api-keys
-  api_key: System.get_env("OPENAI_API_KEY"),
-  # find it at https://platform.openai.com/account/org-settings under "Organization ID"
-  organization_key: System.get_env("OPENAI_ORG_KEY"),
-  # optional, passed to [HTTPoison.Request](https://hexdocs.pm/httpoison/HTTPoison.Request.html) options
-  http_options: [recv_timeout: :infinity, stream_to: self(), async: :once]
```
nicocirio, line 208: Based on[ the logs](https://github.com/Simon-Initiative/oli-torus/actions/runs/6578197033/job/17871405286), it seems `self()` is causing the error. I am not sure if it will work, but we could try with `inspect(self())`, wdyt?

PR #4333, opened 2023-10-19
[CHORE] elixir format and auto format check

bitbldr:
Add a CI check to auto format if possible and also require correct formatting in build step.

Most of the changes here are a result of simon-bot running the auto formatter.

This PR also includes a fix for a failing, non-deterministic tests for ingest_live, where the test would wait 1 second for the JSON validation to complete before checking the rendered html. Because of the validation performance issues, depending on the resources of the system and size of the digest, this could take longer than 1 second to complete and fail. I set the default of bypass JSON validation to be true and removed to toggle JSON validation, since this feature is not working correctly, we haven't kept up with the schemas, and it likely is only a burden on the system more-so than it is helpful.

Previewing the merge conflicts that will be caused in `nextgen-ux`, the conflicts appear to be manageable as long as we integrate master in as soon as we land this one.

Screenshot 2023-10-19 at 2 44 23 PM


nicocirio (approved):
This is really great! 😄

PR #4334, opened 2023-10-20
Nextgen: improve migration and use different dev db

nicocirio:
This PR improves the migration by:

- separating `change` into `up` and `down` to be able to correctly rollback if necessary
- populates existing sections with default page_prompt_template (if not, liveview crashes when student visits a page)

I am aware that we should not modify migrations but create new ones. However, since I understand this branch has not yet been deployed, I think we are still on time to make a modification in the migration.

I think that, since we are probably going to switch frequently between v26 and NG23, it would be handy to work with different dev databases (to avoid having to rollback or run migrations every time we switch). This PR renames the dev db into `oli_ng_dev`.

Finally, the top border to the Prompt Template group in overview was added:

### Before
Screenshot 2023-10-20 at 09 43 57

### After
Screenshot 2023-10-20 at 09 40 47


lib/oli_web/live/sections/overview_view.ex
```diff
@@ -487,13 +496,15 @@ defmodule OliWeb.Sections.OverviewView do
...
-    Oli.Delivery.Sections.update_section(section, %{page_prompt_template: socket.assigns.page_prompt_template})
+    Oli.Delivery.Sections.update_section(section, %{
+      page_prompt_template: socket.assigns.page_prompt_template
+    })
 
-    socket = socket
-    |> put_flash(:info, "Prompt successfully saved")
+    socket =
+      socket
+      |> put_flash(:info, "Prompt successfully saved")
```
gastonabella, line 507: IMO it's also more readable to avoid pipes when calling only one function.
I'd do `put_flash(socket, :info, "Prompt successfully saved")` instead.

[Reference](https://hexdocs.pm/credo/Credo.Check.Readability.SinglePipe.html)

gastonabella (approved):
🚀

darrensiegel (approved):

PR #4373, opened 2023-10-31
[BUGFIX][MER-2557] Precision not working as expected

simonchoxx:
[MER-2557](https://eliterate.atlassian.net/browse/MER-2557)

This PR modifies the way to check the precision of an activity input result.

To check the precision, we take the definition of `precision` given in the stack overflow link in the [ticket description](https://eliterate.atlassian.net/browse/MER-2557).

- As an example, the correct result for the following activity should be 0.36, with a precision of 2.

Captura de pantalla 2023-10-31 a la(s) 16 30 27

- Activity resolution.

https://github.com/Simon-Initiative/oli-torus/assets/16328384/5cd56647-1de4-4baa-a8fa-ad3ed14a10fe

[MER-2557]: https://eliterate.atlassian.net/browse/MER-2557?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli/delivery/evaluation/rule.ex
+++ b/lib/oli/delivery/evaluation/rule.ex
@@ -146,21 +146,23 @@ defmodule Oli.Delivery.Evaluation.Rule do
         ["0", decimal] ->
           decimal
           |> String.replace_leading("0", "")
-          |> String.split("")
-          |> Enum.filter(&is_digit?/1)
-          |> Enum.count()
+          |> count_digits()
 
         [_integer, _decimal] ->
-          str
-          |> String.split("")
-          |> Enum.filter(&is_digit?/1)
-          |> Enum.count()
+          count_digits(str)
 
         [integer] ->
-          integer |> String.split("") |> Enum.filter(&is_digit?/1) |> Enum.count()
+          count_digits(integer)
       end
 
     digit_count == count
+  end
+
+  defp count_digits(string_number) do
+    string_number
+    |> String.split("")
+    |> Enum.filter(&is_digit?/1)
+    |> Enum.count()
   end
 
   defp is_range?(str), do: String.starts_with?(str, ["[", "("])
--- a/lib/oli/delivery/evaluation/rule.ex
+++ b/lib/oli/delivery/evaluation/rule.ex
@@ -139,7 +139,7 @@ defmodule Oli.Delivery.Evaluation.Rule do
   # if a precision is not specified (nil) then always evaluate to true
   defp check_precision(_value, nil), do: true
 
-  # checks the precision, now interpreted as # of significant figures
+  # checks the precision, now interpreted as number of significant figures
   defp check_precision(str, count) when is_binary(str) do
     sigfigs =
       case String.split(String.downcase(str), "e") do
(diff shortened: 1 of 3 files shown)


lib/oli/delivery/evaluation/rule.ex
```diff
@@ -142,10 +142,23 @@ defmodule Oli.Delivery.Evaluation.Rule do
...
-      |> String.split("")
-      |> Enum.filter(&is_digit?/1)
-      |> Enum.count()
+      case String.split(str, ".") do
+        ["0", decimal] ->
+          decimal
+          |> String.replace_leading("0", "")
+          |> String.split("")
+          |> Enum.filter(&is_digit?/1)
+          |> Enum.count()
```
gastonabella, line 151: Nit but these functions are called in the three cases, so what do you think about moving them to a separate function? It'd result in something like:
```elixir
defp count_digits(string_number) do
  string_number
  |> String.split("")
  |> Enum.filter(&is_digit?/1)
  |> Enum.count()
end

...

case String.split(str, ".") do
  ["0", decimal] ->
    decimal
    |> String.replace_leading("0", "")
    |> count_digits()

  [_integer, _decimal] ->
    count_digits(str)

  [integer] ->
    count_digits(integer)
end
```

gastonabella (approved):
Looks great!

andersweinstein:
This fix duplicates legacy "precision". While that is one approach to this issue, legacy "precision" is a very odd and unmotivated concept. What our chemistry instructors really want ultimately is correct significant figures. In discussions the OLI team decided it would be better to go straight to true significant figures on this issue, though this was not recorded on the ticket. That would break legacy parity in a couple of cases that would need to be fixed, but give chemistry and other courses the feature they really want.

The difference from legacy 'precision' arises on integers with trailing zeros: for 3400 legacy 'precision' is 4, but significant figures is 2 because trailing zeros are presumed to be there only because required as placeholders. To indicate a value of 3400 to four significant figures you must write 3400. (including a decimal point) or use scientific notation 3.400e3

bitbldr:
Agree with @andersweinstein. We should implemented actual Significant Figures and update the instances where this might break backward compatibility with legacy, they should be few.

We should also change the label in authoring from "Precision" to "Significant Figures"

bitbldr (changes requested):
See discussion from @andersweinstein around implementing sig figs

andersweinstein:
I got into trying to code this so I've checked in my version. This still has an issue with user input of the form 3400. with decimal point, which should have 4 significant figures rather than 2. The decimal point is apparently getting dropped from the user input somewhere on the way to the precision evaluation, but have not investigated where. 

Also uncertain on desired behavior in the edge case of a string of all zero digits, e.g. 0.00. This code gives it three. It is unlikely any instructor will ever  want to test significant figures on an all zero value so it will not matter very much, but have a question out to one Chem instructor to check.

darrensiegel (approved):
We are landing this as-is.  Any needed tweaks for edge case handling will come in future tickets.

PR #4397, opened 2023-11-06
RTL text bug fixes [MER-2703]

marc-hughes:
There were three issues:

1. The `<div dir="rtl">` tag was being logically inserted at the wrong level of abstraction.
2. The popup element was being rendered as a block element instead of inline
3. The new div with the direction, was causing content within paragraph elements to be styled incorrectly

### DIFF
--- a/assets/src/components/common/DeliveryElementRenderer.tsx
+++ b/assets/src/components/common/DeliveryElementRenderer.tsx
@@ -28,11 +28,18 @@ interface DeliveryElementRendererProps {
 export const DeliveryElementRenderer: React.FC<DeliveryElementRendererProps> = ({
   element,
   context,
-  inline
+  inline,
 }) => {
   const writerContext = defaultWriterContext(context);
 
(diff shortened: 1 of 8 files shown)


assets/src/components/common/DeliveryElementRenderer.tsx
```diff
@@ -6,6 +6,7 @@ import { HtmlContentModelRenderer } from 'data/content/writers/renderer';
 interface DeliveryElementRendererProps {
```
marc-hughes, line 6: Adding an inline option here supports our popup element

lib/oli/rendering/content.ex
```diff
@@ -121,11 +121,8 @@ defmodule Oli.Rendering.Content do
         %{"type" => "content", "children" => children} = element,
```
marc-hughes, line 121: This fixes the direction at the wrong level of abstraction.

Note: the new markdown renderer that @darrensiegel is working on will need to add a content writer function.

darrensiegel (approved):

PR #4412, opened 2023-11-08
[BUG FIX] [MER-2676] Section creation from product fails

bitbldr:
Fixes 2 issues related to database requests exceeding maximum parameters:
1. On section creation from a product with a large amount of section resources (specifically, the REAL chem course pilot)
2. When applying updates to sections with a large amount of section resources

This PR extends the original `chunk_by` record insertion approach by generalizing for all cases where section resource records are being created/recreated. A new function called `batch_insert_all` is introduced which takes a list of row maps, and batches them into `insert_all` operations according to some maximum parameter value (for now, it is just hardcoded to the postgres limit we experienced)

Verified it fixes both of these issues on heliotron using a production snapshot


gastonabella (approved):
LGTM 👏

darrensiegel (approved):
Really nice work generalizing this into the reusable `batch_insert_all`

PR #4423, opened 2023-11-10
[NG23-16] Render explorations container header and sort

bitbldr:
Renders the header for the container in which the exploration originates, according to page links, ordered by their position within the hierarchy. Adds a default exploration image if one is not set. If an exploration has been started, it will show a "Continue" button instead of "Lets Begin"

Screenshot 2023-11-10 at 12 22 38 PM

I also noticed an issue with the content shadows flashing when scrolling beyond zero (0) position (iOS/macOS rubber band effect) and fixed that.


darrensiegel (changes requested):
We need to change this so that `Sections.get_explorations_by_containers` is not called each time that a user visits the Explorations view.  Internally, `get_explorations_by_containers` makes some fairly expensive calls to JSONB path queries to find extract things like page to page links.  

We need to instead call `get_explorations_by_containers` upfront, during course section creation, after remix, and after application of new publication update. Its result should be stored in some new field in the section.  Moreover, `get_explorations_by_containers` and what it calculates is too specific to store.  What we really want (and will need for other views in NG23 as well) is to calculate and store a map of page resource_ids to container_resource_id.  Basically the "which container this page is referenced from, whether directly because it exists in the hierarchy there, or indirectly because it "relates_to" a page there, or is linked from a page there").

At that point, the call from Explorations view can query to find all Exploration pages, then using the struct within this new field map these to their relevant container.  Another call to get container information (title, ordering number).

bitbldr:
> We need to change this so that `Sections.get_explorations_by_containers` is not called each time that a user visits the Explorations view. Internally, `get_explorations_by_containers` makes some fairly expensive calls to JSONB path queries to find extract things like page to page links.
> 
> We need to instead call `get_explorations_by_containers` upfront, during course section creation, after remix, and after application of new publication update. Its result should be stored in some new field in the section. Moreover, `get_explorations_by_containers` and what it calculates is too specific to store. What we really want (and will need for other views in NG23 as well) is to calculate and store a map of page resource_ids to container_resource_id. Basically the "which container this page is referenced from, whether directly because it exists in the hierarchy there, or indirectly because it "relates_to" a page there, or is linked from a page there").
> 
> At that point, the call from Explorations view can query to find all Exploration pages, then using the struct within this new field map these to their relevant container. Another call to get container information (title, ordering number).

Updated the implementation to be a more generic computation of map from resource_id to container_id and persist in the section record as cached JSON blob, which is computed when a section is created and updated when section curriculum is rebuilt. Finally, I optimized the exploration specific pieces to only issue a single query to fetch container titles and ordering of containers within the hierarchy

PR #4439, opened 2023-11-14
Fixed list styling issue in definition meanings [MER-2705]

marc-hughes:
Steps to reproduce:

1. On a basic authoring page, add a definition element
2. Edit that definition
3. In the first meaning, author a list - do not author a second meaning
4. Preview the definition

Expected: A definition where the first meaning visually has a list inside it.
Actual: A definition where the first meaning has no visual list.

 

Sample Input:

image

Before fix:

image

After fix:

image

### DIFF
--- a/assets/styles/common/elements.scss
+++ b/assets/styles/common/elements.scss
@@ -249,7 +249,7 @@ $element-margin-bottom: 1.5em;
     .meanings-single {
       display: inline;
       padding-left: 3px;
-      li {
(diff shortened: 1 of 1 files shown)


andersweinstein (approved):
Looks good on real Logic and Proofs definitions, several of which contain lists and including one with nested lists. 
Screenshot 2023-11-14 at 10 01 05 AM

PR #4448, opened 2023-11-15
[CHORE] Quickstart improvements

Writhe:
This PR introduces a couple of minor improvements to the initial dev set-up. The only real issue was the missing FE build step - all the rest are nitpicks.

**Changes:**
- added `&& yarn deploy` to `devmode.sh`
- added analogous step to `developer.md` ("Mostly Manual Steps")
- rephrased the part about replacing `DB_HOST` value in `oli.env` as the value doesn't currently need replacing
- added information about `/dev/sent_emails` and default admin credentials to "Notes"
- replaced "npm" with "yarn" in "Running Tests" section, for consistency's sake (also verified that FE test suite does pass when run by Yarn)

**Alternate designs:**
- I considered adding a post-install script to build FE assets, but decided against it. I feel it's more readable to have install and build steps entirely separate.


devmode.sh
```diff
@@ -7,7 +7,7 @@ echo "## "
 if [ ! -f .devmode ]; then
   echo "## It looks like this is your first time using devmode. Lets gather some prerequisites..." && sleep 1
   mix deps.get
-  cd assets && yarn
+  cd assets && yarn && yarn deploy
```
bitbldr, line 10: I don't think yarn deploy is actually necessary here, these assets will be built when `mix phx.server` is run

Writhe: They weren't accessible for me after I followed the start-up guide. I tried it once with mostly automatic and once with mostly manual setup. Each time I got an unstyled login page and a nice 404 instead of `app.css`. Once I ran `yarn build` and restarted the server - it was 👌.

Could it be that the server doesn't serve static assets from directories that weren't present on start-up? It finishes spinning up while frontend stuff is still building, creates an index of static assets before they're all there and then ignores requests for them? If that's the case, it would have been enough for me to simply restart the server 🤔. I'll check.

Writhe: @bitbldr 
Alright, so reloading the page does nothing, but restarting the server helps (which lends some credibility to my theory about static assets), but I think I prefer running `yarn build` to telling people to run `mix phx.server`, kill it and then run `mix phx.server` again ;)

Writhe: Here's the error (when you run the server for the first time):
Screenshot 2023-11-28 at 18 46 00

And here's the login screen in all its unstyled glory:
Screenshot 2023-11-28 at 18 47 41

bitbldr (approved):

PR #4458, opened 2023-11-20
[ENHANCEMENT] support legacyBlockRendered flag on MathML formulas [MER-2739]

andersweinstein:
Torus uses the legacyBlockRendered attribute on formula to force block rendering for inline LaTex formulas only. This is used to handle migrated formulas embedded inline in paragraph text using delimiters indicating block formula is desired.

This adds support for the same attribute on MathML formulas. This is needed to migrate inline MathML formulas with attribute display=”block”, which indicates the formula should be rendered as blocks. This style seen in formulas in Stem Readiness/Foundations courses.


marc-hughes:
I tested by creating a mathml & latex formula_inline, then manually adding a legacyBlockRendered param in, worked great in both  the server-rendered and client-rendered versions.

image

marc-hughes (approved):

PR #4460, opened 2023-11-20
[Feature] Directed Discussion Activity [MER-1472]

marc-hughes:
Adds a new directed-discussion activity.

**Open Question** - Admins previewing a course who are not instructors can not participate. Is that ok?

**Open question before we merge**: Right now, I manually added an activity_registration record into my local system, what's the proper way to do that so it shows up for everyone. The ecto.rest / test.ecto.reset does seem to add it, is there an automatic process on rollouts?

https://github.com/Simon-Initiative/oli-torus/assets/333265/a7555e0a-39d1-4223-b09f-a1b0bdf9af49


assets/src/components/activities/directed-discussion/utils.ts
```diff
@@ -0,0 +1,29 @@
...
+      version: 1,
+      transformations: [],
+      parts: [
+        makePart(
+          [makeResponse(matchRule('.*'), 0, 'Default Part')],
+          [makeHint(''), makeHint(''), makeHint('')],
+          '1',
+        ),
+      ],
+      targeted: [],
```
darrensiegel, line 26: This seems like a left over from a copy and paste, as it isn't defined in the `authoring` key from the schema

lib/oli_web/controllers/api/directed_discussion_controller.ex
```diff
@@ -0,0 +1,120 @@
...
+      :status => :approved,
+      :user_id => current_user.id,
+      :section_id => section.id,
+      :resource_id => resource_id,
+      :parent_post_id => parent_post_id,
+      :thread_root_id => parent_post_id,
+      :replies_count => 0,
+      :anonymous => false,
+      :content => %{"message" => content}
+    })
```
darrensiegel, line 33: There is a shortcut syntax for maps with atom keys that can be used which looks like:

```
%{
  status: :approved,
  user_id: current_user.id,
  ...
}
```

lib/oli_web/controllers/api/directed_discussion_controller.ex
```diff
@@ -0,0 +1,120 @@
...
+      })
+    else
+      json(conn, %{
+        "result" => "failure",
+        "error" => "User does not have permission to delete this post."
+      })
+    end
+  end
+
+  defp preload_post_user(post) do
```
darrensiegel, line 93: A more "Elixir" way of doing this would be to pattern match at the function definition level:

```
defp preload_post_user({:ok, post}), do: {:ok, Repo.preload(post, :user)}
defp preload_post_user(error), do: error
```

lib/oli_web/controllers/api/directed_discussion_controller.ex
```diff
@@ -0,0 +1,120 @@
...
+
+  alias Oli.Resources.Collaboration
+  alias Phoenix.PubSub
+  alias Oli.Resources.Collaboration.Post
+
+  def create_post(conn, %{"resource_id" => resource_id, "section_slug" => section_slug}) do
+    content = conn.body_params["content"]
+    parent_post_id = Map.get(conn.body_params, "parent_post_id", nil)
+    current_user = Map.get(conn.assigns, :current_user)
+    section = Sections.get_section_by_slug(section_slug)
```
darrensiegel, line 21: We should also check that the current user is enrolled (and not suspended) in this specific course section, to protect against malicious client-side attacks to post into discussions that they are not (or no longer) enrolled in.

lib/oli_web/controllers/api/directed_discussion_controller.ex
```diff
@@ -0,0 +1,120 @@
...
+  def delete_post(conn, %{
+        "resource_id" => resource_id,
+        "section_slug" => section_slug,
+        "post_id" => post_id
+      }) do
+    current_user = Map.get(conn.assigns, :current_user)
+
+    {post_id, ""} = Integer.parse(post_id)
+
+    post = Collaboration.get_post_by(%{id: post_id})
```
darrensiegel, line 71: Add a check to ensure user is enrolled (and not suspended).  This will prevent a suspended student from deleting a post that they had authored (the contents of which maybe had gotten them suspended, lol)

lib/oli_web/controllers/api/directed_discussion_controller.ex
```diff
@@ -0,0 +1,120 @@
...
+      {:ok, post} -> {:ok, Repo.preload(post, :user)}
+      error -> error
+    end
+  end
+
+  def get_discussion(conn, %{"resource_id" => resource_id, "section_slug" => section_slug}) do
+    section = Sections.get_section_by_slug(section_slug)
+    current_user = Map.get(conn.assigns, :current_user)
+
+    posts =
```
darrensiegel, line 104: Again, just add the enrollment as non-suspended here

assets/src/components/activities/directed-discussion/discussion/DirectedDiscussion.tsx
```diff
@@ -0,0 +1,92 @@
...
+  const hasSearchTerm = !!searchTerm && searchTerm.length > 0;
+  const hasFocusId = focusId !== null;
+
+  const displaySearchResults = !hasFocusId && hasSearchTerm;
+  const displayThreads = !displaySearchResults;
+
+  return (
+    <div className="activity mc-activity">
+      <div className="activity-content relative">
+        <h2>Discussion</h2>
```
darrensiegel, line 51: I think we should remove this `<h2>Discussion</h2>` as it hardcodes in a label that will not be changeable by an author.  The author can add a header within the stem which labels the discussion any way that they would like.

assets/src/components/activities/directed-discussion/discussion/participation-util.tsx
```diff
@@ -0,0 +1,46 @@
...
+  posts: number;
+  replies: number;
+  canPost: boolean;
+  canReply: boolean;
+}
+
+const MAX_POSTS = 1000;
+const MAX_REPLIES = 1000;
+
+export const calculateParticipation = (
```
darrensiegel, line 15: It seems like we could add in here a check against the current date and time vs `postDeadline` and `replyDeadline` which could factor into setting `canPost` and `canReply`.  That will at least give us an initial (but a bit weak) form of scheduling for the directed discussions.  We'd just have to also ensure that the PostDeadline and ReplyDeadlines are displayed to the student when these are set.

lib/oli_web/controllers/api/directed_discussion_controller.ex
```diff
@@ -0,0 +1,120 @@
...
+      :parent_post_id => parent_post_id,
+      :thread_root_id => parent_post_id,
+      :replies_count => 0,
+      :anonymous => false,
+      :content => %{"message" => content}
+    })
+    |> preload_post_user
+    |> case do
+      {:ok, post} ->
+        IO.puts("Sending broadcast to #{topic_name(section_slug, resource_id)}")
```
darrensiegel, line 37: Probably want to change `IO.puts` to `Logger.info`

darrensiegel (changes requested):
Really nice work! Tested locally and works well. There's a few things that we should fix, and I pointed just a couple of Elixir style things that don't need to be addressed really.

darrensiegel (approved):

PR #4469, opened 2023-11-27
[Chore] UX audit - Drag-and-drop styles

Writhe:
This PR aims to solve the problem of light/dark mode in the context of drag-and-drop activities. Right now, the draggable bits are utterly unreadable in dark mode. As I understand it, the styles for them are part of each lesson's configuration, so the actual fix would be to use the authoring environment to edit said styles. An additional complication is that DnD has its own shadow DOM - one has to take it into account when writing styles. My proposition is to provide better defaults that will also serve as a hint (because `:host-context()` is not a pseudoselector one uses every day - I know I don't).

An alternative approach would be to inject a prop (`mode: Mode`) into all custom elements like DnD and let them handle it as they see fit.

### DIFF
--- a/assets/src/components/activities/custom_dnd/utils.ts
+++ b/assets/src/components/activities/custom_dnd/utils.ts
@@ -71,6 +71,11 @@ const DEFAULT_LAYOUT_STYLES = `
   border-color: black;
 }
 
+:host-context(.dark) .initiator {
+  background-color: white;
+  color: black;
+}
+
 .dragdropspacer {
   height: 30px;
 }


bitbldr:
I actually was not able to verify this fixes the readability issue in dark mode
Screenshot 2023-11-29 at 3 35 11 PM

Writhe:
@bitbldr 
It is my understanding that, once saved, these styles become part of the... page? (not sure about the terminology, here, sorry). My change only provides the default content for the editor. To apply the actual change, one would have to manually add the CSS rules using the authoring UI.

andersweinstein:
There was an earlier change to address this (for freshly authored DNDs only) by including `color: black` in the defaut `.initiator `styles. This fix may be better, but not sure what happened to the earlier one merged into master Oct 11.

https://github.com/Simon-Initiative/oli-torus/pull/4303

PR #4472, opened 2023-11-27
[Fix] UX audit - Image and caption rendering

Writhe:
This PR makes minor adjustments to figcaption styles. Additionally, it modifies `captioned_content()` to (optionally) add a class name to its resulting `<figure />` wrapper, to indicate what is being wrapped. In this case - `caption-wrapper-image`. This makes it easier to target styles. Here, it is used to override the bottom margin of an image if that image happens to have a caption.

**Before:**
original style

**After:**
improved style

### DIFF
--- a/assets/styles/common/elements.scss
+++ b/assets/styles/common/elements.scss
@@ -419,7 +419,6 @@ $element-margin-bottom: 1.5em;
     display: block;
     margin-left: auto;
     margin-right: auto;
-    // margin-bottom: $element-margin-bottom;
   }
 
   .embed-responsive {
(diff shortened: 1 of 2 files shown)


bitbldr (changes requested):
@Writhe it appears the build is failing because of a formatting issue. Can you run `mix format` on your branch and push it to correct the build error? The simon-bot auto formatter does not run on forked repos unfortunately.

I would also recommend installing elixir-ls and enabling the "format on save" option if you are using vscode to prevent this issue in the future

Writhe:
@bitbldr 
Done and done. My bad - I'm so used to my TS toolchain I no longer think about setting up linters, formatters etc.

bitbldr (approved):

PR #4474, opened 2023-11-28
Mer 2704 bump lti libreries

tomasferok:
The dependency {:httpoison, "~> 1.6"} was updated to {:httpoison, "~> 2.0"}.

{:httpoison_retry, "~> 1.1"} was deleted.

{:lti_1p3, "~> 0.4.6"} was updated to {:lti_1p3, "~> 0.5.0"}.

{:lti_1p3_ecto_provider, "~> 0.3.3"} was updated to {:lti_1p3_ecto_provider, "~> 0.5.0"}.

A new module, autoretry, was added and used in the constellation module (to replace the retry plugin).


gastonabella (commented):
Looks good! I just left some comments.

Also, I noticed that the [retry.ex](https://github.com/mgwidmann/httpoison_retry/blob/master/lib/httpoison/retry.ex) module in `httpoison_retry` has a good `@moduledoc` that describes pretty well what it does. I think we could copy that into our new module for some context.

darrensiegel (approved):

PR #4484, opened 2023-11-30
[FEATURE][MER-2707] Survey Activities Tab

tomasferok:
[MER-2707 ](https://eliterate.atlassian.net/browse/MER-2707) 

This PR adds the Survey Activities tab to the Instructor Dashboard Overview.

Initially, a table is rendered to lists the pages that contain activities that are found in surveys.

When a user clicks on a page, the details of the activities contained in the selected page will be displayed (below the table of pages).

If the section has the V2 analytics version, the results of the students' answers to the activity details for the Multi Input and Likert activity types are added.

- The Multi Input activity type will show a table with the students' answers if the activity is of type "Text" or "Numeric", and will show the frequency of answers if the type is "Dropdown".

- The Likert activity type will add a chart showing the results of the answers given by the students.


darrensiegel (changes requested):
There seems to be something wrong with how this PR was branched from master.  It appears to delete the entire "Deliberate Practice" feature, and it also replicates portions of the "Practice Activities" feature (but preserving the `chart.js` dependency, which we don't need)

Here is what I recommend doing to fix this:
1. With this branch checked out, copy the files that only contain "Survey Activities" work out to a temporary location. 
2. Then checkout master and pull to get the latest. 
3. Then create a new branch. 
4. Then copy the files from your temporary location back into project. 
5. Add, commit and push, open a PR to see where we are at.

PR #4495, opened 2023-12-04
[MER-2647] add support for legacy response mult

rgachuhi:
This is a draft PR for work in progress that added support for question parts that target multiple inputs in multi-input activities.


bitbldr (approved):
Changes look good. Tested the separate activity implementation and works as expected so I think we should take that approach

assets/src/data/activities/utils.ts
```diff
@@ -38,6 +38,7 @@ export const safelySelectInputs = (activityState: ActivityState | undefined): Ma
 export const safelySelectStringInputs = (
   activityState: ActivityState | undefined,
 ): Maybe<PartInputs> => {
+  console.log(activityState);
```
bitbldr, line 41: Remove console log

assets/src/data/activities/DeliveryState.ts
```diff
@@ -32,6 +32,12 @@ export type AppThunk<ReturnType = void> = ThunkAction<
   unknown,
   AnyAction
 >;
+
+// export type StudentInputMult = {
```
bitbldr, line 36: Remove commented code

bitbldr (changes requested):

bitbldr (approved):

PR #4509, opened 2023-12-12
[BUGFIX][MER-2821] Incorrect message when section has no surveys

simonchoxx:
[MER-2821](https://eliterate.atlassian.net/browse/MER-2821)

This PR modifies the text displayed when section has no surveys to list in the `surveys` tab of the instructor dashboard. 

Captura de pantalla 2023-12-12 a la(s) 11 50 07

Captura de pantalla 2023-12-12 a la(s) 11 50 14

[MER-2821]: https://eliterate.atlassian.net/browse/MER-2821?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli_web/components/delivery/surveys/surveys.ex
+++ b/lib/oli_web/components/delivery/surveys/surveys.ex
@@ -111,26 +111,21 @@ defmodule OliWeb.Components.Delivery.Surveys do
             </form>
           </div>
         </div>
-        <%= if @total_count > 0 do %>
-          <PagedTable.render
-            table_model={@table_model}
-            total_count={@total_count}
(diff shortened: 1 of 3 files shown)


nicocirio:
Santi, what do you think about generalizing this for all paged tables? We could set a new attribute to that component with a default value, so anyone in the future could set a custom message without having to wrap the component with an `if` statement. For example:

```
attr :no_records_message, :string, default: "None exist"
```

bitbldr (approved):

PR #4516, opened 2023-12-13
Wait a second before closing a popup [MER-2777]

marc-hughes:
It was impossible to click on links in popups because the popup would close before you could get to the link with the cursor.

Now:

- The popup stays open for 1 second after mousing off the text anchor
- Mousing onto the popup keeps it open

popup-wait


bitbldr (approved):
Nice solution

PR #4528, opened 2023-12-20
[PERFORMANCE] Optimize contained objectives creation

darrensiegel:
The query that extracted the "contained objectives" from within `Sections.build_contained_objectives` was far too complex (a total cost of over 760 according to the postgres explain).  When run locally in dev environments with little data, the query executed quickly, but when run against realistic data sets (tokamak, heliotron) and over large courses (chem) this query was taking several minutes to execute.  On tokamak, it took 7 minutes. 

This PR breaks apart that single query into individual queries.  One to fetch the section id, one to fetch current contained pages, another to fetch the page-to-activity mapping, and a fourth to fetch the activity-to-objectives mapping.  The two that we really care about are the last two, and now on Tokamak these run very quickly:

```
Dec 20 13:06:22 tokamak oli: 13:06:22.357 [info] build_contained_objectives pages: 33.093438ms
Dec 20 13:06:22 tokamak oli: 13:06:22.406 [info] build_contained_objectives activities: 49.320548ms
```

The overall process to create a course section is now much quicker for large courses (seconds instead of several minutes).

### DIFF
--- a/lib/oli/delivery/sections.ex
+++ b/lib/oli/delivery/sections.ex
@@ -2108,63 +2108,88 @@ defmodule Oli.Delivery.Sections do
     page_type_id = ResourceType.get_id_by_type("page")
     activity_type_id = ResourceType.get_id_by_type("activity")
 
-    section_resource_pages =
-      from(
-        [sr: sr, rev: rev, s: s] in DeliveryResolver.section_resource_revisions(section_slug),
-        where: not rev.deleted and rev.resource_type_id == ^page_type_id
-      )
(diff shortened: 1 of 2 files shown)


gastonabella (approved):
This is an impressive improvement! 👏

For this kind of complex queries that require JSONB queries across model content, which can be expensive, it would be great from now on to always keep in mind to test with the CHEM course at least locally, so that we can catch these performance issues beforehand.

Even though this improvement was huge, it'd be also necessary to keep an eye on this function to see if there are even more improvements to be made (such as adding indexes), since it's being used in several places in important flows of the app.

PR #4530, opened 2023-12-20
Mer 2758 cannot add enrollment of a student

Francisco-Castro:
This PR covers the bug reported in ticket [MER-2758](https://eliterate.atlassian.net/browse/MER-2758) where students cannot be enrolled by an instructor/admin.

The endpoint `post("/:section_slug/enrollments", InviteController, :create_bulk)` was moved from an admin scope to the `/sections/:section_slug/instructor_dashboard` one, otherwise only admins can do the operation and an instructor might not have that type of role.

I would like to understand more the role of the `invited_by_id` field. The current implementation does not have that field set up. It would be helpful to know the function of this field to know best how to set it up. Do they want/need to know whether it was another user or an author? Is it a requirement to identify either the category (user or author) or the specific individual. Is it important to the completion of the enrollment operation? We can make a better recommendation if we have this information.

[MER-2758]: https://eliterate.atlassian.net/browse/MER-2758?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli_web/components/delivery/students/students.ex
+++ b/lib/oli_web/components/delivery/students/students.ex
@@ -3,6 +3,8 @@ defmodule OliWeb.Components.Delivery.Students do
 
   alias Phoenix.LiveView.JS
 
+  alias Oli.Accounts.Author
+  alias Oli.Accounts.User
   alias OliWeb.Common.{PagedTable, SearchInput, Params, Utils}
   alias OliWeb.Delivery.Sections.EnrollmentsTableModel
   alias OliWeb.Router.Helpers, as: Routes
@@ -58,7 +60,10 @@ defmodule OliWeb.Components.Delivery.Students do
        add_enrollments_step: :step_1,
        add_enrollments_selected_role: :student,
        add_enrollments_emails: [],
-       add_enrollments_users_not_found: []
+       add_enrollments_users_not_found: [],
+       inviter: "user",
+       current_user: ctx.user,
+       current_author: ctx.author
      )}
   end
 
@@ -182,6 +187,10 @@ defmodule OliWeb.Components.Delivery.Students do
   attr(:add_enrollments_selected_role, :atom, default: :student)
   attr(:add_enrollments_emails, :list, default: [])
   attr(:add_enrollments_users_not_found, :list, default: [])
+  attr(:current_user, :any, required: false)
+  attr(:current_author, :any, required: false)
+  attr(:inviter, :string, required: false)
+  attr(:myself, :string, required: false)
 
   def render(assigns) do
     ~H"""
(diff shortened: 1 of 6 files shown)


lib/oli_web/components/delivery/students/students.ex
```diff
@@ -375,16 +388,70 @@ defmodule OliWeb.Components.Delivery.Students do
...
+      <fieldset class="mt-2">
+        <div class="ml-2">
+          <input
+            type="radio"
+            id="author"
+            name="inviter"
+            phx-value-inviter="author"
+            phx-click="select_inviter"
+            phx-target={@myself}
+            checked={if @inviter == "author", do: true, else: false}
```
gastonabella, line 430: Nit: you could just do `checked={@inviter == "author"}`

lib/oli_web/components/delivery/students/students.ex
```diff
@@ -375,16 +388,70 @@ defmodule OliWeb.Components.Delivery.Students do
...
+        </div>
+        <div class="ml-2">
+          <input
+            type="radio"
+            id="user"
+            name="inviter"
+            phx-value-inviter="user"
+            phx-click="select_inviter"
+            phx-target={@myself}
+            checked={if @inviter == "user", do: true, else: false}
```
gastonabella, line 442: Same here :)

lib/oli_web/components/delivery/students/students.ex
```diff
@@ -375,16 +388,70 @@ defmodule OliWeb.Components.Delivery.Students do
...
       <input name="role" value={@add_enrollments_selected_role} />
       <input name="section_slug" value={@section_slug} />
+      <input name="inviter" value={@inviter} />
       <button type="submit" class="hidden" />
     </.form>
     <div class="px-4">
       <p>
         Are you sure you want to enroll <%= "#{if length(@add_enrollments_emails) == 1, do: "one user", else: "#{length(@add_enrollments_emails)} users"}" %>?
       </p>
+      <.render_inviter
```
gastonabella, line 398: Supernit, but in general we don't name components with a `render_` prefix or similar, since it's implicit that this call renders its content. I think just `inviter` is clear enough 😄

gastonabella (approved):
Great work! 🚀

lib/oli_web/components/delivery/students/students.ex
Francisco-Castro: Thanks! It's nice to know we have a convention.

lib/oli_web/components/delivery/students/students.ex
Francisco-Castro: Great! I'll do the changes then.

lib/oli_web/components/delivery/students/students.ex
Francisco-Castro: Got it!

darrensiegel (approved):

PR #4536, opened 2023-12-22
Mer 2788 remove unbounded serial db reads from practice survey views

tomasferok:
[MER-2788](https://eliterate.atlassian.net/browse/MER-2788?filter=10069&jql=project%20%3D%20MER%20AND%20issuetype%20in%20(Bug%2C%20Story)%20AND%20status%20in%20(%22In%20Progress%22%2C%20PR%2C%20Ready%2C%20%22To%20Do%22)%20AND%20labels%20%3D%20wyeworks%20ORDER%20BY%20assignee%20ASC%2C%20status%20ASC%2C%20priority%20DESC%2C%20created%20DESC) 
-This pull request enhances the efficiency of the **get_preview_render** function in **surveys** and **practice activities** by extracting activity details. The improvement lies in optimizing the enum map. Additionally, the get_activities function has been refactored to selectively choose only the necessary labels

**Observations:**

- Several queries exist within the functions in the enum map. While they seem to follow a consistent logic, I attempted to refactor some of them, encountering complexity in certain cases. Please review those instances.
- The functions containing queries are currently housed in a liveview module. Would it be beneficial to separate them into a context module?
- Certain private functions seem to encapsulate logic that could be tested. Consider decoupling these functions and introducing tests for the underlying logic.
I propose breaking down these tasks into separate tickets for better organization and tracking, how about it?.


lib/oli_web/components/delivery/practice_activities/practice_activities.ex
```diff
@@ -88,6 +88,8 @@ defmodule OliWeb.Components.Delivery.PracticeActivities do
   end
 
   def render(assigns) do
+    assigns.activities
+
```
simonchoxx, line 92: Maybe you forgot to delete this?

simonchoxx (commented):
I think the changes for a first improvement are fine 👏

Maybe we could clean up the `get_activities` function also in the `practice_activities` module, as it was done in the `surveys` module.

I don't understand why a file called `logfile` was added in the commits.

darrensiegel (changes requested):
This isn't yet where it needs to be, as there is still an unbounded, data-driven serial DB read issue. Each of `add_single_response_details`, `add_choices_frequencies`, `add_multi_input_details`, and `add_likert_details` all issue DB reads.  These functions are called within an `Enum.map` over all activity attempts.  

Instead, replace the query for ActivityAttempt records found in `get_activities_details` with one that fetches all V2 analytics records (i.e.,  `ResourceSummary`, `StudentResponse`, `ResourcePartResponse`) for the **activities** found on that page.   Then post process to arrange the data in a way that can be displayed on a per-activity basis.   I do not believe that you need to involve ActivityAttempt or ResourceAttempt records in any place here. It is sufficient to simply query for activity resource ids which are found on this page id from the ResourceSummary.  In a separate query you can resolve the latest Revisions of those activities that you find.

darrensiegel:
Also, please retarget this to the `hotfix-v0.26.1` branch

darrensiegel (approved):

PR #4538, opened 2023-12-26
[BUGFIX][MER-2760] Cannot change password when editing account

simonchoxx:
[MER-2760](https://eliterate.atlassian.net/browse/MER-2760)

This PR adds support for making it possible for users to effectively change the password. 
This enhancement was made for both **Student/Instructor** and **Author** users. 

It checks that when changing the password, the new password has the required length.
Also, it's checked that when changing the password the new password matches the password confirmation.

https://github.com/Simon-Initiative/oli-torus/assets/16328384/a39c7ac3-a0e6-4ef1-8d15-1105f31b6659

Captura de pantalla 2023-12-26 a la(s) 11 54 56
Captura de pantalla 2023-12-26 a la(s) 11 54 36


lib/oli/accounts.ex
```diff
@@ -252,6 +252,30 @@ defmodule Oli.Accounts do
     |> Repo.insert()
   end
 
+  def update_user(
```
Francisco-Castro, line 255: I'm wondering if we could make some changes to this function to make it more compact. One suggestion is as follows:

```elixir
def update_user(
        %User{} = user,
        %{"current_password" => _, "password" => _, "password_confirmation" => _} = attrs
      ) do
    user
    |> User.update_user_changeset(attrs)
    |> Repo.update()
    |> case do
      {:ok, %User{id: user_id}} ->
        AccountLookupCache.delete("user_#{user_id}")

      error ->
        error
    end
  end
```

This way we put in one-line the matching we want. Also the `case` is a continuation of the 2 pipes, saving one binding.

Something similar can be done for the author one.

Another comment is that now that we have 2 heads for the `update_user`, the documentation needs to be updated.

lib/oli/accounts/schemas/author.ex
```diff
@@ -91,6 +91,23 @@ defmodule Oli.Accounts.Author do
     |> maybe_name_from_given_and_family()
   end
 
+  @doc """
+  Creates a changeset that is used to update an author's profile
+  """
+
+  def update_author_changeset(author, attrs \\ %{}) do
```
Francisco-Castro, line 98: What do you think about changing the name of this function to `update_changeset` because we are in the Author context.

Same for the User side.

lib/oli/accounts/schemas/author.ex
```diff
@@ -91,6 +91,23 @@ defmodule Oli.Accounts.Author do
...
+    author
+    |> pow_changeset(attrs)
+    |> cast(attrs, [
+      :given_name,
+      :family_name,
+      :email
```
Francisco-Castro, line 104: What do you think setting this four lines just in one?

Same for the User side.

test/oli/accounts_test.exs
```diff
@@ -104,6 +104,63 @@ defmodule Oli.AccountsTest do
 
       assert author.preferences.timezone == "America/Los_Angeles"
     end
+
+    test "update author password successfully" do
```
Francisco-Castro, line 108: I'm wondering if we could check more that the password here because we are handling different parts of information like: given_name, family_name, email, and passwords that also could be checked. Practically what the user chan change from the edit form.


I think this test would be

test/oli/accounts_test.exs
```diff
@@ -104,6 +104,63 @@ defmodule Oli.AccountsTest do
...
+      author = insert(:author)
+
+      attrs = %{
+        "current_password" => "password_1",
+        "email" => author.email,
+        "family_name" => author.family_name,
+        "given_name" => author.given_name,
+        "password" => "password_22",
+        "password_confirmation" => "password_2"
+      }
```
Francisco-Castro, line 133: What do you think of shrinking this map to only have the concerning pieces of data: `current_password`, `password` and `password_confirmation`?The others parts can be skipped because we are added by the factory.

test/oli/accounts_test.exs
```diff
@@ -104,6 +104,63 @@ defmodule Oli.AccountsTest do
...
+        "given_name" => author.given_name,
+        "password" => "pass",
+        "password_confirmation" => "pass"
+      }
+
+      assert {:error, changeset} = Accounts.update_author(author, attrs)
+      assert changeset.valid? == false
+
+      {:password, {error_message, [count: 8, validation: :length, kind: :min, type: :string]}} =
+        List.first(changeset.errors)
```
Francisco-Castro, line 160: Because there is only one element in `errors` you could match the list directly. Something like:

```
[{:password_confirmation, {error_message, [validation: :confirmation]}}] = changeset.errors
```

Francisco-Castro (commented):
Nice job! 🏅 

I'll put this review as comments and wait for your response to make another loop.

BTW, Checking the Author module I found [this function](https://github.com/Simon-Initiative/oli-torus/pull/4538/files#diff-dfadc1e7a0a845444a1c1922a9934a945784e7affddf9c4f45daa75e1d9d1798R123) and I think we could make it more readible using something like:


```
  defp default_system_role(
         %Ecto.Changeset{
           valid?: true,
           changes: %{system_role_id: nil},
           data: %Oli.Accounts.Author{system_role_id: nil}
         } = changeset
       ) do
    put_change(changeset, :system_role_id, SystemRole.role_id().author)
  end

  defp default_system_role(changeset), do: changeset
```

It would be great If you could add this change but totally ok if you think is not appropriate.

darrensiegel:
This needs to be retargeted to `master`, given that V26 has already been released

simonchoxx:
@darrensiegel Sure. Let me do a rebase with Master.

darrensiegel (approved):

PR #4547, opened 2024-01-04
[BUGFIX][MER-2641] Project export-import does not preserve language and custom labels

simonchoxx:
[MER-2641](https://eliterate.atlassian.net/browse/MER-2641)

This PR aims to add support for exporting a project to include the `attributes` and `customizations` fields. 

Within the `attributes` field the `learning language` feature will be included if the author made changes to it. 
In addition, `customizations` will include any changes to the customisation of labels for containers, **_units_**, **_modules_** and _**sections**_.

Also, support is added to be able to include these project features(attributes and customizations) when ingesting a project(in both versions of the ingest functionality) and have them reflected in the new imported project.

https://github.com/Simon-Initiative/oli-torus/assets/16328384/036cb1d0-9688-41c3-8283-8c888107d34f

[MER-2641]: https://eliterate.atlassian.net/browse/MER-2641?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli/interop/ingest.ex
+++ b/lib/oli/interop/ingest.ex
@@ -103,7 +103,7 @@ defmodule Oli.Interop.Ingest do
            media_details <- Map.get(resource_map, @media_key),
            hierarchy_details <- Map.get(resource_map, @hierarchy_key),
            {:ok, %{project: project, resource_revision: root_revision}} <-
-             create_project(project_details, as_author),
+             create_project(project_details, as_author, hierarchy_details),
            {:ok, tag_map} <- create_tags(project, resource_map, as_author),
            {:ok, objective_map} <- create_objectives(project, resource_map, tag_map, as_author),
            {:ok, bib_map} <- create_bibentries(project, resource_map, as_author),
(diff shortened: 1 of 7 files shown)


lib/oli/interop/ingest.ex
```diff
@@ -327,27 +327,11 @@ defmodule Oli.Interop.Ingest do
         {:error, :empty_project_title}
 
       title ->
-        labels =
```
darrensiegel, line 330: This code, which detects an object within the `children` array that has a `"type"` attribute equal to `"labels"` has to stay in place.  We have a tool (see https://github.com/Simon-Initiative/course-digest) that converts our legacy system courses into a "Torus Course Digest".  That tool places the label information from the Legacy courses into the hierarchy file.  So we need this to stay in place to read this type of digest structure. 

Maybe the easiest way to support this backwards compatibility would be to change the export of Torus projects to put the label information back into the hierarchy, as it is found here. 

The custom language settings can stay in the project file.

darrensiegel (changes requested):

lib/oli/interop/ingest.ex
simonchoxx: Ok. I understand. I didn't think about legacy system courses when making these changes.

darrensiegel (approved):
Tested, works well.  Nice work!

PR #4551, opened 2024-01-08
Feature/toolbar enhancements [MER-2884 MER-2885 MER-2896 MER-2882]

marc-hughes:
Makes 4 enhancements to the editor toolbar:

1. Adds tooltips to the top level icons
2. Adds an undo button
3. Adds a redo button
4. Moves the text-direction and markdown options from a dropdown menu into the top level toolbar.


darrensiegel (approved):
Tested, looks great

PR #4554, opened 2024-01-09
[FEATURE][MER-2889] All Pages View show whether or not pages are in curriculum

rgachuhi:
This PR adds an indicator for each page of whether or not the page is in the curriculum.

### DIFF
--- a/lib/oli/resources/page_browse.ex
+++ b/lib/oli/resources/page_browse.ex
@@ -5,7 +5,6 @@ defmodule Oli.Resources.PageBrowse do
   alias Oli.Repo
   alias Oli.Authoring.Course.Project
   alias Oli.Resources.Revision
-  alias Oli.Publishing.AuthoringResolver
   alias Oli.Repo.{Paging, Sorting}
 
   @chars_to_replace_on_search [" ", "&", ":", ";", "(", ")", "|", "!", "'", "<", ">"]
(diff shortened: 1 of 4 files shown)


lib/oli/resources/page_browse.ex
```diff
@@ -79,6 +80,13 @@ defmodule Oli.Resources.PageBrowse do
 
     page_type_id = Oli.Resources.ResourceType.get_id_by_type("page")
 
+    container_id = Oli.Resources.ResourceType.get_id_by_type("container")
+    containers = AuthoringResolver.revisions_of_type(slug, container_id)
```
darrensiegel, line 84: From a performance standpoint, this isn't ideal.  This `browse_pages` function is designed to be used by a table based database-driven query.  Which means that each time that a user pages, filters or even sorts the list of all pages, this `AuthoringResolver.revisions_of_type` would execute. 

A more performant and more feature capable solution here would be to leave this `browse_pages` as-is, and instead insert into the `mount` of the AllPages LV, this call to `AuthoringResolver.revisions_of_type(slug, container_id)`.  Invert the returned results to turn it into a map of page resource ids to the parent container revision.  Those two data structures can be passed down into the `pages_table_model` where for each displayed page, we can then display a link to the container that it exists in (if it exists in any).  Clicking the link would route the author to the Curriculum view for that container.    

This probably means we can't support sorting by that column since it isn't being populated by the query - but that isn't the key use case here.  That use case of looking at the hierarchy, in order, will be handled by some other upcoming work.

rgachuhi: Makes sense. I will do the update.

darrensiegel (approved):

PR #4555, opened 2024-01-10
[BUG FIX] [MER-2907] Show Content Browser for LMS sections, fix CollabSpace rendering issue

darrensiegel:
This PR fixes two issues:

1. The student facing "Content Browser", on the home screen, was conditionally being rendered for **only open and free sections**.  It needs to be exist for LMS sections as well. This PR removes the `<%= if @independent_learner do %>` which guarded the rendering of it. 

2. AppSignal has been reporting many instances of a `** (ArgumentError) cannot perform Ecto.Repo.get/2 because the given value is nil` error.  The stack trace points to a problem during the mount of the `CollabSpaceView`, where both `current_user_id` and `current_author_id` attributes of the LV session are `nil`.  My theory is that these session attributes aren't getting populated in certain usage patterns where the user **first** visits a rendered page.  The page rendering comes from a standard controller which directly renders the LV CollabSpaceView.  That direct rendering augments an existing LV session, but there isn't one in this case.  The solution here is to always add in the `current_user_id` value as a session augmentation.

### DIFF
--- a/lib/oli_web/controllers/page_delivery_controller.ex
+++ b/lib/oli_web/controllers/page_delivery_controller.ex
@@ -629,6 +629,7 @@ defmodule OliWeb.PageDeliveryController do
       conn,
       "page.html",
       %{
+        user: user,
         adaptive: adaptive,
         context: context,
         page: context.page,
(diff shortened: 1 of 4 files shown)


lib/oli_web/controllers/page_delivery_controller.ex
```diff
@@ -1096,6 +1101,12 @@ defmodule OliWeb.PageDeliveryController do
     |> render(
       "instructor_page_preview.html",
       %{
+        user:
```
eliknebel, line 1104: nit: A cleaner way to do this might be
`conn.assigns[:current_user]`

eliknebel (approved):

PR #4561, opened 2024-01-12
Mer 1886 discussions remove them from student view if the feature is not turned on

Francisco-Castro:
Reference: [MER-1886](https://eliterate.atlassian.net/browse/MER-1886)

This PR added the ability to show/hide the discussion part from the navbar. The logic that controls it is based on the `status` field on the `collab_space_config` column in the `section` table.

The flag is trigger by 3 cases: creation of a section, remix of a project or defined by an instructor.

[MER-1886]: https://eliterate.atlassian.net/browse/MER-1886?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli/delivery/sections.ex
+++ b/lib/oli/delivery/sections.ex
@@ -1476,6 +1476,7 @@ defmodule Oli.Delivery.Sections do
       join: sr in assoc(s, :section_resources),
       where: s.id == ^section.id,
       where: fragment("?->>'status' = ?", sr.collab_space_config, "enabled"),
+      limit: 1,
       select: sr.id
     )
     |> Oli.Repo.all()
--- a/lib/oli/delivery/sections.ex
+++ b/lib/oli/delivery/sections.ex
@@ -975,9 +975,11 @@ defmodule Oli.Delivery.Sections do
       {:error, %Ecto.Changeset{}}
   """
   def update_section(%Section{} = section, attrs) do
-    section
-    |> Section.changeset(attrs)
-    |> Repo.update()
+    section |> Section.changeset(attrs) |> Repo.update()
+  end
+
+  def update_section!(%Section{} = section, attrs) do
+    section |> Section.changeset(attrs) |> Repo.update!()
   end
 
   @doc """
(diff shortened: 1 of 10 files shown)


lib/oli/delivery/sections.ex
```diff
@@ -1468,6 +1470,21 @@ defmodule Oli.Delivery.Sections do
     end)
   end
 
+  @spec maybe_update_contains_discusssions(Section.t()) :: Section.t()
+  def maybe_update_contains_discusssions(section) do
+    from(s in Section,
+      join: sr in assoc(s, :section_resources),
+      where: s.id == ^section.id,
+      where: fragment("?->>'status' = ?", sr.collab_space_config, "enabled"),
+      select: sr.id
```
darrensiegel, line 1479: We should add a `limit: 1` here to ensure that this query does not return 500 items in the case when collab spaces are enabled on every page.

darrensiegel (changes requested):
I think that this `maybe_update_contains_discussions` is not being called in all of hte places that it needs to be called.  There is the possibility of changing whether or not a section has pages with discussions in all of the following:

1. Course section creation, both from `new_course.ex` and `open_and_free_controller.ex` - both when being created from a Project and from a Product.
2. Applying new publications from a project to a section. 
3. Instructor performing a remix on a course section (adding in pages that have enabled collab spaces, or removign pages that do)
4. Instructor explicitly enabling / disabling collab spaces in their course section from the Preview view. 

Do a grep through the codebase of where `rebuild_contained_pages` is being called is a good way to see very likely where `maybe_update_contains_discussions` needs to be called: 

```
./oli_web/live/new_course/new_course.ex:           {:ok, _} <- Sections.rebuild_contained_pages(section),
./oli_web/live/new_course/new_course.ex:           {:ok, _} <- Sections.rebuild_contained_pages(section),
./oli_web/controllers/open_and_free_controller.ex:           {:ok, _} <- Sections.rebuild_contained_pages(section),
./oli_web/controllers/open_and_free_controller.ex:           {:ok, _} <- Sections.rebuild_contained_pages(section),
./oli/delivery.ex:      {:ok, _} = Sections.rebuild_contained_pages(section)
./oli/delivery.ex:      {:ok, _} = Sections.rebuild_contained_pages(section)
./oli/delivery/sections.ex:      {:ok, _} = rebuild_contained_pages(section, section_resources)
./oli/delivery/sections.ex:  def rebuild_contained_pages(%{id: section_id} = section) do
./oli/delivery/sections.ex:    rebuild_contained_pages(section, section_resources)
./oli/delivery/sections.ex:  def rebuild_contained_pages(
./oli/delivery/sections.ex:    page_map = rebuild_contained_pages_helper(root, {[nil], %{}, map, container_ids})
./oli/delivery/sections.ex:  defp rebuild_contained_pages_helper(sr, {ancestors, page_map, all, container_ids}) do
./oli/delivery/sections.ex:                       rebuild_contained_pages_helper(
```

lib/oli/delivery/sections.ex
Francisco-Castro: Great idea! I'll do add the limit.

darrensiegel (approved):
Nice work with the new `PostProcessing` module!

PR #4565, opened 2024-01-17
Mer 2929 discussions leftover work

Francisco-Castro:
Ticket: [MER-2929](https://eliterate.atlassian.net/browse/MER-2929)

This is a leftover work from [PR-4561](https://github.com/Simon-Initiative/oli-torus/pull/4561). It contains some improvements in the PostProcessing module and promote it uses over the `maybe_update_contains_exploration` and `maybe_update_section_contains_deliberate_practice` functions that were also get rid of in this PR.

A proposal is to move all side effects into the `PostProcessing` module --`rebuild_contained_pages` and `rebuild_contained_objectives`. Doing this we can centralize them.

### DIFF
--- a/lib/oli/delivery.ex
+++ b/lib/oli/delivery.ex
@@ -105,6 +105,7 @@ defmodule Oli.Delivery do
           )
         )
 
+      section = PostProcessing.apply(section, :discussions)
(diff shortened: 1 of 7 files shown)


darrensiegel (approved):
Nice improvements!

PR #4577, opened 2024-01-19
[NG23-48] Provide application level cache for a computed course hierarchy

gastonabella:
[NG23-48](https://eliterate.atlassian.net/browse/NG23-48)

This PR introduces the following changes:
- Extend the `SectionCache` module to allow storing a course full hierarchy and broadcasting of `put` and `delete` operations to other nodes in the cluster.
- Fix a bug in the `SectionCache` module that was not storing the cache values properly.
- Remove the `Sections.rebuild_full_hierarchy/1` function and all its related code, since the `:full_hierarchy` was removed from the sections table.

[NG23-48]: https://eliterate.atlassian.net/browse/NG23-48?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/oli/delivery/sections/section_cache.ex
```diff
@@ -81,14 +127,51 @@ defmodule Oli.Delivery.Sections.SectionCache do
...
-
-    case Cachex.put(@cache_name, key, value, ttl: ttl) do
+    case Cachex.put(@cache_name, key, value, ttl: @ttl) do
       {:ok, true} ->
         {:reply, :ok, state}
 
       _ ->
         {:reply, :error, state}
     end
   end
```
nicocirio, line 137: Does it make sense to use `handle_cast` for `put` and `delete`? For `delete` Cachex docs say "This will return { :ok, true } regardless of whether a key has been removed or not"

lib/oli_web/live/delivery/student/learn_live.ex
```diff
@@ -20,7 +21,12 @@ defmodule OliWeb.Delivery.Student.LearnLive do
           socket.assigns[:current_user]
         )
 
-    root_node = full_hierarchy(socket.assigns.section)
+    section = socket.assigns.section
+
+    root_node =
+      SectionCache.get_or_compute(section.slug, :full_hierarchy, fn ->
+        full_hierarchy(section)
+      end)
```
nicocirio, line 29: Does it make sense to have a `full_hierarchy/1` function in the Sections context? That function could wrap the get_or_compute logic so anyone could call that function without worrying about whether it was computed or retrieved from the cache.

nicocirio (approved):
LGTM, great work!! I left some NITs

darrensiegel (changes requested):
Maybe this is in here, but I don't see it: Where does a section cache entry get invalidated?  

The full hierarchy in the cache needs to be invalidated when:
1. A new course update is applied
2. The instructor performs a course section Remix

Otherwise, students will continue to see the old hierarchy.

gastonabella:
> Maybe this is in here, but I don't see it: Where does a section cache entry get invalidated?
> 
> The full hierarchy in the cache needs to be invalidated when:
> 
> 1. A new course update is applied
> 2. The instructor performs a course section Remix
> 
> Otherwise, students will continue to see the old hierarchy.

One of the reasons I extended and reused the existing Sections cache is because I can leverage that the cached data is already being invalidated in those places you mentioned.
We `clear` ("invalidate") the full hierarchy at the end of `Sections.rebuild_section_curriculum/3` [right here](https://github.com/Simon-Initiative/oli-torus/blob/ng23-48-provide-application-level-cache-for-a-computed-course-hierarchy/lib/oli/delivery/sections.ex#L2328), and this function is called in three situations:
1. When a **minor** update is applied.
2. When a **major** update is applied.
3. When a section is remixed by an instructor.

darrensiegel (approved):
Okay! Looks great then.

PR #4579, opened 2024-01-19
Mer 2456 Ability to see who created a section in the publish area that shows you the sections that were created off of a project

Francisco-Castro:
Ticket: [MER-2456](https://eliterate.atlassian.net/jira/software/c/projects/MER/issues/MER-2456)

This PR contains the addition of two columns on the publication view: one that shows the creator of each section and another one that shows the instructors per section.

Note: a minor refactor in the `unenroll` function was done.

[MER-2456]: https://eliterate.atlassian.net/browse/MER-2456?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/oli/authoring/publishing.ex
```diff
@@ -0,0 +1,27 @@
...
+  defp lookup_instructors_enrolled_in(section) do
+    from(s in Section,
+      join: e in assoc(s, :enrollments),
+      join: u in assoc(e, :user),
+      where: s.id == ^section.id,
+      where: u.can_create_sections == true,
+      order_by: [asc: e.inserted_at],
+      select: u
+    )
+  end
```
simonchoxx, line 26: Maybe we could use some of the functions that already exist in `Oli.Delivery.Sections` to get the instructors of a section. 

See if they work for you: 

- **fetch_instructors/1**
- **instructors_per_section/1**
- **get_instructors_for_section/1**

test/oli/authoring/publishing_test.exs
```diff
@@ -0,0 +1,47 @@
...
+  use Oli.DataCase
+  import Oli.Factory
+
+  alias Oli.Authoring.Publishing
+
+  describe "find_instructors_enrolled_in/1" do
+    test "searches for enrolled instructors only" do
+      instructor_1 = insert(:user, can_create_sections: true)
+      instructor_2 = insert(:user, can_create_sections: true)
+      student = insert(:user, can_create_sections: false)
```
simonchoxx, line 11: `can_create_sections` is false by default. It would not be necessary to add it here!

simonchoxx (commented):
LGTM 🚀🚀 . Nice work mate 👏👏

lib/oli/authoring/publishing.ex
```diff
@@ -0,0 +1,27 @@
+defmodule Oli.Authoring.Publishing do
```
darrensiegel, line 1: This module name does not make sense given the contents within it.  The functions within it are Delivery functions (e.g. "find instructors enrolled in a section") but the Module name here involves Authoring and Publishing. 

I would eliminate this module and move these functions to a pure Delivery module.

lib/oli_web/live/projects/active_sections_table_model.ex
```diff
@@ -64,6 +75,20 @@ defmodule OliWeb.Projects.ActiveSectionsTableModel do
   def custom_render(assigns, section, %ColumnSpec{name: :base_project_id}),
     do: if(section.base_project_id == assigns.project.id, do: "Base Project", else: "Remixed")
 
+  def custom_render(_assigns, section, %ColumnSpec{name: :creator}) do
+    case Publishing.find_oldest_enrolled_instructor(section) do
```
darrensiegel, line 79: Executing DB calls within the rendering phase is generally not a good idea, as we have zero control over how many times these DB calls will be made.  Every time a user sorts a column, these DB calls will execute. 

Instead, I would recommend creating a specialized "find_oldest_instructor" that took as input either a list of all sections, or maybe the project_id.  That call can then be made one time in the mount of the LiveView, and then that data passed down into the table model.

darrensiegel (dismissed):

darrensiegel (approved):

PR #4582, opened 2024-01-22
[Bugfix] Deleting rows/columns improvement [MER-2500]

marc-hughes:
1. Modified the table menu so there are two columns, making it easier to access all the options.
2. Modified the delete-column and delete-row functionality to take into account the visually displayed grid instead of just the table structure.

Demo:
https://github.com/Simon-Initiative/oli-torus/assets/333265/c08d71a5-0125-446d-b779-9ded53654cab


eliknebel (approved):
Changes look good. Tested and table editor seems much more solid now

PR #4598, opened 2024-01-25
[NG23-65] ordering of exploration pages is not correct

eliknebel:
Fixes a couple of different issues related to the ordering of explorations and practice pages and categorization of which containers they are in:

1. Original ordering determination using a single database query and numbering index and level was flawed. In order to properly sort according to the hierarchy, the hierarchy structure must be loaded and used for sorting. Because this code path still makes use of a section level caching mechanism, this cost only needs to be paid a single time for a course section (and whenever the section is remixed, updated, etc.)
2. The key being used to cache the labels was not matching up with the key expected by the SectionCache module and therefore was not being properly cleared as expected. I fixed that and added a guard so that any key used must match a key expected by the module
3. The cache was originally being stored in the section record as JSON and therefore being loaded anytime a section is queried. This approach was also different than the other ways we were caching section level things so I standardized this with the other items in the section-level cache
4. The all_page_links map being used to find the parent container were also including unnecessary links such as activity_references and relates_to, which was throwing off the algorithm which finds the parent container for resources such as the adaptive pages.
5. Added better unit test coverage to test all these cases
6. `fetch_ordered_container_labels` no longer returns the "Curriculum" root container due to how the full_hierarchy/flatten functions work and changing that seems like more of a risk than updating the latest code paths that now use `fetch_ordered_container_labels`. Really the only place that needed to be updated to handle this change was the discussions view

This solution is a bit inefficient in that it uses the existing full_hierarchy call when it really doesn't need any of the actual content of the revisions. We should update this to use the new MinimalHierarchy module present in hotfix 0.26.2 once that becomes available. Since this call is part of the section cached data, the impact on performance is minimal and only paid when a section is first loaded

Screenshot 2024-01-25 at 4 27 59 PM

### DIFF
--- a/lib/oli/delivery.ex
+++ b/lib/oli/delivery.ex
@@ -161,7 +161,6 @@ defmodule Oli.Delivery do
       enroll(user.id, section.id, lti_params)
 
       {:ok, updated_section} = maybe_update_section_contains_explorations(section)
-      {:ok, _} = Sections.update_page_to_container_map(section)
 
       updated_section
     end)
--- a/lib/oli/delivery.ex
+++ b/lib/oli/delivery.ex
@@ -161,7 +161,7 @@ defmodule Oli.Delivery do
       enroll(user.id, section.id, lti_params)
 
       {:ok, updated_section} = maybe_update_section_contains_explorations(section)
-      {:ok, _} = Sections.update_resource_to_container_map(section)
+      {:ok, _} = Sections.update_page_to_container_map(section)
 
       updated_section
     end)
--- a/lib/oli/delivery/sections.ex
+++ b/lib/oli/delivery/sections.ex
@@ -1323,8 +1323,6 @@ defmodule Oli.Delivery.Sections do
     all_page_links =
       [
         get_all_page_links(publication_ids),
-        get_activity_references(publication_ids),
-        get_relates_to(publication_ids),
         get_hierarchical_parent_links(publication_ids)
       ]
       |> Enum.reduce(MapSet.new(), fn links, acc -> MapSet.union(links, acc) end)
@@ -1368,7 +1366,7 @@ defmodule Oli.Delivery.Sections do
   If the section does not have a precomputed resource_to_container_map, one will be generated.
 
   ## Examples
-      iex> get_resource_to_container_map(section)
+      iex> get_page_to_container_map(section)
       %{
         "21" => 39,
         "22" => 40,
@@ -1380,15 +1378,16 @@ defmodule Oli.Delivery.Sections do
         "28" => 43,
       }
   """
-
-  def get_resource_to_container_map(section) do
+  def get_page_to_container_map(section) do
+    # TODO: Replace with SectionCache cachex-based implementation, remove resource_to_container_map
+    # from section
     case section do
       %Section{resource_to_container_map: nil} ->
         Logger.warning(
           "Section #{section.slug} has no precomputed resource_to_container_map. One will be generated now."
         )
 
-        {:ok, section} = update_resource_to_container_map(section)
+        {:ok, section} = update_page_to_container_map(section)
         section.resource_to_container_map
 
       %Section{resource_to_container_map: resource_to_container_map} ->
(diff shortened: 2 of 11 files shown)


lib/oli/delivery.ex
```diff
@@ -161,7 +161,6 @@ defmodule Oli.Delivery do
       enroll(user.id, section.id, lti_params)
 
       {:ok, updated_section} = maybe_update_section_contains_explorations(section)
-      {:ok, _} = Sections.update_resource_to_container_map(section)
```
eliknebel, line 164: Do we want to pay the cost of warming the cache on product/section creation or first student access?

darrensiegel (approved):

lib/oli/delivery/sections.ex
```diff
@@ -1380,46 +1378,27 @@ defmodule Oli.Delivery.Sections do
...
-    end
-  end
-
-  @doc """
-  Builds a section's resource_to_container_map and updates the section with it.
-  """
-  def update_resource_to_container_map(section) do
-    update_section(section, %{
-      resource_to_container_map: build_resource_to_container_map(section)
-    })
```
eliknebel, line 1405: Change caching strategy from storing in the section record to using standardized SectionCache cachex

lib/oli/delivery/sections.ex
```diff
@@ -1455,15 +1434,19 @@ defmodule Oli.Delivery.Sections do
...
-         resource_id,
-         resource_link_map,
-         container_ids,
-         seen
-       ) do
+  @doc """
+  Finds the first parent container for a given resource. If the resource_id given is a
+  container, then it will be the resource_id returned
+  """
+  def find_parent_container(
```
eliknebel, line 1441: Made this function public to add unit tests

lib/oli/delivery/sections.ex
```diff
@@ -1630,69 +1613,41 @@ defmodule Oli.Delivery.Sections do
...
+  def fetch_ordered_container_labels(section_slug, opts \\ []) do
+    short_label = opts[:short_label] || false
 
-    ordered_containers =
-      SectionCache.get_or_compute(section_slug, :ordered_containers, fn ->
-        fetch_ordered_containers(section_slug)
-      end)
+    # TODO: OPTIMIZATION replace this with a minimal hierarchy query after v26.2 is merged
+    %Section{customizations: customizations} = get_section_by_slug(section_slug)
+    full_hierarchy = DeliveryResolver.full_hierarchy(section_slug)
```
eliknebel, line 1626: This is not optimal, can be improved once the MinimalHierarchy improvements in v0.26.2 land

lib/oli/delivery/sections.ex
```diff
@@ -1630,69 +1613,41 @@ defmodule Oli.Delivery.Sections do
...
-  defp fetch_ordered_containers(section_slug) do
-    container_type_id = Oli.Resources.ResourceType.get_id_by_type("container")
+  defp label_for(numbering_level, numbering_index, title, short_label, customizations) do
+    container_label =
+      get_container_label(
+        numbering_level,
+        customizations || Map.from_struct(CustomLabels.default())
+      )
 
-    SectionResource
```
eliknebel, line 1663: This logic of using a single query and numbering level/index was not quite right. The only way to actually get the correct ordering according to the hierarchy is to load the actual hierarchy and use it to sort

lib/oli/delivery/sections/section_cache.ex
```diff
@@ -52,7 +53,7 @@ defmodule Oli.Delivery.Sections.SectionCache do
         "computed_value"
 
   """
-  def get_or_compute(section_slug, key, fun) do
+  def get_or_compute(section_slug, key, fun) when key in @cache_keys do
```
eliknebel, line 56: Add a guard so that incorrect/unexpected keys cannot be used

lib/oli_web/live/delivery/student/discussions_live.ex
```diff
@@ -1029,12 +1029,20 @@ defmodule OliWeb.Delivery.Student.DiscussionsLive do
               "Page #{post.resource_numbering_index}"
 
             container_resource_id ->
-              container_with_numbering_index =
-                Map.get(ordered_containers_map, container_resource_id)
-                |> String.split(":")
-                |> hd()
-
-              "#{container_with_numbering_index}: Page #{post.resource_numbering_index}"
+              case Map.get(ordered_containers_map, container_resource_id) do
```
eliknebel, line 1032: Handle the case where the container_resource_id is the curriculum resource_id, which is not actually present in the new ordered_containers_map

PR #4603, opened 2024-01-26
[Enhancement] Submit/Reset button UI changes [MER-2512]

marc-hughes:
On practice pages many activities have a submit button.
Before this change, the submit button would dissapear and a reset button would appear in the same location.
A learner who accidently double-clicked might submit and then mistakenly reset the activity.

This change moves us to a model where the submit & reset buttons are always visible, however they are disabled when they shouldn't be used.

resetbutton

This affects:

- Check all that apply
- Short Answer
- LikerT
- Multi Input
- Ordering
- File Upload
- Image Hotspot
- Response Multi Input
- VLab


eliknebel (approved):
Nice approach, works as expected.

IMO the multiple choice should have a Reset button now to be consistent with all the other activities so that students must explicitly reset the activity to start a new attempt. But this was understandably not done here as it was not part of the ticket

PR #4611, opened 2024-01-30
[MER-2959] use macros to generate resource type functions

Francisco-Castro:
Ticket [MER-2959](https://eliterate.atlassian.net/browse/MER-2959)

This PR replaced some of the functions contained in the `Oli.Resources.ResourceType` module using unquote fragments.

It remains the `get_id_by_type/1` but a new set of functions that starts with `id_for_(type)` were also included and promoted in this PR. The reason to let `get_id_by_type` is because there are 2 places [here](https://github.com/Simon-Initiative/oli-torus/blob/f31f9ace39665bc993862f5b9815b109636857f9/lib/oli/publishing.ex#L112) and [here](https://github.com/Simon-Initiative/oli-torus/blob/f31f9ace39665bc993862f5b9815b109636857f9/lib/oli_web/controllers/page_delivery_controller.ex#L1430) that receives the `type` as a variable.

One of the main advantages of using this approach is the autocompletion. By just typing `id_for` lead us to all the possible available options.

If this approach is beneficial we could also implemented for the `Oli.Resources.ScoringStrategy` module.

[MER-2959]: https://eliterate.atlassian.net/browse/MER-2959?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


eliknebel (approved):
nice improvement, makes the code much easier to read and autocomplete is a nice bonus

lib/oli/resources/resource_type.ex
```diff
@@ -2,55 +2,26 @@ defmodule Oli.Resources.ResourceType do
...
 
-  def get_id_by_type("page"), do: Map.get(@by_type, "page")
-  def get_id_by_type("container"), do: Map.get(@by_type, "container")
-  def get_id_by_type("activity"), do: Map.get(@by_type, "activity")
-  def get_id_by_type("objective"), do: Map.get(@by_type, "objective")
-  def get_id_by_type("secondary"), do: Map.get(@by_type, "secondary")
-  def get_id_by_type("tag"), do: Map.get(@by_type, "tag")
-  def get_id_by_type("bibentry"), do: Map.get(@by_type, "bibentry")
-  def get_id_by_type("alternatives"), do: Map.get(@by_type, "alternatives")
+    def unquote(String.to_atom("is_#{resource_type}"))(revision),
```
eliknebel, line 16: Would it be safer to use String.to_existing_atom here?

PR #4614, opened 2024-01-30
[Bugfix] Fix date-handling bug when today's date is later than the target month

marc-hughes:
If today's date has a date-part that's bigger than the number of days in the target month, you could not set the schedule to that date because of the way the Date object validates dates.

Explanation of bug:

const d = new Date();  // ex: Date is 1/30/2024
d.setFullYear(2025) // Date is now 1/30/2025
d.setMonth(2); // Date tries to be 2/30/2025, which doesn't exist, and rolls over to 3/1/2025
d.setDate(15); // Date is now 3/15/2025 instead of 2/15/2025 as expected

### DIFF
--- a/assets/src/apps/scheduler/PageScheduleLine.tsx
+++ b/assets/src/apps/scheduler/PageScheduleLine.tsx
@@ -56,7 +56,7 @@ export const PageScheduleLine: React.FC<ScheduleLineProps> = ({ item, indent, da
 
       // On a drag, need to change the date, but preserve the end time if one exists.
       if (item.endDateTime) {
-        targetEndDate = new Date(2024,1,1);
+        targetEndDate = new Date(2024, 1, 1);
         targetEndDate.setFullYear(endDate.getFullYear());
         targetEndDate.setMonth(endDate.getMonth());
         targetEndDate.setDate(endDate.getDate());
--- a/assets/src/apps/scheduler/PageScheduleLine.tsx
+++ b/assets/src/apps/scheduler/PageScheduleLine.tsx
@@ -56,10 +56,8 @@ export const PageScheduleLine: React.FC<ScheduleLineProps> = ({ item, indent, da
 
       // On a drag, need to change the date, but preserve the end time if one exists.
       if (item.endDateTime) {
-        targetEndDate = new Date(2024, 1, 1);
-        targetEndDate.setFullYear(endDate.getFullYear());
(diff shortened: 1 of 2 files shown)


assets/src/apps/scheduler/PageScheduleLine.tsx
```diff
@@ -36,7 +36,7 @@ export const PageScheduleLine: React.FC<ScheduleLineProps> = ({ item, indent, da
       let targetStartDate: Date | DateWithoutTime | null = startDate;
 
       if (item.startDateTime && startDate) {
-        targetStartDate = new Date();
+        targetStartDate = new Date(2024, 1, 1);
```
eliknebel, line 39: Is there a specific reason this date is hard coded? Or is it just a starting date that is required and will always be overridden so the value doesn't actually matter here?

marc-hughes: I missed this one when I changed it to the 3-param version of setFullYear.

But you will find one like this later because DateWithoutTime doesn't have a 3-param version of setFullYear and we needed to start with a 31 day month.

assets/src/apps/scheduler/scheduler-slice.ts
```diff
@@ -233,7 +235,7 @@ const schedulerSlice = createSlice({
           datesChanged =
             datesChanged || action.payload.endDate.getTime() !== mutableItem.endDateTime?.getTime();
 
-          mutableItem.endDate = new DateWithoutTime();
+          mutableItem.endDate = new DateWithoutTime(2020, 1, 1);
```
eliknebel, line 238: Do we want to update this to use the 3 param setFullYear as well?

eliknebel (changes requested):

assets/src/apps/scheduler/scheduler-slice.ts
marc-hughes: The DateWithoutTime doesn't have the 3-param version of setYear.

darrensiegel (approved):

tzs:
Possible dumb question since I have had very little occasion to use the Date class and so am probably missing something obvious, but it looks like all these places are creating a new Date object and initializing it to the same date/time as an existing date object.

Couldn't they all be replaced by using the form of the Date constructor that takes an existing date object? E.g., targetStartDate = new Date(startDate)?

PR #4616, opened 2024-01-31
[BUG] [MER-2925] | Variables not recalling on screen refresh

dtiwarATS:
Hey @bsparks,  can you please look at the PR and approved it. Thanks

### DIFF
--- a/assets/src/apps/delivery/layouts/deck/DeckLayoutFooter.tsx
+++ b/assets/src/apps/delivery/layouts/deck/DeckLayoutFooter.tsx
@@ -3,6 +3,7 @@ import { useDispatch, useSelector } from 'react-redux';
 import { applyState, templatizeText } from 'adaptivity/scripting';
 import { savePartState } from 'apps/delivery/store/features/attempt/actions/savePart';
 import { updateGlobalUserState } from 'data/persistence/extrinsic';
+import { writePageAttemptState } from 'data/persistence/state/intrinsic';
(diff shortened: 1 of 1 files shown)


assets/src/apps/delivery/layouts/deck/DeckLayoutFooter.tsx
```diff
@@ -328,6 +333,24 @@ const DeckLayoutFooter: React.FC = () => {
         }
       }
     });
+
+    //when lesson 'variables' were getting update via mutate state, we were not sending the updated values to server
```
bsparks, line 337: my recollection was that variables were author time settings only that never changed, not something that was persisted even.

bsparks (approved):
I'll approve if you want to change, but I thought that variables were intentionally not persisted

assets/src/apps/delivery/layouts/deck/DeckLayoutFooter.tsx
dtiwarATS: I confirmed with **@Mia** and she said -> " I think in most cases we didn't change variable values, but this was not always the case, several older/migrated lessons change the value of variables throughout. Lately, we've started using variables as a tool to help personalize and randomize the lessons.
So, I think proceeding with the PR to enable variable persistence would be beneficial. Unless there are any potential drawbacks or unforeseen consequences"

PR #4627, opened 2024-02-02
[Bugfix] Could not "add expression" to activity banks in groups [MER-2913]

marc-hughes:
1. Create a group
2. Add an activity bank inside the group
3. Click the add-expression button

Expected: An expression is added
Actual: Nothing is added

Cause of bug: When we save the page, we pass it through PageEditor::adjustContentForConstraints before sending it to the server to make sure we don't have any invalid expressions.

That call to adjustContentForConstraints should have been an immutable operation that generated a new object with any changes. There was a bug in PageEditorContent::updateAll that caused a mutation of the original object instead of purely returning a new one. This caused the expression to be stripped out in-memory as well as in the version we sent to the server, which was not wanted. See notes in the new unit test for clarity

### DIFF
--- a/assets/src/data/editor/PageEditorContent.ts
+++ b/assets/src/data/editor/PageEditorContent.ts
@@ -169,7 +169,7 @@ export class PageEditorContent extends Immutable.Record(defaultParams()) {
  * @param content
  * @returns
  */
-function withDefaultContent(
+export function withDefaultContent(
   content: Immutable.List<ResourceContent>,
 ): Immutable.List<ResourceContent> {
   if (content.size > 0) {
@@ -301,7 +301,11 @@ function updateAll(
 ): Immutable.List<ResourceContent> {
   return items.map((i) => {
     if (isResourceGroup(i)) {
-      (i as ResourceGroup).children = updateAll((i as ResourceGroup).children, fn);
+      const itemWithChildren = {
+        ...i,
+        children: updateAll((i as ResourceGroup).children, fn),
+      } as ResourceGroup;
+      return fn(itemWithChildren);
     }
 
     return fn(i);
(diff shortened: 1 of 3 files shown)


assets/test/editor/page_editor_content_test.ts
```diff
@@ -0,0 +1,146 @@
...
+
+    // At this point we have a PageEditorContent with a group inside it that has an activity bank selection inside the group
+
+    // Sanity check to make sure the child is in the right place before we call adjustContentForConstraints
+    expect((newContent1.model.get(1) as PurposeGroupContent)?.children.get(0)).toEqual(child);
+
+    const newContent2 = PageEditor.adjustContentForConstraints(newContent1);
+
+    // adjustContentForConstraints should not mutate it's input
+    expect((newContent1.model.get(1) as PurposeGroupContent)?.children.get(0)).toEqual(child);
```
marc-hughes, line 127: This line, where it makes sure the input to adjustContentForConstraings was the same as before it was called would fail before the fix was made.

assets/src/data/editor/PageEditorContent.ts
```diff
@@ -301,7 +301,11 @@ function updateAll(
 ): Immutable.List<ResourceContent> {
   return items.map((i) => {
     if (isResourceGroup(i)) {
-      (i as ResourceGroup).children = updateAll((i as ResourceGroup).children, fn);
+      const itemWithChildren = {
```
marc-hughes, line 304: `(i as ResourceGroup).children = updateAll((i as ResourceGroup).children, fn);` here was causing the mutation. We should have been creating a new object instead.

darrensiegel (approved):
Nice work tracking this down.

PR #4628, opened 2024-02-02
[BUGFIX] [MER-2915] Additional activity summarization only visible when answering more than once

simonchoxx:
[MER-2915](https://eliterate.atlassian.net/browse/MER-2915)

This PR seeks to correct the error mentioned in the ticket above. 

To do so, it was necessary to add to the actions that are executed when pressing the `Reset` button (to reset the activity) the `survey_id` information that ends up propagating to the Activity Attempts table in the database. 

It's necessary to set the `survey_id` whenever the activity is reset as this data must be added to the records created for that activity in the `Activity Attempts` table.

This change was necessary because the `Practice Activities` tab should NOT show the details of the activities that are contained within a survey.

[MER-2915]: https://eliterate.atlassian.net/browse/MER-2915?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/assets/src/phoenix/activity_bridge.ts
+++ b/assets/src/phoenix/activity_bridge.ts
@@ -74,7 +74,7 @@ export const initActivityBridge = (elementId: string) => {
       makeRequest(
         `/api/v1/state/course/${e.detail.sectionSlug}/activity_attempt/${e.detail.attemptGuid}`,
         'POST',
-        {},
+        { survey_id: e.detail.props.context.surveyId },
         e.detail.continuation,
       );
     },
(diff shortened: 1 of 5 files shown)


gastonabella (approved):
Looks good!
What do you think about adding a simple test case to the `AttemptController` that ensures that the survey id is correctly passed when resetting the activity?

darrensiegel (approved):

PR #4632, opened 2024-02-08
[ENHANCEMENT] show multi-input feedbacks in order of inputs in stem [MER-2995]

andersweinstein:
For multi-part questions, the torus Evaluation component shows multiple feedbacks in order of occurrence of parts in the question model. However, migrated questions may list parts in arbitrary order (in fact, order corresponding to order in the inputs section of the legacy XML), leading to confusion for students in matching feedbacks to parts on these questions. For example, if a question has 3 inputs on successive lines 1, 2, 3, feedbacks may be listed in order 3,1,2.

This PR sorts multi-input feedbacks in order of occurrence of the input_refs in the question stem. This is taken to be order in which they are listed by elementsOType function which effectively does an in-order traversal of the stem content tree. This generally finds them in reading order from left-to-right then top-to-bottom. Makes no attempt to account for different text direction. 

In cases of tabular layout, left-to-right top-to-bottom might not be the most obvious order, but most questions seen are not of this form, and there is no unambiguous order to use for those in any case. 

Implementation uses an optional partOrder prop which is only set by multiInputDelivery, so should leave other activity's feedback unchanged. 

This also moves a small utility function `isDefined` into a common module.

### DIFF
--- a/assets/src/components/activities/multi_input/MultiInputDelivery.tsx
+++ b/assets/src/components/activities/multi_input/MultiInputDelivery.tsx
@@ -33,6 +33,7 @@ import { configureStore } from 'state/store';
 import { DeliveryElementProvider, useDeliveryElementContext } from '../DeliveryElementProvider';
 import { SubmitResetConnected } from '../common/delivery/SubmitReset';
 import { initializePersistence } from '../common/delivery/persistence';
+import { orderedPartIds } from './utils';
 
 export const MultiInputComponent: React.FC = () => {
   const {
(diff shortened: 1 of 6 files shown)


assets/src/components/activities/multi_input/MultiInputDelivery.tsx
```diff
@@ -344,6 +345,7 @@ export const MultiInputComponent: React.FC = () => {
           }
           attemptState={uiState.attemptState}
           context={writerContext}
+          partOrder={orderedPartIds(model)}
```
darrensiegel, line 348: `orderedPartIds` gets called on every render here, which is not needed.  This should be run one time and the results stored.

darrensiegel (changes requested):

assets/src/components/activities/multi_input/MultiInputDelivery.tsx
andersweinstein: OK, done.

darrensiegel (approved):

PR #4644, opened 2024-02-14
[NG23-73] Schedule view show student progress (part 2)

eliknebel:
https://eliterate.atlassian.net/browse/NG23-73

This PR follows up the original schedule view work to fill in the remaining gaps from the designs. It add dynamic styling and indicators to the schedule according to the active week, page purpose specific icons, and attempts dropdown to scored pages so that attempts can be easily seen and reviewed from the schedule view.

Screenshot 2024-02-14 at 11 42 33 AM


lib/oli/delivery/settings.ex
```diff
@@ -29,6 +31,27 @@ defmodule Oli.Delivery.Settings do
     combine(resolved_revision, section_resource, nil)
   end
 
+  def get_combined_settings_for_all_resources(section_id, user_id) do
+    section = Oli.Delivery.Sections.get_section!(section_id)
+
+    all_pages = DeliveryResolver.all_pages(section.slug)
```
darrensiegel, line 37: `get_combined_settings_for_all_resources` seems like it is fetching a lot of information.  The call to `DeliveryResolver.all_pages/1` in a large course can return 500 or so full revisions, containing all JSON content.  What is actually needed here from the Revision? Only the title?  I think instead this call and perhaps the call to `get_section_resources` should be combined into new query that fetches the absolute minimum pieces of information needed.

darrensiegel (changes requested):

lib/oli/delivery/settings.ex
eliknebel: Reimplemented this query to be more efficient, returning the section resource and only the required settings

darrensiegel (approved):

PR #4646, opened 2024-02-16
Mer 2980 incorrect sorting in due date for scored activities

Francisco-Castro:
Ticket: [MER-2980](https://eliterate.atlassian.net/browse/MER-2980)

This PR fixed the sorting issue mentioned on the aforementioned ticket (see commit [1](https://github.com/Simon-Initiative/oli-torus/pull/4646/commits/23195d168edd0d27515d8f009b8d66a7189275ad) and [2](https://github.com/Simon-Initiative/oli-torus/pull/4646/commits/1bf5fddb7292de27c2ae74a3d80299e6de6b075c)). There is also a refactor over the `ScoreActivies` module to improve readability (see commit [3](https://github.com/Simon-Initiative/oli-torus/pull/4646/commits/8bfd93f5b36d400cc8c44eabc6382402784afb85) and [4](https://github.com/Simon-Initiative/oli-torus/pull/4646/commits/464608c355904f93f7e09229b203463cd41555af)).

[MER-2980]: https://eliterate.atlassian.net/browse/MER-2980?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli_web/components/delivery/scored_activities/scored_activities.ex
+++ b/lib/oli_web/components/delivery/scored_activities/scored_activities.ex
@@ -493,13 +493,7 @@ defmodule OliWeb.Components.Delivery.ScoredActivities do
   defp sort_by(assessments, sort_by, sort_order) do
     case sort_by do
       :due_date ->
-        Enum.sort_by(
-          assessments,
-          fn a ->
-            if a.scheduling_type != :due_by, do: 0, else: Map.get(a, :end_date)
-          end,
-          sort_order
-        )
+        Enum.sort_by(assessments, & &1.end_date, sort_order)
 
       sb when sb in [:avg_score, :students_completion, :total_attempts] ->
         Enum.sort_by(assessments, fn a -> Map.get(a, sb) || -1 end, sort_order)
--- a/lib/oli_web/components/delivery/scored_activities/scored_activities.ex
+++ b/lib/oli_web/components/delivery/scored_activities/scored_activities.ex
@@ -499,11 +499,7 @@ defmodule OliWeb.Components.Delivery.ScoredActivities do
         Enum.sort_by(assessments, fn a -> Map.get(a, sb) || -1 end, sort_order)
 
       :title ->
-        Enum.sort_by(
-          assessments,
-          fn a -> Map.get(a, :title) |> String.downcase() end,
-          sort_order
-        )
+        Enum.sort_by(assessments, &String.downcase(&1.title), sort_order)
     end
   end
 
--- a/lib/oli_web/components/delivery/scored_activities/scored_activities.ex
+++ b/lib/oli_web/components/delivery/scored_activities/scored_activities.ex
@@ -496,7 +496,7 @@ defmodule OliWeb.Components.Delivery.ScoredActivities do
         Enum.sort_by(
           assessments,
           fn a ->
-            if a.scheduling_type != :due_by, do: 0, else: Map.get(a, :due_date)
+            if a.scheduling_type != :due_by, do: 0, else: Map.get(a, :end_date)
           end,
           sort_order
         )
--- a/lib/oli_web/components/delivery/scored_activities/scored_activities.ex
+++ b/lib/oli_web/components/delivery/scored_activities/scored_activities.ex
@@ -427,6 +427,15 @@ defmodule OliWeb.Components.Delivery.ScoredActivities do
       :due_date ->
         Enum.sort_by(assessments, & &1.end_date, sort_order)
 
+        Enum.sort_by(
+          assessments,
+          fn
+            %{scheduling_type: :due_by} = assessment -> assessment.end_date
+            _ -> nil
+          end,
+          sort_order
+        )
+
       sb when sb in [:avg_score, :students_completion, :total_attempts] ->
         Enum.sort_by(assessments, fn a -> Map.get(a, sb) || -1 end, sort_order)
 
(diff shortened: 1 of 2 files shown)


lib/oli_web/components/delivery/scored_activities/scored_activities.ex
```diff
@@ -483,81 +418,45 @@ defmodule OliWeb.Components.Delivery.ScoredActivities do
...
     case sort_by do
       :due_date ->
-        Enum.sort_by(
-          assessments,
-          fn a ->
-            if a.scheduling_type != :due_by, do: 0, else: Map.get(a, :due_date)
-          end,
-          sort_order
-        )
+        Enum.sort_by(assessments, & &1.end_date, sort_order)
```
nicocirio, line 428: I think this fix would sort by end date instead of doing it by due date. My understanding is that we only want to consider assessments with `scheduling_type == :due_by` in the sorting (and group all other cases in one same "fake date"). 
I think the following code will work:
```
     Enum.sort_by(
          assessments,
          &if(&1.scheduling_type != :due_by, do: 0, else: Map.get(&1, :end_date)),
          sort_order
        )
```
Note that this code is almost the same as the original one, but I fixed it to get `:end_date` of the assessment instead of `:due_date` (that key does not exist in the map)

## Before

https://github.com/Simon-Initiative/oli-torus/assets/74839302/7b9658e4-f725-4815-9d7e-c7e7074fb9bb



## After

https://github.com/Simon-Initiative/oli-torus/assets/74839302/29610ebd-9467-4f50-a0af-7c04935c70a7

nicocirio (commented):
I left a suggestion that may work for the fix (sorting by due date and grouping all other cases, instead of sorting by end date). Let me know if this behavior is the expected one.
And what do you think about adding a test to the sorting?

lib/oli_web/components/delivery/scored_activities/scored_activities.ex
Francisco-Castro: Thanks @nicocirio for the suggestion and the explanation. I can see now that the Due Date column depends on two values: `scheduling_type` and the `section_resource.end_date`. I'll do that changes then and let you know.

Francisco-Castro:
@nicocirio Could you please take a look again?

test/oli_web/live/delivery/instructor_dashboard/scored_activities/scored_activities_tab_test.exs
```diff
@@ -1241,6 +1241,64 @@ defmodule OliWeb.Delivery.InstructorDashboard.ScoredActivitiesTabTest do
...
+      Sections.get_section_resource(section.id, page_1.resource_id)
+      |> Sections.update_section_resource(%{end_date: end_date_1, scheduling_type: :due_by})
+
+      # Only section resources of scheduling_type = due_by are considered when sorting by Due Date
+      # The next SectionResource has scheduling_type: :read_by and it shouldn't be considered
+      Sections.get_section_resource(section.id, page_2.resource_id)
+      |> Sections.update_section_resource(%{end_date: end_date_2})
+
+      Sections.get_section_resource(section.id, page_3.resource_id)
+      |> Sections.update_section_resource(%{end_date: end_date_3, scheduling_type: :due_by})
```
nicocirio, line 1265: NIT: I think this might improve the readability of the test

```Elixir
      Sections.get_section_resource(section.id, page_1.resource_id)
      |> Sections.update_section_resource(%{
        end_date: ~U[2000-01-21 12:00:00.00Z],
        scheduling_type: :due_by
      })

      # Only section resources of scheduling_type = due_by are considered when sorting by Due Date
      # The next SectionResource has scheduling_type: :read_by and it shouldn't be considered
      Sections.get_section_resource(section.id, page_2.resource_id)
      |> Sections.update_section_resource(%{end_date: ~U[2000-01-22 12:00:00.00Z]})

      Sections.get_section_resource(section.id, page_3.resource_id)
      |> Sections.update_section_resource(%{
        end_date: ~U[2000-01-23 12:00:00.00Z],
        scheduling_type: :due_by
      })
```

lib/oli_web/components/delivery/scored_activities/scored_activities.ex
```diff
@@ -427,6 +427,15 @@ defmodule OliWeb.Components.Delivery.ScoredActivities do
       :due_date ->
         Enum.sort_by(assessments, & &1.end_date, sort_order)
 
+        Enum.sort_by(
+          assessments,
+          fn
+            %{scheduling_type: :due_by} = assessment -> assessment.end_date
+            _ -> nil
+          end,
```
nicocirio, line 435: ❤️ this approach instead of the `if` I suggested

nicocirio (approved):
Nice work Fran! 🚀

test/oli_web/live/delivery/instructor_dashboard/scored_activities/scored_activities_tab_test.exs
Francisco-Castro: Interesting approach...

Another way to do it would be something like:

```
[
  {end_date: ~U[2000-01-21 12:00:00.00Z], scheduling_type: :due_by}
  {end_date: ~U[2000-01-22 12:00:00.00Z], scheduling_type: :read_by}
  {end_date: ~U[2000-01-23 12:00:00.00Z], scheduling_type: :due_by}
] 
|> Enum.each( fn {datetime, scheduling_type} ->
  Sections.get_section_resource(section.id, page_3.resource_id)
  |> Sections.update_section_resource(datetime, scheduling_type)
end
)
```

and maybe put the comment on top.

PR #4648, opened 2024-02-19
[NG23-70] Support adaptive pages on NG23

gastonabella:
[NG23-70](https://eliterate.atlassian.net/browse/NG23-70)

Support adaptive (exploration) pages on NG23:
- Extend `Learn` and `Lesson` live views to support adaptive pages.
- Support navigation to adaptive pages. Navigation from an adaptive page will be addressed in a separate PR.
- Update the `previous_next_index` field for Sections to store whether the page is adaptive.

On a separate PR:
- Support attempts review on exploration pages.
- Allow back-and-forth navigation from adaptive pages.

Still pending on this PR:
- ~Rebuild `previous_next_index` on existing sections.~
- ~Fix review links for adaptive pages.~

[NG23-70]: https://eliterate.atlassian.net/browse/NG23-70?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/oli_web/router.ex
```diff
@@ -978,6 +978,21 @@ defmodule OliWeb.Router do
           live("/attempt/:attempt_guid/review", Delivery.Student.ReviewLive)
         end
       end
+
+      scope "/adaptive_lesson/:revision_slug" do
```
darrensiegel, line 982: I'm not understanding why we need a special route for adaptive pages. Specializing the route between basic and adaptive pages is likely going to make some other things break (e.g. embedding hyperlinks in page content will also have to be specialized) and is going to add some ongoing maintenance burden.  

Can we explore a different solution to this problem, one that does not involve a specialized route?

darrensiegel (changes requested):

lib/oli_web/router.ex
gastonabella: The main reason why we need another route is the `root_layout`. As you might know, the root layout for an adaptive page (`chromeless.html.heex`) differs from the standard layout, and according to the [Guides](https://hexdocs.pm/phoenix_live_view/live-layouts.html#root-layout) this layout is only rendered in the initial request and can only be set in the router via `plugs` or `live sessions`. Therefore, we need to differentiate the "normal" lesson route from the adaptive one.

One option I didn't explore because I found it a bit streaky, is to have one single generic layout that includes all the assets required for both scenarios and renders them conditionally in the initial request depending on the page type. It has some downsides (it affects live navigation between lessons) and it doesn't sound like a "LiveView solution". Do you have any other ideas?

darrensiegel: Not all adaptive pages render with the `chromeless.html.heex`.  Only those adaptive page that have a "fullscreen" setting active render with that as the root layout.   

But I see now how the chromeless option is being handled.  There is already a `full_screen` route in the router and a function impl in `page_delivery_controller.ex`.  

We can move forward with a solution that inserts this new `adaptive_lesson` scope *only* if the existing generic routes that render pages will still work, and will still ultimately render the page correctly when a student clicks on a generic link of the form `/page/:revision_slug`, for all three cases where the page is:

1. A basic page
2. An adaptive page in full screen mode
3. An adaptive page NOT in full screen mode

gastonabella: @darrensiegel I ended up coding your solution using a plug that handles the redirection to an adaptive page so that we don't have to worry about deciding which route we should call. Let me know if it makes sense to you. Thank you!

darrensiegel (approved):

PR #4649, opened 2024-02-19
[MER-2163] archive media items

rgachuhi:
The following changes have been implemented in this PR to provide support for archiving media items from the media library:

1. Authors can now label a media item as “archived” which will remove it from the media library.
2. When an author tries to archive a media item in the library, there will be a prompt asking them to confirm the process. The prompt will inform them that archived media items will no longer be available in the media library and the archiving process is not reversible.
3. The archived media item will still remain in the database and can be used for reference in older sections.
4. Once the media item has been archived, the author can upload a new media item with the same name.


darrensiegel (approved):
Really nice work

PR #4658, opened 2024-02-26
[ENHANCEMENT] Ensure unique ids are present on next publication

darrensiegel:
This PR adds the infrastructure that will assign unique ids to the necessary page and activity content.  

Module `Oli.Publishing.UniqueIds` is introduced which encapsulates the entirety of the implementation, providing a public facing `add_unique_ids/1` function which is called during the publication process.  The idea here is to simply do this "uniqueify" operation on course publication.  This has the advantage that it requires no special, admin facing "one off" invocation. 

This is a higher risk change as it introduces a programmatic change of all page and activity content.  This is high risk in that something could be changed which breaks existing page or activity functionality - and high risk in that it could introduce a large load on the server when it goes to change 3000 revisions (roughly the number of pages and activities in the Chem course). 

To prevent functional problems, the "traversal" of the page and activity content is designed to stop at the paragraph level and to ignore `input_ref` elements.  The activity specific traversal only touches the "content" portions of the stem, choices, hints and feedbacks. 

Performance risk is mitigated through a streaming and batching implementation.  For the CHEM course for 3000 revisions, the batching of size 50 will require 60 serial DB writes.  This should be okay as the entirety of the publishing process is already done async - and it is only a one time process.


lib/oli/authoring/locks.ex
```diff
@@ -193,6 +193,23 @@ defmodule Oli.Authoring.Locks do
     |> Repo.update_all(set: [locked_by_id: nil, lock_updated_at: nil])
   end
 
+  @doc """
+  Acquires all locks for revisions in the supplied publication for the given user.
+
+  Returns:
+  .`{number, nil | returned data}` where number is the number of rows updated
+  """
+  @spec release_all(binary) :: {number, nil | term()}
```
eliknebel, line 202: This spec seems to be a copy/paste error

eliknebel, line 202: This spec seems to be a copy/paste error

eliknebel (approved):
Changes look good. I just noticed an incorrect spec, but I'll leave it up to you if you want to fix or not

PR #4660, opened 2024-02-27
[NG23-94] Change assistant enabled default value

gastonabella:
[NG23-94](https://eliterate.atlassian.net/browse/NG23-94)

Update the default value of the `assistant_enabled` field in the Section schema to `false`.
Add changeset tests to the Section schema.

**Note:**
The AC states: _"Also, only an Admin should be able to access the UX to enable and disable the Agent. "_.
However, that is the current behavior according to https://github.com/Simon-Initiative/oli-torus/blob/nextgen-ux/lib/oli_web/live/sections/overview_view.ex#L411. No changes should be required here.

[NG23-94]: https://eliterate.atlassian.net/browse/NG23-94?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli/delivery/sections/section.ex
+++ b/lib/oli/delivery/sections/section.ex
@@ -18,6 +18,13 @@ defmodule Oli.Delivery.Sections.Section do
     Section
   }
 
+  @required_fields [
+    :type,
+    :title,
+    :registration_open,
+    :base_project_id
+  ]
+
   schema "sections" do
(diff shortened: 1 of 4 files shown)


nicocirio (commented):
LGTM! I saw there are two tests from other module that failed and I think it has to do with the assistant not being enabled by default.

darrensiegel (approved):

PR #4661, opened 2024-02-28
[MER-2356] student added as instructor

rgachuhi:
This PR addresses an issue where, when logged in as a student and admin in the same browser and creating a course section as admin, the student is added as an instructor to the new course.

### DIFF
--- a/lib/oli/publishing.ex
+++ b/lib/oli/publishing.ex
@@ -392,46 +392,42 @@ defmodule Oli.Publishing do
       []
   """
   def retrieve_visible_sources(user, institution) do
-    case user do
-      nil ->
-        []
+    sources =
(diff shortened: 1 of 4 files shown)


lib/oli/publishing.ex
```diff
@@ -392,39 +392,43 @@ defmodule Oli.Publishing do
       []
   """
   def retrieve_visible_sources(user, institution) do
```
darrensiegel, line 394: The more "Elixir" way to handle user of `nil` is to simply add a pattern match impl and leave the original impl untouched: 

`def retrieve_visible_sources(nil, _), do: []`

darrensiegel (changes requested):

PR #4662, opened 2024-02-28
[Bugfix] Fix whitespace issues around short answer MER-3020

marc-hughes:
Bug in the short answer around whitespace.

>`When the correct answer contains a trailing space ("sábado domingo "), 
> entering the correct answer with or without a trailing space = incorrect.
>
> When the correct answer does not contain a trailing space ("sábado domingo"), 
> entering the correct answer with or without a trailing space = correct (as expected).
>
> The trailing space seems to be trimmed in the student answer but not in authoring.

Now both of these scenarios work as expected. I ended up putting the fix down in makeRule so no future components would run into this as well since the whitespace is baked in more at that level.

### DIFF
--- a/assets/src/data/activities/model/rules.ts
+++ b/assets/src/data/activities/model/rules.ts
@@ -195,17 +195,18 @@ export const notRangeRule = (
 
 export const makeRule = (input: Input): string => {
   if (input.kind === InputKind.Text) {
+    const trimmedValue = input.value.trim();
     switch (input.operator) {
       case 'contains':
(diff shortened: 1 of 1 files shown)


marc-hughes:
> It occurs to me that because (I believe) torus rule evaluation applies whitespace normalization to student input, the same issue would occur for multiple internal spaces in an answer string. That may not be as likely a case, though.

@andersweinstein yup, you were right. I just updated the code to account for that too. Thanks!

darrensiegel (approved):

PR #4663, opened 2024-02-28
[Bugfix] Cut off formulas [MER-2961]

marc-hughes:
Bug: Exceptionally long formulas would get truncated in the display.

I went through a ton of examples and found a few more than reported. Unfortunately, there wasn't a single global fix for this one, so there may be undiscovered combinations of formatting that still cause this to happen.

## Formulas in body:

Before:
image

After:
image

## In the body of a group

Before:
image

After:
image

## Formulas in hints inside groups:

Before:
image

After:
image

## Formulas in MCQ options inside groups:

Before:
image

After:
image

## In Feedback

Before:
image

After:
image

## In Explanations

Before:
image

After:
image


darrensiegel (approved):
Nice work, this is great to finally have a fix for this which doesn't simply involve reducing the font size.

PR #4671, opened 2024-03-01
Mer 2748 no license present

Francisco-Castro:
Ticket: [MER-2748](https://eliterate.atlassian.net/browse/MER-2748)

This PR added license. The license is set up at Project level but it's rendered at Revision level because take into account Remixed course sections that contain material from different projects. 

#### When clicked on the cc logo it opens a new tab pointing to the corresponding cc webpage
image

#### Renders `License:` follows by the license entered by the user. When the text is too long then an overflow-y is applied.
image

## Footer without license renders as always
image

[MER-2748]: https://eliterate.atlassian.net/browse/MER-2748?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/oli_web/views/layout_view.ex
```diff
@@ -131,4 +132,58 @@ defmodule OliWeb.LayoutView do
...
+
+    ~H"""
+    <div
+      id="license"
+      class="container mx-auto flex flex-row px-10 items-center overflow-y-auto h-[40px] max-h-[40px] relative top-0"
+    >
+      <p class="h-full">License: <%= @custom_license_details %></p>
+    </div>
+    """
+  end
```
nicocirio, line 147: I think we can avoid assigning a new variable and directly access the custom license details from the license struct
```suggestion
  def render_license(%{license: %{license_type: :custom}} = assigns) do
    ~H"""
    <div
      id="license"
      class="container mx-auto flex flex-row px-10 items-center overflow-y-auto h-[40px] max-h-[40px] relative top-0"
    >
      <p class="h-full">License: <%= @license.custom_license_details %></p>
    </div>
    """
  end
```

lib/oli_web/views/layout_view.ex
```diff
@@ -131,4 +132,58 @@ defmodule OliWeb.LayoutView do
...
+    ~H"""
+    <div
+      id="license"
+      class="container mx-auto flex flex-row px-10 items-center overflow-y-auto h-[40px] max-h-[40px] relative top-0"
+    >
+      <div class="flex gap-2 items-center">
+        <a href={@cc_url} , target="_blank">
+          <img
+            class="w-[100px]"
+            src={"#{Routes.static_path(OliWeb.Endpoint, "/images/cc_logos/#{@logo_name}.svg")}"}
```
nicocirio, line 174: What do you think about using `~p` instead of `Routes`?

lib/oli_web/views/layout_view.ex
```diff
@@ -131,4 +132,58 @@ defmodule OliWeb.LayoutView do
...
+      Enum.find(CreativeCommons.cc_options(), fn cc_option -> cc_option.id == cc_license end)
+
+    logo_name =
+      cc_data.id |> Atom.to_string() |> String.replace("cc_", "") |> String.replace("_", "-")
+
+    cc_text = cc_data |> Map.get(:text) |> String.split(":") |> Enum.at(1)
+
+    assigns =
+      assigns
+      |> Map.put(:cc_data, cc_data)
```
nicocirio, line 160: I think we do not need to assign `cc_data` as we do not use it in the template.

lib/oli_web/templates/layout/_delivery_footer.html.eex
```diff
@@ -17,6 +18,7 @@
     <div class="text-right">
     </div>
   </div>
+    <%= render_license(assigns) %>
```
nicocirio, line 21: Instead of passing all the assigns to this function component, we could do `<%= render_license(assigns[:license]) %>` and define the fc as follows:
```elixir
def render_license(%{license: %{license_type: cc_license}} = assigns) do
   ...
end

def render_license(nil), do: nil
```

nicocirio (commented):
Nice work Fran! I ❤️  the use of embedded schemas!
I suggest testing the other license cases and testing the prologue page as well.
I left some NIT, please tell me your thoughts about them 😄

lib/oli_web/views/layout_view.ex
Francisco-Castro: Thanks for that suggestion! I put it like you said and it's working fine. For some reason I added the Map.put because I got a warning but it's not happening anymore. 🎉

lib/oli_web/views/layout_view.ex
Francisco-Castro: Good catch!

Francisco-Castro:
@nicocirio could you please take another look at this at your convenience?. I think I covered most of your suggestions. Really appreciate your review on this! ⭐

test/oli_web/controllers/page_delivery_controller_test.exs
```diff
@@ -636,6 +637,118 @@ defmodule OliWeb.PageDeliveryControllerTest do
...
+    # } do
+    #   enroll_as_student(%{section: section, user: user})
+
+    #   conn =
+    #     conn
+    #     |> get(Routes.page_delivery_path(conn, :page, section.slug, revision.slug))
+
+    #   assert html_response(conn, 200) =~ "Page one"
+    #   assert html_response(conn, 200) =~ section.title
+    # end
```
nicocirio, line 654: Should we uncomment this test or remove it?

nicocirio (commented):
Great work Fran, thanks for addressing those changes! There is one commented test that I think should be erased or uncommented 😄

test/oli_web/controllers/page_delivery_controller_test.exs
Francisco-Castro: Good catch! Removed!

lib/oli/authoring/course.ex
```diff
@@ -576,4 +578,34 @@ defmodule Oli.Authoring.Course do
...
+      distinct: true,
+      select: p.attributes
+    )
+    |> Repo.one()
+    |> case do
+      %{license: %{license_type: :custom} = license} ->
+        Map.from_struct(license)
+
+      %{license: %{license_type: license_type} = license} when license_type in @cc_options ->
+        Map.from_struct(license)
```
gastonabella, line 605: Is it necessary to distinguish these two cases when the output is the same? What do you think about something like:

```elixir
|> case do
      %{license: %{license_type: license_type} = license} when license_type in [:custom | @cc_options] ->
        Map.from_struct(license)
        
      nil ->
        nil
    end
```

lib/oli_web/views/layout_view.ex
```diff
@@ -131,4 +132,55 @@ defmodule OliWeb.LayoutView do
...
+    <div
+      id="license"
+      class="container mx-auto flex flex-row px-10 items-center overflow-y-auto h-[40px] max-h-[40px] relative top-0"
+    >
+      <p class="h-full">License: <%= @custom_license_details %></p>
+    </div>
+    """
+  end
+
+  def render_license(%{license_type: cc_license} = assigns) do
```
gastonabella, line 147: Should we handle the `:none` case differently since it doesn't have a url or image associated?

gastonabella (commented):
Awesome work! 
I love what it looks like with the embedded schemas 🚀  

Just left a nit comment and a question

lib/oli/authoring/course.ex
Francisco-Castro: Nice catch! We don't even need the guard because the `@cc_options` includes `:custom`.

lib/oli_web/views/layout_view.ex
Francisco-Castro: Good point! I'll use a guard to be more precise. Thanks!

lib/oli_web/views/layout_view.ex
```diff
@@ -131,4 +132,55 @@ defmodule OliWeb.LayoutView do
...
   defp show_collab_space?(%CollabSpaceConfig{status: :disabled}), do: false
   defp show_collab_space?(_), do: true
+
+  def render_license(%{license_type: :custom} = assigns) do
+    ~H"""
+    <div
+      id="license"
+      class="container mx-auto flex flex-row px-10 items-center overflow-y-auto h-[40px] max-h-[40px] relative top-0"
+    >
+      <p class="h-full">License: <%= @custom_license_details %></p>
```
darrensiegel, line 142: Hardcoding "License: " here probably isn't a good idea.  The user can always include that string in the custom portion of the license.

darrensiegel (changes requested):
I see two things that we need to fix here:

A) Removal of the hardcoded "License: " string in the "Custom" option rendering.

B) The handling of the default value for license is not consistent.  Here are steps to reproduce: 

1. After this branch gets applied (even with running `mix ecto.reset`) I see from the project page the default value of "Non-CC / Copyrighted / Other" correctly selected.  When I view a page in a course section in Delivery for that course I see nothing in the footer.  
2. Then from the project page change the license to another CC license and save it.  Verify that from Delivery that license text and image correctly show.
3. Then go back to Project overview and change it back to "Non-CC / Copyrighted / Other".  Now if you go to Delivery instead of seeing nothing like in step 1, you actually see:  

Screenshot 2024-03-11 at 8 47 45 AM

The issue here is the handling of the default value of `nil` and the patterns that exist in `render_license/1` in `layout_view.ex`.  I think we need to explicitly handle the `:none` and the `nil` case in both ways (i.e. by rendering `nil`)

darrensiegel:
I still see the default value issue.  The key here is that there is a requirement that an author can make a license selection which results in "displaying nothing at all".  Currently, the only way to do this is by never changing the license for an existing course section.  Once you change it, and then change it back to "None ..." the footer displays "Non-CC / Copyrighted / Other".

darrensiegel:
I pushed a couple of commits to:
1. Tweak the CSS styles to eliminate scrolling of the license portion and generally to allow it to display better on small form factors.
2. Removed the requirement that the custom license string is required.  There is a requirement here that an author has a way to literally display no license information.  They can now do this by selecting "Custom" and leaving the custom license text field blank.

darrensiegel (approved):

PR #4673, opened 2024-03-05
[NG23-51] Add support into Learn view for top level pages

nicocirio:
[Link](https://eliterate.atlassian.net/browse/NG23-51) to the ticket

## Before
https://github.com/Simon-Initiative/oli-torus/assets/74839302/659e4901-4a13-4597-8eb9-3ae2f1066c19

## After
https://github.com/Simon-Initiative/oli-torus/assets/74839302/108cfa64-6625-4ff6-a2e2-4ca4171135b0

### DIFF
--- a/lib/oli_web/live/delivery/student/learn_live.ex
+++ b/lib/oli_web/live/delivery/student/learn_live.ex
@@ -451,8 +451,8 @@ defmodule OliWeb.Delivery.Student.LearnLive do
   def unit(%{unit: %{"resource_type_id" => 1}} = assigns) do
     ~H"""
     <div id={"top_level_page_#{@unit["resource_id"]}"}>
-      <div class="md:p-[25px] md:pl-[50px]" role={"unit_#{@unit["numbering"]["index"]}"}>
-        <div class="flex flex-col md:flex-row md:gap-[30px]">
+      <div class="md:p-[25px] md:pl-[50px]" role={"top_level_page_#{@unit["numbering"]["index"]}"}>
+        <div role="header" class="flex flex-col md:flex-row md:gap-[30px]">
           <div class="text-[14px] leading-[19px] tracking-[1.4px] uppercase mt-[7px] mb-1 whitespace-nowrap opacity-60">
             <%= "PAGE #{@unit["numbering"]["index"]}" %>
           </div>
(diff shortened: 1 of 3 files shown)


lib/oli_web/live/delivery/student/learn_live.ex
```diff
@@ -433,6 +447,80 @@ defmodule OliWeb.Delivery.Student.LearnLive do
   attr :unit_raw_avg_score, :map
   attr :assistant_enabled, :boolean, required: true
 
+  # top level page as a card with title and header
+  def unit(%{unit: %{"resource_type_id" => 1}} = assigns) do
```
gastonabella, line 451: We probably need to think again about all the naming on this LiveView, but it could be a little bit confusing to keep this function component as "`unit`" when it could be also a page. I don't have the ideal name either, but maybe something more generic like `top_level_resource` or a descriptive name that explains more about its structure on the page (such as **row** or **slider**).

lib/oli_web/live/delivery/student/learn_live.ex
```diff
@@ -433,6 +447,80 @@ defmodule OliWeb.Delivery.Student.LearnLive do
...
+                    d="M10.0496 17.9996L4.34961 12.2996L5.77461 10.8746L10.0496 15.1496L19.2246 5.97461L20.6496 7.39961L10.0496 17.9996Z"
+                    fill="#0CAF61"
+                  />
+                </svg>
+              </div>
+            </div>
+          </div>
+        </div>
+        <div class="w-[288px]">
+          <.module_card
```
gastonabella, line 505: Same with this component which is no longer a module exclusively.

gastonabella (commented):
Looks great! Just left a couple of comments regarding function components naming.

lib/oli_web/live/delivery/student/learn_live.ex
nicocirio: Yes, I agree. I think `row` is very descriptive considering how the learn page structure is shown.

lib/oli_web/live/delivery/student/learn_live.ex
nicocirio: makes sense. I renamed it to `card`

gastonabella (approved):
Looks great! Thanks for making the changes 🙌

I noticed there are some conflicts to solve

PR #4678, opened 2024-03-08
[MER-3002] Number Enrolled Should be Students Only

tomasferok:
This pr [MER-3002](https://eliterate.atlassian.net/browse/MER-3002?filter=-1) add a subquery to only count the number of student enrollments

[MER-3002]: https://eliterate.atlassian.net/browse/MER-3002?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli/delivery/sections/browse.ex
+++ b/lib/oli/delivery/sections/browse.ex
@@ -1,6 +1,7 @@
 defmodule Oli.Delivery.Sections.Browse do
   import Ecto.Query, warn: false
 
+  alias Oli.Delivery.Sections.EnrollmentContextRole
   alias Lti_1p3.Tool.ContextRoles
   alias Oli.Accounts.User
   alias Oli.Delivery.Sections.{Section, Enrollment, BrowseOptions}
@@ -98,11 +99,14 @@ defmodule Oli.Delivery.Sections.Browse do
         },
         group_by: [e.section_id]
 
+    student_role_id =
+      ContextRoles.get_role(:context_learner).id
+
     student_enrollments =
       from e in Enrollment,
-        join: u in User,
-        on: e.user_id == u.id,
-        where: u.can_create_sections == false,
+        join: ecr in EnrollmentContextRole,
+        on: ecr.enrollment_id == e.id,
+        where: ecr.context_role_id == ^student_role_id,
         select: %{
           id: e.id,
           section_id: e.section_id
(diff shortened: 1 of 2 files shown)


gastonabella (approved):
Looks good!

lib/oli/delivery/sections/browse.ex
```diff
@@ -98,9 +98,19 @@ defmodule Oli.Delivery.Sections.Browse do
...
 
+    student_enrollments =
+      from e in Enrollment,
+        join: u in User,
+        on: e.user_id == u.id,
+        where: u.can_create_sections == false,
+        select: %{
+          id: e.id,
+          section_id: e.section_id
+        }
```
simonchoxx, line 109: Looking at the query I think that maybe instead of querying for the `can_create_sections` field to find out if a user is a Student or an Instructor, you could add a join in this query with the `enrollments_context_roles` table, which contains the information of what type of enrollment the user has. I comment it just to think about it...

simonchoxx (commented):
LGTM!!! Nice work Tomi!! 🚀

Adding a test to prove that these changes are working properly would not be a bad thing!

lib/oli/delivery/sections/browse.ex
nicocirio: I agree with Santi. A user who can create sections does not necessarily has to have an instructor role in all sections. A user can have the ability to create sections and be enrolled as a student in many courses. 
So, I think the right way to address the ticket is the way Santi suggests

nicocirio (changes requested):

test/oli_web/live/sections/admin_index_live_test.exs
```diff
@@ -283,5 +283,28 @@ defmodule OliWeb.Sections.AdminIndexLiveTest do
...
+      instructor = insert(:user)
+
+      enroll_user_to_section(student, section, :context_learner)
+      enroll_user_to_section(instructor, section, :context_instructor)
+
+      {:ok, view, _html} = live(conn, @live_view_index_route)
+
+      assert view
+             |> element("table > tbody > tr:first-child > td:nth-child(3) > div")
+             |> render() =~ "1"
```
simonchoxx, line 307: Maybe it would be more effective to try to match here against some function that returns the number of enrollments of the section, such as `Oli.Delivery.Sections.count_enrollments`, WDYT?

simonchoxx (approved):

darrensiegel (approved):

PR #4680, opened 2024-03-11
[BUG FIX] [MER-3055] Restrict zero activity score override to basic pages

darrensiegel:
The logic that was added to be able to score basic pages which contain zero activities did not take into account the fact that adaptive page evaluation only looks at "non active" activities.  So, if a user submits an attempt on the first screen in an adaptive page, that page attempt will contain zero "non active" activities and thus trip this case and get scored 1 out of 1.  Instead, we always want adaptive pages to go thru the standard scoring flow where it will likely get a content driven score override.


gastonabella (approved):
LGTM!

PR #4682, opened 2024-03-12
[BUG FIX] [SIM - CAPI Issue] | variable added in configData section of a SIM not getting replaced before sending it to SIM

dtiwarATS:
Hey @bsparks, please look at the PR.

To enhance performance, we've introduced an `"conditionsNeedEvaluation"` array in the authoring system. This array contains a list of variables that require evaluation. During delivery, the system checks if a given variable is present in this list to determine whether it should be evaluated or ignored.

In the authoring process, if a configuration data variable has an expression, it is added to the `"conditionsNeedEvaluation"` list as `'stage.simId.simVariable'`. However, when passing the `'Key'` from the external activity page, we were only passing 'simVariable' as the key to the `'templatizeText'` function. As a result, when the system checked if '`simVariable'` was present in '`conditionsNeedEvaluation`', it couldn't find it. Therefore, we need to send the entire key.

### DIFF
--- a/assets/src/components/parts/janus-capi-iframe/ExternalActivity.tsx
+++ b/assets/src/components/parts/janus-capi-iframe/ExternalActivity.tsx
@@ -163,6 +163,7 @@ const ExternalActivity: React.FC<PartComponentProps<CapiIframeModel>> = (props) 
       setScriptEnv(env);
     }
     simLife.domain = initResult.context.domain || 'stage';
+    simLife.simId = id;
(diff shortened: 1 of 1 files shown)


darrensiegel (approved):

darrensiegel (changes requested):
Please file a MER ticket and include a [MER-xxxx] reference to it here please

dtiwarATS:
> Please file a MER ticket and include a [MER-xxxx] reference to it here please

Done

PR #4686, opened 2024-03-13
[MER-2885]-[MER-2896] disable undo and redo

rgachuhi:
This PR enables or disables the undo or redo buttons to match the slate editor

### DIFF
--- a/assets/src/components/editing/toolbar/buttons/CommandButton.tsx
+++ b/assets/src/components/editing/toolbar/buttons/CommandButton.tsx
@@ -31,4 +31,5 @@ export const CommandButton = (props: Props) => {
       <ButtonContent description={props.description} />
     </button>
   );
+  //
 };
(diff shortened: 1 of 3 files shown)


assets/src/components/editing/toolbar/buttons/CommandButton.tsx
```diff
@@ -16,11 +17,16 @@ export const CommandButton = (props: Props) => {
 
   return (
     <button
-      className={classNames(styles.toolbarButton, active?.(editor) && styles.active)}
+      className={classNames(
+        'disabled:opacity-50',
+        styles.toolbarButton,
+        active?.(editor) && styles.active,
```
eliknebel, line 23: There appear to still be a hover style applied even though the  button is disabled, I think if possible we should remove the hover background change as well when disabled is true.

I think to accomplish this you need to add a CSS rule in `assets/src/components/editing/toolbar/Toolbar.modules.scss` line 89, something like

```
  &[disabled]:hover {
    background-color: transparent;
  }

```

eliknebel (changes requested):

PR #4688, opened 2024-03-14
MER-2979-admin-view-incorrect-sorting-on-sections-of-institutions

tomasferok:
This pr [MER-2979](https://eliterate.atlassian.net/browse/MER-2979) correct the sorting by cost in section and students view


simonchoxx (approved):
LGTM 🚀

PR #4689, opened 2024-03-14
[NG23-105] Create input for poster image

nicocirio:
[Link](https://eliterate.atlassian.net/browse/NG23-105) to the ticket

## Uploading
An admin can select an already uploaded image or upload a new one, for a container or for a page (in this case from both the curriculum and the all-pages view).

https://github.com/Simon-Initiative/oli-torus/assets/74839302/d169f580-851b-4f72-82d7-e9af72d392d2

## Validations
Size and file type validations prevent upload and disable the save button. The user can still upload a new poster image after validations are shown.

https://github.com/Simon-Initiative/oli-torus/assets/74839302/b2dd1650-ab45-4af5-8f67-b1674933c532

This PR also addresses this [ticket](https://eliterate.atlassian.net/browse/NG23-104). A test was added to validate that the `duration_minutes` is properly set into the revision when the options form is submitted.


eliknebel (approved):
Looks great, works as expected. Just a remaining failing test that needs looked at

PR #4692, opened 2024-03-15
[MER-3071] Fix datashop export database timeout issue

eliknebel:
https://eliterate.atlassian.net/browse/MER-3071

This is a fix for the datashop generation error related to database connection timing out on large courses. There are 2 changes here:

1. Add the ability to configure this timeout system-wide with an environment variable
2. Set the specific datashop query transaction timeout to infinity, thereby using as much time as needed to generate.

When tested on heliotron with the Chemistry course with ~300 enrollments, this operation took about 3 hours. While this effectively consumes a database connection for that period, we can limit the number of simultaneous datashop jobs that run at a time so that the system doesn't starve itself of database connections (currently, the `datashop_export` queue is configured to 3).

These changes were tested against v0.26.5 but rebased with master for this PR

### DIFF
--- a/lib/oli/analytics/datashop_export_worker.ex
+++ b/lib/oli/analytics/datashop_export_worker.ex
@@ -60,32 +60,35 @@ defmodule Oli.Analytics.DatashopExportWorker do
     datashop_snapshot_path = build_filename(project)
 
     {:ok, result} =
-      Oli.Repo.transaction(fn ->
-        Oli.Utils.use_tmp(fn tmp_dir ->
-          temp_file_name =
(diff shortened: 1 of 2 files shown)


lib/oli/analytics/datashop_export_worker.ex
```diff
@@ -60,32 +60,35 @@ defmodule Oli.Analytics.DatashopExportWorker do
...
+              timestamp
+            )
 
-          {full_upload_url, timestamp}
-        end)
-      end)
+            {full_upload_url, timestamp}
+          end)
+        end,
+        timeout: :infinity
```
eliknebel, line 90: Allow datashop export jobs to run indefinitely

darrensiegel (approved):

PR #4720, opened 2024-03-27
[BUGFIX][MER-3000] Lost quiz answers second tab

simonchoxx:
[MER-3000](https://eliterate.atlassian.net/browse/MER-3000)

This PR aims to show an error message if a student would like to make an attempt at an activity that has already been submitted or evaluated.

https://github.com/Simon-Initiative/oli-torus/assets/16328384/400e0287-28f3-4163-b124-c1ce22e38999

[MER-3000]: https://eliterate.atlassian.net/browse/MER-3000?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/assets/src/phoenix/activity_bridge.ts
+++ b/assets/src/phoenix/activity_bridge.ts
@@ -28,7 +28,9 @@ function makeRequest(
 
 const nothingTransform = (result: any) => Promise.resolve(result);
 const submissionTransform = (key: string, result: any) => {
-  return Promise.resolve({ actions: result[key] });
+  return result.error
+    ? Promise.reject({ message: result.message })
+    : Promise.resolve({ actions: result[key] });
 };
 
 export const initActivityBridge = (elementId: string) => {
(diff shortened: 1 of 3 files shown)


darrensiegel (changes requested):
This new logic should go inside of `ActivityEvaluation.evaluate_activity` and not in the controller, for two reasons:

1. We want to keep as much application business logic **out** of controllers, and instead push that logic down into contexts and modules that do not depend on `OliWeb`.   
2. Pushing this logic down into `ActivityEvaluation.evaluate_activity` allows us to eliminate the extra DB call to fetch the activity attempt. 

You will want to change `ActivityEvaluation.evaluate_activity` to return `{:error, :already_submitted}` in this case, and then set the appropriate JSON error and message in the controller.

darrensiegel (approved):

PR #4724, opened 2024-03-29
[NG23-81] user reacts to an existing note

eliknebel:
**NOTE:** Changes are currently based on [NG23-82 branch](https://github.com/Simon-Initiative/oli-torus/pull/4721) and includes those changes as well

https://eliterate.atlassian.net/browse/NG23-81

Implements the ability for a user to react to an existing note. Currently the only supported reaction is "like" however this could be expanded to other reaction types in the future.

Screenshot 2024-03-29 at 9 36 52 AM


darrensiegel (approved):
Nice work, I like the structure of the UserReactionPost - easy to extend to other reaction types in the future

PR #4725, opened 2024-04-02
[ENHANCEMENT] [MER-2932] Filter insights by product course section

tomasferok:
This pr [MER-2932](https://eliterate.atlassian.net/browse/MER-2932) correct the filter by section and product and solve the tab by page.

[MER-2932]: https://eliterate.atlassian.net/browse/MER-2932?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli/analytics/by_activity.ex
+++ b/lib/oli/analytics/by_activity.ex
@@ -2,6 +2,7 @@ defmodule Oli.Analytics.ByActivity do
   import Ecto.Query, warn: false
   alias Oli.Publishing.DeliveryResolver
   alias Oli.Repo
+  alias Oli.Resources.ResourceType
   alias Oli.Analytics.Common
   alias Oli.Publishing
 
@@ -17,7 +18,10 @@ defmodule Oli.Analytics.ByActivity do
   defp get_base_query(project_slug, filtered_sections) do
     subquery =
       if filtered_sections != [] do
-        DeliveryResolver.revisions_filter_by_section_ids(filtered_sections, 3)
+        DeliveryResolver.revisions_filter_by_section_ids(
+          filtered_sections,
+          ResourceType.id_for_activity()
+        )
       else
         Publishing.query_unpublished_revisions_by_type(
           project_slug,
--- a/lib/oli/analytics/by_activity.ex
+++ b/lib/oli/analytics/by_activity.ex
@@ -1,6 +1,6 @@
 defmodule Oli.Analytics.ByActivity do
   import Ecto.Query, warn: false
-  alias Oli.Delivery.Sections.SectionResource
+  alias Oli.Publishing.DeliveryResolver
   alias Oli.Repo
   alias Oli.Analytics.Common
   alias Oli.Publishing
@@ -11,18 +11,13 @@ defmodule Oli.Analytics.ByActivity do
   def query_against_project_slug(project_slug, filtered_sections) do
     project_slug
     |> get_base_query(filtered_sections)
-    |> get_query_with_join_filter(filtered_sections)
     |> Repo.all()
   end
 
   defp get_base_query(project_slug, filtered_sections) do
     subquery =
       if filtered_sections != [] do
-        Publishing.query_unpublished_revisions_by_type_and_section(
-          project_slug,
-          "activity",
-          filtered_sections
-        )
+        DeliveryResolver.revisions_filter_by_section_ids(filtered_sections, 3)
       else
         Publishing.query_unpublished_revisions_by_type(
           project_slug,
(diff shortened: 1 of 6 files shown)


lib/oli/analytics/by_objective.ex
```diff
@@ -14,18 +14,13 @@ defmodule Oli.Analytics.ByObjective do
...
 
   defp get_base_query(project_slug, activity_objectives, filtered_sections) do
     subquery =
       if filtered_sections != [] do
-        Publishing.query_unpublished_revisions_by_type_and_section(
-          project_slug,
-          "objective",
-          filtered_sections
-        )
+        DeliveryResolver.revisions_filter_by_section_ids(filtered_sections, 4)
```
simonchoxx, line 23: Same as above, but changing the type of resource.

lib/oli/analytics/by_page.ex
```diff
@@ -15,26 +15,13 @@ defmodule Oli.Analytics.ByPage do
...
-
   defp get_base_query(project_slug, activity_pages, filtered_sections) do
     subquery =
       if filtered_sections != [] do
-        Publishing.query_unpublished_revisions_by_type_and_section(
-          project_slug,
-          "page",
-          filtered_sections
-        )
+        DeliveryResolver.revisions_filter_by_section_ids(filtered_sections, 1)
```
simonchoxx, line 24: Same as above

lib/oli/analytics/by_page.ex
```diff
@@ -44,11 +31,7 @@ defmodule Oli.Analytics.ByPage do
 
     subquery_activity =
       if filtered_sections != [] do
-        Publishing.query_unpublished_revisions_by_type_and_section(
-          project_slug,
-          "activity",
-          filtered_sections
-        )
+        DeliveryResolver.revisions_filter_by_section_ids(filtered_sections, 3)
```
simonchoxx, line 34: Same as above

lib/oli/analytics/by_activity.ex
```diff
@@ -11,18 +11,13 @@ defmodule Oli.Analytics.ByActivity do
...
 
   defp get_base_query(project_slug, filtered_sections) do
     subquery =
       if filtered_sections != [] do
-        Publishing.query_unpublished_revisions_by_type_and_section(
-          project_slug,
-          "activity",
-          filtered_sections
-        )
+        DeliveryResolver.revisions_filter_by_section_ids(filtered_sections, 3)
```
simonchoxx, line 20: I think that here, in the second argument, we should use something like Oli.Resources.ResourceType.id_for_activity to not hardcode that id. WDYT?

NOTE: Oli.Resources.ResourceType should be an alias of our module so that we don't repeat it everywhere we have to call it.

simonchoxx (commented):
LGTM 🚀

lib/oli_web/live/insights/insights.ex
```diff
@@ -533,6 +531,34 @@ defmodule OliWeb.Insights do
...
+  defp get_by_page_row(socket, :by_objective) do
+    if socket.assigns.is_product do
+      Oli.Analytics.ByObjective.query_against_project_slug(
+        socket.assigns.project.slug,
+        socket.assigns.product_ids
+      )
+    else
+      Oli.Analytics.ByObjective.query_against_project_slug(
+        socket.assigns.project.slug,
+        socket.assigns.section_ids
```
Francisco-Castro, line 543: I'd like to clarify: How do you differentiate between a product_id and a section_id when they are passed as the second argument to the query_against_project_slug/2 function? Both are integers, right?

Francisco-Castro (changes requested):
You're making progress, but there are some critical points in your code that need to be addressed.

lib/oli_web/live/insights/insights.ex
tomasferok: section_ids and product_ids are both lists of integers. Yes, there are two handle_events: section_change and product_change. One handles the list of section_ids corresponding to what the user selects on the multiselect component, and the other handles the product_ids. Additionally, there is a variable called is_product, which is a boolean indicating which list (product_ids or section_ids) has to be passed in the query.

lib/oli/publishing/delivery_resolver.ex
```diff
@@ -263,6 +263,63 @@ defmodule Oli.Publishing.DeliveryResolver do
     |> emit([:oli, :resolvers, :delivery], :duration)
   end
 
+  def revisions_of_type_filtered_by_section_ids(section_ids, resource_type_id) do
```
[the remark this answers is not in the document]
tomasferok: I think that function is not being used anywhere. revisions_filter_by_section_ids is the function that is actually being used. Thanks for the observation, Fran!!!

lib/oli_web/live/insights/insights.ex
Francisco-Castro: As we discussed previously, both products and sections utilize the same struct, which is why they can be used interchangeably for the query_against_project_slug/2 function without needing a specific distinction.

darrensiegel (approved):

PR #4728, opened 2024-04-04
[BUG FIX] [MER-3001] Fix  incorrect sorting in student reports

tomasferok:
this pr correct the sorting by last_interaction [MER-3001](https://eliterate.atlassian.net/browse/MER-3001) 

[MER-3001]: https://eliterate.atlassian.net/browse/MER-3001?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli_web/components/delivery/students/students.ex
+++ b/lib/oli_web/components/delivery/students/students.ex
@@ -145,7 +145,7 @@ defmodule OliWeb.Components.Delivery.Students do
         Enum.sort_by(students, fn student -> student.email end, sort_order)
 
       :last_interaction ->
-        Enum.sort_by(students, & &1.last_interaction, {sort_order, Date})
+        Enum.sort_by(students, & &1.last_interaction, {sort_order, DateTime})
 
       :progress ->
         Enum.sort_by(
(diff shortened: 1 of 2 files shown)


lib/oli_web/components/delivery/students/students.ex
```diff
@@ -145,7 +145,19 @@ defmodule OliWeb.Components.Delivery.Students do
...
+              fn x, y -> DateTime.compare(x.last_interaction, y.last_interaction) != :gt end
+
+            :desc ->
+              fn x, y -> DateTime.compare(x.last_interaction, y.last_interaction) != :lt end
+          end
+
+        Enum.sort(
+          students,
+          comparator
+        )
```
Francisco-Castro, line 160: There is something weird happening here... I run the test with the previous code `Enum.sort_by(students, fn student -> student.last_interaction end, sort_order)` and your new code and both are making the test pass. Could you please check that?

Francisco-Castro (changes requested):
It appears that both the original and the updated code are successfully passing the test. 🤨

lib/oli_web/components/delivery/students/students.ex
tomasferok: the reason why this test is passing in both is because the previous code only compare with hour and ignore the year, if you put for example ~U[2019-04-05 13:28:42Z] on the set_interation for user_4, the previous code is going to fail.

Francisco-Castro: @tomasferok could you please alter the test to reflect your changes. Maybe add a test failing and a test passing so that you cover the two possible cases.

PR #4735, opened 2024-04-09
MER-3048-objective-filter-in-activity-bank-works-incorrectly

tomasferok:
This pr [MER-3028](https://eliterate.atlassian.net/browse/MER-3048) correct the filter by does_not_equal in Activity Bank


darrensiegel (approved):
Tested this locally and it works well and fixes the problem.  Nice work!

PR #4736, opened 2024-04-09
Mer 3007 generate reset password link

Francisco-Castro:
Ticket: [MER-3007](https://eliterate.atlassian.net/browse/MER-3007)

This PR added the ability to generate a Reset Password Link for user --students- in order to be share by an admin.

### DIFF
--- a/lib/oli_web/live/users/user_actions.ex
+++ b/lib/oli_web/live/users/user_actions.ex
@@ -7,6 +7,7 @@ defmodule OliWeb.Users.Actions do
   attr(:for_author, :boolean, default: false)
   attr(:user, :any, required: true)
   attr(:csrf_token, :any, required: true)
+  attr(:password_reset_link, :string, default: "")
 
   def render(assigns) do
     {resend, reset} =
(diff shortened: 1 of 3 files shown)


lib/oli_web/live/users/user_actions.ex
```diff
@@ -26,6 +27,42 @@ defmodule OliWeb.Users.Actions do
...
     <div>
+      <div class="form-group">
+        <label for="reset_link_input">Generate Reset Password Link</label>
+        <div class="input-group" style="max-width: 100%;">
+          <input
+            readonly
+            type="text"
+            id="invite-link-1"
+            class="form-control"
+            placeholder="Section Invite Link"
```
darrensiegel, line 38: Change "Section Invite Link" for the aria-label to "Password Reset Link Text Input" and since this is a read only input, simply remove the `placeholder` attribute altogether.

darrensiegel (changes requested):
Tested and works well, just one thing to change in the UI

darrensiegel (approved):

PR #4742, opened 2024-04-11
[BUG FIX] [MER-3152] Remove prevention of submission on already submitted attempts

darrensiegel:
Reverts the changes from MER-3000 as this causes problems for adaptive pages.

Note, in `evaluate.ex` the only real change is the removal of the outermost `if...then`, to remove:

```
if activity_attempt.lifecycle_state in [:submitted, :evaluated] do
      Logger.info(
        "These changes could not be saved as this attempt may have already been submitted"
      )
     {:error, :already_submitted}
else
```

The rest of the diff is unfortunately just simply whitespace change. 

To verify this change:

**Without this fix in place**

1. Ingest the attached course project, publish and create an Open and Free course
2. As a student, visit the adaptive page that is in the course section (the only one).
3. Visit any screen, then use Table Of Contents to navigate back to first screen. Click another screen.
Expect: View transitions to selected screen
Actual: 403 failure is thrown by server and screen does not transition

Then **with this fix in place**
1. As a student, visit the adaptive page that is in the course section (the only one).
2. Visit any screen, then use Table Of Contents to navigate back to first screen. Click another screen.
Expect: View transitions to selected screen
Actual: View transitions to selected screen

[export_component_library_for_testing (1).zip](https://github.com/Simon-Initiative/oli-torus/files/14949223/export_component_library_for_testing.1.zip)

### DIFF
--- a/lib/oli/delivery/attempts/activity_lifecycle/evaluate.ex
+++ b/lib/oli/delivery/attempts/activity_lifecycle/evaluate.ex
@@ -28,78 +28,71 @@ defmodule Oli.Delivery.Attempts.ActivityLifecycle.Evaluate do
     activity_model = select_model(activity_attempt)
     part_attempts = get_latest_part_attempts(activity_attempt_guid)
 
-    if activity_attempt.lifecycle_state in [:submitted, :evaluated] do
-      Logger.info(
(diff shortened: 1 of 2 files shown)


eliknebel (approved):
Tested, works as expected

PR #4751, opened 2024-04-16
[FEATURE] [NG23-83] user searches within notes panel

eliknebel:
https://eliterate.atlassian.net/browse/NG23-83

Implements the ability for a user to search their notes and posts in the sidebar. When a search result it clicked, the point block that contains the note is selected and the replies for the particular post are loaded.

Screenshot 2024-04-16 at 5 05 06 PM

This PR also includes some changes specified by the designers to allow a bubble to be deselected, showing all notes for a page regardless of point block annotations, however a block must be selected in order to create a note. Seen in the mockup video below.

https://github.com/Simon-Initiative/oli-torus/assets/6248894/64cb0c9f-c258-404d-af24-7f1cb4aa04fa

It's worth mentioning this PR does not pre-compute an index column for the ts_vector used by postgres to facilitate search. This can be added later if performance becomes a concern or if we deem it to be necessary as part of this work.

### DIFF
--- a/assets/css/app.css
+++ b/assets/css/app.css
@@ -140,3 +140,7 @@ https://tailwindcss.com/docs/utility-first
   -ms-overflow-style: none; /* IE and Edge */
   scrollbar-width: none; /* Firefox */
 }
+
+.search-result em {
+  @apply not-italic bg-yellow-50;
+}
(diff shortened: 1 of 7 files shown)


darrensiegel (approved):
This all looks good, the debounce is key.

PR #4763, opened 2024-04-22
[BUG FIX] [NG23-165] My notes filter not working correctly

eliknebel:
https://eliterate.atlassian.net/browse/NG23-165

Fixes an issue where the "my notes" query would return all notes for all users in a particular section and resource.

"My Notes" are intended to be private. This PR fixes the query so that when visibility is :private, only the notes for the current user are returned.

This PR also fixes an issue where search results within replies for an anonymous user/guest would crash the view.

### DIFF
--- a/lib/oli/resources/collaboration.ex
+++ b/lib/oli/resources/collaboration.ex
@@ -1197,6 +1197,15 @@ defmodule Oli.Resources.Collaboration do
           dynamic([p], p.annotated_block_id == ^point_block_id)
       end
 
+    filter_by_visibility =
+      case visibility do
+        :private ->
+          dynamic([p], p.visibility == ^visibility and p.user_id == ^user_id)
+
+        _ ->
+          dynamic([p], p.visibility == ^visibility)
+      end
+
     Repo.all(
       from(
         post in Post,
@@ -1212,6 +1221,7 @@ defmodule Oli.Resources.Collaboration do
             (post.status in [:approved, :archived] or
                (post.status == :submitted and post.user_id == ^user_id)),
         where: ^filter_by_point_block_id,
+        where: ^filter_by_visibility,
         where:
           fragment(
             "to_tsvector('english', ?) @@ websearch_to_tsquery('english', ?)",
--- a/lib/oli/resources/collaboration.ex
+++ b/lib/oli/resources/collaboration.ex
@@ -1129,6 +1129,15 @@ defmodule Oli.Resources.Collaboration do
           dynamic([p], p.annotated_block_id == ^point_block_id)
       end
 
+    filter_by_visibility =
+      case visibility do
+        :private ->
+          dynamic([p], p.visibility == ^visibility and p.user_id == ^user_id)
+
+        _ ->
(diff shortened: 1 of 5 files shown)


lib/oli/resources/collaboration.ex
```diff
@@ -1306,14 +1315,23 @@ defmodule Oli.Resources.Collaboration do
...
   """
   def list_post_counts_for_user_in_section(section_id, resource_id, user_id, visibility) do
+    filter_by_visibility =
+      case visibility do
+        :private ->
+          dynamic([p], p.visibility == ^visibility and p.user_id == ^user_id)
+
+        _ ->
+          dynamic([p], p.visibility == ^visibility)
+      end
```
nicocirio, line 1325: We should implement this same filter clause in `search_posts_for_user_in_point_block/6` to prevent a student from searching through other students' private notes.

https://github.com/Simon-Initiative/oli-torus/assets/74839302/ef7a28db-83ad-46b2-a089-94d43af4becf

nicocirio (changes requested):
Nice work! After implementing the `filter_by_visibility` in the search function this PR should be ready to go!

NIT: While testing the feature locally I saw that if we search in one tab and switch to the other one, the search does not re-run. 
If we think that feature is nice-to-have, we could make a small modification to the "select_tab" handle event to address that use-case:

```elixir
def handle_event("select_tab", %{"tab" => tab}, socket) do
    tab =
      case tab do
        "my_notes" -> :my_notes
        "all_notes" -> :all_notes
        _ -> :my_notes
      end

    if socket.assigns.annotations.search_term not in [nil, ""] do
      %{
        current_user: current_user,
        section: section,
        page_context: %{
          page: %{resource_id: resource_id}
        },
        annotations: %{
          selected_point: selected_point,
          search_term: search_term
        }
      } = socket.assigns

      async_search_annotations(
        section,
        resource_id,
        current_user,
        visibility_for_active_tab(tab),
        selected_point,
        search_term
      )

      {:noreply,
       socket
       |> assign_annotations(search_results: :loading, active_tab: tab)}
    else
      async_load_annotations(
        socket.assigns.section,
        socket.assigns.page_context.page.resource_id,
        socket.assigns.current_user,
        socket.assigns.course_collab_space_config,
        visibility_for_active_tab(tab),
        socket.assigns.annotations.selected_point
      )

      {:noreply, assign_annotations(socket, active_tab: tab, posts: nil)}
    end
  end
```

### Before

https://github.com/Simon-Initiative/oli-torus/assets/74839302/d05b9c6a-8257-4c13-9d8f-d7d4edacabab


### After

https://github.com/Simon-Initiative/oli-torus/assets/74839302/35fba097-8fc7-4f84-be8c-22f78949b6f2

lib/oli/resources/collaboration.ex
eliknebel: Good catch. Fixed

eliknebel:
Nice suggestions, added in the logic for switching tabs 👍

nicocirio (approved):
🙌 🚀

PR #4765, opened 2024-04-23
[BUG FIX] [MER-3153] Re-implement lost quiz answers second tab

simonchoxx:
[MER-3153](https://eliterate.atlassian.net/browse/MER-3153)

This PR adds checks to see if an activity or part has already been submitted or evaluated at the time of saving. 
A message will be displayed to the student if this happens.

Feel free to suggest any changes to the texts shown to the student in case he/she tries to submit or save an activity or part that has been previously submitted or evaluated, as well as suggest other test cases.

[MER-3153]: https://eliterate.atlassian.net/browse/MER-3153?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli/delivery/attempts/activity_lifecycle.ex
+++ b/lib/oli/delivery/attempts/activity_lifecycle.ex
@@ -20,6 +20,7 @@ defmodule Oli.Delivery.Attempts.ActivityLifecycle do
   alias Oli.Delivery.Evaluation.{Explanation, ExplanationContext}
 
   import Oli.Delivery.Attempts.Core
+  require Logger
 
   @doc """
   Retrieve a hint for an attempt.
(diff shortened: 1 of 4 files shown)


lib/oli/delivery/attempts/activity_lifecycle.ex
```diff
@@ -332,21 +333,31 @@ defmodule Oli.Delivery.Attempts.ActivityLifecycle do
...
-        updated_at = NOW()
-      FROM (
-          VALUES
-          #{part_input_values}
-      ) AS batch_values (attempt_guid, response)
-      WHERE part_attempts.attempt_guid = batch_values.attempt_guid
-    """
-
-    Ecto.Adapters.SQL.query(Oli.Repo, sql, params)
+    part_attempt = get_part_attempt_by(attempt_guid: hd(params))
```
darrensiegel, line 336: The `save_part`  endpoint is a high throughput endpoint, and one that has caused performance problems in the past.  The proposed change here would add another DB call (the read to get the part attempt).  We need a solution that does not add any more DB reads or writes.  Perhaps adding into the `UPDATE` query a constraint `and lifecycle_state = :active` does the trick, handling the case of updated zero records as ""These changes could not be saved as this attempt may have already been submitted"

darrensiegel (changes requested):

darrensiegel (approved):

PR #4774, opened 2024-04-25
[FEATURE] [MER-3137] Skip grade passback jobs when host is changed

Francisco-Castro:
Ticket [MER-3137](https://eliterate.atlassian.net/browse/MER-3137)

This PR implements the logic required to ensure that GradeUpdateWorkers execute tasks solely on the origin host. When run on a different host, the workers simply return `:ok` without performing any work.

[MER-3137]: https://eliterate.atlassian.net/browse/MER-3137?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


test/oli/delivery/attempts/page_lifecycle/grade_update_worker_test.exs
```diff
@@ -0,0 +1,57 @@
...
+          section: section,
+          last_grade_update_id: last_grade_update_id
+        )
+
+      {:ok, updated_resource_access} =
+        perform_job(GradeUpdateWorker, %{
+          section_id: section.id,
+          resource_access_id: resource_access_id,
+          type: :manual_batch
+        })
```
nicocirio, line 27: NIT: we could add the host here

eliknebel:
I think the approach here is fine, we just need to make sure we are handling the other host-dependent jobs such as xapi upload and email sending

test/oli/delivery/attempts/page_lifecycle/grade_update_worker_test.exs
Francisco-Castro: I'm going to pass on this because we are not using the host inside the body. Please let me know if I'm missing something.

nicocirio: The `host` is not being pattern-matched in the args of the `perform` function, but that does not imply that the Oban job will not have a host (since we modified the `create` function to add one)
So my idea was to explicitly pass to the `perform` function an Oban job with a host that matches the host of the test env, and assert that the Oban job is executed as expected. In this way, the test will end up mimicking better what will end up happening (receiving an Oban job with a host that matches the current one and executing the job)

Francisco-Castro: I see your point. It looks more explicit then. This PR has been paused because it looks like we are going to use another approach. But really appreciate your feedback in here. 👍

priv/repo/seeds.exs
```diff
@@ -18,6 +18,14 @@ alias Oli.Accounts
 alias Oli.Accounts.{User, Author}
 alias Oli.Repo
 alias Oli.Utils.DataGenerators.NameGenerator
+alias Oli.Host.HostIdentifier
+import Ecto.Query
+
+unless Repo.exists?(HostIdentifier) do
```
eliknebel, line 24: I think this is going in the right direction, but we still need a check here to actually read the value and compare with `System.get_env("HOST")` to determine whether to drop all oban jobs or not.

This implementation as it stands will only drop the oban jobs if the record in the DB doesn't exist. It may (and will) exist in DB clones and therefore this will not execute when it needs to

eliknebel (changes requested):

priv/repo/seeds.exs
Francisco-Castro: Thanks for pointing that out. I just sent a [commit](https://github.com/Simon-Initiative/oli-torus/pull/4774/commits/45a03eb071e012a764333ed903cad089c1bc7a56) to handle it.

eliknebel (approved):

PR #4775, opened 2024-04-25
[BUG FIX] [MER-2796] LTI integration requests no longer trigger Slack notification

rgachuhi:
This PR fixes an issue with the message format sent to the Slack hook for notification. The issue was due to an error in URL formatting for one of the Slack action blocks.


eliknebel (approved):
Tested, appears to have fixed the issue

PR #4777, opened 2024-04-29
[BUG FIX] [MER-2932] filter-insights-by-product-course-section

tomasferok:
This pull request ([MER-2932](https://eliterate.atlassian.net/browse/MER-2932)) addresses two main issues:

It corrects the filters on the tab-by-page feature.
It enhances the remix functionality by ensuring that when a section is removed, any associated activities are also deleted along with the revision.
this video show the previous behavior of remix course

https://github.com/Simon-Initiative/oli-torus/assets/80487824/a76ddd04-9f2a-4a99-a29d-f4ec75b8bb80

and this the behavior with this bug fix 

https://github.com/Simon-Initiative/oli-torus/assets/80487824/98daac31-a592-4392-9ac5-2367472a22f1

the number of the activity corresponds to the number of the assessment (page) that are contained

[MER-2932]: https://eliterate.atlassian.net/browse/MER-2932?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/oli/analytics/by_page.ex
```diff
@@ -20,9 +20,11 @@ defmodule Oli.Analytics.ByPage do
   end
 
   defp get_base_query(project_slug, activity_pages, filtered_sections) do
+    Repo.all(activity_pages)
```
gastonabella, line 23: I believe you forgot to remove this line since it's performing a query that is never used nor stored in a variable.

lib/oli/analytics/by_page.ex
```diff
@@ -33,11 +35,13 @@ defmodule Oli.Analytics.ByPage do
         )
       end
 
+    Repo.all(subquery)
```
gastonabella, line 38: Same here

lib/oli/delivery/sections.ex
```diff
@@ -2586,6 +2590,61 @@ defmodule Oli.Delivery.Sections do
...
   end
 
+  defp build_list_children_activity_ids([]), do: []
+
+  defp build_list_children_activity_ids(children_activities) do
+    tuple_of_children_activities =
+      children_activities
+      |> process_activity_ids()
+
+    Enum.concat(Tuple.to_list(tuple_of_children_activities))
```
gastonabella, line 2600: Nit: good chance to use the pipe operator here :)

lib/oli/delivery/sections.ex
```diff
@@ -2586,6 +2590,61 @@ defmodule Oli.Delivery.Sections do
     end)
   end
 
+  defp build_list_children_activity_ids([]), do: []
+
+  defp build_list_children_activity_ids(children_activities) do
+    tuple_of_children_activities =
+      children_activities
+      |> process_activity_ids()
```
gastonabella, line 2598: Contrary to the example below, here we could avoid the pipe since we're just chaining one argument.

gastonabella (commented):
Great work!

lib/oli/analytics/by_page.ex
tomasferok: I remove the IO.inspect but I forgot the repo all 😅

lib/oli/delivery/sections.ex
tomasferok: nice suggestion

lib/oli/delivery/sections.ex
```diff
@@ -2994,6 +3056,8 @@ defmodule Oli.Delivery.Sections do
         {:ok}
       end)
 
+    IO.inspect(result, label: "diferencesss")
```
eliknebel, line 3059: Remove IO.inspect

eliknebel (changes requested):

lib/oli/delivery/sections.ex
```diff
@@ -3378,11 +3442,13 @@ defmodule Oli.Delivery.Sections do
...
          skip_resource_ids: skip_resource_ids,
-         required_survey_resource_id: required_survey_resource_id
+         required_survey_resource_id: required_survey_resource_id,
+         list_children_activity_ids: list_children_activity_ids
        ) do
     published_resources_by_resource_id =
-      MinimalHierarchy.published_resources_map(publication_ids)
+      build_published_resources_by_resource_id(publication_ids, list_children_activity_ids)
 
+    IO.inspect(published_resources_by_resource_id, label: "wetree")
```
eliknebel, line 3451: remove IO.inspect

lib/oli/delivery/sections.ex
```diff
@@ -2586,6 +2592,62 @@ defmodule Oli.Delivery.Sections do
     end)
   end
```
eliknebel, line 2594: This is a really critical part of the system. I'm not exactly sure what the intention of these
changes are, at the very least we need to see some unit tests around this change that clearly
demonstrate what it is trying to accomplish. These changes are also not listed in the original
ticket, so I think we need a better explaination in the PR/ticket of why these are being made

eliknebel (changes requested):
I believe I understand the issue identified and fixed here, but again because the sections module is such a critical part of the system, I think we need to do the following before we can land this:
1. Update the PR description or create a separate ticket that clearly explains here why this change is being made, since this issue is not described in the attached ticket or PR description. I.e., Why exactly are we addressing this as part of these changes? What is the main negative consequence of leaving these records around, if any. It seems to make sense to clean these up, but is there another reason why this is being done here as part of this PR? An acceptable explanation might be: "these records were affecting the insights query" but I want to make sure the changes here are actually intended since this is a crucial part of the system. I'm trying to understand why this function needed to be changed here.
2. ~~We need to codify this as a unit test (as demonstrated in the "After" video). The test should exercise the `create_section_resources` function by removing an existing graded page that contains activities and then query the database, similar to show in the video, to ensure the contained activities are removed as expected~~

Edit: Apologies, I missed the last commit. It seems there is a unit test which does exercise this function. So we just need to explain in the description how this change relates to the referenced ticket

lib/oli/delivery/sections.ex
tomasferok: This code was developed to remove the section resources associated with a deleted page. I've also incorporated a test to validate this action.

The function "build_reference_activity_ids" generates a list of IDs based on the hierarchy. Subsequently, within the function "build_published_resources_by_resource_id" (which is nested inside "create_nonstructural_section_resources"), it retrieves the publications related to this list of IDs. The condition "if length(processed_resource_ids) != 1" checks if there is more than one processed resource ID (implying that it's not just the curriculum resource itself), and if so, it proceeds to create the section resources.

eliknebel (approved):

PR #4788, opened 2024-05-06
[FEATURE] [MER-3096] User enables AB testing with Upgrade

Francisco-Castro:
Ticket: [MER-3096](https://eliterate.atlassian.net/browse/MER-3096)

This PR introduces the capability to enable/disable A/B testing. When enabled for the first time, a revision is created. Subsequent actions on the checkbox only toggle the ability to edit the content of the revision.

Endpoint: 
`http://localhost/authoring/project/__SOME_PROJECT__/experiments`

https://github.com/Simon-Initiative/oli-torus/assets/47334502/575ac111-d7d1-476c-829a-48bbb001e3cf

Note: A significant portion of code was copied and adapted for the OliWeb.Resources.AlternativesEditor module. I recognize that some refactoring could reduce complexity. However, I refrained from doing so as this PR is currently blocking others.

[MER-3096]: https://eliterate.atlassian.net/browse/MER-3096?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli/authoring/editing/resource_editor.ex
+++ b/lib/oli/authoring/editing/resource_editor.ex
@@ -7,9 +7,29 @@ defmodule Oli.Authoring.Editing.ResourceEditor do
   import Ecto.Query, warn: false
   import Oli.Authoring.Editing.Utils
 
+  alias Oli.Resources.ResourceType
+  alias Oli.Authoring.Course.Project
+  alias Oli.Authoring.Course.ProjectResource
   alias Oli.Resources.Revision
   alias Oli.Authoring.Course
   alias Oli.Publishing.AuthoringResolver
+
+  @alternatives_type_id ResourceType.id_for_alternatives()
+  @experiment_id "upgrade_decision_point"
+
+  def get_experiment(project_slug) do
+    from(pr in ProjectResource,
+      join: p in Project,
+      on: p.id == pr.project_id,
+      join: r in Revision,
+      on: r.resource_id == pr.resource_id,
+      where: p.slug == ^project_slug,
+      where: r.resource_type_id == @alternatives_type_id,
+      where: fragment("?->>'strategy' = ?", r.content, @experiment_id),
+      select: r
+    )
+    |> Oli.Repo.one()
+  end
 
   @doc """
   Retrieves a list of all (non-deleted) resources of the specified type for a given project.
--- a/lib/oli/authoring/editing/resource_editor.ex
+++ b/lib/oli/authoring/editing/resource_editor.ex
@@ -7,31 +7,9 @@ defmodule Oli.Authoring.Editing.ResourceEditor do
   import Ecto.Query, warn: false
   import Oli.Authoring.Editing.Utils
 
-  alias Oli.Resources.ResourceType
-  alias Oli.Authoring.Course.Project
-  alias Oli.Authoring.Course.ProjectResource
-  alias Oli.Resources.Revision
   alias Oli.Authoring.Course
   alias Oli.Publishing.AuthoringResolver
-
-  @alternatives_type_id ResourceType.id_for_alternatives()
-  @experiment_id "upgrade_decision_point"
-
-  def get_latest_experiment(project_slug) do
-    from(pr in ProjectResource,
-      join: p in Project,
-      on: p.id == pr.project_id,
-      join: r in Revision,
-      on: r.resource_id == pr.resource_id,
-      where: p.slug == ^project_slug,
-      where: r.resource_type_id == @alternatives_type_id,
-      where: fragment("?->>'strategy' = ?", r.content, @experiment_id),
-      select: r,
-      limit: 1,
-      order_by: [desc: r.inserted_at]
-    )
-    |> Oli.Repo.one()
-  end
+  alias Oli.Resources.Revision
 
   @doc """
   Retrieves a list of all (non-deleted) resources of the specified type for a given project.
--- a/lib/oli/authoring/editing/resource_editor.ex
+++ b/lib/oli/authoring/editing/resource_editor.ex
@@ -17,7 +17,7 @@ defmodule Oli.Authoring.Editing.ResourceEditor do
   @alternatives_type_id ResourceType.id_for_alternatives()
   @experiment_id "upgrade_decision_point"
 
-  def get_experiment(project_slug) do
+  def get_latest_experiment(project_slug) do
     from(pr in ProjectResource,
       join: p in Project,
       on: p.id == pr.project_id,
@@ -26,7 +26,9 @@ defmodule Oli.Authoring.Editing.ResourceEditor do
       where: p.slug == ^project_slug,
       where: r.resource_type_id == @alternatives_type_id,
       where: fragment("?->>'strategy' = ?", r.content, @experiment_id),
-      select: r
+      select: r,
+      limit: 1,
+      order_by: [desc: r.inserted_at]
     )
     |> Oli.Repo.one()
   end
--- a/lib/oli_web/live/experiments/experiments_live.ex
+++ b/lib/oli_web/live/experiments/experiments_live.ex
@@ -14,6 +14,9 @@ defmodule OliWeb.Experiments.ExperimentsView do
   alias OliWeb.Common.Modal.DeleteModal
   alias OliWeb.Common.Modal.FormModal
 
+  on_mount {OliWeb.LiveSessionPlugs.SetUser, :default}
+  on_mount {OliWeb.LiveSessionPlugs.SetProject, :default}
+
   @title "Experiments"
   @alternatives_type_id ResourceType.id_for_alternatives()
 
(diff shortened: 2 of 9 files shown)


lib/oli_web/live/experiments/experiments_live.ex
```diff
@@ -1,14 +1,35 @@
...
+  @title "Experiments"
+  @alternatives_type_id ResourceType.id_for_alternatives()
 
   def mount(_params, _session, socket) do
-    {:ok, assign(socket, title: "Experiments")}
+    experiment = ResourceEditor.get_latest_experiment(socket.assigns.project.slug)
+
+    socket = assign(socket, is_upgrade_enabled: false)
+    socket = assign(socket, title: @title)
+    socket = assign(socket, experiment: experiment)
```
nicocirio, line 25: What do you think about piping this?

lib/oli_web/router.ex
```diff
@@ -417,7 +417,15 @@ defmodule OliWeb.Router do
...
     # Experiment management
-    live("/:project_id/experiments", Experiments.ExperimentsView)
+
+    live_session :another_load_projects,
+      on_mount: [
+        OliWeb.LiveSessionPlugs.SetUser,
+        OliWeb.LiveSessionPlugs.SetProject
+      ] do
+      live("/:project_id/experiments", Experiments.ExperimentsView)
+    end
```
nicocirio, line 427: Do we plan to expand other routes in this `live_session`? If not, perhaps we could define the `on_mount` within the Experiments liveview.
If we are going to expand it, what do you think about another name for the live session rather than "another_load_projects"? Maybe just "experiments"?

lib/oli_web/live/experiments/experiments_live.ex
```diff
@@ -26,8 +47,423 @@ defmodule OliWeb.Experiments.ExperimentsView do
...
+
+    {:noreply, show_modal(socket, modal, modal_assigns: modal_assigns)}
+  end
+
+  def handle_event(
+        "create_option",
+        %{"params" => %{"id" => option_id, "name" => name, "resource_id" => resource_id}},
+        socket
+      ) do
+    ###
```
nicocirio, line 132: NIT: remove this if we are not going to comment

lib/oli_web/live/experiments/experiments_live.ex
```diff
@@ -26,8 +47,423 @@ defmodule OliWeb.Experiments.ExperimentsView do
...
+    resource_id = ensure_integer(resource_id)
+
+    %{content: %{"options" => options} = content} = socket.assigns.experiment
+
+    new_options = [%{"id" => option_id, "name" => name} | options]
+
+    %{project: project, ctx: ctx} = socket.assigns
```
nicocirio, line 139: what do you think about getting the `options`, the `project` and the `ctx` in one single pattern matching with `socket.assigns`?

lib/oli_web/live/experiments/experiments_live.ex
```diff
@@ -26,8 +47,423 @@ defmodule OliWeb.Experiments.ExperimentsView do
...
+
+  def handle_event(
+        "delete_option",
+        %{"resource-id" => resource_id, "option-id" => option_id},
+        socket
+      ) do
+    resource_id = ensure_integer(resource_id)
+    %{project: project, ctx: ctx} = socket.assigns
+
+    %{content: %{"options" => options} = content} = experiment = socket.assigns.experiment
```
nicocirio, line 209: same here

lib/oli_web/live/experiments/experiments_live.ex
```diff
@@ -26,8 +47,423 @@ defmodule OliWeb.Experiments.ExperimentsView do
...
+
+  def handle_event(
+        "edit_group",
+        %{"params" => %{"resource_id" => resource_id, "title" => title}},
+        socket
+      ) do
+    %{project: project, ctx: ctx} = socket.assigns
+    resource_id = ensure_integer(resource_id)
+
+    experiment = socket.assigns.experiment
```
nicocirio, line 274: and here

lib/oli_web/live/experiments/experiments_live.ex
Francisco-Castro: I changed for this format, otherwise the extraction is going to look odd.


```
%{project: project, ctx: ctx, experiment: experiment} = socket.assigns
%{content: %{"options" => options} = content} = experiment
new_options = [%{"id" => option_id, "name" => name} | options]
```

lib/oli_web/live/experiments/experiments_live.ex
Francisco-Castro: I did something similar:

```
%{project: project, ctx: ctx, experiment: experiment} = socket.assigns
%{content: %{"options" => options} = content} = experiment
```

lib/oli_web/live/experiments/experiments_live.ex
Francisco-Castro: Changed for

```
%{project: project, ctx: ctx, experiment: experiment} = socket.assigns
```

lib/oli_web/router.ex
Francisco-Castro: Nice suggestion! I learned something today! Thank!

nicocirio (commented):
Thanks Fran, nice work! We tested that the user can "check, then uncheck when no previous experiment exists", and reading the ticket I thought there were other cases yet to test. For instance:
"no new alternative resource is created if it already exists"
"can edit alternative title"
"can add a new option"
"can edit an existing option"
"can delete an option"

(Perhaps the latter ones could be added to "OliWeb.Resources.AlternativesEditorTest")

Francisco-Castro:
Good idea! I'll add those tests then.

darrensiegel (dismissed):

test/oli_web/live/experiments_live_test.exs
```diff
@@ -91,133 +88,220 @@ defmodule OliWeb.ExperimentsLiveTest do
...
+    test "when checkbox is off then disable the ability to modify experiment", %{view: view} do
+      {view, %{}}
+      |> step(:check_on_checkbox)
+      |> step(:check_on_checkbox)
+      |> step(:test_has_button_show_edit_group_modal, :refute)
+      |> step(:test_has_button_show_edit_option_1_modal, :refute)
+      |> step(:test_has_button_show_edit_option_2_modal, :refute)
+      |> step(:test_has_button_show_delete_option_1_modal, :refute)
+      |> step(:test_has_button_show_delete_option_2_modal, :refute)
+      |> step(:test_has_new_option_link, :refute)
```
nicocirio, line 131: If I understood correctly this way of piping the test with steps, I think it would be impossible to find all those elements in the view since the assertion depends on having the options and resource ids in the context map. I mean, we are trying to find some elements in the view with a selector that depends on those ids (phx-value-option-id and phx-value-resource-id)

nicocirio: I think we could get the needed resource and options ids by fetching the experiment in the db, or maybe change the way we define those element selectors to not depend on their ids.

test/oli_web/live/experiments_live_test.exs
```diff
@@ -94,39 +88,220 @@ defmodule OliWeb.ExperimentsLiveTest do
...
-             )
+      {view, %{}}
+      |> step(:test_has_alternatives_group)
+      |> step(:test_has_options)
+      |> step(:test_has_button_show_edit_group_modal, :refute)
+      |> step(:test_has_button_show_edit_option_1_modal, :refute)
+      |> step(:test_has_button_show_edit_option_2_modal, :refute)
+      |> step(:test_has_button_show_delete_option_1_modal, :refute)
+      |> step(:test_has_button_show_delete_option_2_modal, :refute)
+      |> step(:test_has_new_option_link, :refute)
```
nicocirio, line 153: same here

test/oli_web/live/experiments_live_test.exs
Francisco-Castro: It makes sense. We need the `resource_id` to get or not a match. Let me see how can I adapt it.

Francisco-Castro: @nicocirio It was relatively easy to fix using the piping approach... Check out this [commit](https://github.com/Simon-Initiative/oli-torus/pull/4788/commits/ee230ea1457283adab272158341b37f3b7d55dcb). I'm starting to like it.

test/oli_web/live/experiments_live_test.exs
Francisco-Castro: Done.

lib/oli/authoring/editing/resource_editor.ex
```diff
@@ -7,10 +7,32 @@ defmodule Oli.Authoring.Editing.ResourceEditor do
...
+  alias Oli.Authoring.Course.ProjectResource
   alias Oli.Resources.Revision
   alias Oli.Authoring.Course
   alias Oli.Publishing.AuthoringResolver
 
+  @alternatives_type_id ResourceType.id_for_alternatives()
+  @experiment_id "upgrade_decision_point"
+
+  def get_latest_experiment(project_slug) do
+    from(pr in ProjectResource,
```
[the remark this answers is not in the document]
Francisco-Castro: It makes sense. I pushed some commit to handle both comments.

darrensiegel (approved):

PR #4796, opened 2024-05-08
[FEATURE] [NG23-89] add a link to go to a content page from a post

eliknebel:
Adds a "Go to Page" link to notes that links to the page where a note was created. Also adds a virtual field to facilitate link creation and updates the queries used to populate the new field.

This feature adds a bit of overhead to the simplified queries to gather the necessary slug in order to facilitate link creation. Now these queries must also join through the section resource, spp, published resource to revision records. This query however, is similar to the queries that were already in place before the social annotation work.

Screenshot 2024-05-08 at 2 31 03 PM

### DIFF
--- a/lib/oli/resources/collaboration.ex
+++ b/lib/oli/resources/collaboration.ex
@@ -633,6 +633,14 @@ defmodule Oli.Resources.Collaboration do
     results =
       from(
         post in Post,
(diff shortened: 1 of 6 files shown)


darrensiegel (approved):
Tested, works.

PR #4808, opened 2024-05-14
[ENHANCEMENT] [MER-3179] Set navigation menu bar at the top of the page on vertical scroll

Francisco-Castro:
Ticket [MER-3179](https://eliterate.atlassian.net/browse/MER-3179).

This PR put the navigation menu bar at the top of the page even when scrolling down.

### New looking

https://github.com/Simon-Initiative/oli-torus/assets/47334502/ccf5832d-a548-4922-80b2-c9773160f0e8

[MER-3179]: https://eliterate.atlassian.net/browse/MER-3179?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli_web/components/delivery/instructor_dashboard.ex
+++ b/lib/oli_web/components/delivery/instructor_dashboard.ex
@@ -169,7 +169,7 @@ defmodule OliWeb.Components.Delivery.InstructorDashboard do
 
   def header(assigns) do
     ~H"""
-    <div class="w-full bg-delivery-instructor-dashboard-header text-white border-b border-slate-600">
+    <div class="w-full bg-delivery-instructor-dashboard-header text-white border-b border-slate-600 sticky top-0 z-50">
       <div class="container mx-auto flex flex-row">
         <div class="flex items-center">
           <a
--- a/lib/oli_web/components/delivery/instructor_dashboard.ex
+++ b/lib/oli_web/components/delivery/instructor_dashboard.ex
@@ -169,7 +169,7 @@ defmodule OliWeb.Components.Delivery.InstructorDashboard do
 
   def header(assigns) do
     ~H"""
-    <div class="w-full bg-delivery-instructor-dashboard-header text-white border-b border-slate-600 sticky top-0 z-50">
+    <div class="w-full bg-delivery-instructor-dashboard-header text-white border-b border-slate-600 fixed top-0 z-50">
       <div class="container mx-auto flex flex-row">
         <div class="flex items-center">
           <a
--- a/lib/oli_web/components/delivery/instructor_dashboard.ex
+++ b/lib/oli_web/components/delivery/instructor_dashboard.ex
@@ -169,7 +169,7 @@ defmodule OliWeb.Components.Delivery.InstructorDashboard do
 
   def header(assigns) do
     ~H"""
-    <div class="w-full bg-delivery-instructor-dashboard-header text-white border-b border-slate-600 fixed top-0 z-50">
+    <div class="w-full bg-delivery-instructor-dashboard-header text-white border-b border-slate-600 sticky top-0 z-50">
       <div class="container mx-auto flex flex-row">
         <div class="flex items-center">
           <a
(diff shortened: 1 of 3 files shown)


nicocirio:
Nice work! I would suggest considering flash messages, since now the sticky header will overlap them. I am not a CSS expert, but I think using "fixed" combined with z-index instead of "sticky" might solve that overlapping issue.

https://github.com/Simon-Initiative/oli-torus/assets/74839302/1e5c2ea1-baeb-4ac6-abc9-31e5e3bd107e

Francisco-Castro:
Nice catch @nicocirio! I sent a [commit](https://github.com/Simon-Initiative/oli-torus/pull/4808/commits/62ef98899882cd9353890a11cfab76d4ea196d4b) handling the issue with the flash message. Please feel free to take a look at it at your convenience.

lib/oli_web/components/delivery/instructor_dashboard.ex
```diff
@@ -169,7 +169,7 @@ defmodule OliWeb.Components.Delivery.InstructorDashboard do
 
   def header(assigns) do
     ~H"""
-    <div class="w-full bg-delivery-instructor-dashboard-header text-white border-b border-slate-600">
+    <div class="w-full bg-delivery-instructor-dashboard-header text-white border-b border-slate-600 fixed top-0 z-50">
```
nicocirio, line 172: If we take the `fixed` class approach we should add a margin-top to `instructor_dashboard.html.heex` to avoid having the course title being overlapped by the header.

https://github.com/Simon-Initiative/oli-torus/assets/74839302/07935833-e161-482e-ba08-c6ec87a30f20



With this approach, we get an animation when scrolling all the way up to the top, where a white background is seen by a fraction of a sec. (maybe [overscroll-behaviour ](https://developer.mozilla.org/en-US/docs/Web/CSS/overscroll-behavior) can fix it, not sure TBH).


Another approach may be using the `sticky` class, ensuring adapting the top position of the flash message to avoid being overlapped by the header (you've already adjusted that top position).

Francisco-Castro: I have made another iteration. This time, the only minor issue is that when reloading the page, the Cookie Preferences briefly appear at the top for a fraction of a second before disappearing. What do you think? Is that or the white space as you mentioned earlier in between the menu bar and the content.

nicocirio: No, the white space I was refering to is the one at second 23 in the attached video. Here is a screenshot

image

I pulled the changes and saw the issue with the cookie preferences. It looks really weird. I think we should fix that for sure

lib/oli_web/components/layouts/instructor_dashboard.html.heex
```diff
@@ -1,4 +1,4 @@
-<div class="flex-1 flex flex-col h-screen">
+<div class="flex-1 flex flex-col h-auto">
```
nicocirio, line 1: `h-auto` seems to be the reason why we see the cookie preferences for a second. I think it may have to do with liveview lifecycle, when the "dead" view gets rendered the height of the screen is shorter, since the other content has not yet been rendered. 
Why do we need this change? reverting it seems to fix the "cookie preferences" issue

Francisco-Castro: That's the point. We could go with `h-auto` that is going to bring the Cookie Preferences issue, or what you mentioned:

> With this approach, we get an animation when scrolling all the way up to the top, where a white background is seen by a fraction of a sec. (maybe [overscroll-behaviour ](https://developer.mozilla.org/en-US/docs/Web/CSS/overscroll-behavior)can fix it, not sure TBH).

Either way, something looks weird.

Francisco-Castro: I [tried](https://github.com/Simon-Initiative/oli-torus/pull/4808/commits/aec24579f2b22d229ac03b43e1755aa80cc8d5a9) using `min-h-screen`, and it did the trick.

darrensiegel (approved):
Nice work!

PR #4820, opened 2024-05-17
[FEATURE] [MER-3219] Allow page progress trigger practice gating

simonchoxx:
[MER-3219](https://eliterate.atlassian.net/browse/MER-3219)

This RP adds a new type of `Gating Condition` so that it can be defined in a product.  

It adds a `Progress` type that will evaluate the progress of a source resource of type `Practice Page`.

This will evaluate the progress that is stored in ResourceAccess for the source resource to determine if the learner has access to a target resource defined in the Gating Condition.

https://github.com/Simon-Initiative/oli-torus/assets/16328384/40476b49-f340-437e-81ed-17e1068fd562

Captura de pantalla 2024-05-17 a la(s) 10 08 45 a  m

[MER-3219]: https://eliterate.atlassian.net/browse/MER-3219?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


darrensiegel (approved):
Very nice work!

PR #4827, opened 2024-05-21
[FEATURE] [MER-3217] Instructors back button assessments view

simonchoxx:
[MER-3217](https://eliterate.atlassian.net/browse/MER-3217)

This PR modifies the button to return to the list of assessments, which is displayed in the activity detail view, for the `Scored Activities` view.

- Before

Captura de pantalla 2024-05-21 a la(s) 10 49 16 a  m

- After

Captura de pantalla 2024-05-21 a la(s) 10 46 57 a  m

[MER-3217]: https://eliterate.atlassian.net/browse/MER-3217?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/oli_web/components/delivery/scored_activities/scored_activities.ex
```diff
@@ -135,6 +136,18 @@ defmodule OliWeb.Components.Delivery.ScoredActivities do
   def render(assigns) do
     ~H"""
     <div>
+      <%= unless is_nil(@current_assessment) do %>
+        <button class="whitespace-nowrap" phx-click="back" phx-target={@myself}>
```
nicocirio, line 140: What do you think about  ```<button :if={!is_nil(@current_assessment)} ...>...</button>``` ?

To be honest, I would love if we could just do 

```<button :unless={is_nil(@current_assessment)} ...>...</button>```

but I think that [HEEX special attribute](https://hexdocs.pm/phoenix_live_view/Phoenix.Component.html#sigil_H/2-heex-extension-special-attributes) is not available yet

**Review by nicocirio — COMMENTED** [review]
Nice work Santi. I left a small NIT

File: lib/oli_web/components/delivery/scored_activities/scored_activities.ex (hunk 1 above)
Comment (simonchoxx) [reply lib/oli_web/components/delivery/scored_activities/scored_activities.ex:140]: I like your proposal. I often forget to use the if statement inside the tag!

Comment (Francisco-Castro) [reply lib/oli_web/components/delivery/scored_activities/scored_activities.ex]: Yeah.. the `unless` definitely sounds appealing. @nicocirio you could propose that in the liveview repo.

I'm wondering if we could use something like

```
<button :if{@current_assessment not in [nil]}
```

That way my  brain doesn't need to flip the value when I reach the `!`

Also this other way

```
<button :if{@current_assessment != nil}
```

but I like the `not in` because it reads as a sentence.

darrensiegel (approved):

PR #4828, opened 2024-05-22
[BUG FIX] [MER-3146] Advanced Author and Simple Author/Flowchart: Keyboard Shortcuts Request on Canvas

dtiwarATS:
Hey @bsparks please look at the PR. Thanks

While creating the original PR - https://github.com/Simon-Initiative/oli-torus/pull/4810 , while making the suggested changes, I didn't count for the Mac OS hence adding the fix for that as well as fixing the https://eliterate.atlassian.net/browse/MER-3266.

### DIFF
--- a/assets/src/components/activities/adaptive/components/authoring/LayoutEditor.tsx
+++ b/assets/src/components/activities/adaptive/components/authoring/LayoutEditor.tsx
@@ -404,7 +404,7 @@ const LayoutEditor: React.FC<LayoutEditorProps> = (props) => {
     {},
     [selectedPartAndCapabilities, configurePartId],
   );
-  useKeyDown(handleCopyComponent, ['KeyC'], { ctrlKey: true, metaKey: true }, [
+  useKeyDown(handleCopyComponent, ['KeyC'], { ctrlKey: true }, [
     selectedPartAndCapabilities,
     parts,
     handleCopyComponent,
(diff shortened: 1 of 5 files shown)


assets/src/components/activities/adaptive/components/authoring/LayoutEditor.tsx
```diff
@@ -394,8 +394,17 @@ const LayoutEditor: React.FC<LayoutEditorProps> = (props) => {
     [dragSize, isDragging, selectedPartId],
   );
 
-  useKeyDown(() => setShowConfirmDelete(true), ['Delete', 'Backspace'], {}, []);
-  useKeyDown(handleCopyComponent, ['KeyC'], { ctrlKey: true }, [
+  useKeyDown(
+    () => {
+      if (selectedPartAndCapabilities && !configurePartId.length) {
```
dtiwarATS, line 399: This is to make sure that the delete only gets triggered when a part is selected, and it is not in Edit mode.

assets/src/hooks/useKeyDown.tsx
```diff
@@ -11,7 +11,7 @@ export function useKeyDown(
 
     const ctrlMatch = ctrlRequired === undefined || ctrlKey === ctrlRequired;
     const metaMatch = metaRequired === undefined || metaKey === metaRequired;
-    if (ctrlMatch && metaMatch && keyCodes.includes(code)) {
+    if ((ctrlMatch || metaMatch) && keyCodes.includes(code)) {
```
bsparks, line 14: I don't think this makes sense to do. It shouldn't be an OR and I think that the ctrlKey should map to Command or something on a Mac  already

dtiwarATS: Done.

bsparks (approved):

darrensiegel (approved):

PR #4847, opened 2024-05-30
[FEATURE] [NG23-140] Learning progress views

nicocirio:
[Link](https://eliterate.atlassian.net/browse/NG23-140) to the ticket

## Disable scrolling when module is expanded

https://github.com/Simon-Initiative/oli-torus/assets/74839302/8bbe609d-9334-4a40-aed5-50676a3f1b83

## Collapse module with bottom bar

https://github.com/Simon-Initiative/oli-torus/assets/74839302/6b0bfd47-de91-4c79-be86-e74590953a09

## ~Auto scroll to first unfinished resource (at level 1 of hierarchy)~ => This feature was finally removed (see [comment](https://github.com/Simon-Initiative/oli-torus/pull/4847#issuecomment-2146396265))

https://github.com/Simon-Initiative/oli-torus/assets/74839302/e742717e-93bd-43da-9fe5-2b0de4163534


lib/oli_web/live/delivery/student/learn_live.ex
```diff
@@ -133,6 +137,42 @@ defmodule OliWeb.Delivery.Student.LearnLive do
...
+  After scrolling, we set the scroll_to_first_unfinished_level_1_resource flag to false
+  to prevent auto-scrolling from being re-triggered with every user interaction.
+  """
+
+  defp maybe_scroll_to_first_unfinished_level_1_resource(
+         %{assigns: %{scroll_to_first_unfinished_level_1_resource: false}} = socket
+       ),
+       do: socket
+
+  defp maybe_scroll_to_first_unfinished_level_1_resource(socket) do
```
darrensiegel, line 153: I'm concerned about the complexity of the underlying query here, given that we are running it within a Phoenix LiveView event handler.  Instead of only asychronously initiating the scroll, we may want to asychronously execute the query and do the scroll so that we do not introduce notieable delays within the UI.

darrensiegel (changes requested):

darrensiegel:
I'm concerned not only about run-time cost of the implementation of "Scroll to first not-complete", I'm also worried that this is a quite intrusive bit of UX.   There is no way to bypass it, it will always scroll and pulse, everytime a user visits the Learn page.   

This design doesn't consider that some courses may simply have "optional" material that students do not need to work through. It also doesn't take into account gated material.  So, there may literally be need - or no way - for a student to make progress in certain portions of a course, yet this impl will always scroll to that content.

darrensiegel:
We are going to remove the "scroll to first 'not completed' level 1 resource" implementation.

assets/src/hooks/scroller.ts
```diff
@@ -102,6 +103,16 @@ export const Scroller = {
       }
     });
 
+    window.addEventListener('phx:pulse-target', (e: Event) => {
+      console.log('estoy en el handler correcto');
```
eliknebel, line 107: Remove console log statements

assets/src/hooks/scroller.ts
```diff
@@ -102,6 +103,16 @@ export const Scroller = {
...
+      const target = document.getElementById((e as CustomEvent).detail.target_id);
+
+      console.log((e as CustomEvent).detail.target_id);
```
eliknebel, line 110: Remove console log statements

eliknebel (changes requested):

darrensiegel (approved):

PR #4849, opened 2024-05-31
[FEATURE] [MER-3104] Page-level: A/B Test Button Creation

Francisco-Castro:
Ticket [MER-3104](https://eliterate.atlassian.net/browse/MER-3104).

This PR adds a feature to enable or disable the AB Testing button at a page level. The state of the experiment is retrieved from the context.

To turn on/off the AB button, we use the enable/disable checkbox in the experiments page. However, since the context is at a page level, a refresh is required to update the context state.

https://github.com/Simon-Initiative/oli-torus/assets/47334502/f4bd77a2-d0eb-4cac-9cf8-6ebd7ca05b9b

[MER-3104]: https://eliterate.atlassian.net/browse/MER-3104?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/oli_web/router.ex
```diff
@@ -513,6 +513,7 @@ defmodule OliWeb.Router do
     delete("/:project/lock/:resource", Api.LockController, :release)
 
     get("/:project/alternatives", Api.ResourceController, :alternatives)
+    get("/:project/is_experiment_enabled", Api.ResourceController, :is_experiment_enabled)
```
darrensiegel, line 516: We do not need this, given that `has_experiments` lives on the project, and can be passed down via context to the page editor

darrensiegel (changes requested):
This can be a lot simpler.  There is no need to create a new "experiments" schema, no need to create a API to allow fetching of the state of experiments, and no need for the "A/B Test" button to react to changes from another view. 

The "Projects" schema already has a `has_experiments` flag on it.  That flag should be toggled when the toggles the "Enable A/B Testing" checkbox from the "Experiments" view.   Then, when the page editor is launched, the value of that flag can be passed down to the app, obviating the need to make an API call back to the server.

lib/oli_web/router.ex
Francisco-Castro: Thanks for pointing the field `has_experiments` on the `project` model. I didn't know how to persist the state so I created a table for that. I'll restructure that part to use it then.

lib/oli_web/controllers/resource_controller.ex
```diff
@@ -22,8 +22,9 @@ defmodule OliWeb.ResourceController do
     author = conn.assigns[:current_author]
     is_admin? = Accounts.at_least_content_admin?(author)
 
-    case PageEditor.create_context(project_slug, revision_slug, conn.assigns[:current_author]) do
+    case PageEditor.create_context(project_slug, revision_slug, author) do
       {:ok, context} ->
+        context = Map.put(context, :hasExperiments, conn.assigns.project.has_experiments)
```
Francisco-Castro, line 27: I didn't want to modify `PageEditor.create_context/3` because other parts of the code use it. Instead, I added the field after retrieving the context.

darrensiegel (commented):
Nice work!

darrensiegel (approved):

PR #4856, opened 2024-06-04
[FEATURE] [MER-3182] Summary card highlights filter content - Part 1

simonchoxx:
[MER-3182](https://eliterate.atlassian.net/browse/MER-3182)

This PR adds cards to the `Content` and `Student` views in the instructor dashboard, which will allow to filter the content of the tables by different metrics. 

Each of the cards corresponds to a different metric and they can be selected and deselected in order to filter the content of the table showing the data corresponding to each tab.

https://github.com/Simon-Initiative/oli-torus/assets/16328384/096bca4d-b96f-4737-ada4-ea5039071702

[MER-3182]: https://eliterate.atlassian.net/browse/MER-3182?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli_web/components/delivery/content/card_highlights.ex
+++ b/lib/oli_web/components/delivery/content/card_highlights.ex
@@ -1,0 +1,39 @@
+defmodule OliWeb.Components.Delivery.CardHighLights do
+  use Phoenix.Component
+
+  attr :title, :string, required: true
+  attr :count, :integer, required: true
+  attr :is_selected, :boolean, default: false
+  attr :value, :string, required: true
+  attr :on_click, :map, required: true
+  attr :container_filter_by, :atom
+
+  def render(assigns) do
+    ~H"""
+    <div
+      phx-click={@on_click}
+      phx-value-selected={@value}
+      class={"w-56 h-auto rounded-md #{if @is_selected, do: "shadow border-2 border-blue-500 bg-slate-50", else: "bg-white border border-blue-500/30 hover:border-blue-500/80"} flex-col justify-start items-start px-4 py-3 hover:cursor-pointer"}
+    >
+      <div class="text-slate-500 text-xs font-normal leading-none">
+        <%= @title %>
+      </div>
+      <div class="flex items-baseline space-x-2 mt-2">
+        <div class={"text-3xl font-semibold leading-10 #{if @is_selected, do: "text-blue-500", else: "text-slate-800"}"}>
+          <%= @count %>
+        </div>
+        <div class="text-gray-400 text-xs font-normal leading-none">
+          <%= case @container_filter_by do %>
+            <% :units -> %>
+              Units
+            <% :modules -> %>
+              Modules
+            <% _ -> %>
+              Students
+          <% end %>
+        </div>
+      </div>
+    </div>
+    """
+  end
+end
--- a/lib/oli_web/components/delivery/content/card_highlights.ex
+++ b/lib/oli_web/components/delivery/content/card_highlights.ex
@@ -1,4 +1,4 @@
-defmodule OliWeb.Components.Delivery.CardHighLights do
+defmodule OliWeb.Components.Delivery.CardHighlights do
   use Phoenix.Component
 
   attr :title, :string, required: true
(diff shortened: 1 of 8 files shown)


lib/oli_web/components/delivery/content/card_highlights.ex
```diff
@@ -0,0 +1,39 @@
+defmodule OliWeb.Components.Delivery.CardHighLights do
```
gastonabella, line 1: Nit but should it be `CardHighlights`?

lib/oli_web/components/delivery/content/content.ex
```diff
@@ -321,6 +377,22 @@ defmodule OliWeb.Components.Delivery.Content do
     end)
   end
 
+  defp maybe_filter_by_card(containers, nil), do: containers
+  defp maybe_filter_by_card(containers, ""), do: containers
+
+  defp maybe_filter_by_card(containers, selected_card_value) do
+    case selected_card_value do
+      "1" ->
```
gastonabella, line 385: It would be easier for me to map and understand if we used a more explicit value instead of a string representation of a number. Something like the atoms you've already used above (`:high_progress_low_proficiency`, `zero_student_progress`) would make this code easier to understand IMO. You could also use pattern matching on this function definition instead of the case, but that's just a nitpick :)

You could end up with something like:
```elixir
defp maybe_filter_by_card(containers, :high_progress_low_proficiency) ...
defp maybe_filter_by_card(containers, :zero_student_progress) ...
defp maybe_filter_by_card(containers, _), do: containers
```

lib/oli_web/components/delivery/students/students.ex
```diff
@@ -93,6 +119,38 @@ defmodule OliWeb.Components.Delivery.Students do
...
+  defp maybe_filter_by_card(students, "", _), do: students
+
+  defp maybe_filter_by_card(students, selected_card_value, filter_by) do
+    case selected_card_value do
+      "1" ->
+        Enum.filter(students, fn student ->
+          Metrics.progress_range(student.progress) == "Low" ||
+            (is_nil(student.progress) and
+               (student.enrollment_status == filter_by and
+                  student.user_role_id == 4))
```
gastonabella, line 132: Wdyt about using an existing function or defining a new one that makes the role check more readable? At first glance, it is not easy to realize what role we are using to filter here.

lib/oli_web/components/delivery/students/students.ex
```diff
@@ -93,6 +119,38 @@ defmodule OliWeb.Components.Delivery.Students do
...
+        end)
+
+      "2" ->
+        Enum.filter(students, fn student ->
+          Metrics.proficiency_range(student.overall_proficiency) == "Low" ||
+            (is_nil(student.overall_proficiency) and
+               (student.enrollment_status == filter_by and
+                  student.user_role_id == 4))
```
gastonabella, line 140: Same here :)

lib/oli_web/components/delivery/students/students.ex
```diff
@@ -48,6 +48,30 @@ defmodule OliWeb.Components.Delivery.Students do
...
 
+    selected_card_value = Map.get(assigns.params, :selected_card_value, nil)
+    students_count = students_count(students, params.filter_by)
+
+    card_props = [
+      %{
+        title: "Low Progress",
+        count: Map.get(students_count, :low_progress),
+        is_selected: selected_card_value == "1",
+        value: "1"
```
gastonabella, line 59: Similar to the other comment, I would use a more descriptive value here

gastonabella (commented):
Looks really good!

Besides the comments, I realized there's a small issue on dark mode (see screenshot). 

Screenshot 2024-06-04 at 4 54 20 PM

`Course Modules` is barely readable, and the cards have a light background while it should probably have a grayish one. WDYT?

lib/oli_web/components/delivery/content/content.ex
simonchoxx: I like your approach. Thanks

simonchoxx:
Thanks @gastonabella for your comments. I have already improved the dark mode and made some changes based on your suggestions!

Captura de pantalla 2024-06-05 a la(s) 2 18 30 p  m

lib/oli_web/components/delivery/students/students.ex
```diff
@@ -524,7 +517,10 @@ defmodule OliWeb.Components.Delivery.Students do
   end
 
   def handle_event("select_card", %{"selected" => value}, socket) do
-    value = if value == Map.get(socket.assigns.params, :selected_card_value), do: nil, else: value
+    value =
+      if String.to_atom(value) == Map.get(socket.assigns.params, :selected_card_value),
```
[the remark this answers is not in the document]
simonchoxx: Done!

gastonabella (approved):

PR #4862, opened 2024-06-06
[FEATURE] [MER-3010] Activity reports display component 

rgachuhi:
This PR adds a new feature to Torus that allows any activity to generate a report, which can be displayed on any other page or page via the new "Report" component. 

Example: The legacy student cognition toolbox uses a component in which students take a survey on one page and then will see their results on the next. This is called a survey report in legacy.

https://github.com/Simon-Initiative/oli-torus/assets/11138398/b5b2ffbb-0e5c-46e5-816d-921dbff90355

https://github.com/Simon-Initiative/oli-torus/assets/11138398/acd3a744-31b8-44c2-bf03-d90a2527fb80


lib/oli/rendering/report.ex
```diff
@@ -0,0 +1,34 @@
...
+  @moduledoc """
+  This modules defines the rendering functionality for report.
+  """
+  import Oli.Utils
+
+  alias Oli.Rendering.Context
+
+  @callback report(%Context{}, %{}) :: [any()]
+  @callback error(%Context{}, %{}, {Atom.t(), String.t(), String.t()}) :: [any()]
+
```
eliknebel, line 11: `Atom.t()` shows as an unknown type. Use `atom()` instead

```suggestion
  @callback error(%Context{}, %{}, {atom(), String.t(), String.t()}) :: [any()]
```

assets/src/components/resource/editors/createEditor.tsx
```diff
@@ -58,6 +59,9 @@ export const createEditor = (editorProps: EditorProps): JSX.Element => {
       return <PurposeGroupEditor {...editorProps} contentItem={contentItem} />;
     case 'survey':
       return <SurveyEditor {...editorProps} contentItem={contentItem} />;
+    case 'report':
+      console.log('creating report editor');
```
eliknebel, line 63: Remove console log

eliknebel (commented):
Implementation looks good, tested and works as expected. Just a few comments about databse calls in the rendering codepath and console log statements

assets/src/components/misc/VegaLiteRenderer.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { VegaLite, VisualizationSpec } from 'react-vega';
+
+export interface VegaLiteSpec {
+  spec: VisualizationSpec;
+}
+
+export const VegaLiteRenderer = (props: VegaLiteSpec) => {
+  console.log(' sdhiere');
```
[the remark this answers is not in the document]
rgachuhi: Done

lib/oli/rendering/report/html.ex
```diff
@@ -0,0 +1,113 @@
...
+      report,
+      "</div></div>\n"
+    ]
+  end
+
+  defp determine_parent_page(section_slug, activity_id) do
+    pub_ids = DeliveryResolver.section_publication_ids(section_slug) |> Oli.Repo.all()
+
+    Enum.reduce(pub_ids, %{}, fn a, c ->
+      case Map.get(Oli.Publishing.determine_parent_pages([activity_id], a), activity_id) do
```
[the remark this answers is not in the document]
rgachuhi: I could not find a sensible way to precompute the function, but I have refactored it to be a more efficient query

lib/oli/rendering/report.ex
rgachuhi: done

assets/src/components/resource/editors/createEditor.tsx
rgachuhi: Removed

lib/oli/delivery/attempts/core.ex
```diff
@@ -589,6 +589,26 @@ defmodule Oli.Delivery.Attempts.Core do
     |> Repo.preload(revision: [:activity_type])
   end
 
+  def get_latest_activity_attempt(section_id, user_id, resource_id) do
+    Repo.one(
```
darrensiegel, line 593: This does not take into account that a resource (page) can have multiple attempts.  This should be reworked to ensure that you are fetching the latest attempt for this activity WITHIN the latest attempt for the page.

lib/oli/activities/reports/providers/OliLikert.ex
```diff
@@ -0,0 +1,166 @@
...
+  defp process_attempt_data(section_id, user_id, activity_id) do
+    activity_attempt =
+      Core.get_latest_activity_attempt(section_id, user_id, activity_id)
+      |> Oli.Repo.preload(:part_attempts)
+
+    part_attempts_by_id =
+      Enum.reduce(activity_attempt.part_attempts, %{}, fn a, c -> Map.put(c, a.part_id, a) end)
+
+    data =
+      Map.get(activity_attempt.revision.content, "items")
```
darrensiegel, line 68: The `revision` field should not be directly used here, as doing so would not take into account dynamic questions.  Instead, one must select the model, taking into account the `transformed_model` attribute of the activity attempt.  If it is not `nil`, use that.  If it is `nil`, using the content of the `revision`.

lib/oli/publishing.ex
```diff
@@ -1489,7 +1489,8 @@ defmodule Oli.Publishing do
   Returns a map of activity_ids to a map containing the slug and resource id of the
   page that encloses it
   """
-  def determine_parent_pages(activity_resource_ids, publication_id) do
+  def determine_parent_pages(activity_resource_ids, publication_ids)
```
darrensiegel, line 1492: List support is added here, but it doesn't seem to be used anywhere.  If it isn't being used, we should revert these changes.

darrensiegel (changes requested):
This looks really good.  There are a few things that have to be addressed.  After that is done I'll do a second pass review where I'll test this out.

darrensiegel:
Also, I see in the screenshot that the VegaLite actions menu is present. We want to disable that: https://stackoverflow.com/questions/63715640/how-to-hide-the-three-dot-dropdown-menu-on-a-vega-chart

rgachuhi:
> Screenshot 2024-06-11 at 7 28 53 PM
> I noticed that Vegalite has a menu that allows you to open in an editor. Is there any way to disable this menu?
> 
> Also, is this number scale intentional? The numbers seem unnecessarily large

Not sure why the ranges are that high, but it all depends on the values the likert activity sets

rgachuhi:
> Also, I see in the screenshot that the VegaLite actions menu is present. We want to disable that: https://stackoverflow.com/questions/63715640/how-to-hide-the-three-dot-dropdown-menu-on-a-vega-chart

I was able to disable those.

lib/oli/publishing.ex
rgachuhi: That is used in the `defp determine_parent_link(section_slug, activity_id) do` at `defmodule Oli.Rendering.Report.Html do`

assets/src/components/misc/VegaLiteRenderer.tsx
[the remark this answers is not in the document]
rgachuhi: Removed

lib/oli/delivery/attempts/core.ex
rgachuhi: Not sure I understand this. From what I tell, the query fetches the last activity attempt based on the greater activity attempt id, which means that for a particular activity, user and section combo, it will get the greatest activity attempt regardless of what page attempt that belongs to.

lib/oli/activities/reports/providers/OliLikert.ex
rgachuhi: Issue addessed

lib/oli/activities/reports/providers/OliLikert.ex
```diff
@@ -0,0 +1,172 @@
...
+  alias Oli.Delivery.Attempts.Core
+
+  @impl Oli.Activities.Reports.Renderer
+  def render(
+        %Oli.Rendering.Context{enrollment: enrollment, user: user} = context,
+        %{"activityId" => activity_id} = _element
+      ) do
+    with %{choices: choices, data: data} <-
+           process_attempt_data(enrollment.section_id, user.id, activity_id) do
+      data_url = "/api/v1/activity/report/#{enrollment.section_id}/#{activity_id}"
```
darrensiegel, line 13: We shouldn't hardcode URLs in code.  Instead, use the `Routes` module to determine the url.  But doing so here would lead to a dependency from `Oli` to `OliWeb`, which we don't want.   So you will need to pass to this function a function that can build the URL.

lib/oli/activities/reports/providers/OliLikert.ex
```diff
@@ -0,0 +1,172 @@
...
+          response: r,
+          color: color_matcher(Map.get(a, "group")),
+          group: Map.get(a, "group"),
+          index: idx
+        }
+
+        c ++ [value]
+      end)
+
+    %{choices: Map.get(activity_attempt.revision.content, "choices"), data: data}
```
darrensiegel, line 104: We shouldn't use the revision directly, as this will not honor dynamic question content.

lib/oli/delivery/attempts/core.ex
```diff
@@ -589,6 +589,26 @@ defmodule Oli.Delivery.Attempts.Core do
...
+      from(aa in ActivityAttempt,
+        left_join: aa2 in ActivityAttempt,
+        on:
+          aa.resource_attempt_id == aa2.resource_attempt_id and
+            aa.id < aa2.id,
+        join: ra in ResourceAttempt,
+        on: ra.id == aa.resource_attempt_id,
+        join: rac in ResourceAccess,
+        on: rac.id == ra.resource_access_id,
```
darrensiegel, line 602: Maybe this actually should be querying to find the latest "evaluated" activity?

darrensiegel (changes requested):
Two things to change and a question around the "get_latest_activity_attempt". 

Also, perhaps I am just mis-remembering, but I thought that the legacy OLI activity report showed the student not "their" response, but instead the aggregation of their classmates responses for the activity in question?  As in "Here is what your peers answered to this question".

darrensiegel:
I tested this out locally with a new course. I first created a practice page and added a Likert activity.  It contained three choices "Agree", "Neither Agree or Disagree" and "Disagree".  I added two Items "Item one" and "Item two".  I then added an activity report to a second page, published and accessed the course as a student. 

Visiting the page with the report first correctly showed me the message that I hadn't answered the question on the other page. 

After answering the question "Disagree" for Item 1 and "Agree" for item 2, then clicking "Submit" for the Likert, I went back to the page with the report and see this:
Screenshot 2024-06-14 at 7 56 29 AM

The visualization doesn't seem to be correct in a few ways.  First, it isn't showing all choices in the Y axis, only the two that I picked.  Second, the Y axis scale doesn't make sense.  And finally, I'm not sure what good using a horizontal bar chart here is.  It immediately gives me the sense that P1 is "more" or "better" than P2.    Much better would be something along the lines of https://vega.github.io/vega-lite/examples/layer_likert.html

But as I also pointed out in the other comment, I thought that the Legacy impl of this report showed the "Aggregate" results of this activity across all students in the course?  

I then reset the activity on the page and went back to the page with the report and now see this:
Screenshot 2024-06-14 at 7 57 46 AM

So I think we probably should be only querying for the latest "evaluated" activity

lib/oli/activities/reports/providers/OliLikert.ex
darrensiegel: The purpose of splitting the data apart from the visualization and instead embedding a URL to the controller was so that we do not have to do any DB calls during the rendering phase.  But `process_attempt_data` ends up making three DB queries.  One for the "get_latest_attempt", then two for the external `preloads`.  

We need to remove all of these DB calls here.

rgachuhi: No hardcoded URL in the backend. Moved logic into a new likert frontend component

I re-implemented it with a new solution that avoids most db calls except for a minor one on the render function.

lib/oli/delivery/attempts/core.ex
rgachuhi: Latest "evaluated" activity check added to the query

rgachuhi:
> I tested this out locally with a new course. I first created a practice page and added a Likert activity. It contained three choices "Agree", "Neither Agree or Disagree" and "Disagree". I added two Items "Item one" and "Item two". I then added an activity report to a second page, published and accessed the course as a student.
> 
> Visiting the page with the report first correctly showed me the message that I hadn't answered the question on the other page.
> 
> After answering the question "Disagree" for Item 1 and "Agree" for item 2, then clicking "Submit" for the Likert, I went back to the page with the report and see this: Screenshot 2024-06-14 at 7 56 29 AM
> 
> The visualization doesn't seem to be correct in a few ways. First, it isn't showing all choices in the Y axis, only the two that I picked. Second, the Y axis scale doesn't make sense. And finally, I'm not sure what good using a horizontal bar chart here is. It immediately gives me the sense that P1 is "more" or "better" than P2. Much better would be something along the lines of https://vega.github.io/vega-lite/examples/layer_likert.html
> 
> But as I also pointed out in the other comment, I thought that the Legacy impl of this report showed the "Aggregate" results of this activity across all students in the course?
> 
> I then reset the activity on the page and went back to the page with the report and now see this: Screenshot 2024-06-14 at 7 57 46 AM
> 
> So I think we probably should be only querying for the latest "evaluated" activity

I have addressed the three issues commented on above. One; improved the visualization to something that makes more sense, two; ensured that the likert scalling works correctly, three; ensured that the last "evaluated" attempt is respected

The legacy impl is indeed reporting for individual users not the "Aggregate" :-)

lib/oli/activities/reports/providers/OliLikert.ex
rgachuhi: Changed to honor the transformed model and fallback to content if model not transformed.

rgachuhi:
> Screenshot 2024-06-11 at 7 28 53 PM
> I noticed that Vegalite has a menu that allows you to open in an editor. Is there any way to disable this menu?
> 
> Also, is this number scale intentional? The numbers seem unnecessarily large

assets/src/components/activities/likert/reports/LikerReportRenderer.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+export type Report = {
+  type: 'success';
+  spec?: VisualizationSpec;
+  prompts?: string;
+  message?: string;
+  parent: Parent;
+};
+
+export const LikerReportRenderer = (props: LikerReportRendererProps) => {
```
darrensiegel, line 24: I just noticed this is misspelled `Liker` instead of `Likert`

darrensiegel (changes requested):
One minor thing with a incorrect spelling. 

Also, it appears as all of the DB work is now done during rendering, but it looks like the activity_report_data_controller and the LikerReportRenderer component are still in place.  Can we now delete those?

assets/src/components/activities/likert/reports/LikerReportRenderer.tsx
rgachuhi: I shifted all DB work from the `Oli.Rendering.Report.Html` report function. That way, rendering of any page containing the report is fast. The actual contents of the activity are rendered when the client-side LikertReportRenderer calls back to the `activity_report_data_controller` to fetch data in addition to other content that requires DB calls. I figured that I may as well bundle the data for the VegaLite with the spec for this activity report. Other activities may choose to use the `activity_report_data_controller` otherwise.

darrensiegel (approved):

darrensiegel (approved):

PR #4863, opened 2024-06-07
[BUG FIX] [NG23-159] Adaptive pages back button

nicocirio:
[Link](https://eliterate.atlassian.net/browse/NG23-159) to the ticket

With this fix, the adaptive page's back button redirects the student to the page from which they accessed the adaptive page.
Before the fix the student was always redirected to the explorations page, not considering that an adaptive page may not necessarily be an exploration.

## Adaptive page (not an exploration)

https://github.com/Simon-Initiative/oli-torus/assets/74839302/77ca1491-3089-49a8-a454-89cf31b19c36

_**Note:**_ It may lead to confusion that when accessing the page we read at the center of the screen a title that says "Climate Exploration" and we were supposed to be accessing an adaptive page that was **not** an exploration. The reason is that I built an adaptive page (not an exploration) by grabbing the "Climate Exploration" page and changing its purpose from `:application` to `:foundation`.
You may validate this by looking at the card icon on the home page, that has a flag icon (graded page icon) instead of a world icon (exploration icon)

## Adaptive page (an exploration)

https://github.com/Simon-Initiative/oli-torus/assets/74839302/2ec14c7b-ed63-4326-ad3f-128197826665

### DIFF
--- a/assets/src/apps/delivery/layouts/deck/DeckLayoutHeader.tsx
+++ b/assets/src/apps/delivery/layouts/deck/DeckLayoutHeader.tsx
@@ -16,6 +16,7 @@ import ReviewModeNavigation from './components/ReviewModeNavigation';
 
 interface DeckLayoutHeaderProps {
   pageName?: string;
+  backUrl?: string;
   activityName?: string;
   showScore?: boolean;
   themeId?: string;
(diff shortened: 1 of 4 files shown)


gastonabella (approved):
Looks great!

PR #4867, opened 2024-06-11
[FEATURE] [MER-3182] Summary card highlights filter content - Part 2

simonchoxx:
[MER-3182](https://eliterate.atlassian.net/browse/MER-3182)

This PR adds two new functionalities to the tab that shows the metrics of the students per container. 

- A navigation component to browse through the containers and see the metrics of the students for each of them. You can choose to browse all containers or those that have been previously filtered in the `insights >> Content` tab.

- The button to go back to the Insights >> Content view, which can show `Back to Units` or `Back to Modules`, will go back but show the results with the filters that had been applied at the time of entering the current tab. 

On the other hand, an ETS has been implemented, allowing to keep the state of the liveview and that both components can share information without losing it if a page reload happens. 

https://github.com/Simon-Initiative/oli-torus/assets/16328384/ed41dd29-4d7f-492d-aa96-d260b0e7cfb3

https://github.com/Simon-Initiative/oli-torus/assets/16328384/7c9f4239-9a50-4959-bbe1-f52e7bacd33a

[MER-3182]: https://eliterate.atlassian.net/browse/MER-3182?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


darrensiegel (changes requested):
While I appreciate the attempt the cache the navigation data with the ETS table impl, we should remove that from this PR as there are multiple problems with that approach:

1. Using only the `user_id` for key uniqueness will lead to errors as soon as an instructor that is managing multiple course sections uses this feature.  It also doesn't seem to support admin usage (admins do not have a `current_user`. 
2. On multi app server scenarios, there is no guarantee that a page reload event will end up hitting the same app server, where that ETS table data is stored.  
3. Adding what is effectively a cache of user state at this point in the development cycle is too high risk.

simonchoxx:
@darrensiegel Thanks for your review. I agree with what you say. I think it would just solve the problem of sharing necessary information between live components through URL params. Besides that it will help us (I think) to prevent data loss when refreshing the page. WDYT?

darrensiegel (approved):

PR #4872, opened 2024-06-12
[FEATURE] [MER-3106] Insert existing ab testing point

rgachuhi:
This PR adds a feature that allows an author to add UpGrade experiment options to a page and author contents for each corresponding option. It also displays an error message when UpGrade is disabled. 

https://github.com/Simon-Initiative/oli-torus/assets/11138398/a6b801ba-0998-484a-97c7-195c551436e3


darrensiegel (changes requested):
Unlike regular Alternatives group elements, Upgrade decision points *must* contain all options for the experiment to function properly.  Therefore, we need a UX impl here where when the user clicks "A/B Test", the system should create and insert a decision point *with all of its existing options already defined*.  This was maybe not clear from the MER ticket where it says "Then, the decision point created on the “Experiments” page will appear in the page with its corresponding options (The user does not choose from a drop-down like in V27)"

Also, there is a flickering of a modal upon insertion of the decision point.  That should be removed as well.

rgachuhi:
> Unlike regular Alternatives group elements, Upgrade decision points _must_ contain all options for the experiment to function properly. Therefore, we need a UX impl here where when the user clicks "A/B Test", the system should create and insert a decision point _with all of its existing options already defined_. This was maybe not clear from the MER ticket where it says "Then, the decision point created on the “Experiments” page will appear in the page with its corresponding options (The user does not choose from a drop-down like in V27)"
> 
> Also, there is a flickering of a modal upon insertion of the decision point. That should be removed as well.

Those two issues should now be resolved. One question, what happens when a user adds a new decision point to the experiment after the ab test has been placed on a page?

darrensiegel:
| One question, what happens when a user adds a new decision point to the experiment after the ab test has been placed on a page?

I'm not sure that I understand. But with A/B testing enabled there can be only 1 decision point resource in the project. Instances of that decision point can be placed any number of times in any number of pages throughout a course. 

If an author adds a new "Option" to the decision point after placing an item on the page, likely the entire AB test setup has to be reworked (i.e., new Segment and Experiment JSON files have to downloaded and ingested into Upgrade).  We probably should disable the "on page" Option adding UI for decision points, which forces the user to go back to the Experiments page to add an option.  From there, a message indicating that "Adding additional options will require rebuilding the experiment in Upgrade" or something like that.

rgachuhi:
> | One question, what happens when a user adds a new decision point to the experiment after the ab test has been placed on a page?
> 
> I'm not sure that I understand. But with A/B testing enabled there can be only 1 decision point resource in the project. Instances of that decision point can be placed any number of times in any number of pages throughout a course.
> 
> If an author adds a new "Option" to the decision point after placing an item on the page, likely the entire AB test setup has to be reworked (i.e., new Segment and Experiment JSON files have to downloaded and ingested into Upgrade). We probably should disable the "on page" Option adding UI for decision points, which forces the user to go back to the Experiments page to add an option. From there, a message indicating that "Adding additional options will require rebuilding the experiment in Upgrade" or something like that.

Ok I will hide the add option button in the Alternatives UI of upgrade experiments. In addition to what you said, if a user does add a new option in the experiments UI, they will need to delete any experiments place on pages and re-add them in order have all the options.

rgachuhi:
Add option button in the Alternatives UI now hidden when upgrade experiment selected

darrensiegel (approved):

PR #4882, opened 2024-06-14
[BUG FIX] [MER-3337] Hitting backspace in many places in adaptive pages causes delete popup

dtiwarATS:
Hey @bsparks / @darrensiegel please review the PR. Thanks.

### DIFF
--- a/assets/src/components/parts/janus-input-number/InputNumber.tsx
+++ b/assets/src/components/parts/janus-input-number/InputNumber.tsx
@@ -275,7 +275,7 @@ const InputNumber: React.FC<PartComponentProps<InputNumberModel>> = ({
       {showLabel && (
         <React.Fragment>
           <label htmlFor={`${id}-number-input`} className="inputNumberLabel">
-            {label.length > 0 ? label : ''}
+            {label?.length > 0 ? label : ''}
           </label>
           <br />
         </React.Fragment>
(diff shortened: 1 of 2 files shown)


assets/src/components/parts/janus-input-number/InputNumber.tsx
```diff
@@ -275,7 +275,7 @@ const InputNumber: React.FC<PartComponentProps<InputNumberModel>> = ({
       {showLabel && (
         <React.Fragment>
           <label htmlFor={`${id}-number-input`} className="inputNumberLabel">
-            {label.length > 0 ? label : ''}
+            {label?.length > 0 ? label : ''}
```
dtiwarATS, line 278: When the user clears the label of the number input, the label's `value `is set to `undefined`. As a result, attempting to execute `label.length` throws an error because label is `undefined`, and we're trying to get its length.

assets/src/components/parts/janus-input-number/InputNumberAuthor.tsx
```diff
@@ -30,7 +30,7 @@ const InputNumberAuthor: React.FC<AuthorPartComponentProps<InputNumberModel>> =
       {showLabel && (
         <React.Fragment>
           <label htmlFor={`${id}-number-input`} className="inputNumberLabel">
-            {label.length > 0 ? label : ''}
+            {label?.length > 0 ? label : ''}
```
dtiwarATS, line 33: When the user clears the label of the number input, the label's `value `is set to `undefined`. As a result, attempting to execute `label.length` throws an error because label is `undefined`, and we're trying to get its length.

darrensiegel (approved):

PR #4909, opened 2024-06-24
[BUG FIX] [NG23-227] Fix Contained Objectives Builder

gastonabella:
[NG23-227](https://eliterate.atlassian.net/browse/NG23-227)

Fixes the issue where many contained objectives records were not inserted when running the Contained Objectives creation task or remixing a course section.
The cause was a bad assumption of the page's content structure: the logic assumed that objects with the `activity-reference` type were only at the top level of the model, and they could be at any level.

## Technical notes:
The bugfix consists of a recursive function written in PostgreSQL, that receives a `jsonb` representing a page's content and recursively iterates over the nested structure to find all the activity ids. Another solution could be to implement the same function in Elixir, which would force the query to load all the page contents for a course in memory, and in big courses that could not be optimal. The SQL function not only avoids that, but it also reduces the query's processing time.

Here are the results when running the query for the RealCHEM course locally, before and after  the solution. Note that the processing times decrease after the fix, except for the contained-objectives-creation time that obviously increased since we are inserting 800 more rows.

 ### **Without `psql` function**
 build_contained_objectives pages: **717.81075ms**
 build_contained_objectives activities: **199.85220800000002ms**
 build_contained_objectives contained_objectives: **0.147708ms**
contained_objectives_count: **48**

 ### **With `psql` function**
 build_contained_objectives pages: **155.04484200000002ms**
 build_contained_objectives activities: **24.20246ms**
 build_contained_objectives contained_objectives: **1.038736ms**
 contained_objectives_count: **848**

## Testing notes:
For testing and running the fix in prod, someone with console access should run the mix task again but with the `--all` flag to remove all the existing contained objectives and create new ones.

`mix create_contained_objectives --all`

[NG23-227]: https://eliterate.atlassian.net/browse/NG23-227?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/mix/tasks/create_contained_objectives.ex
+++ b/lib/mix/tasks/create_contained_objectives.ex
@@ -1,5 +1,8 @@
 defmodule Mix.Tasks.CreateContainedObjectives do
-  @shortdoc "Create contained objectives for all sections that were not migrated"
+  @shortdoc """
+  Create contained objectives for all sections that have not yet been migrated.
+  Use --all to drop existing objectives and create contained objectives for all sections.
+  """
 
   use Mix.Task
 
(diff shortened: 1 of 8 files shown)


darrensiegel (approved):
This all looks great.  I really appreciate the diligence paid towards testing the speed of the new postgres function.

PR #4910, opened 2024-06-24
[BUG FIX] [NG23-252] Learning objective proficiency at page does not consider child learning objective

nicocirio:
[Link](https://eliterate.atlassian.net/browse/NG23-252) to the ticket.

Take into account that this PR points to `NG23-246-content-on-page-is-not-visible-on-begin-attempt` instead of `prerelease-v0.28.0` because the target branch has changes applied to the lesson and prologue view that was needed for this ticket (to prevent future merge conflicts)

### Before
image

### After
image

### DIFF
--- a/lib/oli/delivery/metrics.ex
+++ b/lib/oli/delivery/metrics.ex
@@ -665,6 +665,96 @@ defmodule Oli.Delivery.Metrics do
     |> Enum.into(%{})
   end
 
+  @doc """
+  Given a list of objective revisions it calculates the aggregated proficiency for each of those objectives
+  for a given student in a given course.
+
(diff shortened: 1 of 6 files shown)


lib/oli/delivery/metrics.ex
```diff
@@ -665,6 +665,96 @@ defmodule Oli.Delivery.Metrics do
     |> Enum.into(%{})
   end
 
+  @doc """
+  Given a list of objective revisions it calculates the aggregated proficiency for each of those objectives
```
darrensiegel, line 669: Fantastic work!

darrensiegel (approved):

PR #4915, opened 2024-06-25
[BUG FIX] [NG23-222] Fix Due Date in Homepage

gastonabella:
[NG23-222](https://eliterate.atlassian.net/browse/NG23-222)

Fixes how the Due Date description is calculated on the student's homepage.
There was an edge case where a resource's end date and the current date were on the same day in the UTC timezone, but in the user's timezone, they were not. With the fix, we now consider the user's timezone for the calculation. 

[NG23-222]: https://eliterate.atlassian.net/browse/NG23-222?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli/datetime.ex
+++ b/lib/oli/datetime.ex
@@ -1,8 +1,13 @@
 defmodule Oli.DateTime do
   @callback utc_now :: DateTime.t()
+  @callback now!(String.t()) :: DateTime.t()
 
   def utc_now() do
     date_time().utc_now()
+  end
+
+  def now!(timezone) do
+    date_time().now!(timezone)
   end
 
   defp date_time(), do: Application.get_env(:oli, :date_time_module, DateTime)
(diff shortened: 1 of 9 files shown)


lib/oli/datetime.ex
```diff
@@ -1,9 +1,14 @@
...
   @callback utc_now :: DateTime.t()
+  @callback now!(String.t()) :: DateTime.t()
 
   def utc_now() do
     date_time().utc_now()
   end
 
+  def now!(timezone) do
+    date_time().now!(timezone)
+  end
```
nicocirio, line 11: 👏

nicocirio (approved):
Nice work! 🚀

eliknebel (approved):

PR #4918, opened 2024-06-26
[BUG FIX] [NG23-228] social annotation point blocks not working for some pages

eliknebel:
https://eliterate.atlassian.net/browse/NG23-228

This PR fixes an issue where social annotation point markers were not showing up for some pages, specifically identified in the chemistry course.

The root cause of the issue was due to the `ids_added` attribute on the publication not taking into account pages which may not have been processed as part of the publication process when ids were fixed, leading to the possibility that certain pages added in after the fact may not have been processed to ensure these ids exist.

The fix here is to remove the `ids_added` attribute from the publication record and instead add it to the revision record, and run the `ensure_ids` logic on every publish. Once a revision has been processed and has this flag set, it no longer needs to be processed going forward because the system guarantees at that point that these ids exists at the authoring level.

The risk here is medium since it involves adding an attribute to the revisions table, but special care was used to ensure that this attribute is also included in the revision_from_previous logic and there is a decent amount of unit test coverage around revisions. We will want to spend some extra time in QA testing the publishing process, authoring then republishing to ensure everything is still working correctly here.

This is an important fix for social annotation feature, but also in general to close this gap where the content is assumed to have the required ids based on this flag, however in the current state that assumption is not always correct.

### DIFF
--- a/lib/oli/resources/revision.ex
+++ b/lib/oli/resources/revision.ex
@@ -37,6 +37,7 @@ defmodule Oli.Resources.Revision do
     field :title, :string
     field :slug, :string
     field :deleted, :boolean, default: false
+    field :ids_added, :boolean, default: false
     belongs_to :author, Oli.Accounts.Author
     belongs_to :resource, Oli.Resources.Resource
     belongs_to :previous_revision, Oli.Resources.Revision
@@ -96,6 +97,7 @@ defmodule Oli.Resources.Revision do
       :title,
       :slug,
       :deleted,
+      :ids_added,
       :author_id,
       :resource_id,
       :primary_resource_id,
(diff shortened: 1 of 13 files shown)


lib/oli/resources/revision.ex
```diff
@@ -37,6 +37,7 @@ defmodule Oli.Resources.Revision do
     field :title, :string
     field :slug, :string
     field :deleted, :boolean, default: false
+    field :ids_added, :boolean, default: false
```
nicocirio, line 40: NIT: Adding a comment on the meaning of this field might be helpful in the future. In the context of this PR we understand its purpose but probably in some months, it won't be easy to understand why and where we need this field.

nicocirio (approved):

PR #4923, opened 2024-06-27
[FEATURE] [MER-3065] Track assessment settings audit log

simonchoxx:
[MER-3065](https://eliterate.atlassian.net/browse/MER-3065)

This PR implements a tracking of all changes made in the `Assessment Settings` view, both for the individual changes of each setting and for the existing bulk apply option.

This new functionality will store the changes in a new table called `settings_changes` in the database.

[MER-3065]: https://eliterate.atlassian.net/browse/MER-3065?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli/delivery/settings/settings_changes.ex
+++ b/lib/oli/delivery/settings/settings_changes.ex
@@ -8,6 +8,7 @@ defmodule Oli.Delivery.Settings.SettingsChanges do
     belongs_to(:section, Oli.Delivery.Sections.Section)
     belongs_to(:user, Oli.Accounts.User)
 
+    field(:user_type, Ecto.Enum, values: [:author, :instructor])
     field(:key, :string)
     field(:new_value, :string)
     field(:old_value, :string)
@@ -21,6 +22,7 @@ defmodule Oli.Delivery.Settings.SettingsChanges do
       :user_id,
       :section_id,
       :resource_id,
+      :user_type,
       :key,
       :new_value,
       :old_value
--- a/lib/oli/delivery/settings/settings_changes.ex
+++ b/lib/oli/delivery/settings/settings_changes.ex
@@ -6,8 +6,8 @@ defmodule Oli.Delivery.Settings.SettingsChanges do
   schema "settings_changes" do
     belongs_to(:resource, Oli.Resources.Resource)
     belongs_to(:section, Oli.Delivery.Sections.Section)
-    belongs_to(:user, Oli.Accounts.User)
 
+    field(:user_id, :integer)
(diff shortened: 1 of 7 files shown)


darrensiegel (changes requested):
Changes look good, only one thing that we have to consider:   If I am *only* logged in as an administrator, the `user.id` that gets written is the `id` of that Authoring account and not the id of an Instructor account.   We need to handle this and store whether the `user` is an Author or an Instructor.   This can be done via a new field `user_type` which is `:author` or `:instructor`

darrensiegel:
There is also a problem (which already existed) that we should fix.  The LiveFlash appears twice (see screenshot below). Please eliminate the larger of the two.

Screenshot 2024-06-28 at 8 34 02 AM

lib/oli/delivery/settings/settings_changes.ex
```diff
@@ -0,0 +1,32 @@
+defmodule Oli.Delivery.Settings.SettingsChanges do
+  use Ecto.Schema
+
+  import Ecto.Changeset
+
+  schema "settings_changes" do
+    belongs_to(:resource, Oli.Resources.Resource)
+    belongs_to(:section, Oli.Delivery.Sections.Section)
+    belongs_to(:user, Oli.Accounts.User)
```
darrensiegel, line 9: We should change this from `belongs_to` since it creates a foreign key relationship with the `users` table, when we will sometimes be putting author ids into this.  

Just make `user_id` an integer field

priv/repo/migrations/20240620202825_create_settings_changes.exs
```diff
@@ -0,0 +1,17 @@
+defmodule Oli.Repo.Migrations.CreateSettingsChanges do
+  use Ecto.Migration
+
+  def change do
+    create table(:settings_changes) do
+      add :resource_id, references(:resources)
+      add :section_id, references(:sections)
+      add :user_id, references(:users)
```
darrensiegel, line 8: Remove the `references` here and make `user_id` just a integer field

darrensiegel (changes requested):

lib/oli/delivery/settings/settings_changes.ex
simonchoxx: Nice catch! You are right. I will fix it.

darrensiegel (approved):

PR #4926, opened 2024-06-27
[BUG FIX] [MER-3393] Time restricted assessments no longer work

simonchoxx:
[MER-3393](https://eliterate.atlassian.net/browse/MER-3393)

This PR adds the functionality to show a countdown on graded pages, as long as the `Time Limit` option has a value greater than zero. 

Also, if a graded page has an active countdown and the countdown ends, it will redirect to the page review.

Another case that is handled is if a graded page has an active countdown, if the user closes the tab or the browser, the page submit will be executed anyway.

https://github.com/Simon-Initiative/oli-torus/assets/16328384/b9942f27-fbc8-4d67-8eed-aa5b0ce09282

[MER-3393]: https://eliterate.atlassian.net/browse/MER-3393?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nicocirio:
I am not sure about the expected behavior. By watching the video I was wondering what should happen when a student closes the page:
- if the user reopens the page before the time ends, then I would expect the user to be allowed to continue with the in-progress attempt. -> By watching the video I understood the attempt is finished.
- if the user does not return to the page and the time ends, then the attempt should be auto-submitted. -> I understand this is what is happening.

darrensiegel (approved):

PR #4948, opened 2024-07-04
[BUG FIX] [MER-3393] Time restricted assessments no longer work rework

simonchoxx:
[MER-3393](https://eliterate.atlassian.net/browse/MER-3393)

This PR adds checks necessary for the countdown on a graded page to work as expected. 

Necessary changes were made so that if the student exits the page and re-enters the page, the countdown or grace period status is maintained.

In addition, it adds the grace period time that is executed once the countdown of the set time limit for that page is over. 

Finally, we fix the bug that existed when the page was refreshed and the page submit was executed automatically.

[MER-3393]: https://eliterate.atlassian.net/browse/MER-3393?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/assets/src/hooks/countdown_timer.ts
+++ b/assets/src/hooks/countdown_timer.ts
@@ -1,66 +1,105 @@
 export const CountdownTimer = {
   mounted() {
-    const { timerId, submitButtonId, timeOutInMins, startTimeInMs, effectiveTimeInMs, autoSubmit } =
-      this.el.dataset;
-    const parsedTimeOutInMins = parseInt(timeOutInMins, 10);
-    const parsedStartTimeInMs = parseInt(startTimeInMs, 10);
-    const parsedEffectiveTimeInMs = parseInt(effectiveTimeInMs, 10);
-    const parsedAutoSubmit = autoSubmit === 'true';
-
-    countdownTimer(
+    const {
       timerId,
       submitButtonId,
-      parsedTimeOutInMins,
-      parsedStartTimeInMs,
-      parsedEffectiveTimeInMs,
-      parsedAutoSubmit,
-    );
-    window.addEventListener('beforeunload', this.handleBeforeUnload.bind(this, submitButtonId));
+      timeOutInMins,
+      startTimeInMs,
+      effectiveTimeInMs,
+      gracePeriodInMins,
+      autoSubmit,
+    } = this.el.dataset;
+
+    // Parse the dataset values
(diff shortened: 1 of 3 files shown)


assets/src/hooks/countdown_timer.ts
```diff
@@ -1,68 +1,107 @@
...
     window.removeEventListener('beforeunload', this.handleBeforeUnload.bind(this));
   },
-  handleBeforeUnload(submitButtonId: string) {
-    const submitButton = document.getElementById(submitButtonId);
-    submitButton ? submitButton.click() : console.error('Submit button not found');
+  handleVisibilityChange() {
+    this.isPageHidden = document.hidden;
+  },
+  handleBeforeUnload() {
+    if (this.isPageHidden) {
```
eliknebel, line 54: This logic seems to indicate that this will only be auto submitted when the page is closed and the page was hidden, is that the correct behavior?

I would have expected this to instead check if time is expired and submit if so, regardless of page visibility state.

Or even better - do nothing when page visibility or unload happens and let the backend handle the auto submit if a student leaves. There is already logic here to handle when the page is loaded and past due to auto submit, so that should handle the case of reloading the page past the due date

eliknebel (changes requested):

assets/src/hooks/countdown_timer.ts
darrensiegel: Yes, there already is backend code that will auto submit an assessment after the effective deadline (with about 1 minute of added slack).  

The code in the frontend should ONLY auto submit when the effective deadline is hit and the page is still visible.

The changes that need to be made here are to remove all of the event handlers for `beforeUnload` and `handleVisibilityChange` and then ONLY auto submit from the front end when the effective deadline is reached.  
Also remove the code on initialization which checks the effective deadline and auto submits.

darrensiegel (changes requested):

assets/src/hooks/countdown_timer.ts
simonchoxx: Ok, I got it. I will remove all references to `beforeUnload` and `handleVisibilityChange ` (in the `mounted` should only call the function with its corresponding parameters). Now I wonder if it makes sense the logic I built into the `countdownTimer `function to take into account not only the `effective time`, but also the `grace period` time.

eliknebel (approved):
Changes look good

darrensiegel (approved):

PR #4956, opened 2024-07-09
[ENHANCEMENT] [MER-3294] User can create an account through a course enrollment link

manelli:
Add `Sign up` button in enrollment page.

User flow preview:

https://github.com/Simon-Initiative/oli-torus/assets/5324858/695d0478-8ab8-456e-ad31-7797b419beec

See: https://eliterate.atlassian.net/browse/MER-3294

### DIFF
--- a/lib/oli_web/templates/delivery/enroll.html.heex
+++ b/lib/oli_web/templates/delivery/enroll.html.heex
@@ -14,7 +14,7 @@
         </div>
       </div>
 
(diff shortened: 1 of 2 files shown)


Francisco-Castro (commented):
This looks like it's doing the job. Just a few suggestions:

- Would it be possible to add a test for this?
- Could you change the route to a p sigil one? I see that the others endpoints use the `Routes` so you could change them too.

Additionally, I suggest adding a link to the ticket in the PR description for quick reference.

manelli:
@Francisco-Castro hey, just added tests and changed the view to use the verified routes sigil. Thanks for the review!

Francisco-Castro (approved):
Thanks for adding the tests and changing the routes. I tested it locally, and it works as expected.

darrensiegel (approved):

PR #4958, opened 2024-07-10
[ENHANCEMENT] [MER-3396] Async project export to fix gateway timeout

eliknebel:
https://eliterate.atlassian.net/browse/MER-3396

This PR fixes an issue with exporting very large courses where the export processing takes longer than a browser connection (and default Phoenix/cowboy) can handle, resulting in a Gateway Timeout error and no way to download the export for a course.

Adjusting the HAProxy setting to a longer time window did not fix the issue on stellarator, so this required a code fix.

Now the export process is handled as a asynchronous background job using the existing AsyncExporter module. It will be important that users understand that in order to get the latest version of their project, they must first "Regenerate" before clicking the download button. This is likely not the best UX, but will work for the short term and unblocks the ability to download large projects.

When a job is in progress, a spinner is shown .
Screenshot 2024-07-10 at 12 27 22 PM

When finished, LiveViews are notified and the exported project can be download via S3 link.
Screenshot 2024-07-10 at 12 27 30 PM

### DIFF
--- a/lib/oli/authoring/course.ex
+++ b/lib/oli/authoring/course.ex
@@ -563,46 +563,6 @@ defmodule Oli.Authoring.Course do
   end
 
   @doc """
-  Generates a project export for the given project if one is not already in progress
-  """
-  def generate_project_export(project) do
-    case project_export_status(project) do
-      {:in_progress} ->
-        {:error, "Project export is already in progress"}
-
-      _ ->
-        generate_project_export!(project)
-    end
-  end
-
-  defp generate_project_export!(project) do
-    %{project_slug: project.slug}
-    |> Oli.Authoring.ProjectExportWorker.new()
-    |> Oban.insert()
-  end
-
-  def project_export_status(project) do
-    if export_in_progress?(project.slug, "project_export") do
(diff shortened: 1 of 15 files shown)


lib/oli/authoring/course.ex
```diff
@@ -550,6 +562,46 @@ defmodule Oli.Authoring.Course do
...
+    case project_export_status(project) do
+      {:in_progress} ->
+        {:error, "Project export is already in progress"}
+
+      _ ->
+        generate_project_export!(project)
+    end
+  end
+
+  defp generate_project_export!(project) do
```
darrensiegel, line 578: This function introduces a bidirectional dependency between this module and the `ProjectExportWorker` module.  Instead, I'd suggest just placing this function directly within the `ProjectExportWorker` module.

darrensiegel (changes requested):

lib/oli/authoring/course.ex
eliknebel: Good catch, fixed

darrensiegel (approved):

PR #4964, opened 2024-07-11
[ENHANCEMENT] [MER-3503] Add support for accessing research consent form

eliknebel:
https://eliterate.atlassian.net/browse/MER-3503

Adds support for accessing research consent form from the user menu in delivery.

System admin can configure whether or not to use the consent form for direct delivery at the system level under Feature Flags. LTI users consent form will still be driven by the institution-specific setting.
Screenshot 2024-07-11 at 5 16 21 PM

If enabled, a "Research Consent" link will show in the user menu, which takes the user to the consent form where the previous set value for the user selection is shown. If not set, the value will default to "I Agree".
Screenshot 2024-07-11 at 5 18 00 PM

I tested the different scenarios including direct delivery student, LTI launch from Canvas as both instructor and student.


darrensiegel (approved):
Excellent work!

PR #4966, opened 2024-07-12
[BUG FIX] [MER-3010] fixes sparse likert and multiple likert attempts

rgachuhi:
This PR addresses two issues:

- Resetting and retaking a survey likert causes the report to throw an internal server error.
- If you miss an answer in the Likert, the report throws an internal server error.


eliknebel (approved):
Looks good. Tested, resolves the issues identified

PR #4976, opened 2024-07-22
[FEATURE] [MER-3320] Hide content during course remix

simonchoxx:
[MER-3320](https://eliterate.atlassian.net/browse/MER-3320)

This PR allows authors and instructors to show and hide materials from the hierarchy during the course remix.

Materials that are hidden should not be accessible by students within a course. 

_*If you think there is somewhere else where access to hidden materials should be restricted, please let me know._

https://github.com/user-attachments/assets/461928bb-441a-4c7d-a38f-c73b2fd0bd97

[MER-3320]: https://eliterate.atlassian.net/browse/MER-3320?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli_web/live/delivery/remix/hide_resource_modal.ex
+++ b/lib/oli_web/live/delivery/remix/hide_resource_modal.ex
@@ -2,6 +2,8 @@ defmodule OliWeb.Delivery.Remix.HideResourceModal do
   use Phoenix.LiveComponent
   use Phoenix.HTML
 
+  import OliWeb.Components.Common
+
   alias Oli.Delivery.Hierarchy.HierarchyNode
 
   def render(%{node: %HierarchyNode{}} = assigns) do
@@ -28,22 +30,23 @@ defmodule OliWeb.Delivery.Remix.HideResourceModal do
             Are you sure you want to <%= get_label_action(@node) %> <b><%= @node.revision.title %></b>?
           </div>
           <div class="modal-footer">
-            <button
+            <.button
               type="button"
-              class="btn btn-secondary"
+              variant={:info}
               data-bs-dismiss="modal"
               phx-click="HideResourceModal.cancel"
             >
               Cancel
-            </button>
-            <button
-              phx-click="HideResourceModal.toggle"
-              phx-key="enter"
+            </.button>
+            <.button
+              type="button"
+              variant={:danger}
               phx-value-uuid={@node.uuid}
-              class="btn btn-danger"
+              phx-key="enter"
+              phx-click="HideResourceModal.toggle"
             >
               <%= get_label_action(@node) |> String.capitalize() %>
-            </button>
+            </.button>
           </div>
         </div>
       </div>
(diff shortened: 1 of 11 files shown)


lib/oli_web/live/delivery/remix_section.ex
```diff
@@ -848,6 +848,49 @@ defmodule OliWeb.Delivery.RemixSection do
...
+       modal_assigns: modal_assigns
+     )}
+  end
+
+  def handle_event("HideResourceModal.toggle", %{"uuid" => uuid}, socket) do
+    %{hierarchy: hierarchy, section: section} = socket.assigns
+
+    section_resource = Hierarchy.find_in_hierarchy(hierarchy, uuid).section_resource
+
+    Sections.update_section_resource(section_resource, %{hidden: !section_resource.hidden})
```
darrensiegel, line 880: The toggling of `hidden` cannot be directly made to the section resource record.  Take a look at how reordering and how removing (aka "deleting") resources is done.  The change is made "in memory" to the `hierarchy` assign, and the `has_unsaved_changes` flag is then set to `true`.   Toggling `hidden` should fit into this same framework.

darrensiegel (changes requested):

lib/oli_web/live/delivery/remix/hide_resource_modal.ex
```diff
@@ -0,0 +1,55 @@
...
+              data-bs-dismiss="modal"
+              phx-click="HideResourceModal.cancel"
+            >
+              Cancel
+            </button>
+            <button
+              phx-click="HideResourceModal.toggle"
+              phx-key="enter"
+              phx-value-uuid={@node.uuid}
+              class="btn btn-danger"
```
eliknebel, line 43: Going forward, we should avoid introducing new buttons using any bootstrap classes, e.g. `btn btn-danger`. Instead, we should be using tailwind and for higher-level common components such as buttons, make use of the `<.button />` component defined in `OliWeb.Components.Common`

lib/oli_web/live/delivery/remix_section.ex
```diff
@@ -848,6 +848,50 @@ defmodule OliWeb.Delivery.RemixSection do
...
+       modal_assigns: modal_assigns
+     )}
+  end
+
+  def handle_event("HideResourceModal.toggle", %{"uuid" => uuid}, socket) do
+    %{hierarchy: hierarchy, active: active} = socket.assigns
+
+    hierarchy =
+      Hierarchy.find_and_toggle_hidden(hierarchy, uuid)
+      |> Hierarchy.finalize()
```
eliknebel, line 880: Finalize is a somewhat expensive operation that updates the numberings, etc. While eventually required before persisting, we likely don't need to call it here since it isn't necessary for rendering this view and will be called upon save

eliknebel (changes requested):

lib/oli_web/live/delivery/remix/hide_resource_modal.ex
simonchoxx: Thanks for you review @eliknebel. I'll change those buttons to use the button component.

eliknebel (approved):

darrensiegel (approved):

PR #4979, opened 2024-07-23
[BUG FIX] [MER-3506] After completing a lesson the back arrow does not work

Francisco-Castro:
Ticket: [MER-3506](https://eliterate.atlassian.net/browse/MER-3506)

This PR adds a fix for an issue that occurred after completing a lesson where the back arrow didn't work. The problem was that it failed to consider cases when the `request_path` was `""`. After completing an adaptive page, React called an endpoint in our backend that redirects twice to different endpoints, and we never filled out the `request_path`. The back link wasn't prepared for this scenario.

[MER-3506]: https://eliterate.atlassian.net/browse/MER-3506?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli_web/components/layouts/student_delivery_lesson.html.heex
+++ b/lib/oli_web/components/layouts/student_delivery_lesson.html.heex
@@ -62,7 +62,11 @@
         <%= react_component("Components.OfflineDetector") %>
       </div>
       <.back_arrow
-        to={assigns[:request_path]}
+        to={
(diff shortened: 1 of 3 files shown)


darrensiegel (approved):
Nice work to track this down.

PR #4982, opened 2024-07-24
[ENHANCEMENT] [MER-3298] Workspace nav bar and new instructor dashboard - part 2

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-3298) to the ticket

This PR is the second part of [this one](https://github.com/Simon-Initiative/oli-torus/pull/4963). It does not introduce any changes in the UI and only makes a refactor to handle the 3 workspaces (course author, instructor, and student) in different routes and liveviews.
Anticipating and considering how those three workspaces will evolve in the ["Clear delineation of which type of account"](https://eliterate.atlassian.net/jira/software/c/projects/MER/boards/2?issueParent=18503) epic, this refactor was necessary to avoid managing all required features in a unique liveview (that would end up being extremely big and difficult to maintain)


eliknebel (approved):
Looks great, nice work

PR #4984, opened 2024-07-25
[FEATURE] [MER-3504] one question at a time feature

rgachuhi:
This PR adds a new attribute, “assessment_mode,” to the Revision and SectionResource schemas. This attribute will be an enum with the values “traditional” and “one_at_a_time.” Ensures that the relevant code for creating a new course section and applying publication updates to course sections copies these values from the underlying Revision to the corresponding SectionResource. Extends the Assessment Settings UX and Student Exceptions UX to allow Instructors to change these settings for a course section and for individual students


lib/oli/resources.ex
```diff
@@ -339,6 +339,7 @@ defmodule Oli.Resources do
           time_limit: previous_revision.time_limit,
           scope: previous_revision.scope,
           retake_mode: previous_revision.retake_mode,
+          assessment_mode: previous_revision.assessment_mode,
```
eliknebel, line 342: Nice catch remembering to update this function. This merging of attrs really exploded and been a previous source of bugs, I wonder if we should take the opportunity to change the approach here and instead of explicitly specifying each attr we want to carry forward, maybe we should just merge all attrs from `previous_revision` and only drop any that we don't want to carry forward (which currently, I think is none since we are merging all of them). Thoughts?

eliknebel (changes requested):
Looks good. Though I noticed there is a wierd underscore happening in the selection text, maybe it is converting from an atom? Should we change so that this text reads with spaces instead of underscores?

Screenshot 2024-07-26 at 3 17 02 PM

lib/oli/resources.ex
rgachuhi: Makes sense. I will apply the change

rgachuhi:
> Looks good. Though I noticed there is a wierd underscore happening in the selection text, maybe it is converting from an atom? Should we change so that this text reads with spaces instead of underscores?
> 
> Screenshot 2024-07-26 at 3 17 02 PM

Updated to remove underscores

eliknebel (approved):

PR #4987, opened 2024-07-26
[FEATURE] [MER-3302] Course author new projects view

simonchoxx:
[MER-3302](https://eliterate.atlassian.net/browse/MER-3302)

This PR implements the projects view for an author in the new workspace. 

If an authenticated author exists, the already existing projects view will be displayed when positioned in the `Course Author` tab.

Since it is not yet implemented to redirect path when an author logs into the application, it is necessary to have an authenticated author and then go to the path `/sections/workspace/course_author` to be able to see the changes.

- Dark Mode

Captura de pantalla 2024-07-26 a la(s) 3 56 00 p  m

Captura de pantalla 2024-07-26 a la(s) 3 58 31 p  m

- Light Mode

Captura de pantalla 2024-07-26 a la(s) 3 55 53 p  m

Captura de pantalla 2024-07-26 a la(s) 3 58 38 p  m

[MER-3302]: https://eliterate.atlassian.net/browse/MER-3302?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/oli_web/router.ex
```diff
@@ -789,9 +796,8 @@ defmodule OliWeb.Router do
   scope "/sections/workspace/", OliWeb do
```
eliknebel, line 796: I just realized this routing is not quite right. This scope should not include 'sections', Instead it should simply be `/workspaces`. We can address this in a separate ticket

eliknebel (approved):

PR #4998, opened 2024-08-02
[FEATURE] [MER-3523] Add "dark" mode to adaptive pages

dtiwarATS:
Hey @bsparks  / @eliknebel  / @darrensiegel Could you please look at the PR? Thanks

### DIFF
--- a/assets/src/apps/delivery/Delivery.tsx
+++ b/assets/src/apps/delivery/Delivery.tsx
@@ -154,7 +154,7 @@ const Delivery: React.FC<DeliveryProps> = ({
   const dialogMessage = content?.custom?.logoutMessage;
   const fullscreen = !content?.displayApplicationChrome;
 
-  // this is something SS does...
+  // this is something SS does....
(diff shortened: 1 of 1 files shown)


assets/src/apps/delivery/Delivery.tsx
```diff
@@ -158,7 +158,7 @@ const Delivery: React.FC<DeliveryProps> = ({
   const { width: windowWidth } = useWindowSize();
   const isLessonEnded = useSelector(selectLessonEnd);
   return (
-    <div className={parentDivClasses.join(' ')}>
+    <div className={`${parentDivClasses.join(' ')} ${localStorage?.theme || ''}`}>
```
dtiwarATS, line 161: I check and found that Torus saves a `theme` key in` local storage` that contains the value of the theme student selected on the attempt page so I am just passing adding that in the class list so that external CSS can target this and add their CSS etc.

eliknebel (approved):

PR #5006, opened 2024-08-08
[BUG FIX] [MER-3566] Grade Passback Bad Sync Issue

eliknebel:
https://eliterate.atlassian.net/browse/MER-3566

Removes the grade pass back optimization to filter out students that are no longer enrolled in the LMS course.

### DIFF
--- a/lib/oli_web/live/grades/grades.ex
+++ b/lib/oli_web/live/grades/grades.ex
@@ -4,7 +4,6 @@ defmodule OliWeb.Grades.GradesLive do
   alias Oli.Grading
   alias Lti_1p3.Tool.Services.AGS
   alias Lti_1p3.Tool.Services.AGS.LineItem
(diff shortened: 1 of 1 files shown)


Francisco-Castro (approved):
I don't fully understand the workflow of this optimization, but it seems reasonable to comment out the lookup optimization function. However, if we plan to use `fetch_students/2` again in the future, it might be better to add a feature flag instead of just commenting it out.

eliknebel:
> I don't fully understand the workflow of this optimization, but it seems reasonable to comment out the lookup optimization function. However, if we plan to use `fetch_students/2` again in the future, it might be better to add a feature flag instead of just commenting it out.

Good point. There is some follow-up outlined in https://eliterate.atlassian.net/browse/MER-3574 to uncomment this function and fix the underlying issue such that a feature flag will not be required.

PR #5007, opened 2024-08-08
[BUG FIX] [MER-3573] torus learning proficiency algorithm leads to undesired results

eliknebel:
https://eliterate.atlassian.net/browse/MER-3573

This PR adjusts the way proficiency is calculated by requiring a minimum of 3 total number of first attempts before showing a proficiency or either "Low", "Medium", or "High" and otherwise returning "Not enough data".

### DIFF
--- a/lib/oli/delivery/metrics.ex
+++ b/lib/oli/delivery/metrics.ex
@@ -718,7 +718,7 @@ defmodule Oli.Delivery.Metrics do
     end)
   end
 
-  defp aggregate_raw_proficiency([]), do: proficiency_range(nil)
+  defp aggregate_raw_proficiency([]), do: proficiency_range(nil, 0)
 
   defp aggregate_raw_proficiency(raw_values) do
     {total_correct, total_count} =
@@ -728,7 +728,7 @@ defmodule Oli.Delivery.Metrics do
 
     proficiency_value = if total_count == 0, do: 0, else: total_correct / total_count
 
-    proficiency_range(proficiency_value)
+    proficiency_range(proficiency_value, total_count)
   end
 
   def raw_proficiency_per_learning_objective(%Section{analytics_version: :v1, slug: section_slug}) do
@@ -945,12 +945,12 @@ defmodule Oli.Delivery.Metrics do
            fragment(
              "CAST(COUNT(CASE WHEN ? THEN 1 END) as float) / CAST(COUNT(*) as float)",
              sn.correct
-           )}
+           ), fragment("CAST(COUNT(*) as float)")}
       )
 
     Repo.all(query)
-    |> Enum.into(%{}, fn {student_id, proficiency} ->
-      {student_id, proficiency_range(proficiency)}
+    |> Enum.into(%{}, fn {student_id, proficiency, num_first_attempts} ->
+      {student_id, proficiency_range(proficiency, num_first_attempts)}
     end)
   end
 
@@ -985,19 +985,19 @@ defmodule Oli.Delivery.Metrics do
             summary.user_id != -1 and
             summary.resource_type_id == ^page_type_id,
         where: ^filter_by_container,
-        group_by: summary.user_id,
+        group_by: [summary.user_id, summary.num_first_attempts],
         select:
           {summary.user_id,
            fragment(
              "CAST(SUM(?) as float) / NULLIF(CAST(SUM(?) as float), 0.0)",
              summary.num_first_attempts_correct,
              summary.num_first_attempts
-           )}
+           ), summary.num_first_attempts}
       )
 
     Repo.all(query)
-    |> Enum.into(%{}, fn {student_id, proficiency} ->
-      {student_id, proficiency_range(proficiency)}
+    |> Enum.into(%{}, fn {student_id, proficiency, num_first_attempts} ->
+      {student_id, proficiency_range(proficiency, num_first_attempts)}
     end)
   end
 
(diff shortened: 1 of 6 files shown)


lib/oli/delivery/metrics.ex
```diff
@@ -945,12 +945,12 @@ defmodule Oli.Delivery.Metrics do
            fragment(
              "CAST(COUNT(CASE WHEN ? THEN 1 END) as float) / CAST(COUNT(*) as float)",
              sn.correct
-           )}
+           ), fragment("CAST(COUNT(*) as float)")}
```
eliknebel, line 948: For v1 analytics (to keep it compatible with the `proficiency_range` function), we use the SQL fragment "CAST(COUNT(*) as float)" to get the total number of first attempts

lib/oli/delivery/metrics.ex
```diff
@@ -985,19 +985,19 @@ defmodule Oli.Delivery.Metrics do
             summary.user_id != -1 and
             summary.resource_type_id == ^page_type_id,
         where: ^filter_by_container,
-        group_by: summary.user_id,
+        group_by: [summary.user_id, summary.num_first_attempts],
```
eliknebel, line 988: `summary.num_first_attempts` is required to be in the group_by clause in order to be in the select clause

lib/oli/delivery/metrics.ex
```diff
@@ -1117,13 +1117,14 @@ defmodule Oli.Delivery.Metrics do
             "CAST(? as float) / NULLIF(CAST(? as float), 0.0)",
             summary.num_first_attempts_correct,
             summary.num_first_attempts
-          )
+          ),
+          summary.num_first_attempts
```
eliknebel, line 1121: v2 analytics now return the `summary.num_first_attempts` as the 3rd element in the tuple

lib/oli/delivery/metrics.ex
```diff
@@ -1243,19 +1244,20 @@ defmodule Oli.Delivery.Metrics do
...
+    |> Enum.into(%{}, fn {page_id, proficiency, num_first_attempts} ->
+      {page_id, proficiency_range(proficiency, num_first_attempts)}
     end)
   end
 
-  def proficiency_range(nil), do: "Not enough data"
-  def proficiency_range(proficiency) when proficiency <= 0.5, do: "Low"
-  def proficiency_range(proficiency) when proficiency <= 0.8, do: "Medium"
-  def proficiency_range(_proficiency), do: "High"
+  def proficiency_range(_, num_first_attempts) when num_first_attempts < 3, do: "Not enough data"
```
eliknebel, line 1256: This is the core of the logic to determine if `num_first_attempts` is less than 3, then simply return "Not enough data"

lib/oli_web/components/delivery/students/students.ex
```diff
@@ -231,7 +231,7 @@ defmodule OliWeb.Components.Delivery.Students do
 
   defp maybe_filter_by_card(students, :low_proficiency, filter_by) do
     Enum.filter(students, fn student ->
-      Metrics.proficiency_range(student.overall_proficiency) == "Low" ||
+      student.overall_proficiency == "Low" ||
```
eliknebel, line 234: This was a bit strange to me that `student.overall_proficiency` was being run through `Metrics.proficiency_range` again, when it should already be a value of. "Low", "Medium", "High", or "Not enough data" at this point. Further, the way this function is pattern matched, if an invalid value was being passed to `Metrics.proficiency_range`, it would simply return "High" which seems like a bug

lib/oli_web/components/delivery/students/students.ex
```diff
@@ -1292,8 +1292,8 @@ defmodule OliWeb.Components.Delivery.Students do
         end),
       low_proficiency:
         Enum.count(students, fn student ->
-          (Metrics.proficiency_range(student.overall_proficiency) == "Low" ||
-             is_nil(student.overall_proficiency)) and is_learner_selected(student, filter_by)
+          student.overall_proficiency == "Low" ||
+            (is_nil(student.overall_proficiency) and is_learner_selected(student, filter_by))
```
eliknebel, line 1296: I'm not sure why we are checking if `student.overall_proficiency` is nil here, I don't see any way in which it could be null since the values default to "Not enough data" on initialization. I tried to leave this logic in tact though because it seemed to break some unit tests and I also want to limit the changes here if possible

test/oli/delivery/metrics/learning_proficiency_test.exs
```diff
@@ -397,25 +404,25 @@ defmodule Oli.Delivery.Metrics.LearningProficiencyTest do
...
-      set_snapshot(section, page_3.resource, page_3_obj.resource, student_1, false)
-      set_snapshot(section, page_3.resource, page_3_obj.resource, student_2, true)
-      set_snapshot(section, page_3.resource, page_3_obj.resource, student_3, false)
-      set_snapshot(section, page_3.resource, page_3_obj.resource, student_4, false)
-
-      set_snapshot(section, page_4.resource, page_4_obj.resource, student_1, false)
-      set_snapshot(section, page_4.resource, page_4_obj.resource, student_2, true)
-      set_snapshot(section, page_4.resource, page_4_obj.resource, student_3, false)
-      set_snapshot(section, page_4.resource, page_4_obj.resource, student_4, false)
+      set_snapshots(section, page_1.resource, page_1_obj.resource, student_1, true)
```
eliknebel, line 407: In order to show any proficiency for v1, we might have at least 3 snapshot records so I changed this function being called to `set_snapshots` which will create 3 records of the same value to meet this new requirement

lib/oli/delivery/metrics.ex
eliknebel: Changed to use `sum` aggregation instead, this also fixed the intermittently broken unit test and is the more correct approach

rgachuhi (approved):

lib/oli_web/components/delivery/students/students.ex
nicocirio: Nice catch!

nicocirio (commented):
Nice work!

PR #5008, opened 2024-08-09
[BUG FIX] [MER-3572] Delivery student grace period pay banner

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-3572) to the ticket

https://github.com/user-attachments/assets/f858fa56-a074-40bb-935b-5b86578b23f0

### DIFF
--- a/test/oli_web/live/delivery/student/index_live_test.exs
+++ b/test/oli_web/live/delivery/student/index_live_test.exs
@@ -680,6 +680,42 @@ defmodule OliWeb.Delivery.Student.IndexLiveTest do
       assert has_element?(view, "div", "Upcoming Agenda")
     end
 
+    test "renders paywall message when grace period is not over (or gets redirected when over)",
+         %{
+           conn: conn,
+           section: section
+         } do
+      stub_current_time(~U[2024-10-15 20:00:00Z])
+
+      {:ok, product} =
+        Sections.update_section(section, %{
+          type: :blueprint,
+          registration_open: true,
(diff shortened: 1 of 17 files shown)


test/oli_web/live/delivery/student/index_live_test.exs
```diff
@@ -680,6 +680,42 @@ defmodule OliWeb.Delivery.Student.IndexLiveTest do
       assert has_element?(view, "div", "Upcoming Agenda")
     end
 
+    test "renders paywall message when grace period is not over (or gets redirected when over)",
+         %{
+           conn: conn,
+           section: section
+         } do
+      stub_current_time(~U[2024-10-15 20:00:00Z])
```
Francisco-Castro, line 688: Not a blocker but it would be great to check the boundaries of the grace period by doing something like:

```
dt = ~U[2024-10-15 20:00:00Z]
stub_current_time(dt)
grace_period_days = 18
...
...
# Grace period is over
stub_current_time(DateTime.add(dt, grace_period_days + 1, :day))
```

I tested this locally, and the test passed. However, when doing something like this:

```
passed_grace_period = dt 
  |> DateTime.add(grace_period_days, :day) 
  |> DateTime.add( 1, :second)

stub_current_time(passed_grace_period)
```

the test failed. 🤔

I'm guessing it's failing because of the way the grace period is computed.

Francisco-Castro (approved):
Great job! 🏅 

Nice job using mocking for the time.

PR #5010, opened 2024-08-09
[BUG FIX] [MER-3571] fixes explanation strategy ui fields bug

rgachuhi:
This PR fixes an issue where, when an author tries to set the Explanation Strategy for a page in authoring, the inputs disappear when a value is input.


Francisco-Castro (commented):
I'll add this review as a comment for now until I get an answer to the following question:

@rgachuhi I noticed that when selecting 'After Set Num Attempts,' the input for attempts can be left blank and is still considered valid. Is that the intended behavior, or should we set up a default numeric value?

Francisco-Castro (approved):

Francisco-Castro:
A new ticket will be created to address the edge case mentioned above.

PR #5022, opened 2024-08-14
[BUG FIX] [MER-3626] Fix exporting project breaking if revision is nil

manelli:
Apply fixes from https://github.com/Simon-Initiative/oli-torus/pull/4902 to `hotfix-v0.28.3` branch


nicocirio (approved):
🚀

PR #5026, opened 2024-08-15
[BUG FIX] [MER-3625] duplicated pages retain previous revision id to the parent resource

eliknebel:
Fixes an issue where duplicated pages retain the previous revision id by setting this field to nil on duplication. This PR also addresses the fact that there are instances of this out in the wild and the revision history tool is now more resilient to handling these cases


nicocirio (approved):
Nice work!

PR #5027, opened 2024-08-15
[BUG FIX] [MER-3630] ensure all required non hierarchical resource records are created when remixing

eliknebel:
https://eliterate.atlassian.net/browse/MER-3630

Fixes an issue where delivery resolver was not properly including all required resources remixed into a section, such as objectives and nested activities. This was resulting in resolver failures in delivery, cause 500 errors for students.

The fix here is to revert commit https://github.com/Simon-Initiative/oli-torus/commit/32498051478aaab4ccd54d27525b2a10899cc336

Once deployed, an admin will have to take the additional step of "remixing" the affected course section(s). It is sufficient to drag/move an item around, then move it back and click "Save" which will trigger the required rebuild_section_resources step and resolve the issue.

Finally, this PR reverts the changes made in https://eliterate.atlassian.net/browse/MER-3624 since they are no longer necessary.

### DIFF
--- a/lib/oli/delivery/page/objectives_rollup.ex
+++ b/lib/oli/delivery/page/objectives_rollup.ex
@@ -20,9 +20,7 @@ defmodule Oli.Delivery.Page.ObjectivesRollup do
   end
 
   defp rollup(attached_objective_ids, resolver, section_slug) do
-    all =
-      resolver.from_resource_id(section_slug, attached_objective_ids)
-      |> Enum.reject(&is_nil(&1))
+    all = resolver.from_resource_id(section_slug, attached_objective_ids)
 
     parents = resolver.find_parent_objectives(section_slug, attached_objective_ids)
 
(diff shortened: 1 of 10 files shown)


lib/oli/delivery/page/objectives_rollup.ex
```diff
@@ -20,9 +20,7 @@ defmodule Oli.Delivery.Page.ObjectivesRollup do
   end
 
   defp rollup(attached_objective_ids, resolver, section_slug) do
-    all =
-      resolver.from_resource_id(section_slug, attached_objective_ids)
-      |> Enum.reject(&is_nil(&1))
+    all = resolver.from_resource_id(section_slug, attached_objective_ids)
```
eliknebel, line 23: Revert https://eliterate.atlassian.net/browse/MER-3624

darrensiegel (approved):

PR #5029, opened 2024-08-16
[BUG FIX] [MER-3631] Fix rendering issues with cognito/open and free routes

eliknebel:
https://eliterate.atlassian.net/browse/MER-3631

Fixes a rendering issue with cognito workflow in the open and free routes, resulting in 500 error. I added some basic test coverage here around these routes as well.

I suspect some of the `resources` routes that exist for `/sections/independent` such as `:show` and `:edit` are now obsolete with the new system workflows, as section viewing/editing now takes place from the instructor dashboard live view. I left these in for now but we likely want to remove them in the future.

### DIFF
--- a/lib/oli_web/templates/open_and_free/form.html.eex
+++ b/lib/oli_web/templates/open_and_free/form.html.eex
@@ -6,7 +6,7 @@
   <% end %>
 
   <% # Only show source if creating a new section (not editing) %>
-  <%= @action == Routes.independent_sections_path(@conn, :create) do %>
+  <%= if @action == Routes.independent_sections_path(@conn, :create) do %>
     <%= hidden_input f, :source_id, value: @source_id %>
 
     <div class="form-label-group">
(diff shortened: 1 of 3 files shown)


lib/oli_web/templates/open_and_free/form.html.eex
```diff
@@ -6,7 +6,7 @@
   <% end %>
 
   <% # Only show source if creating a new section (not editing) %>
-  <%= @action == Routes.independent_sections_path(@conn, :create) do %>
+  <%= if @action == Routes.independent_sections_path(@conn, :create) do %>
```
eliknebel, line 9: Missing `if` was the root cause of the rendering error

lib/oli_web/templates/open_and_free/show.html.eex
```diff
@@ -21,7 +21,7 @@
       </tr>
       <tr>
         <td><strong>URL:</strong></td>
-        <td><%= external_link Routes.page_delivery_url(@conn, :index, @section.slug), to: Routes.page_delivery_url(@conn, :index, @section.slug) %></td>
+        <td><%= url(~p"/sections/#{@section.slug}") %></td>
```
eliknebel, line 24: This page_delivery route no longer exists so I updated it to the new route, even though this `show` view likely isn't being used anymore anyway

nicocirio (approved):
LGTM. I tested it locally and I could create a section as expected

PR #5033, opened 2024-08-19
[BUG FIX] [MER-3631] Fix rendering issues with cognito/open and free routes (Re-targeted to v0.28.4)

eliknebel:
**Re-targeted to v0.28.4**

https://eliterate.atlassian.net/browse/MER-3631

Fixes a rendering issue with cognito workflow in the open and free routes, resulting in 500 error. I added some basic test coverage here around these routes as well.

I suspect some of the resources routes that exist for /sections/independent such as :show and :edit are now obsolete with the new system workflows, as section viewing/editing now takes place from the instructor dashboard live view. I left these in for now but we likely want to remove them in the future.

### DIFF
--- a/lib/oli_web/templates/open_and_free/form.html.eex
+++ b/lib/oli_web/templates/open_and_free/form.html.eex
@@ -6,7 +6,7 @@
   <% end %>
 
   <% # Only show source if creating a new section (not editing) %>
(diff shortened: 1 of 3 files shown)


nicocirio (approved):
LGTM. This is the same fix previously implemented on [this PR](https://github.com/Simon-Initiative/oli-torus/pull/5029), but retargeted to v0.28.4

PR #5040, opened 2024-08-21
[BUG FIX] [MER-3670] Every page is loaded twice

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-3670) to the ticket

When a liveview is mounted the `mount` callback is called twice; one time when the HTTP request is received and then when the websocket connection is established (see [docs](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.html#module-life-cycle)).
So, to avoid triggering the access track twice, I adapted the logic to do it when the HTTP request is first received.

### Before

https://github.com/user-attachments/assets/091cae4c-3b2c-4419-a2ca-3fabdcd0d839

### After

https://github.com/user-attachments/assets/e0baa8ca-8478-4b64-834f-c1ceb745658f

While working on the ticket I found that we were having a similar issue with the XAPI Events being sent twice. To fix this I wrapped the execution of that function in the `connected?(socket)` conditional.

### DIFF
--- a/lib/oli/delivery/page/page_context.ex
+++ b/lib/oli/delivery/page/page_context.ex
@@ -166,12 +166,16 @@ defmodule Oli.Delivery.Page.PageContext do
   and activities that may be present in the content of the page. This
   information is collected and then assembled in a fashion that can be given
   to a renderer.
+
+  The `opts` parameter is a keyword list that can contain the following options:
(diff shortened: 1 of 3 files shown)


eliknebel (approved):
Looks good! Can we retarget to v0.28.4?

PR #5044, opened 2024-08-22
[BUG FIX] [MER-3671] Adaptive page styles broken after lesson publishing - part 2

nicocirio:
[Link to the comment](https://eliterate.atlassian.net/browse/MER-3671?focusedCommentId=24029) in the ticket

#### Before

https://github.com/user-attachments/assets/8d2d38f3-046b-4a63-bb39-f6852fa19e22

#### After
https://github.com/user-attachments/assets/3e1bb083-26fa-441b-b628-bc9b72ca1fd3

The issue was identified after editing and publishing a page. So, when that event occurs, all active attempts of the edited page change their state from `:in_progress` to `revised`.
Since the `RedirectAdaptiveChromeless` livesession plug was not considering `:revised` states to perform the redirect, the user ended up in the `/lesson` route instead of the `/adaptive_lesson` one. Both layout templates for those routes are very similar, but the first one includes a modal, and that hidden modal was the one preventing the click from being correctly targeted to the dropdown.

The same fix was extended to the `RedirectToLesson` livesession plug.

All other cases were already addressed in [this PR](https://github.com/Simon-Initiative/oli-torus/pull/5036)

### DIFF
--- a/lib/oli_web/live_session_plugs/redirect_adaptive.ex
+++ b/lib/oli_web/live_session_plugs/redirect_adaptive.ex
@@ -53,8 +53,9 @@ defmodule OliWeb.LiveSessionPlugs.RedirectAdaptiveChromeless do
             "displayApplicationChrome" => false
           }
         },
-        progress_state: :in_progress
-      } ->
+        progress_state: progress_state
+      }
+      when progress_state in [:revised, :in_progress] ->
         {:halt,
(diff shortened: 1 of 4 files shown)


simonchoxx (approved):
LGTM 🚀

eliknebel (approved):

PR #5045, opened 2024-08-22
[BUG FIX] [MER-3669] date groups do not match the dates rendered on the page

eliknebel:
https://eliterate.atlassian.net/browse/MER-3669

Fixes an issue where date groupings were not being localized causing inconsistent due dates.

When grouping pages and assignments by due dates in the module index, the grouping was being determined by a non-localized datetime -> date conversion, losing the timezone information in the process. This resulted in some dates rendering off by a day, effectively rendering the UTC date, when actually it should have matched the other dates being displayed, rendered in the localized timezone.

The fix here is to localize the datetime before converting to naive date for grouping such that date groupings accurately match the actual dates being rendered on the page.

### DIFF
--- a/lib/oli_web/live/delivery/student/learn_live.ex
+++ b/lib/oli_web/live/delivery/student/learn_live.ex
@@ -1391,7 +1391,7 @@ defmodule OliWeb.Delivery.Student.LearnLive do
   attr :module, :map
   attr :student_raw_avg_score_per_page_id, :map
   attr :student_end_date_exceptions_per_resource_id, :map
-  attr :ctx, :map
+  attr :ctx, :map, required: true
   attr :student_id, :integer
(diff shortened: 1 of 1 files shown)


gastonabella (approved):
LGTM

lib/oli_web/live/delivery/student/learn_live.ex
```diff
@@ -2490,14 +2505,21 @@ defmodule OliWeb.Delivery.Student.LearnLive do
...
+          show_completed_pages,
+          ctx
         )
 
       _ ->
         []
     end)
   end
 
+  defp to_localized_date(datetime, ctx) do
```
manelli, line 2517: Maybe we can use the existing `OliWeb.Common.FormatDateTime.maybe_localized_datetime/2` function?

PR #5054, opened 2024-08-26
[ENHANCEMENT] [MER-3672] Page context is loaded twice

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-3672) to the ticket

- [x] early redirect -> avoid passing through many plugs and live_session plugs to end up being redirected.
- [x] ~separate previous-next index from the need of having `page_context` already calculated~ => this was not necessary since we render it when socket is already connected
- [x] create a new `prologe_context` for the prologue (needs less info than the `page_context` one)
- [x] when accessing a page (graded or practice) avoid calculating the page_context twice ( mount callback )
- [x] handle adaptive pages with controller instead of liveview: assigning the `page_context` when the websocket connection is established allowed me to render the content of the page, but it crashes when interacting with it (I understand it has to do with the `OLI.initActivityBridge` wiring and with when the activity scripts needed are loaded)
- [x] update tests: about 60 tests were updated, since now we have to "wait" for the websocket to be connected and the scripts loaded to make assertions. On the other hand, some tests now need to be handled by a common HTTP get request instead of a liveview get (after migrating the adaptive pages to a controller approach) 
- [x] fix rendering of maths expressions: the MathJax script seems to have an issue

### "Homemade" Benchmark
I navigated through some pages of the Chemistry course to count how many queries were executed while accessing that page (by counting the amount of `QUERY OK` printed in the console). This may not be a robust benchmark, but it will provide us with a sense of the impact of this enhancement.

image

There is more room to continue reducing the queries (or even improve the ones that are the most expensive ones in terms of execution time). This could be addressed with better insights that this [AppSignal PR](https://github.com/Simon-Initiative/oli-torus/pull/5067) will provide us.


darrensiegel:
This all looks good, I pushed one commit to move a bit more code into the "connected?" block, which saves another query. 

Tested locally and works well.

darrensiegel (approved):

PR #5055, opened 2024-08-26
[BUG FIX] [MER-3686] Student index_live crashes when lesson end date is not set (nil)

eliknebel:
This PR handles case where lesson end date might be nil when the start by date is set to some point in the future. The results was a 500 crash on the main student index page.

### DIFF
--- a/lib/oli_web/live/delivery/student/index_live.ex
+++ b/lib/oli_web/live/delivery/student/index_live.ex
@@ -522,7 +522,11 @@ defmodule OliWeb.Delivery.Student.IndexLive do
       <div class="pr-2 pl-1 self-end">
         <div class="flex items-end gap-1">
           <div class="text-right dark:text-white text-opacity-90 text-xs font-semibold">
-            <%= Utils.days_difference(@lesson.end_date, @ctx) %>
+            <%= case @lesson do %>
+              <% %{end_date: nil} -> %>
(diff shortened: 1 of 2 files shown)


lib/oli_web/live/delivery/student/index_live.ex
```diff
@@ -522,7 +522,11 @@ defmodule OliWeb.Delivery.Student.IndexLive do
       <div class="pr-2 pl-1 self-end">
         <div class="flex items-end gap-1">
           <div class="text-right dark:text-white text-opacity-90 text-xs font-semibold">
-            <%= Utils.days_difference(@lesson.end_date, @ctx) %>
+            <%= case @lesson do %>
+              <% %{end_date: nil} -> %>
+              <% %{end_date: end_date} -> %>
+                <%= Utils.days_difference(end_date, @ctx) %>
```
gastonabella, line 528: This function is also being used in the [ScheduleComponent](https://github.com/Simon-Initiative/oli-torus/blob/hotfix-v0.28.4/lib/oli_web/live/delivery/student/home/components/schedule_component.ex#L292), so we should fix it there too.
We could also make the fix only in one place, and update the function to return `nil` if the `end_date` is `nil`

eliknebel: Good catch!

darrensiegel (approved):

PR #5061, opened 2024-08-28
[FEATURE] [MER-3596] Migrate authoring bibliography live view to new workspace layout and live session

Francisco-Castro:
Ticket: [MER-3596](https://eliterate.atlassian.net/browse/MER-3596)

This PR migrated the authoring bibliography page to LiveView for use as part of the course author space.

[MER-3596]: https://eliterate.atlassian.net/browse/MER-3596?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli_web/live/workspaces/course_author/bibliography_live.ex
+++ b/lib/oli_web/live/workspaces/course_author/bibliography_live.ex
@@ -1,30 +1,64 @@
 defmodule OliWeb.Workspaces.CourseAuthor.BibliographyLive do
   use OliWeb, :live_view
 
+  alias Oli.Accounts
+  alias OliWeb.Common.React
+
+  on_mount {OliWeb.LiveSessionPlugs.AuthorizeProject, :default}
+
   @impl Phoenix.LiveView
   def mount(_params, _session, socket) do
     project = socket.assigns.project
+    author = socket.assigns.current_author
+    is_admin? = Accounts.has_admin_role?(author)
+    ctx = socket.assigns.ctx
 
-    {:ok,
-     assign(socket,
-       resource_slug: project.slug,
-       resource_title: project.title,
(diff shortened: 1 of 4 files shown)


lib/oli_web/live/workspaces/course_author/bibliography_live.ex
```diff
@@ -1,30 +1,64 @@
 defmodule OliWeb.Workspaces.CourseAuthor.BibliographyLive do
   use OliWeb, :live_view
 
+  alias Oli.Accounts
+  alias OliWeb.Common.React
+
+  on_mount {OliWeb.LiveSessionPlugs.AuthorizeProject, :default}
```
eliknebel, line 7: This on_mount seems general enough that it could maybe live in the router as part of the live_session?

Francisco-Castro: Good observation @eliknebel. That `on_mount` is only necessary when working with a project, but the instructor is also in the scope of the workspace's **live_session**, as shown in the snippet below. I'm working on a fix for that in PR https://github.com/Simon-Initiative/oli-torus/pull/5069 where we will have 2 live_sessions: one for `course_author` and another one for the `instructor`. Would you like me to include it here, or should I wait until the fix is complete?

https://github.com/Simon-Initiative/oli-torus/blob/85ad23911ef9f8c31b3a1aec043f6d04f60e9305/lib/oli_web/router.ex#L789-L827

eliknebel: We can handle it in the follow-up PR

eliknebel (approved):

PR #5063, opened 2024-08-28
[FEATURE] [MER-3600] Migrate Review liveview

simonchoxx:
[MER-3600](https://eliterate.atlassian.net/browse/MER-3600)

This PR migrates the `Review` liveview to be accessible from the new menu for the course author workspace.

In addition to the liveview, the necessary components used by the main liveview are migrated, as well as the existing logic tests.

https://github.com/user-attachments/assets/6d5935e7-3901-4860-9216-b8c84670b800

[MER-3600]: https://eliterate.atlassian.net/browse/MER-3600?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli_web/live/workspaces/course_author/qa/state.ex
+++ b/lib/oli_web/live/workspaces/course_author/qa/state.ex
@@ -1,6 +1,5 @@
 defmodule OliWeb.Workspaces.CourseAuthor.Qa.State do
   alias OliWeb.Common.SessionContext
-  alias OliWeb.Common.Breadcrumb
 
   @default_filters MapSet.new(["pedagogy", "content", "accessibility", "equity"])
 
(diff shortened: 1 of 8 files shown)


eliknebel:
Screenshot 2024-08-29 at 4 49 01 PM

While trying to test out these changes, I am seeing this error message, I'm not sure if its related or not

simonchoxx:
@eliknebel I have not received such an error message while working on the ticket. As far as I can see the error is occurring in the sidebar component and I don't think it would have anything to do with the liveview I worked on. 
Have you tried more than one project?

simonchoxx:
@eliknebel We did some research with @nicocirio and @Francisco-Castro about this error and it seems we have a bug that we need to fix in that `set_sidebar` plug. 
Maybe we could address this error in another ticket. 
I think that by deleting the cookies and restarting the server you should stop seeing the error.

eliknebel (approved):

eliknebel:
Filed a followup ticket for the set_sidebar issue. https://eliterate.atlassian.net/browse/MER-3715

PR #5066, opened 2024-08-29
[BUG FIX] [MER-3711] Style tables within feedback to ensure legibility

andersweinstein:
The default tailwind table styles were leading to illegible results for tables in the feedback context, particularly in dark mode. This adds custom styles to tables within feedback to ensure visibility of borders and text against the colored feedback background, and also ensure no dark mode changes (since feedback background/text color doesn't change). 

In theory the table text and borders should use css variables to render with the same color as text in the feedback context. Currently different css color variables are defined in `tailwind.theme.js` that would allow for potentially different text colors in correct, incorrect, and explanation feedback. Keeping to that model would require different table styles for each of these contexts as well. But currently these text colors are all just black. So decided this is excess generality we will likely never need and kept it simple, with just one variable for feedback-table-color. In any case the simple method is no worse than previous code which had no styling, just inheriting tailwind defaults.

### DIFF
--- a/assets/styles/delivery/activity.scss
+++ b/assets/styles/delivery/activity.scss
@@ -112,6 +112,12 @@
         margin-right: 3px;
       }
     }
+
+    // override tailwind styles to ensure borders, text visible against feedback background & no dark mode change
+    table {
+      color: black;
+      border-color: black;
+    }
   }
 }
 
@@ -162,8 +168,6 @@
     /* MER-2961 - Some wide content (specifically formulas) were getting cut off in hints, making it scrollable
     */
     overflow-x: auto;
-
-
   }
 
   p:last-of-type {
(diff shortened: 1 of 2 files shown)


andersweinstein:
So simple method has a problem. In light mode only, the table winds up appearing like this:
Screenshot 2024-08-29 at 3 41 43 PM

This apparently occurs because of the way tailwind implements borders in the "pre-flight" base table styles as hidden (zero-width) ::before or :after elements which get shown by use of tailwind border class. See
https://tailwindcss.com/docs/preflight:

```
In order to make it easy to add a border by simply adding the border class, Tailwind overrides the default border styles for all elements with the following rules:

*,
::before,
::after {
  border-width: 0;
  border-style: solid;
  border-color: theme('borderColor.DEFAULT', currentColor);
}
```
So a different method has to be used.

rgachuhi (approved):

PR #5069, opened 2024-09-02
[FEATURE] [MER-3602] Migrate authoring Products LiveView to new workspace layout and LiveSession

Francisco-Castro:
Ticket: [MER-3602](https://eliterate.atlassian.net/browse/MER-3602)

This PR migrates our product pages to the course_author workspace.

Note 1: [This line](https://github.com/Simon-Initiative/oli-torus/blob/f23eb45602f79b5c92f108564af50b6d21b5d040/lib/oli_web/templates/layout/delivery.html.heex#L159) triggers an additional loop through the `live_session`, resulting in three iterations instead of the expected two from a LiveView page. It looks like this can be avoided by using a function component and a plug to manage the handle_info callbacks.

Note 2: The way the sidebar works is by modifying the URL with a parameter called sidebar_expanded. To get the URL path, we relied on a series of atoms assigned to each page as part of the LiveView migration. This has worked fine for cases with a single level, but products have nested pages, which doesn't work well with our current approach. To address this and make the sidebar functional, this PR parses the URL and finds the correct path. As a result, we now have a hybrid approach, and it’s likely that we will use the URL path as the source of truth, allowing us to eventually get rid of the atoms in each LiveView page. See the following commit: https://github.com/Simon-Initiative/oli-torus/pull/5069/commits/3db95cc8a405467b9345598405011e441b2d352a

Note 3: [This commit](https://github.com/Simon-Initiative/oli-torus/pull/5069/commits/54d4dbd3f10194edd786da5f204eee4d9385ddf4) addresses the [comment](https://github.com/Simon-Initiative/oli-torus/pull/5061#discussion_r1742676361) made by @eliknebel:

> This on_mount seems general enough that it could maybe live in the router as part of the live_session?

Note 4: This PR fixes a module that was duplicated and incorrectly named: https://github.com/Simon-Initiative/oli-torus/blob/f23eb45602f79b5c92f108564af50b6d21b5d040/test/oli_web/live/workspace/course_author/bibliography_test.exs#L1

Note 5: [This commit](https://github.com/Simon-Initiative/oli-torus/pull/5069/commits/16a297ea82b01b133541740bfffcc83f09baad7c) fixes the name of a folder that needed to be pluralized to align with our folder structure consistency.

https://github.com/user-attachments/assets/46652abf-fc96-46db-ab87-c06207ea9b98

[MER-3602]: https://eliterate.atlassian.net/browse/MER-3602?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/oli_web/router.ex
```diff
@@ -789,15 +789,15 @@ defmodule OliWeb.Router do
   scope "/workspaces/", OliWeb.Workspaces do
     pipe_through([:browser, :authoring_and_delivery, :set_sidebar])
 
-    live_session :workspaces,
+    live_session :course_author_workspace,
```
eliknebel, line 792: Since this session contains instructor and student views, I'm not sure it makes sense to rename it
to course_author_workspace

Francisco-Castro: Good catch! The `:course_author_workspace` name was a leftover from splitting the workspaces. Now that all workspaces are combined, it makes sense to revert back to `:workspaces`.

eliknebel (approved):
Nice work, tested and works as expected

PR #5071, opened 2024-09-03
[BUG FIX] [MER-3716] Ensure same promise is used for all MathJax typesetting

darrensiegel:
28.6 hotfix added a new spot of MathJAX typesetting, but didn't honor the requirement that the same global promise is used (as was fixed in https://eliterate.atlassian.net/browse/MER-1782?atlOrigin=eyJpIjoiYjdlOGFmMDhhYTRhNGNlMzk4MjMxZTNmY2E5YTAzNzQiLCJwIjoiamlyYS1zbGFjay1pbnQifQ)

This PR applies the same logic in this new phoenix liveview hook

### DIFF
--- a/assets/src/hooks/evaluate_mathjax_expressions.ts
+++ b/assets/src/hooks/evaluate_mathjax_expressions.ts
@@ -5,6 +5,27 @@ export const EvaluateMathJaxExpressions = {
   mounted() {
     const elements = document.querySelectorAll('.formula');
 
-    window.MathJax.typesetPromise(Array.from(elements) as HTMLElement[]);
+    const getGlobalLastPromise = () => {
+      /* istanbul ignore next */
+      let lastPromise = window?.MathJax?.startup?.promise;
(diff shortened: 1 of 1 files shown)


andersweinstein:
Maybe good to break out that common code into a single utility module, like MathJaxUtils or whatever? I mean GetGlobalLastPromise, setGlobalLastPromise which are duplicated. Could also add a common routine to do the three-line typesetting, like:
   ```
 function safeTypesetMathJax(nodes: HTMLElement[] ) {
    let lastPromise = getGlobalLastPromise();
    lastPromise = lastPromise.then(() =>
      window.MathJax.typesetPromise(nodes),
    );
    setGlobalLastPromise(lastPromise)
   }
```

andersweinstein:
The hotfix also lost typesetting of inline formulas because it does not look for elements of class `formula-inline`

eliknebel (approved):

PR #5074, opened 2024-09-04
[ENHANCEMENT] [MER-3597] Curriculum workspace migration links backwards compatibility

manelli:
Ensure that when accessing curriculum links to sections from the old-style view, we link to the old-style section/containers and not the ones in the course author workspace

See: https://eliterate.atlassian.net/browse/MER-3597

### DIFF
--- a/lib/oli_web/live/workspaces/course_author/curriculum/entry.ex
+++ b/lib/oli_web/live/workspaces/course_author/curriculum/entry.ex
@@ -1,4 +1,4 @@
-defmodule OliWeb.Curriculum.Entry do
+defmodule OliWeb.Workspaces.CourseAuthor.Curriculum.Entry do
   @moduledoc """
   Curriculum item entry component.
   """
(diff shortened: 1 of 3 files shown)


lib/oli_web/live/workspaces/course_author/curriculum/entry.ex
nicocirio, line 1: I think we do not need this component for the fix

lib/oli_web/live/workspaces/course_author/curriculum_live.ex
```diff
@@ -12,13 +12,14 @@ defmodule OliWeb.Workspaces.CourseAuthor.CurriculumLive do
...
     DropTarget,
-    Entry,
     OptionsModalContent,
     DeleteModal,
     NotEmptyModal,
     HyperlinkDependencyModal
   }
 
+  alias OliWeb.Workspaces.CourseAuthor.Curriculum.Entry
+
```
nicocirio, line 22: If we do not need the component, we should then remove this alias

nicocirio (commented):
I tested it locally and it worked as expected.
However, I am unsure if we need the `Entry` component for this fix. Maybe it is necessary for the next ticket?

lib/oli_web/live/workspaces/course_author/curriculum/entry.ex
manelli: I added a separate Entry component (different from the one in the old UI) because it will be needed for the next parts of the migration.
Without this new Entry, the current fix does not work

nicocirio (approved):

PR #5076, opened 2024-09-05
[BUG FIX] [MER-3336] Make Adaptive Pages work with MER-3000/MER-3153

dtiwarATS:
Hey @bsparks, could you please look at this PR? It's a big one which is changing the core of how the Adaptive delivery works when saving the student attempt. 

We used to save the attempt data to the owner activity of the part but now they will be saved with the current active screen instead of their respective owner.

There is a possibility of an active attempt where some screen data is stored in older format but when they resume the lesson, the new data will be stored in the new approach. To identify these lessons, I added 2 session variables to handle that.

Also, there will be existing attempts that will be opened for review mode, so I have handled that as well. 

I have added comment in code so that we know why it was written.

Hey @eliknebel / @darrensiegel, we also need to be merged this https://github.com/Simon-Initiative/oli-torus/pull/4765 in this branch. This is the PR that locks out the attempt save if it's already submitted. We reverted it back then using this https://github.com/Simon-Initiative/oli-torus/pull/4813


eliknebel:
@dtiwarATS I reintegrated the changes from https://github.com/Simon-Initiative/oli-torus/pull/4765 into this base branch [MER-3336-Make-Adaptive-Pages-work-with-MER-3000-MER-3153](https://github.com/Simon-Initiative/oli-torus/tree/MER-3336-Make-Adaptive-Pages-work-with-MER-3000-MER-3153). However, it appears that there are 2 tests that were skipped in `test/oli_web/controllers/api/attempt_controller_test.exs` when this was reverted, that are now failing when the skip tag is removed.

Since @simonchoxx was the original contributor here, can you work with him to figure out why these tests are failing and get them back into a passing state please?

dtiwarATS:
I have reached out to @simonchoxx. Once he is back from his vacation, he will look at it.

simonchoxx:
@dtiwarATS I was checking some PR's (#4765 and #4813) and those tests were skipped because we did a rollback of the implementation. I think that as the elixir code has not changed, it would not be necessary to keep them.  
What do you think about removing them @eliknebel?

eliknebel:
> @dtiwarATS I was checking some PR's (#4765 and #4813) and those tests were skipped because we did a rollback of the implementation. I think that as the elixir code has not changed, it would not be necessary to keep them. What do you think about removing them @eliknebel?

Yes, we should be good to just remove them, thanks!

simonchoxx:
Great! Then it would be enough for @dtiwarATS to delete the entire  block [describe](https://github.com/Simon-Initiative/oli-torus/blob/25f6d672a078b3b08e771aed6f59232affcb2cde/test/oli_web/controllers/api/attempt_controller_test.exs#L80) containing the tests that have been skipped in the `attempt_controller_test.exs` file. Thanks @eliknebel!

dtiwarATS:
Hey @eliknebel, Since I don't have write access on Torus repo, I have created another PR for removing the test - https://github.com/Simon-Initiative/oli-torus/pull/5091

PR #5093, opened 2024-09-12
[CHORE] [MER-3786] Allow setting of the interval and target for db_connection

darrensiegel:
This PR exposes the DBConnection library (which `Ecto` uses) config values for `queue_target` and `queue_interval`. 

The default value of `50` (milliseconds) for `queue_target` is how long a request for a connection from the pool will queue before it times out and throws the dreaded error: 

```
** (DBConnection.ConnectionError) connection not available and request was dropped from queue after 101ms. This means requests are coming in and your connection pool cannot serve them fast enough. You can address this by:

  1. Ensuring your database is available and that you can connect to it
  2. Tracking down slow queries and making sure they are running fast enough
  3. Increasing the pool_size (although this increases resource consumption)
  4. Allowing requests to wait longer by increasing :queue_target and :queue_interval
```

That limit will get doubled (to 100ms) if all connections in a period of `queue_interval` take longer than 50ms. 

We can avoid these connection errors by simply increasing the `queue_target` up from the default of `50` to something like `5000` (5 seconds).  This will simply queue requests a lot longer when the DBConnection pool is full.  

This, in conjunction with increase pool size and in conjunction with tracking down long running queries (ongoing work) and streamlining specific endpoints (I'm looking at you, `OliWeb.Delivery.Student.IndexLive`) will give us much stronger guarantees against these DBConnection errors, increase overall throughput, etc., etc.

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -61,6 +61,8 @@ if config_env() == :prod do
     database: System.get_env("DB_NAME", "oli"),
     pool_size: String.to_integer(System.get_env("POOL_SIZE") || "10"),
     timeout: db_timeout,
+    queue_target: String.to_integer(System.get_env("DB_QUEUE_TARGET") || "50"),
+    queue_interval: String.to_integer(System.get_env("DB_QUEUE_INTERVAL") || "1000"),
     ownership_timeout: 600_000,
     socket_options: maybe_ipv6
 


nicocirio:
> I'm looking at you, `OliWeb.Delivery.Student.IndexLive`

🤣 🤣

PR #5099, opened 2024-09-12
[BUG FIX] [MER-3797] Fix Youtube XAPI emitting 

darrensiegel:
When a Youtube video was present in the page (as opposed to being in an activity) it was NOT correctly emitting XAPI events from the front end for play, pause, stop, etc.  It was missing the page attempt guid, which broke the bundle creation on the server and prevented the XAPI bundle from going to S3. 

This PR fixes the problem by mirroring the solution present for HTML5 video (which does work), namely to simply specify and pass down the attempt guid as a React param.

### DIFF
--- a/lib/oli_web/controllers/api/xapi_controller.ex
+++ b/lib/oli_web/controllers/api/xapi_controller.ex
@@ -22,6 +22,8 @@ defmodule OliWeb.Api.XAPIController do
 
       {:error, e} ->
         Logger.error("Error constructing xapi bundle: #{inspect(e)}")
+        Oli.Utils.Appsignal.capture_error("Error constructing xapi bundle: #{inspect(e)}")
+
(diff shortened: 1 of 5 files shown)


lib/oli_web/controllers/api/xapi_controller.ex
```diff
@@ -22,6 +22,8 @@ defmodule OliWeb.Api.XAPIController do
 
       {:error, e} ->
         Logger.error("Error constructing xapi bundle: #{inspect(e)}")
+        Oli.Utils.Appsignal.capture_error("Error constructing xapi bundle: #{inspect(e)}")
```
eliknebel, line 25: It seems like there should be some way to notify AppSignal any time Logger.error is called, so that we don't need to explicitly call it separately. But that is something we can maybe explore in a non-hotfix release

eliknebel (approved):

PR #5103, opened 2024-09-16
[BUG FIX] [MER-3826] Update numbering index and level in SectionResource records during major update

darrensiegel:
Adds in a missing step of updating the `numbering_index` and `numbering_level` attributes of the section resource records, in the case of a `:major` update.  

This missing step is what was causing a number of issues specified in https://eliterate.atlassian.net/browse/TRIAGE-998

### DIFF
--- a/lib/oli/delivery/sections/updates.ex
+++ b/lib/oli/delivery/sections/updates.ex
@@ -10,6 +10,8 @@ defmodule Oli.Delivery.Sections.Updates do
   alias Oli.Publishing.Publications.PublicationDiff
   alias Oli.Delivery.Updates.Broadcaster
   alias Oli.Delivery.Sections.PostProcessing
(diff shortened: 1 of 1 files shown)


lib/oli/delivery/sections/updates.ex
```diff
@@ -268,12 +270,75 @@ defmodule Oli.Delivery.Sections.Updates do
         "perform_update.MAJOR: section[#{section.slug}] #{Oli.Timing.elapsed(mark) / 1000 / 1000}ms"
       )
 
+      renumber_hierarchy(section)
```
gastonabella, line 273: Should we add this function call to the `with` clause above? It performs a DB query and might return 
```elixir 
{:ok, query_result} | {:error, Exception.t()}
```

eliknebel (approved):

PR #5104, opened 2024-09-16
[CHORE] [MER-3717] Load polyfill script from cloudflare cdn 

andersweinstein:
[Ticket](https://eliterate.atlassian.net/browse/MER-3717)
This changes the URLs from which polyfill.js script is loaded from the now insecure polyfill.io to the cloudflare cdn, which implements a compatible service.

### DIFF
--- a/lib/oli_web/templates/layout/chromeless.html.heex
+++ b/lib/oli_web/templates/layout/chromeless.html.heex
@@ -96,7 +96,7 @@
     <% end %>
 
     <%= unless Map.get(assigns, :vr_agent_active, false) do %>
-      <script src="https://polyfill.io/v3/polyfill.min.js?features=es6">
+      <script src="https://cdnjs.cloudflare.com/polyfill/v3/polyfill.min.js?features=es6">
       </script>
     <% end %>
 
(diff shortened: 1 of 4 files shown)


ChristianMurphy (commented):
I would highly recommend not using _public_ CDN at all.
Torus is also used in the EU, including this would require a new subprocessor agreement for all adopters there, regardless of if they use CloudFlare otherwise or not. https://wideangle.co/blog/content-delivery-network-cdn-and-gdpr

In addition with improvements to site isolation built into browsers, public CDN's don't offer as much benefit in the first place https://httptoolkit.com/blog/public-cdn-risks/

Consider installing any polyfills needed through NPM and bundling/serving them along side Torus.

_edit_ 📓  Note: Private CDNs in front of apps are still fine, those are usually procured directly by adopters, and have better defined agreements.

eliknebel:
Thanks for the review @ChristianMurphy. We will investigate alternative approaches to using public CDNs.

For now, this will suffice as a short term fix intended to address the immediate issue of using the potentially insecure polyfill.io library

eliknebel (approved):

PR #5108, opened 2024-09-17
[BUG FIX] [MER-3619] Fix an issue where account creation silently fails in certain cases

eliknebel:
https://eliterate.atlassian.net/browse/MER-3619

This PR fixes 2 issues related to LTI and direct delivery accounts being intermingled if they have the same email address.

The issue is that the system is not properly constraining the existing account checks to `independent_learner: true` in 2 important places in the code:
- `lib/oli_web/pow/user_context.ex`
- `lib/oli_web/pow/user_identities_context.ex`

This is preventing torus from creating a new account when one already exists in the LTI domain with the same email.

~~Further, my suspicion is that the callback to user_identities_context is being triggered during an LTI launch. This callback pattern matches on “email_verified: true” assuming that only social login providers can meet that condition, however on LTI launch this email_verified is explicity set to true before the Pow session is created, effectively triggering this callback codepath for every LTI launch in. Furthermore, when this condition is met, an upsert is performed with the LTI details, which effectively converts the direct delivery user to an LTI user.~~ I verified this is not the case. However, I have left in the change here to ensure this function only runs for independent learners, as we never want a user to be able to add a social login capability to their existing LMS account. It could be that this was part of the confusion with the one-off account mentioned in the ticket.

I also updated the email template here to be better client compatible (tailwind is not available in email renderers) and match the existing email styles. This email message especially should match the rest of the email styles to not help assure a user that the email is valid.

One other note to make here: I noticed that the identification of a unique LTI account relies solely on the uniqueness of a given "sub" field. In most (if not all) cases, this is a UUID supplied by an LMS. However, because this is technically  outside of our control, it would be better to constrain this `insert_or_update_lms_user` function to also only consider users for a given institution. I didn't make this change here though, as it seemed a bit more risky. I filed a followup ticket https://eliterate.atlassian.net/browse/MER-3824

### DIFF
--- a/lib/oli_web/pow/messages.ex
+++ b/lib/oli_web/pow/messages.ex
@@ -58,17 +58,5 @@ defmodule OliWeb.Pow.Messages do
   end
 
   def pow_assent_login_with_provider(conn),
-    do:
-      interpolate("Continue with %{provider}", provider: Naming.humanize(conn.params["provider"]))
-
-  defp interpolate(msg, opts) do
-    Enum.reduce(opts, msg, fn {key, value}, msg ->
-      token = "%{#{key}}"
-
-      case String.contains?(msg, token) do
-        true -> String.replace(msg, token, to_string(value), global: false)
-        false -> msg
-      end
-    end)
-  end
+    do: "Continue with #{Naming.humanize(conn.params["provider"])}"
 end
--- a/lib/oli_web/templates/email/account_already_exists.html.eex
+++ b/lib/oli_web/templates/email/account_already_exists.html.eex
@@ -1,12 +1,23 @@
-<div class="flex flex-column gap-y-4">
-  <h3>Are you trying to create a new account?</h3>
-  <p>
-    Someone tried to create an account with this email but an account already exists.
-    If this was you, you can login <%= link "here", to: @url, target: "_blank" %> with your existing email and password.
-    If you forgot your password, you can reset it by clicking <%= link "Forgot Password?", to: @forgot_password, target: "_blank" %>.
-  </p>
-  <p>If this was not you, you can disregard this email.</p>
-</div>
-
-
-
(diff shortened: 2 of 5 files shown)


lib/oli_web/pow/messages.ex
```diff
@@ -58,17 +58,5 @@ defmodule OliWeb.Pow.Messages do
...
-    Enum.reduce(opts, msg, fn {key, value}, msg ->
-      token = "%{#{key}}"
-
-      case String.contains?(msg, token) do
-        true -> String.replace(msg, token, to_string(value), global: false)
-        false -> msg
-      end
-    end)
-  end
+    do: "Continue with #{Naming.humanize(conn.params["provider"])}"
```
eliknebel, line 61: This was some odd code I stumbled upon that clearly intended to be a simple string interpolation, so I simplified it

lib/oli_web/templates/email/account_already_exists.html.eex
```diff
@@ -1,12 +1,23 @@
...
-    Someone tried to create an account with this email but an account already exists.
-    If this was you, you can login <%= link "here", to: @url, target: "_blank" %> with your existing email and password.
-    If you forgot your password, you can reset it by clicking <%= link "Forgot Password?", to: @forgot_password, target: "_blank" %>.
-  </p>
-  <p>If this was not you, you can disregard this email.</p>
-</div>
-
-
-
+<table align="center" role="presentation" cellspacing="0" cellpadding="0" border="0" width="100%" style="margin: auto;">
```
eliknebel, line 1: Email clients typically don't render html in the same sane way that browsers do, further they certainly do not have access to tailwind classes. I updated this template to be email friendly like the other templates and better match the typical email style since this in particular is an email we want a user to see as authentic

darrensiegel (approved):

PR #5109, opened 2024-09-18
[BUG FIX] [MER-3823] Convert numeric to text input

darrensiegel:
Converts the input from type `number` to type `text` and adds in regex based validation that what the user has entered is a number.  When it is not a number, displays a red border around the input to indicate the validation value.   In all cases, whatever they have entered into the input will round trip to the server and save.

### DIFF
--- a/assets/src/components/activities/common/delivery/inputs/NumericInput.tsx
+++ b/assets/src/components/activities/common/delivery/inputs/NumericInput.tsx
@@ -16,7 +16,6 @@ interface Props {
 
 export const NumericInput: React.FC<Props> = (props) => {
   const numericInputRef = createRef<HTMLInputElement>();
-  const [uniqueId] = useState(Math.random().toString(36).substring(7));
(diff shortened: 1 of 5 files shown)


andersweinstein:
1. HTML standard allows numbers of the form `.333` with no digits to the left of the decimal point. The regexp here does not appear to allow that.

2. We have a request to allow input of the form `300.` with no digits to the `right` of the decimal place: https://eliterate.atlassian.net/browse/MER-3012 (and see linked Triage ticket)
That is not a valid floating-point number according to the HTML spec, but is a notation used in practice to indicate the value is known to 3 significant figures. If the validation were changed to allow this as well, that would knock off that ticket. (I believe the significant figures check handles this correctly)

Allowing for absence of digits on either side of the decimal place requires a more involved regexp using a disjunction. Some discussion here:
https://stackoverflow.com/questions/12643009/regular-expression-for-floating-point-numbers/42629198#42629198

eliknebel (approved):

PR #5116, opened 2024-09-19
[BUG FIX] [MER-3832] Preserve other revision settings

darrensiegel:
There were four other settings not being pulled forward from the revisions during course section update application.

### DIFF
--- a/lib/oli/delivery/sections/updates.ex
+++ b/lib/oli/delivery/sections/updates.ex
@@ -214,7 +214,16 @@ defmodule Oli.Delivery.Sections.Updates do
           scoring_strategy_id: r.scoring_strategy_id,
           slug: Oli.Utils.Slug.generate("section_resources", r.title),
           inserted_at: {:placeholder, :timestamp},
-          updated_at: {:placeholder, :timestamp}
+          updated_at: {:placeholder, :timestamp},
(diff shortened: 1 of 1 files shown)


lib/oli/delivery/sections/updates.ex
```diff
@@ -214,7 +214,16 @@ defmodule Oli.Delivery.Sections.Updates do
...
           inserted_at: {:placeholder, :timestamp},
-          updated_at: {:placeholder, :timestamp}
+          updated_at: {:placeholder, :timestamp},
+          collab_space_config: r.collab_space_config,
+          max_attempts:
+            if is_nil(r.max_attempts) do
+              0
+            else
+              r.max_attempts
+            end,
```
simonchoxx, line 224: I prefer the short way of writing the `IF` statement, such as: 

`if is_nil(r.max_attempts), do: 0, else: r.max_attempts`

but that's just a personal taste!

simonchoxx (approved):
LGTM!

lib/oli/delivery/sections/updates.ex
Francisco-Castro: @simonchoxx what about: `r.max_attempts || 0`

PR #5118, opened 2024-09-19
[ENHANCEMENT] [MER-3750] admin masquerade as user

eliknebel:
https://eliterate.atlassian.net/browse/MER-3750

This PR adds the capability to "Act as User", allowing an admin to masquerade as any user account in the system (excluding author accounts). The intent here is to given system and account admin users, specifically support teams, better access to user views for troubleshooting and taking action on behalf of a user. Access to this feature can easily be restricted to system admins only if that is desired.

https://github.com/user-attachments/assets/2085843d-3f4b-4b67-9f5b-d55242ad6873

There are a couple of important things to note:
- In this implementation, "Acting as a User" means that all actions including event generation such as xapi messages such as viewing a page will occur and attempts can be made on behalf of a user. This is a powerful feature but also one that demands careful consideration by an admin to prevent unintended side effects.
- Adaptive pages do not show the colorful banner message and border (yet)
- Learn and Lesson view layouts had to be reworked to get the banner and border working properly. This was necessary for v29 anyways and if we don't take this PR for 29, we should carve out this work and take it anyways (commit [6d4a956](https://github.com/Simon-Initiative/oli-torus/pull/5118/commits/6d4a956c847c95fc0b4cf6a52111f2e56a046b45))

### DIFF
--- a/assets/src/apps/authoring/components/Flowchart/onboard-wizard/OnboardWizard.tsx
+++ b/assets/src/apps/authoring/components/Flowchart/onboard-wizard/OnboardWizard.tsx
@@ -97,7 +97,7 @@ const Step3Advanced: React.FC<{
         <h2>Note</h2>
 
         <p>
-          Projects created in Advanced Authoring do not open in Basic Authoring. This requires
+          Projects created in Advanced Authoring do not open in Simple Authoring. This requires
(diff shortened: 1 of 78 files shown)


eliknebel:
This will have to wait until v30

PR #5126, opened 2024-09-24
[FEATURE] [MER-3598] Migrate Revision History to workspaces/course_author

manelli:
Migrate the `Revision History` to a new liveview under the workspaces/course_author scope.

https://github.com/user-attachments/assets/0300e948-cf1d-472a-93c3-cf24c32ca8d2

See: https://eliterate.atlassian.net/browse/MER-3598


lib/oli_web/live/workspaces/course_author/history_live.ex
```diff
@@ -0,0 +1,733 @@
...
+        </div>
+      <% end %>
+    </div>
+
+    <script
+      id="keep-alive"
+      type="text/javascript"
+      src={Routes.static_path(OliWeb.Endpoint, "/js/keepalive.js")}
+    >
+    </script>
```
gastonabella, line 359: Do you think this code block is necessary? It is already included in `live.html.heex` which is included in the second line of the module with `use OliWeb, :live_view`. I might be missing something, but I'm concerned that we might be showing flash messages twice in this liveview.

gastonabella (commented):
Looks great!

I noticed from the attached recording that when visiting this new authoring view, it still selects `Student` in the workspace sidebar on the left. Is this expected or should we have `Course Author` selected instead?

manelli:
> Looks great!
> 
> I noticed from the attached recording that when visiting this new authoring view, it still selects `Student` in the workspace sidebar on the left. Is this expected or should we have `Course Author` selected instead?

This was because I branched out from an outdated master branch that had that problem with the sidebar, I rebased and it is now showing the correct tab.

image

lib/oli_web/live/workspaces/course_author/history_live.ex
manelli: Nice catch, yeah sorry I was testing something and forgot to remove it.
Removed in 7e65a8789babca228d4cd9e467004a37997d41e8

gastonabella (approved):

PR #5131, opened 2024-09-25
[BUG FIX] [MER-3847] Sort attempts by date_evaluated

darrensiegel:
The comments in the code explain what this PR does - but essentially it strengthens the guarantee that we take the "most recently evaluated page attempt" for the `most_recent` scoring strategy.

### DIFF
--- a/lib/oli/delivery/attempts/scoring.ex
+++ b/lib/oli/delivery/attempts/scoring.ex
@@ -71,7 +71,7 @@ defmodule Oli.Delivery.Attempts.Scoring do
     # the natural database order - or somehow the query doesn't return
     # in database order.
     [most_recent | _] =
-      Enum.filter(items, & &1.date_evaluated != nil)
+      Enum.filter(items, &(&1.date_evaluated != nil))
       |> Enum.sort_by(& &1.date_evaluated, &(DateTime.compare(&1, &2) != :lt))
 
     %Result{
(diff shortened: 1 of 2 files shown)


lib/oli/delivery/attempts/scoring.ex
```diff
@@ -63,7 +63,14 @@ defmodule Oli.Delivery.Attempts.Scoring do
...
   def calculate_score("most_recent", items) do
-    most_recent = Enum.reverse(items) |> hd
+    # Sort the resource_attemmpts by the date_evaluated field, so that
+    # the most recent evaluated attempt is the first item in the list.
+    #
+    # This makes this scoring strategy a little more robust in the face of
+    # attempts where somehow the most recent attempt does not match
+    # the natural database order - or somehow the query doesn't return
+    # in database order.
+    [most_recent | _] = Enum.sort_by(items, & &1.date_evaluated, &>=/2)
```
nicocirio, line 73: I understand this approach may not sort as expected. While comparing datetimes [we should instead use `compare/2`](https://hexdocs.pm/elixir/1.12.3/DateTime.html)

```suggestion
    [most_recent | _] = Enum.sort_by(items, & &1.date_evaluated, &(DateTime.compare(&1, &2) != :lt))
```

nicocirio (changes requested):

PR #5135, opened 2024-09-25
[BUG FIX] [MER-3715] Fix sidebar nil error

simonchoxx:
[MER-3715](https://eliterate.atlassian.net/browse/MER-3715)

 This PR tries to fix the error that occasionally happens in the `set_sidebar` plugin, when trying to match a nil value. 

This seems to be an edge case and apparently happens when in the session structure we have a value for the `section_slug` key for which no root container is found. 

The changes introduced try to avoid the case mentioned above, by handling more securely values that may be nil.

[MER-3715]: https://eliterate.atlassian.net/browse/MER-3715?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli_web/live_session_plugs/set_sidebar.ex
+++ b/lib/oli_web/live_session_plugs/set_sidebar.ex
@@ -12,10 +12,16 @@ defmodule OliWeb.LiveSessionPlugs.SetSidebar do
   alias Oli.Resources.Collaboration
   alias Oli.Publishing.{DeliveryResolver}
 
-  def on_mount(:default, _params, session, socket) do
+  def on_mount(:default, params, session, socket) do
+    section_slug =
+      case params do
+        %{"section_slug" => section_slug} -> section_slug
+        _ -> nil
+      end
+
     socket =
       socket
-      |> assign_notes_and_discussions_enabled(session["section_slug"])
+      |> assign_notes_and_discussions_enabled(section_slug)
       |> assign(sidebar_expanded: session["sidebar_expanded"])
       |> assign(disable_sidebar?: false)
       |> assign(header_enabled?: true)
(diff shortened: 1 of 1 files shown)


lib/oli_web/live_session_plugs/set_sidebar.ex
```diff
@@ -52,18 +52,23 @@ defmodule OliWeb.LiveSessionPlugs.SetSidebar do
...
+        case Collaboration.get_collab_space_config_for_page_in_section(
+               revision_slug,
+               section_slug
+             ) do
+          {:ok, %CollabSpaceConfig{status: :enabled}} -> true
+          _ -> false
+        end
+      else
+        false
       end
```
Francisco-Castro, line 72: This is an excellent opportunity to use a `with` statement:

```
    with %{slug: revision_slug} <- DeliveryResolver.root_container(section_slug),
         {:ok, %CollabSpaceConfig{status: :enabled}} <-
           Collaboration.get_collab_space_config_for_page_in_section(revision_slug, section_slug) do
      true
    else
      _ -> false
    end
```

When there is a well-defined happy path, `with` is a good building block to use.

Note: I haven't tested the suggested block; use it as a reference.

Francisco-Castro (changes requested):
I did some research on this issue and found that we are storing the`section_slug` in the session, most likely through the  `RequireSection` plug. Storing this value in the session persists across every request, as mentioned in the [documentation](https://hexdocs.pm/plug/Plug.Conn.html#module-session-vs-assigns):

> In [Plug](https://hexdocs.pm/plug/Plug.html), a "session" is a place to store data that persists from one request to the next. Typically, this data is stored in a cookie using [Plug.Session.COOKIE](https://hexdocs.pm/plug/Plug.Session.COOKIE.html).

In cases where we store this value in the session, we are always making extra queries, even though we don’t need the data half of the time—we use it only when in the `instructor` space. For the `course_author` space, we don't need it. Some tweaks to the `OliWeb.LiveSessionPlugs.SetSidebar` module would fix that issue. Take a look at the following piece of code:

```
def on_mount(:default, params, session, socket) do
    section_slug =
      case params do
        %{"section_slug" => section_slug} -> section_slug
        _ -> nil
      end

    socket =
      socket
      |> assign_notes_and_discussions_enabled(section_slug)
      ...
```

Would you be willing to add these changes here to avoid the extra queries?

Francisco-Castro (approved):

PR #5141, opened 2024-09-27
[ENHANCEMENT] [MER-3570] Optimize Student/Instructor Workspace queries

manelli:
Make less queries when gathering the sections for a user when is student/instructor

Related: https://eliterate.atlassian.net/browse/MER-3570


lib/oli_web/live/workspaces/student.ex
```diff
@@ -24,21 +34,17 @@ defmodule OliWeb.Workspaces.Student do
...
       when not is_nil(current_user) do
-    all_sections =
-      Sections.list_user_open_and_free_sections(current_user)
-      |> add_user_role(current_user)
-
     sections =
-      all_sections
-      |> filter_by_role(:student)
+      current_user.id
+      |> sections_where_user_is_student()
```
manelli, line 39: **sections_where_user_is_student:** 1 query in total
```sql
SELECT * FROM sections AS s0
INNER JOIN enrollments" AS e1 ON s0.id = e1.section_id
INNER JOIN enrollments_context_roles AS e2 ON e1.id = e2.enrollment_id
INNER JOIN users_platform_roles AS u3 ON e1.user_id = u3.user_id
WHERE
  (e1.user_id = $1) AND
  (s0.open_and_free = TRUE) AND
  (s0.status = 'active') AND
  (e2.context_role_id IN ($2) OR u3.platform_role_id IN ($3))

```

lib/oli_web/live/workspaces/student.ex
```diff
@@ -24,21 +34,17 @@ defmodule OliWeb.Workspaces.Student do
...
       |> add_instructors()
```
manelli, line 40: **add_instructors part 1:** 1 query in total
```sql
SELECT e0.*, s1.id FROM enrollments AS e0
INNER JOIN sections AS s1 ON e0.section_id = s1.id
INNER JOIN enrollments_context_roles AS e2 ON e0.id = e2.enrollment_id
WHERE (
  (s1.id IN ($1) AND (e0.status = 'enrolled')) AND
  (e2.context_role_id = $2)
)

```
**add_instructors part 2:** 1 query per each instructor of section
```sql
SELECT * FROM users AS WHERE (id = $1)
```

lib/oli_web/live/workspaces/student.ex
```diff
@@ -24,21 +34,17 @@ defmodule OliWeb.Workspaces.Student do
...
       |> add_sections_progress(current_user.id)
```
manelli, line 41: **add_sections_progress:** 2 queries per section


```sql
SELECT count(c0.id) FROM contained_pages AS c0
WHERE
  (c0.section_id = $1) AND
  (c0.container_id IS NULL)

```

```sql
SELECT r1.user_id, SUM(r1.progress) / $1 FROM contained_pages AS c0
INNER JOIN resource_accesses AS r1 ON ((c0.page_id = r1.resource_id) AND (c0.section_id = r1.section_id)) AND r1.user_id = ANY($2)
WHERE
  (c0.section_id = $3) AND
  (c0.container_id IS NULL)
GROUP BY r1.user_id

```

lib/oli_web/live/workspaces/student.ex
```diff
@@ -24,21 +34,17 @@ defmodule OliWeb.Workspaces.Student do
...
 
     {:ok,
      assign(socket,
        sections: sections,
        params: params,
-       disable_sidebar?: user_is_only_a_student?(all_sections),
+       disable_sidebar?: user_is_only_a_student?(current_user.id),
```
manelli, line 47: **user_is_only_student?:** 1 query in total


```sql
SELECT TRUE FROM enrollments AS e0
INNER JOIN enrollments_context_roles AS e1 ON e0.id = e1.enrollment_id
INNER JOIN users_platform_roles AS u2 ON e0.user_id = u2.user_id
WHERE
  (e0.user_id = $1) AND
  (NOT (e1.context_role_id = ANY($2)) OR NOT (u2.platform_role_id = ANY($3)))
LIMIT 1

```

lib/oli_web/live/workspaces/student.ex
manelli: This was previously 1 query + 4 queries per section

lib/oli_web/live/workspaces/student.ex
manelli: This was previously 1 query + 4 queries per section

lib/oli_web/live/workspaces/student.ex
manelli: This was previously 1 query per section

lib/oli_web/live/workspaces/student.ex
```diff
@@ -429,8 +416,38 @@ defmodule OliWeb.Workspaces.Student do
...
+    !Repo.exists?(
+      from e in Enrollment,
+        left_join: ecr in EnrollmentContextRole,
+        on: e.id == ecr.enrollment_id,
+        left_join: upr in "users_platform_roles",
+        on: e.user_id == upr.user_id,
+        where: e.user_id == ^user_id,
+        where:
+          ecr.context_role_id not in ^@context_student_roles_ids or
+            upr.platform_role_id not in ^@platform_student_roles_ids
```
gastonabella, line 431: I wonder if this part of the query works as we expect. Let's say a student has only a student platform role but **not** a student context role. In that case, line `430` will return `true` regardless that the user is platform student, and the whole query will return `false` which I understand is not the desired outcome. I think changing the `or` with an `and` would fix this edge case. Wdyt?

lib/oli_web/live/workspaces/instructor/index_live.ex
```diff
@@ -433,4 +426,23 @@ defmodule OliWeb.Workspaces.Instructor.IndexLive do
         Params.get_boolean_param(params, "sidebar_expanded", @default_params.sidebar_expanded)
     }
   end
+
+  defp sections_where_user_is_instructor(user_id) do
```
gastonabella, line 430: Nit, but what about moving this and other queries to a separate context module?

gastonabella (commented):
Great work! 🚀

lib/oli_web/live/workspaces/instructor/index_live.ex
eliknebel: Agree, we should not really be calling Oli.Repo from OliWeb context or making queries. This would be better suited in the Oli.Delivery.Sections module

lib/oli_web/live/workspaces/student.ex
```diff
@@ -429,8 +416,38 @@ defmodule OliWeb.Workspaces.Student do
...
+        left_join: ecr in EnrollmentContextRole,
+        on: e.id == ecr.enrollment_id,
+        left_join: upr in "users_platform_roles",
+        on: e.user_id == upr.user_id,
+        where: e.user_id == ^user_id,
+        where: s.open_and_free == true,
+        where: s.status == :active,
+        where:
+          ecr.context_role_id in ^@context_student_roles_ids or
+            upr.platform_role_id in ^@platform_student_roles_ids,
```
eliknebel, line 449: I think we should only be considering context roles here, we don't need to check platform roles. By including platform roles, this will return true for all sections because a student will always have a student platform role

lib/oli_web/live/workspaces/instructor/index_live.ex
```diff
@@ -4,22 +4,30 @@ defmodule OliWeb.Workspaces.Instructor.IndexLive do
...
+  ]
+
   @impl Phoenix.LiveView
   def mount(_params, _session, %{assigns: %{current_user: current_user}} = socket)
       when not is_nil(current_user) do
     sections =
-      Sections.list_user_open_and_free_sections(current_user)
-      |> add_user_role(current_user)
+      current_user.id
+      |> sections_where_user_is_instructor()
```
Francisco-Castro, line 29: Unfortunately, we don't have tests for the logic of pulling the section, except for one that I found in

https://github.com/Simon-Initiative/oli-torus/blob/e8ffbeaa5732bda895bde6414d1295e52751057a/test/oli/delivery/sections_test.exs#L2368

Calling this new `sections_where_user_is_instructor/1` in test mentioned I would expected to have the same result but I don't. See the following snippet:


```
      [s1, s3, s2] =
        OliWeb.Workspaces.Instructor.IndexLive.sections_where_user_is_instructor(user.id)
        |> Enum.filter(fn s -> s.user_role == "instructor" end)

      assert s1.title == "Elixir"
      assert s3.title == "LiveView"
      assert s2.title == "Phoenix"

      [s1, s3, s2] = Sections.list_user_open_and_free_sections(user)

      assert s1.title == "Elixir"
      assert s3.title == "LiveView"
      assert s2.title == "Phoenix"
```

Is that expected?

lib/oli_web/live/workspaces/instructor/index_live.ex
Francisco-Castro: Agree with Gaston. When you move it you can also add some tests to verify the expected behavior.

lib/oli_web/live/workspaces/student.ex
eliknebel: Agree, we should simply not be considering platform roles in the query

```
 from e in Enrollment,
        left_join: ecr in EnrollmentContextRole,
        on: e.id == ecr.enrollment_id and e.user_id == ^user_id,
        where:
          ecr.context_role_id not in ^@context_student_roles_ids
```

eliknebel (changes requested):

lib/oli_web/live/workspaces/student.ex
```diff
@@ -429,8 +416,38 @@ defmodule OliWeb.Workspaces.Student do
...
+        select: s
+    )
+  end
```
Francisco-Castro, line 452: This query needs a context and some tests.

Francisco-Castro (changes requested):
Your PR is making good progress! Just a few areas that could use some extra attention, but overall it's coming along nicely.

lib/oli_web/live/workspaces/student.ex
manelli: Thanks for the input!
Now the query is:
```elixir
      from sections in Section,
        join: enrollments in assoc(sections, :enrollments),
        join: context_roles in EnrollmentContextRole,
        on: enrollments.id == context_roles.enrollment_id,
        where: enrollments.user_id == ^user_id,
        where: sections.open_and_free == true,
        where: sections.status == :active,
        where: context_roles.context_role_id in ^context_role_ids,
        select: sections

```

lib/oli_web/live/workspaces/instructor/index_live.ex
manelli: 👍 Moved them to the `Oli.Delivery.Sections` module

lib/oli_web/live/workspaces/student.ex
manelli: 👍 added tests in `test/oli/delivery/sections_test.exs`

lib/oli_web/live/workspaces/instructor/index_live.ex
manelli: Yep, because `list_user_open_and_free_sections/1` returns all the sections for a user (enrolled as student, instructor, etc).

Francisco-Castro: @manelli Notice in the code snippet I provided that I added an Enum equivalent to what `filter_by_role(:instructor)` is doing to replicate the current behavior:

https://github.com/Simon-Initiative/oli-torus/blob/9f2888cb308bbbcd619e6c00cc7bbaff87fa65e4/lib/oli_web/live/workspaces/instructor/index_live.ex#L18-L22

We should expect the same result for both your query and the combination provided in the snippet, right?

manelli: The function that should be piped to `Enum.filter(fn s -> s.user_role == "instructor" end)`  is
```elixir
   Sections.list_user_open_and_free_sections(current_user) 
   |> add_user_role(current_user) 
```

not `sections_where_user_is_instructor`

Francisco-Castro: Thanks for pointing that. I'll do another loop then.

PR #5143, opened 2024-09-27
[BUG FIX] [MER-3300] Show sidebar if user can create sections

manelli:
If a user can create sections, then the user is considered and instructor and the navigation sidebar should be displayed.

See: https://eliterate.atlassian.net/browse/MER-3300


Francisco-Castro (approved):
Nice way to add the logic.

PR #5144, opened 2024-09-27
[FEATURE] [MER-3308] Infiniscope returning user enters argos through sso then enters course author

Francisco-Castro:
Ticket: [MER-3308](https://eliterate.atlassian.net/browse/MER-3308)

This PR adds the following feature:

1. When a user comes from InfiniScope to Argos and already has both an instructor and author account, they should be automatically logged into both accounts.

2. For existing InfiniScope users who already have an associated author account, we should preserve that link and not create a new account.

[MER-3308]: https://eliterate.atlassian.net/browse/MER-3308?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gastonabella (approved):
LGTM 🚀

PR #5151, opened 2024-10-01
[BUG FIX] [MER-3708] 500 error with iframed lesson launch links

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-3708) to the ticket

The query params in the URL were causing the iframe to crash. The fix removes the query params from adaptive pages routes and, since the `request_path` query param was needed to provide the back URL path, we now pass it through the session.

### Before
The request_path (used for the back/home button) was passed as an URL query param

https://github.com/user-attachments/assets/e1afcbf5-75ce-44b7-a9d2-ea0204e57d58

### After
The request_path is set to the conn session to keep back/home button functionality working as before

https://github.com/user-attachments/assets/c158de38-1320-4f67-ae83-ccb3c441edf9

### DIFF
--- a/lib/oli_web/plugs/redirect_by_attempt_state.ex
+++ b/lib/oli_web/plugs/redirect_by_attempt_state.ex
@@ -251,7 +251,6 @@ defmodule OliWeb.Plugs.RedirectByAttemptState do
          "/adaptive_lesson/"
        ) do
       conn
-      |> put_session(:request_path, conn.params["request_path"])
       |> assign(:already_been_redirected?, false)
     else
       # adaptive lesson in iframes do not support query params in the url
--- a/lib/oli_web/plugs/redirect_by_attempt_state.ex
+++ b/lib/oli_web/plugs/redirect_by_attempt_state.ex
@@ -222,13 +222,13 @@ defmodule OliWeb.Plugs.RedirectByAttemptState do
     end
   end
 
-  defp ensure_path(conn, rendering_type) when rendering_type in [:adaptive_lesson, :lesson] do
+  defp ensure_path(conn, rendering_type) when rendering_type == :lesson do
     section_slug = conn.params["section_slug"]
     revision_slug = conn.params["revision_slug"]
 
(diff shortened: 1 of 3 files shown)


lib/oli_web/plugs/redirect_by_attempt_state.ex
```diff
@@ -222,22 +222,46 @@ defmodule OliWeb.Plugs.RedirectByAttemptState do
     end
   end
 
-  defp ensure_path(conn, rendering_type) when rendering_type in [:adaptive_lesson, :lesson] do
+  defp ensure_path(conn, rendering_type) when rendering_type == :lesson do
```
simonchoxx, line 225: Here we could use pattern matching in the definition of the function, WDYT? 
I think it should work anyway by doing something like : 

`defp ensure_path(conn, :lesson) do`

lib/oli_web/plugs/redirect_by_attempt_state.ex
```diff
@@ -222,22 +222,46 @@ defmodule OliWeb.Plugs.RedirectByAttemptState do
...
+      conn
+      |> halt()
+      |> assign(:already_been_redirected?, true)
+      |> Phoenix.Controller.redirect(
+        to: "/sections/#{section_slug}/lesson/#{revision_slug}?#{conn.query_string}"
+      )
+    end
+  end
+
+  defp ensure_path(conn, rendering_type) when rendering_type == :adaptive_lesson do
```
simonchoxx, line 245: Same as above

simonchoxx (approved):
Nice solution by using the session to pass the `request_path` for the adaptive pages! 👏

I left some minor comments but they are not a priority to change.

lib/oli_web/plugs/redirect_by_attempt_state.ex
nicocirio: Totally agree! I adapted the code and didn't realize the overhead the `when` clause is adding here

PR #5155, opened 2024-10-02
[BUG FIX] [MER-3570] Fix delay when switching between instructor and student tabs

manelli:
https://github.com/user-attachments/assets/16cea1d2-99e0-406a-8a93-47f5b9045648

The code removed was added in commit [1c7af4d](https://github.com/Simon-Initiative/oli-torus/commit/1c7af4dee57f445a0b2cfd024692d70234091059#diff-281eaaca2196c3d882fddc362b9e96e241cc931eec2591cd6212f9c41147fecaR96). I don't know if it is essential to that feature. But it was also present in the student workspace.

See: https://eliterate.atlassian.net/browse/MER-3570

### DIFF
--- a/lib/oli_web/live/workspaces/student.ex
+++ b/lib/oli_web/live/workspaces/student.ex
@@ -309,7 +309,7 @@ defmodule OliWeb.Workspaces.Student do
       phx-mounted={
         JS.transition(
           {"ease-out duration-300", "opacity-0 -translate-x-1/2", "opacity-100 translate-x-0"},
-          time: 300 + @index * 60
+          time: 300
         )
       }
       class="opacity-0 flex flex-col w-96 h-[500px] rounded-lg border-2 border-gray-700 transition-all overflow-hidden bg-white"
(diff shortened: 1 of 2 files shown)


rgachuhi (approved):

lib/oli_web/live/workspaces/student.ex
```diff
@@ -309,7 +309,7 @@ defmodule OliWeb.Workspaces.Student do
       phx-mounted={
         JS.transition(
           {"ease-out duration-300", "opacity-0 -translate-x-1/2", "opacity-100 translate-x-0"},
-          time: 300 + @index * 60
+          time: 300
```
nicocirio, line 312: I believe we should apply the same fix on line 259

nicocirio (commented):
If we want to keep the card animations, I think we could limit them to the first elements  (`time: if(index < 6, do: 100 + index * 20, else: 240)`, for example), since the other elements are not visible for the user when they enter the page. 
This way the user will still see the animation and I believe we won't have the delay for users with a long list of courses.

https://github.com/user-attachments/assets/2c6bab67-a1ef-48b3-9ac1-32ea711628fc

PR #5157, opened 2024-10-03
[FEATURE] [MER-3415] Changing number of attempts available in simple author/flowchart tool

dtiwarATS:
Hey @bsparks could you please look at the PR. Thanks

The simple author currently defaults to 3 attempts for a screen, with no option to change this. The requirement was to introduce the ability to change the number of allowed attempts. The default should remain at three, but users should be able to specify a different number of attempts (1, 2, 3, 4, or 5). The system should honor the specified number.

### DIFF
--- a/assets/src/apps/authoring/components/Flowchart/rules/create-cata-choice-rules.ts
+++ b/assets/src/apps/authoring/components/Flowchart/rules/create-cata-choice-rules.ts
@@ -26,7 +26,7 @@ import {
   getSequenceIdFromDestinationPath,
   getSequenceIdFromScreenResourceId,
 } from './create-generic-rule';
-import { generateThreeTryWorkflow } from './create-three-try-workflow';
+import { generateMaxTryWorkflow } from './create-three-try-workflow';
 import { RulesAndVariables } from './rule-compilation';
 
 // Check all that apply, aka Multiple Choice Question with multi-selection enabled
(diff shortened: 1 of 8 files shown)


assets/src/apps/authoring/components/Flowchart/rules/create-cata-choice-rules.ts
```diff
@@ -142,13 +142,14 @@ export const generateCATAChoiceRules = (
...
 
-  return generateThreeTryWorkflow(
+  return generateMaxTryWorkflow(
     correct,
     incorrect,
     commonErrorConditionsFeedback,
     setCorrect,
     blankCondition,
     disableAction,
+    { maxAttempt: screen?.content?.custom?.maxAttempt || '2' },
```
bsparks, line 152: I thought the default was 3, why '2' everywhere?

bsparks (approved):

assets/src/apps/authoring/components/Flowchart/rules/create-cata-choice-rules.ts
dtiwarATS: That's how it has been in simple authoring. For maxOrLaterTry which was the default max 3 attempt in the existing behavior, the code was like this: 

```
const maxOrLaterTry = (value: string): ICondition => ({
  fact: 'session.attemptNumber',
  operator: 'greaterThan',
  value: '2',
  type: 1,
  id: guid(),
});
```
hence, I just replaced it with the user enter attempt and default to what it was before.

PR #5160, opened 2024-10-03
[ENHANCEMENT] [MER-3570] Add card transitions back for Instructor/Student index

manelli:
As commented by @nicocirio in https://github.com/Simon-Initiative/oli-torus/pull/5155#pullrequestreview-2345421105, we can add the transitions back but only make them dynamic up to the 6th card which is what the user sees.


nicocirio (approved):
I'm glad it worked as we expected 😄 
Thanks for the implementation!

PR #5161, opened 2024-10-03
[BUG FIX] [MER-3612] Fix set num attempts default value

simonchoxx:
[MER-3612](https://eliterate.atlassian.net/browse/MER-3612)

This PR sets a default value of 2 to the numeric input that is used to set the number of attempts of the `After Set Num Attempts` explanation strategy within the modal that an author/admin uses to set the options for a course curriculum page. 

If the above strategy has no value it will default to a value of 2 and otherwise it will use the value that has been set previously.

https://github.com/user-attachments/assets/256110d8-22b3-4433-8645-5e72ca518f47

[MER-3612]: https://eliterate.atlassian.net/browse/MER-3612?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli_web/live/curriculum/entries/options_modal.ex
+++ b/lib/oli_web/live/curriculum/entries/options_modal.ex
@@ -385,16 +385,12 @@ defmodule OliWeb.Curriculum.OptionsModalContent do
                       }
                     />
                     <div class="ml-2">
+                      <% explanation_strategy =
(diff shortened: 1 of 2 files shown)


manelli (approved):
Looks good 🚀

PR #5175, opened 2024-10-07
[BUG FIX] [MER-3887] Take author to new workspace after login

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-3887) to the ticket

### Before

https://github.com/user-attachments/assets/7ba5633c-a97b-430f-b9cd-d59c7dc3d432

### After

https://github.com/user-attachments/assets/846d5a0c-e7f7-4dad-95fd-5a89b49f4f31

### DIFF
--- a/lib/oli_web/controllers/pow/session_html/new.html.heex
+++ b/lib/oli_web/controllers/pow/session_html/new.html.heex
@@ -42,7 +42,7 @@
           >
             OR
           </div>
-          <%= form_for @changeset, @action, [as: :user], fn f -> %>
+          <%= form_for @changeset, Routes.authoring_pow_session_path(OliWeb.Endpoint, :create, request_path: ~p"/workspaces/course_author"), [as: :user], fn f -> %>
             <div class="flex flex-col gap-y-2">
               <div class="w-80 h-11 m-auto form-label-group border-none">
                 <%= email_input(f, Pow.Ecto.Schema.user_id_field(@changeset),
--- a/lib/oli_web/controllers/pow/session_html/new.html.heex
+++ b/lib/oli_web/controllers/pow/session_html/new.html.heex
@@ -42,7 +42,7 @@
           >
             OR
           </div>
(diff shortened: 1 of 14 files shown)


simonchoxx (approved):
LGTM 🚀

lib/oli_web/controllers/pow/session_html/new.html.heex
```diff
@@ -42,7 +42,7 @@
           >
             OR
           </div>
-          <%= form_for @changeset, @action, [as: :user], fn f -> %>
+          <%= form_for @changeset, Routes.authoring_pow_session_path(OliWeb.Endpoint, :create, request_path: ~p"/workspaces/course_author"), [as: :user], fn f -> %>
```
eliknebel, line 45: This works and is a good solution for this login, however I think the real culprit here is the routes defined in `lib/oli_web/pow/author_routes.ex` for `after_sign_in_path` and `after_registration_path`

eliknebel (changes requested):

eliknebel:
I think there are a few other places that could possibly lead a user into the old views:
- lib/oli_web/controllers/project_plugs.ex
- lib/oli_web/plugs/authorize_project.ex
- lib/oli_web/plugs/community_admin.ex

Replacing the references to OliWeb.Projects.ProjectsLive in these views with the new workspace route should fix it

simonchoxx:
@eliknebel Ok, I will change the references to `OliWeb.Projects.ProjectsLive` in those other three places to go to `/workspaces/course_author`

lib/oli_web/controllers/pow/session_html/new.html.heex
simonchoxx: @eliknebel Do you think we should keep the code as it was and modify the routes in those two functions (after_sign_in_path and after_registration_path) ???

manelli (approved):

eliknebel (approved):

PR #5179, opened 2024-10-08
[BUG FIX] [MER-3841] Make PageContent traversal functions hit content in properties like caption

andersweinstein:
Citations in image captions were found to get broken on ingesting a correctly-migrated course, or on export-ingest of a correctly working course. This happened because the required rewiring process was not hitting cite elements in image captions. This happened because this rewiring used a function from the PageContent module which only traversed through element children, and caption content is in a caption property of the image,  not a child element. 

This PR implements a general solution to modify the relevant PageContent function to visit and process content in properties like the caption property of an image element, and also the relevant properties of a popup element, as is done correctly in the code that rewires links. 

~The relevant function is also used when generating unique ids for publication, so this change could affect that process as well, though if correct, should only improve it.~ Corrected below.

### DIFF
--- a/lib/oli/authoring/editing/util.ex
+++ b/lib/oli/authoring/editing/util.ex
@@ -49,23 +49,6 @@ defmodule Oli.Authoring.Editing.Utils do
     end
   end
 
-  @doc """
-  Returns a MapSet of all bib_entry_id ids found in the page content hierarchy.
-  """
-  def citation_references(content) do
-    case content do
-      %{"model" => _} ->
-        Oli.Resources.PageContent.flat_filter(content, fn %{"type" => type} ->
-          type == "cite"
-        end)
-        |> Enum.map(fn %{"bib_entry_id" => id} -> id end)
-        |> MapSet.new()
-
-      _ ->
-        MapSet.new([])
-    end
-  end
-
   def new_container_name(numberings, parent_container) do
     numbering = Map.get(numberings, parent_container.id)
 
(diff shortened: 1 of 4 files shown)


darrensiegel (changes requested):
We need to prove this works via a unit test

andersweinstein:
> We need to prove this works via a unit test

OK, how extensive a test do you want? Just the case that motivated this, citations in captions? I verified it on ingest/export with a minimal one-page-with-one-image-with-caption course. Could turn the json into a unit test, I suppose.

andersweinstein:
I grasp now that there are two different sets of content traversing `map_reduce` functions in the codebase, one in `Oli.Resources.PageContent`, and another near duplicate set in `Oli.Publishing.UniqueIds` and they behave differently:

-The UniqueId versions are described in the code as differing in order to take "stop" and "ignore" parameters to control their traversal. 
- But another difference is that the top-level PageContent map_reduce function does NOT descend into structured content blocks (type="content") at all. (Presumably it was intended for locating activities, not traversing the content element tree). 

The PageContent module also includes functions like `item_with_children`. This can be used within a suitable outer loop or map function to traverse into structured content where that is desired. It is **this** method that is used by the ingestion/export code to find and process citations.

This seems very unfortunate.  But in any case: the current change fixes `item_with_children` within the the PageContent module so it will visit content elements in properties when used for that purpose. It's possible the same fix ought to be added to unique_ids. But the present issue was the rewiring of citations on ingest.

andersweinstein:
Minimal one-page test course including citation in image caption that can be used to test ingest and export->ingest again.

[anders_citation_test.zip](https://github.com/user-attachments/files/17412114/anders_citation_test.zip)

darrensiegel (approved):

PR #5185, opened 2024-10-11
[PERFORMANCE] [MER-3911] Section Resource Depot

darrensiegel:
This PR introduces `Depot`, a flexible and performant ETS table backed cache for Ecto schemas.  

Specific Depot instances can be created to improve access performance for Ecto schemas that are infrequently written to but frequently read.  The `SectionResource` schema is a perfect example, as there are only a handful of places where changes are made to these schemas (updating the schedule, changing an assessment setting, applying a course update, etc), yet student facing code constantly needs to access `SectionResource` records in various complex (and often expensive) ways. 

This PR introduces the `SectionResourceDepot` which provides instructor and student specific functions for accessing SR records in various ways - all backed by an ETS in memory table.  Cache invalidation and updating is done via a separate interface, the `DepotCoordinator`. 

Developers should plan to add additional functions to `SectionResourceDepot` - particularly when building new end-user facing functionality that they want to be performant from the beginning. 

`SectionResourceDepot` (and `Depot` itself) is faster and more scalable than existing caching strategies such as `SectionCache` and the section schema based `prev_next_index` due to its "same process" reading implementation.  Instead of a single process in the node providing access to the cache (as done in SectionCache), Depot is designed to be accessed directly from any process.  This "same process" strategy results in less process-to-process messaging, faster access to data and overall a more scalable solution (as there is no single process that can bottleneck cache access)

This approach is also more flexible than other caching approaches since it is caching the data in its fundamental form, rather than in a derived, or secondary form (such as the hierarchy cache in SectionCache or the `prev_next_index` cache in the `Section` schema).  This allow quick access to the raw records, and then post processing can take place to transform to the desired derived form.  In additional to flexibility, this approach saves memory space by caching the data in exactly one form - and allowing specialized depot code to re-arrange and post process that into other formats. 

To maximize the benefit of a `SectionResourceCache`, this impl replicates several `Revision` fields into the `SectionResource` schema.  This provides much faster and simpler access to fields like the title, graded, resource type, intro_content which only change during publication updates.   A corresponding inline migration is introduced through the `SectionResourceDepot` to migrate the relevant versions of data into SR records.  The implication here, though, is that these fields cannot be used by code OUTSIDE of the `SectionResourceCache`, otherwise there is no guarantee that the migration has been run. 

The `Depot` based caching approach is also faster at invalidation, since it can completely rebuild its cache with a simple query of the existing set of schema records for a specific table.  In other words, for `SectionResourceDepot`, to invalidate the cache we simply can delete the table, and the next read will trigger a query only over the `SectionResource` table (and not some complex delivery resolver query).

Note, only containers, pages and objectives are included in the `SectionResourceCache` at the moment.  There just wasn't a need yet to include activities and tags or any other resource type. 

Screenshot 2024-10-11 at 9 13 39 AM


lib/oli/delivery/singleton_depot_coordinator.ex
```diff
@@ -0,0 +1,23 @@
+defmodule Oli.Delivery.SingletonDepotCoordinator do
```
nicocirio, line 1: If we are using this module exclusively for testing purposes, should we move the file to the /test/support folder? And/or maybe extend the module doc to explain this.
```suggestion
defmodule Oli.Delivery.SingletonDepotCoordinator do
```

lib/oli/delivery/distributed_depot_coordinator.ex
```diff
@@ -0,0 +1,47 @@
...
+    end
+
+    {:noreply, state}
+  end
+
+  def handle_info({:update_all, depot_desc, entries}, state) do
+    [first | _rest] = entries
+    table_id = Map.get(first, depot_desc.table_id_field)
+
+    if Depot.table_exists?(depot_desc, table_id) do
```
nicocirio, line 38: If the table does not exist, shouldn't we create and populate it? Or do we only want to initialize the table when we explicitly need the data (when calling any of the SectionResourceDepot functions)?

lib/oli_web/live/delivery/student/index_live.ex
```diff
@@ -77,7 +77,7 @@ defmodule OliWeb.Delivery.Student.IndexLive do
                   section
                   |> get_or_compute_full_hierarchy()
                   |> Hierarchy.find_module_ancestor(
-                    page[:resource_id],
+                    Map.get(page, :resource_id),
```
nicocirio, line 80: NIT: I think both approaches will return `nil` if the key does not exist in the map, or the value if it exists. So I was wondering why we made this change here

nicocirio (commented):
LGTM. I left some minor suggestions and a doubt about when to initialize the cached tables. 

I think it would be great to replicate the description of this PR as a new page in our wiki, WDYT?

lib/oli_web/live/delivery/student/index_live.ex
darrensiegel: There is a subtle affect that serializing / deserializing the SectionResource schema has:  `page[:resource_id]` style access of fields no longer works.

lib/oli/delivery/sections/section_resource_depot.ex
```diff
@@ -0,0 +1,150 @@
...
+  in places where code had been calling Oli.Delivery.Hierarchy.full_hierarchy
+  directly.
+  """
+  def get_full_hierarchy(%Section{} = section) do
+    init_if_necessary(section.id)
+
+    page = Oli.Resources.ResourceType.id_for_page()
+    container = Oli.Resources.ResourceType.id_for_container()
+
+    srs = Depot.query(@depot_desc, section.id, [{:resource_type_id, :in, [page, container]}])
```
nicocirio, line 48:
```suggestion
    srs = Depot.query(@depot_desc, section.id, [{:resource_type_id, {:in, [page, container]}}])
```

lib/oli/delivery/sections/section_resource_depot.ex
```diff
@@ -0,0 +1,150 @@
...
+  in places where code had been calling Oli.Publishing.DeliveryResolver.full_hierarchy
+  directly.
+  """
+  def get_delivery_resolver_full_hierarchy(%Section{} = section) do
+    init_if_necessary(section.id)
+
+    page = Oli.Resources.ResourceType.id_for_page()
+    container = Oli.Resources.ResourceType.id_for_container()
+
+    srs = Depot.query(@depot_desc, section.id, [{:resource_type_id, :in, [page, container]}])
```
nicocirio, line 66:
```suggestion
    srs = Depot.query(@depot_desc, section.id, [{:resource_type_id, {:in, [page, container]}}])
```

lib/oli/delivery/sections/section_resource_depot.ex
```diff
@@ -0,0 +1,150 @@
...
+          [{:resource_type_id, {:==, container_type_id}}]
+
+        _ ->
+          [{:resource_type_id, {:in, [container_type_id, page_type_id]}}]
+      end
+
+    Depot.query(@depot_desc, section_id, filter_by_resource_type)
+  end
+
+  def get_lessons(section, graded_only) do
```
nicocirio, line 106:
```suggestion
  @doc """
  Returns a list of SectionResource pages (graded + ungraded) for a given section.
  An optional parameter `graded_only` can be passed to filter only graded pages.
  """
  def get_lessons(section, graded_only \\ false) do
```

lib/oli/publishing/delivery_resolver.ex
```diff
@@ -343,6 +349,49 @@ defmodule Oli.Publishing.DeliveryResolver do
...
+        project_id: sr.project_id,
+        project_slug: sr.proj_slug,
+        section_resource: sr
+      }
+
+      {
+        Map.put(nodes, sr.id, node),
+        if(section.root_section_resource_id == sr.id, do: node, else: root)
+      }
+    end)
```
nicocirio, line 390: We should wrap this in a function in order to be able to pipe it into `run()`. And there was a typo in `project_slug: sr.project_slug`

```suggestion
    fn ->
      Enum.reduce(section_resources, {%{}, nil}, fn sr, {nodes, root} ->
        node = %HierarchyNode{
          uuid: uuid(),
          numbering: %Numbering{
            index: sr.numbering_index,
            level: sr.numbering_level,
            labels: labels
          },
          children: sr.children,
          resource_id: sr.resource_id,
          revision: %{
            id: sr.revision_id,
            resource_id: sr.resource_id,
            title: sr.title,
            slug: sr.revision_slug,
            resource_type_id: sr.resource_type_id,
            objectives: sr.objectives,
            graded: sr.graded,
            relates_to: sr.relates_to,
            purpose: sr.purpose,
            deleted: false
          },
          project_id: sr.project_id,
          project_slug: sr.project_slug,
          section_resource: sr
        }

        {
          Map.put(nodes, sr.id, node),
          if(section.root_section_resource_id == sr.id, do: node, else: root)
        }
      end)
    end
```

nicocirio (changes requested):
I tested `Oli.Delivery.Sections.SectionResourceDepot` locally and found that some minor changes need to be addressed to retrieve the data as expected.

lib/oli/delivery/sections/section_resource_depot.ex
```diff
@@ -0,0 +1,154 @@
...
+      end
+
+    Depot.query(@depot_desc, section.id, conditions)
+  end
+
+  defp init_if_necessary(section_id) do
+    if Depot.table_exists?(@depot_desc, section_id) do
+      {:ok, :exists}
+    else
+      if SectionResourceMigration.requires_migration?(section_id) do
```
gastonabella, line 128: Why are we migrating data in runtime and not as part of a separate migration?

gastonabella (commented):
Looks really nice!

lib/oli/delivery/depot/match_spec_translator.ex
```diff
@@ -0,0 +1,194 @@
...
+  is done to minimize the number of times the guards are evaluated, which optimizes retrieval.
+
+  Example (return only the titles where the duration equals 10):
+
+  > translate([{:title, :string}, {:duration, :integer}], 1, [duration: 10], [:title])
+  {{1, :"$f1", 10}, [], [:"$f1"]}
+
+  Example (return all fields where the duration is between 5 and 10):
+
+  > translate([{:title, :string}, {:duration, :integer}], 1, [{:duration, {:between, 5, 10}}], nil)
```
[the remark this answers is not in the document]
darrensiegel: Ah good catch, that is a bit of out of date documentation from a previous version of the function.

lib/oli/delivery/sections.ex
```diff
@@ -2910,7 +2910,10 @@ defmodule Oli.Delivery.Sections do
       |> Repo.transaction()
 
       # reset any section cached data
-      SectionCache.clear(section.slug)
```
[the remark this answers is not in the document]
darrensiegel: At one point, this work completely eliminated the SectionCache, but I had to bring it back.  We defiitely need to be clearing it here, so I have added this back.

lib/oli/delivery/sections/section_resource_depot.ex
darrensiegel: I went down the path of a just in time migration to avoid a potentially large and expensive migration at deployment time.

nicocirio (approved):
Nice work! 🚀

PR #5189, opened 2024-10-14
[BUG FIX] [MER-3917] set page scoring strategy to "Best" on ingestion, to match new page default

andersweinstein:
Page option settings are not represented in the course archive format; rather, hard-coded defaults are assigned on ingestion. Ingestion code had been assigning a default page scoring strategy of "Average". This does not match the default assigned to newly authored pages, namely "Best". This caused headaches for authors of migrated courses. This PR changes the hard-coded ingestion default to "Best" to match the new page default.

### DIFF
--- a/lib/oli/interop/ingest/processor/pages.ex
+++ b/lib/oli/interop/ingest/processor/pages.ex
@@ -96,7 +96,7 @@ defmodule Oli.Interop.Ingest.Processor.Pages do
       children: {:placeholder, :children},
       resource_type_id: {:placeholder, :resource_type_id},
       activity_type_id: Map.get(state.registration_by_subtype, Map.get(resource, "subType")),
-      scoring_strategy_id: Oli.Resources.ScoringStrategy.get_id_by_type("average"),
+      scoring_strategy_id: Oli.Resources.ScoringStrategy.get_id_by_type("best"),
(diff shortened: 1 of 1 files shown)


nicocirio (approved):
LGTM, now the scoring strategy on ingestion matches the default value for a new page:

image

Francisco-Castro:
@andersweinstein We could add a `ScoringStrategy.default_value()` function that returns the desired default value, which can be used in both scenarios.

andersweinstein:
Yes, it occurred to me it would be better to have a single-source for default option values so as to avoid mismatches like this. But defaults for some options depend on the page type (graded vs ungraded) and this was the only case we know about currently, so I just did the simple thing.

Francisco-Castro:
Right! Having a single source for default option values is a good idea. It improves code readability. Given that we have two different types of default values, we could implement something like `ScoringStrategy.default_for_graded()` and `ScoringStrategy.default_for_ungraded()`.

andersweinstein:
I was thinking something like `DeliverySettings.getDefaultValue(setting_name, gradedFlag) ` so all the different default settings for a page would be specified together in one place. 

Since this PR has already been merged, we could post it as new CHORE type issue to improve the code on this, though it is a low priority.

PR #5192, opened 2024-10-16
[BUG FIX] [MER-3921] Send reset password email

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-3921) to the ticket

### After
https://github.com/user-attachments/assets/a37adddc-3c76-4ea1-9b2a-39e9fea5423b

## Other fix
This PR also fixes the password reset link generator for an author (the page was crashing when clicking on "Generate")

### Before
https://github.com/user-attachments/assets/b59a88ef-3a80-4352-ab61-7e98db581e9a

### After
https://github.com/user-attachments/assets/5ee42ee8-5d84-4715-abf0-af9f9e140693

### DIFF
--- a/lib/oli_web/controllers/pow.ex
+++ b/lib/oli_web/controllers/pow.ex
@@ -11,21 +11,35 @@ defmodule OliWeb.PowController do
   @ttl :timer.minutes(24 * 60)
   @cache_config {PowResetPassword.Store.ResetTokenCache, ttl: @ttl}
 
-  def create_user_password_reset_link(%{"id" => id}) do
-    %{email: email} = _user = Repo.get(User, id)
+  @doc """
+  Create a password reset link for a user or author that will expire in
+  the amount of time defined in @ttl (currently 24 hours).
+
+  id: the id of the user or author
+  account_type: :user or :author
+  """
+  def create_password_reset_link(%{"id" => id}, account_type) do
+    %{email: email} =
+      account_type
+      |> schema_for_type()
+      |> Repo.get(id)
+
     params = %{"user" => %{"email" => email}}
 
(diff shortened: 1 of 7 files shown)


lib/oli_web/controllers/pow.ex
```diff
@@ -11,23 +11,37 @@ defmodule OliWeb.PowController do
...
 
-  def create_user_password_reset_link(%{"id" => id}) do
-    %{email: email} = _user = Repo.get(User, id)
+  @doc """
+  Create a password reset link for a user or author that will expire in
+  the amount of time defined in @ttl (currently 24 hours).
+
+  id: the id of the user or author
+  account_type: :user or :author
+  """
```
Francisco-Castro, line 20: Thanks for adding the docs. Including a spec would also be beneficial for a quick overview.

lib/oli_web/controllers/pow.ex
```diff
@@ -11,23 +11,37 @@ defmodule OliWeb.PowController do
...
+  def create_password_reset_link(%{"id" => id}, account_type) do
```
Francisco-Castro, line 21: I noticed that `send_user_password_reset_link` and `send_author_password_reset_link` are almost identical, except for the model type. It might be interesting to see whether they can be merged into a single function call, although this part will likely be replaced with the upcoming new sign-in system.

Francisco-Castro (approved):
LGTM. Tested locally and works as expected.

Thanks for adding the tests.

lib/oli_web/controllers/pow.ex
nicocirio: Yes, I agree. I didn't want to make more reactors since Pow me be removed from the codebase

PR #5195, opened 2024-10-17
[ENHANCEMENT] [MER-3618] Add Section Information to DOT prompt

manelli:
https://github.com/user-attachments/assets/71c58a22-46bd-4a19-b589-df606fdde311

(The first part of DOT's response 
_"I cannot find specific results related to information about moles in the current course curriculum"_ is because in my local environment I do not have the embeddings for the course)

See: https://eliterate.atlassian.net/browse/MER-3618

### DIFF
--- a/lib/oli/delivery/sections.ex
+++ b/lib/oli/delivery/sections.ex
@@ -5356,19 +5356,19 @@ defmodule Oli.Delivery.Sections do
     SectionCache.get_or_compute(section_slug, :section_prompt_info, fn ->
       instructors =
         section_slug
-        |> Sections.fetch_instructors()
+        |> fetch_instructors()
         |> Enum.map(&"#{&1.name} (#{&1.email})")
         |> Enum.join("\n")
 
       course_layout =
         section_slug
-        |> Sections.fetch_ordered_container_labels()
+        |> fetch_ordered_container_labels()
         |> Enum.map(&elem(&1, 1))
         |> Enum.join("\n")
 
       pages_with_links =
         section_slug
-        |> Sections.fetch_all_pages()
+        |> fetch_all_pages()
         |> Enum.map(
           &"#{&1.title} (#{Routes.page_delivery_path(OliWeb.Endpoint, :page, section_slug, &1.slug)})"
         )
--- a/lib/oli/delivery/sections.ex
+++ b/lib/oli/delivery/sections.ex
@@ -57,6 +57,7 @@ defmodule Oli.Delivery.Sections do
   alias Oli.Delivery.Paywall
   alias Oli.Delivery.Sections.PostProcessing
   alias Oli.Branding.CustomLabels
+  alias OliWeb.Router.Helpers, as: Routes
 
   require Logger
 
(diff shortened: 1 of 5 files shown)


darrensiegel (changes requested):
This should not use the `SectionCache`, we should treat the `SectionCache` as a deprecated approach for caching delivery information.  Instead, this should use the `SectionResourceDepot`.    

Instead of putting all of this in the context for each and every OpenAI API call, we could also take the approach of exposing this information via a function call.

lib/oli/delivery/sections.ex
```diff
@@ -57,6 +57,7 @@ defmodule Oli.Delivery.Sections do
   alias Oli.Delivery.Paywall
   alias Oli.Delivery.Sections.PostProcessing
   alias Oli.Branding.CustomLabels
+  alias OliWeb.Router.Helpers, as: Routes
```
darrensiegel, line 60: We should not have an OliWeb dependency in Oli code.

lib/oli/delivery/sections.ex
```diff
@@ -5349,4 +5350,34 @@ defmodule Oli.Delivery.Sections do
...
+    section_slug = Repo.get(Section, section_id).slug
+
+    instructors =
+      section_slug
+      |> fetch_instructors()
+      |> Enum.map(&%{name: &1.name, email: &1.email})
+
+    layout =
+      section_slug
+      |> fetch_ordered_container_labels()
```
darrensiegel, line 5364: This should be a cached call to a new function within `SectionResourceDepot`

lib/oli/delivery/sections.ex
```diff
@@ -5349,4 +5350,34 @@ defmodule Oli.Delivery.Sections do
...
+      |> Enum.map(&elem(&1, 1))
+
+    content =
+      section_slug
+      |> fetch_all_pages()
```
darrensiegel, line 5369: Same thing here, we should add a new function to `SectionResourceDepot` that returns the needed info for all pages.

darrensiegel (changes requested):

lib/oli/delivery/sections.ex
manelli: Thanks for the feedback, fixed it in f91fe64cc8

lib/oli/delivery/sections.ex
manelli: Fixed in f91fe64cc8

lib/oli/delivery/sections.ex
manelli: Thanks for the feedback, removed it as it is no longer necessary

lib/oli/delivery/sections.ex
```diff
@@ -5349,4 +5350,32 @@ defmodule Oli.Delivery.Sections do
...
+    %Section{customizations: customizations} = section = Repo.get(Section, section_id)
+
+    instructors =
+      section.slug
+      |> fetch_instructors()
+      |> Enum.map(&%{name: &1.name, email: &1.email})
+
+    layout =
+      section_id
+      |> SectionResourceDepot.get_section_resources_by_type_ids([ResourceType.id_for_container()])
```
darrensiegel, line 5364: For consistency sake, make one call to SectionResourceDepot.get_section_resources_by_type_ids for `[ResourceType.id_for_container(), ResourceType.id_for_page()]` then post process to separate the two types. 

Also, `Oli.Delivery.Sections` should not depend on or use `SectionResourceDepot`.   The dependency is the other way around.  So instead, move this entire function out of `Oli.Delivery.Sections` into a different contenxt, perhaps something `Converstation` specific.

darrensiegel (changes requested):

lib/oli/delivery/sections.ex
manelli: 👍 Fixed in f57ed5c4034806a6a70701b2c77afe0526850e50
Made one call to `SectionResourceDepot.get_section_resources_by_type_ids/2` and  moved the function to the `Oli.Conversation.Functions` module

darrensiegel (approved):

PR #5196, opened 2024-10-17
[BUG FIX] [MER-3783] grace period setting in product isnt honored in courses built from the product

Francisco-Castro:
Ticket: [MER-3783](https://eliterate.atlassian.net/browse/MER-3783)

This PR addresses the issue where exporting a project with products does not carry over the product settings, and these settings are not honored when ingested.

A test file that was incorrectly named was also fixed.

The added product fields were confirmed by @eliknebel.

[MER-3783]: https://eliterate.atlassian.net/browse/MER-3783?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/oli/delivery/sections/blueprint.ex
```diff
@@ -159,12 +159,13 @@ defmodule Oli.Delivery.Sections.Blueprint do
...
             "title" => title,
-            "requires_payment" => false,
-            "payment_options" => "direct_and_deferred",
-            "pay_by_institution" => false,
-            "registration_open" => false,
-            "grace_period_days" => 1,
-            "amount" => Money.new(:USD, "25.00"),
+            "requires_payment" => attrs["requires_payment"] || false,
+            "payment_options" => attrs["payment_options"] || "direct_and_deferred",
+            "pay_by_institution" => attrs[""] || false,
```
nicocirio, line 164: Should it be `attrs["pay_by_institution"] || false"? If so, perhaps we need to parse the value if it is not a boolean

lib/oli/delivery/sections/blueprint.ex
```diff
@@ -159,12 +159,13 @@ defmodule Oli.Delivery.Sections.Blueprint do
...
+            "registration_open" => attrs["registration_open"] || false,
+            "grace_period_days" => attrs["grace_period_days"] || 1,
+            "amount" =>
+              Money.new(attrs["amount"]["currency"] || :USD, attrs["amount"]["amount"] || "25.00"),
```
nicocirio, line 168: Should we only assign the default currency and amount values if both are missing? I am imagining the improbable case where the original value is EUR 30, and we end up having USD 30 if the currency value is missing.

nicocirio (commented):
Nice work! I left some comments

lib/oli/delivery/sections/blueprint.ex
Francisco-Castro: Good catch! I added a test to cover the complete cycle—export and import.

lib/oli/delivery/sections/blueprint.ex
Francisco-Castro: There were already default values for the amount, so I just set the context for handling those values during the import. I couldn't find where the currency is defined. @eliknebel, do we use any currency other than 'USD'?

eliknebel: Yes I agree with Nico, we should only assign the default currency if both amount and currency are missing to avoid unintentionally changing the underlying value

darrensiegel (approved):

PR #5197, opened 2024-10-17
[BUG FIX] [MER-3923] Modals do not show while editing pages

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-3923) to the ticket

### After

https://github.com/user-attachments/assets/afb886fd-9018-4705-bc97-0a7dcd25fe40


andersweinstein (approved):
Works for me.

PR #5199, opened 2024-10-18
[BUG FIX] [MER-3552] Cookies header limit exceeded

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-3552) to the ticket

This PR reduces the number of keys in the session cookie and, as discussed in our last standup meeting, it adds a plug to monitor the size of the request/response headers. 
The `HeaderSizeLogger` plug logs a warning and sends details to AppSignal when the header size is bigger than a threshold (threshold = HTTP_MAX_HEADER_VALUE_LENGTH set in env, or 4096 if not provided).

Here is an example of the log warning when I set a small amount to force it:

Screenshot 2024-10-18 at 9 33 32 AM


lib/oli_web/plugs/header_size_logger.ex
```diff
@@ -0,0 +1,76 @@
...
+
+  defp maybe_log_response_header_size(conn, opts) do
+    response_header_size = calculate_headers_size(conn.resp_headers)
+
+    if response_header_size > @threshold do
+      log_to_analytics("Response", response_header_size, conn, opts)
+    end
+
+    conn
+  end
```
gastonabella, line 49: Nit: we could also pass the `type` as a function param and merge these two functions.

lib/oli_web/plugs/header_size_logger.ex
```diff
@@ -0,0 +1,76 @@
...
+
+  defp calculate_headers_size(headers) do
```
gastonabella, line 51: We could move this function to a shared module (i.e. `SessionUtils`) and make it public, so we can reuse it here, in the test module and anywhere else we need it in the future.

gastonabella (commented):
Great work optimizing the session data!
I left some minor comments, let me know what you think about them :)

lib/oli_web/plugs/header_size_logger.ex
```diff
@@ -0,0 +1,76 @@
...
+  their HTTP Header sizes are growing beyond the max configured size.
+  """
+
+  import Plug.Conn
+  require Logger
+
+  @threshold Application.compile_env(:oli, OliWeb.Endpoint, [])
+             |> Keyword.get(:http, [])
+             |> Keyword.get(:protocol_options, [])
+             |> Keyword.get(:max_header_value_length, 4096)
```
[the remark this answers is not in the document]
nicocirio: Yes, it was already added before. But you made me realize we do not need the default here

gastonabella (approved):
LGTM! Thanks for the changes

PR #5209, opened 2024-10-22
[FEATURE] [MER-3905] Admin manage student payment status from paid to unpaid

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-3905) to the ticket

https://github.com/user-attachments/assets/c53330bc-4bd0-4eb9-835b-d8aab0672bdf

### DIFF
--- a/lib/oli_web/components/delivery/actions/actions.ex
+++ b/lib/oli_web/components/delivery/actions/actions.ex
@@ -44,7 +44,8 @@ defmodule OliWeb.Components.Delivery.Actions do
        has_payment: has_payment,
        current_user: current_user,
        is_admin: Accounts.has_admin_role?(current_user),
-       is_suspended?: is_suspended?
+       is_suspended?: is_suspended?,
+       payment_status_uuid: UUID.uuid4()
      )}
   end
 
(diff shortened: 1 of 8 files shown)


lib/oli_web/components/delivery/actions/actions.ex
```diff
@@ -54,100 +54,156 @@ defmodule OliWeb.Components.Delivery.Actions do
...
-      <% end %>
+        <% end %>
+      </div>
     </div>
     """
   end
 
+  def handle_event("force_reset_payment_status", _params, socket) do
+    {:noreply, assign(socket, payment_status_uuid: UUID.uuid4())}
+  end
```
Francisco-Castro, line 190: Any reason why we are resetting the form ID after canceling the modal?

priv/repo/migrations/20241021182426_add_invalidated_by_user_id.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Oli.Repo.Migrations.AddInvalidatedByUserId do
+  use Ecto.Migration
+
+  def change do
+    alter table(:payments) do
+      add :invalidated_by_user_id, :integer, null: true
```
Francisco-Castro, line 6: Why are we setting this as an integer and not as a reference to a user?

Francisco-Castro (approved):
Good job!!

I tested it locally, and it worked as expected.

I left a couple of comments and questions that don’t block the expected criteria, so I'm going to go ahead and approve it.

lib/oli_web/components/delivery/actions/actions.ex
nicocirio: Yes, I wanted to force the browser to set the select input back to its original value in case the admin cancels or closes the confirmation modal. A "hacky" way of addressing this is forcing Phoenix to render the component again (-> changing its id). 
In the attached video you can see this behaviour in action.

priv/repo/migrations/20241021182426_add_invalidated_by_user_id.exs
nicocirio: To be consistent with how we implemented the `bypassed_by_user_id` field.

Francisco-Castro: It makes sense.

lib/oli_web/components/delivery/actions/actions.ex
Francisco-Castro: I got it, otherwise the form is going to stuck with an incorrect value. Nice work.

PR #5174, opened 2024-10-07
[FEATURE] [MER-3876] Add breadcrumbs to Curriculum

Francisco-Castro:
Ticket: [MER-3876](https://eliterate.atlassian.net/browse/MER-3876)

This PR added breadcrumbs to Curriculum.

https://github.com/user-attachments/assets/ef632606-9417-4d82-8e8f-22a256ca72bc

[MER-3876]: https://eliterate.atlassian.net/browse/MER-3876?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5221, opened 2024-10-28
[BUG FIX] [MER-3932] issue when adding activities to activity bank

Francisco-Castro:
Ticket: [MER-3932](https://eliterate.atlassian.net/browse/MER-3932)

This PR fixes an issue where navigating to the activity bank did not load the necessary JavaScript files to render the React components, making the page unusable.

It can be seen in the following video that the JS files necessary to render the React components are loaded only once—when entering the activity bank page. In subsequent visits, the JS files are not loaded again, allowing for a smooth transition between pages within the same live session.

https://github.com/user-attachments/assets/70f1d2d6-d1c3-40c0-9db4-9310a20356ba

[MER-3932]: https://eliterate.atlassian.net/browse/MER-3932?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/oli_web/live/workspaces/utils.ex
```diff
@@ -213,6 +216,13 @@ defmodule OliWeb.Workspaces.Utils do
     end
   end
 
+  def is_navigate_route(route) do
```
darrensiegel, line 219: I'm not sure what this function does.  Can you add some documentation for it, as well as include a unit test?

darrensiegel (dismissed):

lib/oli_web/live/workspaces/utils.ex
Francisco-Castro: This should be a private function. Its purpose is to identify the type of navigation needed for each route: either `href` or `navigate`. This identification is then used to set the appropriate navigation logic. Although it's not the most elegant approach, as it disrupts our smooth navigation between live pages within the same live session, it's necessary. Without this distinction, we won't load the required JS files to run the React components on the view.

In researching this issue, I found a similar case in this post: https://elixirforum.com/t/is-it-possible-to-have-page-specific-javascript-with-liveview/32614. I tried a few experiments based on this but couldn’t get it to work, so I’ve added it to my to-do list.

This issue also affects our breadcrumbs, as note [here](https://github.com/Simon-Initiative/oli-torus/pull/5174/files#r1790755239).

nicocirio: Fran, as we discussed the other day, a way to get the scripts loaded in an async way is reusing the already existing `LoadSurveyScripts` hook (there are some references in the code where we use this strategy). I am almost sure this approach will work but, if it doesn't, what do you think about moving this page out of the live_session? In this way you will "force" to fully reload the page (and scripts should load from scratch), and will remove the need of this `is_navigate_route` helper function.

Francisco-Castro:
I put this PR in draft while I'm exploring what @nicocirio suggested ☝️

lib/oli_web/live/workspaces/course_author/activity_bank_live.ex
```diff
@@ -13,44 +12,55 @@ defmodule OliWeb.Workspaces.CourseAuthor.ActivityBankLive do
...
+        <.loader />
+      <% end %>
     </div>
-    <%= React.component(@ctx, "Components.ModalDisplay", %{}, id: "modal-display") %>
     """
   end
+
+  @impl true
+  def handle_event("survey_scripts_loaded", %{"error" => _}, socket) do
+    {:noreply, assign(socket, error: true)}
```
nicocirio, line 60: What do you think about removing the spinner and informing the user an error has occurred? If not the user will see the spinner forever. Something like:

```suggestion
    {:noreply, socket |> assign(scripts_loaded: true) |> put_flash(:error, "The page had an error while loading")}
```

nicocirio (changes requested):
Nice work Fran! Thanks for taking the async-scripts-loading approach! this will ensure the fastest navigation between pages of the same live_session.

I requested changes since we need to handle the (improbable) case where the scripts can't be loaded and inform the user the page won't work as expected.

assets/src/hooks/load_survey_scripts.ts
```diff
@@ -6,11 +6,19 @@ export const LoadSurveyScripts = {
...
           new Promise<void>((resolve, reject) => {
-            const script = document.createElement('script') as HTMLScriptElement;
-            script.src = source;
-            script.addEventListener('load', () => resolve());
-            script.addEventListener('error', () => resolve());
-            head.appendChild(script);
+            const isLoaded = Array.from(document.getElementsByTagName('script')).some((script) =>
+              script.src.includes(source),
+            );
+            if (!isLoaded) {
```
nicocirio, line 12: Have we tested this new logic does not affect in an unexpected way other pages that use this same hook? A smoke-testing would probably be enough.

Francisco-Castro: Good idea!  tested another part of our code that uses the hook, and it worked well.

Francisco-Castro:
@nicocirio Thanks for reviewing my PR. I added some commits to address your [comment](https://github.com/Simon-Initiative/oli-torus/pull/5221#pullrequestreview-2415311134). Could you please take another look at your convenience?

nicocirio (approved):
Nice work! 🚀

PR #5225, opened 2024-10-29
[FEATURE] [MER-3828] Home screen course progress card

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-3828) to the ticket

### Demo

https://github.com/user-attachments/assets/4bc84d1f-0f3a-4f1d-9b1b-cccbc85c7e30

### Demo (excludes pages with progress < 100% in the card progress calculation)
"Let's practice" is a practice page where the student has answered 1 out of 2 activities, so it's progress for that page is 50%. And has completed another 2 pages (100%). The course has 15 pages in total.

https://github.com/user-attachments/assets/258e7633-294e-4368-b012-db971a876f7e

### Progress match between student's workspace and student's HOME page

https://github.com/user-attachments/assets/fe6816f6-ff1a-4de1-90db-727ef3680eb1


lib/oli/delivery/depot.ex
```diff
@@ -113,4 +113,29 @@ defmodule Oli.Delivery.Depot do
...
+  The list can be a simple keyword list of field values to match on as an AND:
+
+  [resource_type_id: 3, graded: false]
+
+  or a list of tuples with the field name, operator, and value (all of which are
+  ANDed together):
+
+  [{:duration, {:between, 5, 10}}, {:graded, {:=, true}}]
+  """
+  def count(%DepotDesc{} = depot_desc, table_id, conditions) do
```
darrensiegel, line 132: This seems like a nice addition to the core `Depot`, but I don't see it being used anywhere in this PR.  Also, if we want to keep it - we should add some unit tests for it.

darrensiegel (changes requested):

lib/oli/delivery/depot.ex
nicocirio: Yes, you are right. At first, I built this new depot function to use it [here
](https://github.com/Simon-Initiative/oli-torus/pull/5225/commits/8358f26903120a72ce377ffe2cfd6c028529a9be#diff-9cd70d1d862781e5ecd2a468b858c410ad804e68e0bb40ea8d86491a192649d9R106-R116)
But then I took another approach and deleted it in [this commit](https://github.com/Simon-Initiative/oli-torus/pull/5225/commits/55210bebb6046b92f862af35389360a416e741c8), but decided to keep the Depot function because I think it will probably be useful in the future. So, if you agree, I can add some tests for it and keep it, WDYT?

darrensiegel: Yes, let's keep it

darrensiegel (approved):

PR #5226, opened 2024-10-29
[FEATURE] [MER-3971] Create icons page (only for DEV)

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-3971) to the ticket

https://github.com/user-attachments/assets/1ecd2b43-5883-4bdb-a3e1-eb975f612a46

### DIFF
--- a/lib/oli_web/icons.ex
+++ b/lib/oli_web/icons.ex
@@ -1,6 +1,8 @@
 defmodule OliWeb.Icons do
   @moduledoc """
   A collection of icons used in the OliWeb application.
+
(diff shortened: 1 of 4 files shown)


simonchoxx (approved):
This is Awesome!!!! 😍🚀👏

Francisco-Castro (commented):
Good job @nicocirio 🏅 

After watching the video, I noticed that some icons are not visible. After experimenting with the code, I found that in most cases, the `fill` property is set to`"none"` in the `svg` element. We could adjust this to something other than `none` so that we can make the icons visible. Just an idea for the next iteration.

PR #5231, opened 2024-10-31
[BUG FIX] [MER-3412] Image component has fixed aspect ratio based on image dimensions

dtiwarATS:
Hey @bsparks, could you please look at this PR. Thanks

### DIFF
--- a/assets/src/components/parts/janus-image/ImageAuthor.tsx
+++ b/assets/src/components/parts/janus-image/ImageAuthor.tsx
@@ -50,7 +50,15 @@ const ImageAuthor: React.FC<AuthorPartComponentProps<ImageModel>> = (props) => {
 
     const currentWidth = imageContainerRef.current.width;
     const currentHeight = imageContainerRef.current.height;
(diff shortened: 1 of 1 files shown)


assets/src/components/parts/janus-image/ImageAuthor.tsx
```diff
@@ -50,7 +50,15 @@ const ImageAuthor: React.FC<AuthorPartComponentProps<ImageModel>> = (props) => {
 
     const currentWidth = imageContainerRef.current.width;
     const currentHeight = imageContainerRef.current.height;
-    if (naturalWidth <= 0 || naturalHeight <= 0 || currentWidth <= 0 || currentHeight <= 0) {
+
+    const updatedWidth = updatedModel?.width || 0;
```
dtiwarATS, line 54: this will fix the latest comment in the ticket ->

`For some of the lessons, we have custom CSS where we delete the width of an image, then it defaults to full screen. Right now, if we delete the width, it automatically returns to 100. Could we add in a mechanism to not resize when the width is deleted? `

darrensiegel (approved):

PR #5240, opened 2024-11-07
[BUG FIX] [MER-3918] Simple Author and Advanced Author - Slider component - Setting intervals

dtiwarATS:
Hey @bsparks could you please look at the PR? Thanks.

There are 2 changes made for this card:
- We have added a description to the field that says '**_Value cannot be smaller than 1/100 of the range between the min and max values_**'
image
- If user enters a value that increases the threshold, system will assign the max threshold value in the interval field.

https://github.com/user-attachments/assets/8e9eb0bb-11b8-4051-9ebd-143a86ce451a

### DIFF
--- a/assets/src/components/parts/janus-slider/SliderAuthor.tsx
+++ b/assets/src/components/parts/janus-slider/SliderAuthor.tsx
@@ -1,10 +1,11 @@
 import React, { CSSProperties, useEffect, useRef, useState } from 'react';
 import { AuthorPartComponentProps } from 'components/parts/types/parts';
+import { clone } from 'utils/common';
 import './Slider.scss';
 import { SliderModel } from './schema';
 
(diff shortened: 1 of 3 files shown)


assets/src/components/parts/janus-slider/SliderAuthor.tsx
```diff
@@ -71,6 +72,18 @@ const SliderAuthor: React.FC<AuthorPartComponentProps<SliderModel>> = (props) =>
     if (snapInterval) {
       const options = [];
       const numberOfTicks = (maximum - minimum) / snapInterval;
+      const numberOfTricksThreshold = 100;
```
bsparks, line 75: typo: "tricks" instead of "ticks"

eliknebel (approved):

PR #5247, opened 2024-11-12
[BUG FIX] [MER-4004] Fix snapshot creation for manually graded activities when Oban is enabled.

manelli:
When Oban was enabled, scoring a manually graded activity would queue the snapshot creation instead of creating it immediately. This caused a false error message to appear in the UI. 

See: https://eliterate.atlassian.net/browse/MER-4004


Francisco-Castro (approved):
Nice job!

I tested locally and it works expected.

PR #5250, opened 2024-11-12
[FEATURE] [MER-3807] My assignments student view

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-3807) to the ticket

- [X] Light and dark mode
- [X] Back button to return to assignments
- [X] Responsive (added ellipsis when we have an overflow on the title)
- [X] Reduced memory footprint

https://github.com/user-attachments/assets/6f83c068-be96-46a2-99c0-cf57d41293af


lib/oli_web/live/delivery/student/assignments_live.ex
```diff
@@ -0,0 +1,245 @@
...
+      )
+    end)
+  end
+
+  def handle_info(:gc, socket) do
+    # manually garbage collect to reduce memory usage after mount/3
+    :erlang.garbage_collect(socket.transport_pid)
+    :erlang.garbage_collect(self())
+    {:noreply, socket}
+  end
```
gastonabella, line 61: Since this piece of code is starting to repeat across different LiveViews, I'm beginning to think that it might be a good idea to extract it to the `OliWeb` module as an option of the `live_view` macro.

Something like
```elixir
use OliWeb, :live_view, garbage_collect: true
```

We could discuss it further and do the refactor in a separate PR. I'm just opening the discussion :)

gastonabella (approved):
Great work!

lib/oli_web/live/delivery/student/assignments_live.ex
nicocirio: I love the idea! 🚀

PR #5252, opened 2024-11-13
[ENHANCEMENT] [MER-3997] Add CAPI to Track Popup Opening by Student

dtiwarATS:
Hey @bsparks, could you please look at the PR. Thanks

### DIFF
--- a/assets/src/components/parts/janus-popup/Popup.tsx
+++ b/assets/src/components/parts/janus-popup/Popup.tsx
@@ -55,6 +55,11 @@ const Popup: React.FC<PartComponentProps<PopupModel>> = (props) => {
           type: CapiVariableTypes.BOOLEAN,
           value: false,
         },
+        {
+          key: 'userOpened',
+          type: CapiVariableTypes.BOOLEAN,
+          value: false,
+        },
       ],
     });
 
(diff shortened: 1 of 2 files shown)


bsparks:
maybe I'm not remembering right, but this seems like a weird way to use CAPI... maybe it's common already?

assets/src/components/parts/janus-popup/Popup.tsx
```diff
@@ -209,6 +214,21 @@ const Popup: React.FC<PartComponentProps<PopupModel>> = (props) => {
     };
   }, [props.notify]);
 
+  useEffect(() => {
+    if (showPopup) {
```
bsparks, line 218: does this flag actually "prove" that the user did it? seems like we'd need a click event on a button or something?

bsparks (approved):

dtiwarATS:
> maybe I'm not remembering right, but this seems like a weird way to use CAPI... maybe it's common already?

Actually, it's not a CAPI. it's just a new attribute in adaptivitySchema of pop-up. May be the title is confusing.

assets/src/components/parts/janus-popup/Popup.tsx
dtiwarATS: If this flag is true then it means the pop-up was opened and student had read it hence, I set the `userOpened `property to it.

If we  go with the button click then the problem with that it, it the pop-up is opened by default then students will only close the pop-up so the main pop-up button will never get clicked hence I didn't set it on button click.

PR #5262, opened 2024-11-21
[BUG FIX] [MER-4033] fix evaluation of signed floats without leading zero

andersweinstein:
Torus accepts floating point input without leading zero, e.g. `.333`. However, evaluation code was not properly handling this form when a sign was prefixed, e.g. `-.333`, causing answers of this form to be marked incorrect. Error was caused by a step where code converts the zero-less form to a zero-prefixed form; regexp used there was not allowing for possible sign.

This also adds a few unit tests for this case.


gastonabella (approved):
Looks good!

PR #5265, opened 2024-11-23
[ENHANCEMENT] [MER-4012] first access to a large course real chem has a significant wait time

Francisco-Castro:
Ticket: [MER-4012](https://eliterate.atlassian.net/browse/MER-4012)

This PR contains 2 parts:

1. Moves the triggering of the Depot from `Student.Index` to the point where an instructor creates a `CourseSection`. This change aims to prevent the first student from triggering the Depot, potentially improving their experience.

3. Introduces DepotWarmer, a module that preloads entries into the Depot system when the application is deployed.

[MER-4012]: https://eliterate.atlassian.net/browse/MER-4012?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/oli_web/live/new_course/new_course.ex
```diff
@@ -432,13 +434,16 @@ defmodule OliWeb.Delivery.NewCourse do
...
+    end)
 
-    {:noreply,
-     redirect(socket,
-       to: Routes.live_path(OliWeb.Endpoint, OliWeb.Sections.OverviewView, section_slug)
-     )}
+    socket
+    |> put_flash(:info, "Section successfully created.")
+    |> redirect(to: ~p"/sections/#{section_slug}/manage")
+    |> noreply()
```
manelli, line 446: Nitpick: helper functions can sometimes make code look "prettier," they aren't always necessary, especially when the standard approach is already straightforward and widely understood. The current implementation of the `{:no_reply, socket}` follows the standard pattern, which is often preferable for consistency and maintainability
WDYT?

lib/oli/delivery/sections/section_resource_depot.ex
```diff
@@ -178,12 +177,31 @@ defmodule Oli.Delivery.Sections.SectionResourceDepot do
...
+        where: sr.section_id == ^section_id,
+        where: sr.resource_type_id in [^page, ^container, ^objective],
         select: sr
 
     results = Repo.all(query)
 
     Depot.clear_and_set(@depot_desc, section_id, results)
   end
+
+  def fetch_recently_active_sections() do
```
manelli, line 189: Should we move this section to the Sections context since it query information related to the sections, not the depot?
something like `fetch_sections_updated_since(time_ago)`
WDYT

manelli (commented):
Hi @Francisco-Castro left some minor comments. Let me know what you think. Awesome work 💯

lib/oli_web/live/new_course/new_course.ex
Francisco-Castro: It makes sense what you are saying. What you’re saying makes sense. What I’ve found is that we have a combination of two patterns in our codebase:

Type 1:

```
{:noreply,
  assign(socket,
  key_1: :val_1,
  key_2: :val_2,
  ...: ...
  key_N: :val_N,
  )}
```

Type 2:

```
socket = assign(socket,
  key_1: :val_1,
  key_2: :val_2,
  ...: ...
  key_N: :val_N,
  )

{:noreply,, socket}
```

There’s nothing inherently wrong with these approaches, but there are trade-offs. Type 1 is widely used in our codebase. However, for me, it’s not immediately obvious where the second tuple ends, which creates a lot of pain points when working with this pattern. It’s not very reader-friendly. While writing this pattern might feel straightforward, reading it can be challenging. As writers, we have context because we’re starting the sentence, but readers don’t always have the same advantage.

On the other hand, Type 2 is more structured and easier to follow. But it’s sometimes not the preferred choice among developers, perhaps because we approach writing and reading code differently.

The approach I’ve taken here is inspired by similar patterns I’ve seen in other projects. Additionally, [German Velazco](https://www.elixirstreams.com/tips/awesome-ok-noreply-liveview-helpers?utm_source=elixir-merge) has expressed support for this style.

My intention in adding this function is not just to make the code “prettier” but to make it easier to read. Given that we frequently use the Type 1 approach in our project, using `noreply/1` plays well with the pipe operator and doesn’t introduce significant overhead when applied.

Please let me know what you think.

manelli (approved):
Awesome job! 🚀

PR #5272, opened 2024-11-27
[BUG FIX] [MER-3981] Illegible text in dark mode

simonchoxx:
[MER-3981](https://eliterate.atlassian.net/browse/MER-3981)

This PR fixes the color of the texts in the Create Project modal to make them more readable in dark mode. 

These changes include the color of the placeholder text.

The textarea for adding/editing a bibliography entry has been fixed in this [PR](https://github.com/Simon-Initiative/oli-torus/pull/5212).

- Modal to Create Project before

image

- Modal to Create Project after

image

[MER-3981]: https://eliterate.atlassian.net/browse/MER-3981?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/assets/css/app.css
+++ b/assets/css/app.css
@@ -216,17 +216,6 @@ Not supports in Firefox and IE */
   height: 0;
 }
 
-/* set placeholder text color */
-#project-title-placeholder::placeholder {
-  color: #9ca3af;
-}
-
-/* set placeholder text color in dark mode */
-.dark #project-title-placeholder::placeholder {
-  color: #eeebf5;
-  opacity: 0.7;
-}
-
 /* Firefox customized styles */
 
 @-moz-document url-prefix() {
--- a/assets/css/app.css
+++ b/assets/css/app.css
@@ -216,6 +216,17 @@ Not supports in Firefox and IE */
   height: 0;
 }
 
+/* set placeholder text color */
+#project-title-placeholder::placeholder {
+  color: #9ca3af;
+}
+
+/* set placeholder text color in dark mode */
+.dark #project-title-placeholder::placeholder {
+  color: #eeebf5;
+  opacity: 0.7;
+}
+
 /* Firefox customized styles */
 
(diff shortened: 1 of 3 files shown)


lib/oli_web/live/projects/create_project_modal.ex
```diff
@@ -36,10 +39,11 @@ defmodule OliWeb.Projects.CreateProjectModal do
               <div class="form-label-group">
                 <%= text_input(f, :title,
                   required: true,
+                  id: "project-title-placeholder",
                   class: "block min-w-full",
```
nicocirio, line 43: We could get the same result with tailwind:

`class: "block min-w-full placeholder-[#9ca3af] dark:placeholder-[#eeebf5]/70"`

I think it is more "straightforward" and avoids us having to define an id and a specific class for that id in app.css. WDYT?

assets/css/app.css
```diff
@@ -216,6 +216,17 @@ Not supports in Firefox and IE */
...
+#project-title-placeholder::placeholder {
+  color: #9ca3af;
+}
+
+/* set placeholder text color in dark mode */
+.dark #project-title-placeholder::placeholder {
+  color: #eeebf5;
+  opacity: 0.7;
+}
+
```
nicocirio, line 229: This could be omitted if we decide to implement it with Tailwind directly on the input

nicocirio (commented):
LGTM. I believe a pure Tailwind approach might be better

lib/oli_web/live/projects/create_project_modal.ex
simonchoxx: Nice. I didn't know about this tailwind class. Thanks!

nicocirio (approved):
Thanks for taking into account the suggestion! 😄

PR #5273, opened 2024-11-29
[ENHANCEMENT] [MER-3247] Add back button to adaptive page review

gastonabella:
[MER-3247](https://eliterate.atlassian.net/browse/MER-3247)

Fix the back button URL when reviewing adaptive page attempts to redirect back to the previously visited page instead of a hardcoded path.


Francisco-Castro (approved):
LGTM.

I tested it locally, and it worked as expected.

PR #5274, opened 2024-11-29
[FEATURE] [MER-3873] Likert activity for instructors

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-3873) to the ticket

Different likert activities (with different amount of questions and options, even different text length for the options) rendered correctly in :

https://github.com/user-attachments/assets/ca6cfb51-65db-4820-b49d-b728da35cccb

This PR sets the ground for rendering light/dark Vegalite Visualizations that react to changes made by the user in the mode selected ("dark", "light", "auto"). Changes can be seen on `assets/src/components/misc/VegaLiteRenderer.tsx`

### Some refactoring
While working on this feature I found that three tabs (Scored Activities, Practice Activities and Surveys) from the Insights instructor view had duplicated code. So, this PR exctracts common logic of those 3 live_component modules (OliWeb.Components.Delivery.{PracticeActivities, ScoredActivities, Surveys}) into a new `OliWeb.Delivery.ActivityHelpers` module.
And some ecto queries from that migrated code where moved to their corresponding context module as well (`Oli.Analytics.Summary` and `Oli.Delivery.Attempts.Core`)


lib/oli_web/components/delivery/activity_helpers.ex
```diff
@@ -0,0 +1,624 @@
...
+
+  alias Oli.Analytics.Summary
+  alias Oli.Delivery.Attempts.Core
+  alias Oli.Delivery.Sections.Section
+  alias Oli.Publishing.DeliveryResolver
+  alias OliWeb.ManualGrading.RenderedActivity
+
+  def get_activities(assessment_resource_id, section_id, student_ids, filter_by_survey \\ false),
+    do:
+      Core.get_evaluated_activities_for(
```
darrensiegel, line 17: This is a really nice isolation of redundant code that had been spread across Practice, Summative and Survey views.  But we really should be taking this opportunity to move away from querying across ActivityAttempt and ResourceAttempt tables.   Instead, this information can be much more efficiently retrieved by querying across Summary tables and then joining to the corresponding SectionResource record (through section id and resource id) to get the title of the resource.

darrensiegel (changes requested):
We can probably land this as is, but we need to file another ticket (and prioritize) for the work to move away from the resource hierarchy querying.

PR #5275, opened 2024-12-02
[ENHANCEMENT] [MER-4022] Do not show Assignments navigation if section has no assignments.

manelli:
Removes the assignment link in the nav bar for sections that do not have assignments.

https://github.com/user-attachments/assets/aac17c70-945f-439b-9c3e-dc3c76b133dd

See: https://eliterate.atlassian.net/browse/MER-4022


nicocirio (approved):
LGTM 🚀

PR #5283, opened 2024-12-04
[CHORE] [MER-4090] only create activity_attempts_submitted index if it does not already exist

eliknebel:
https://eliterate.atlassian.net/browse/MER-4090

We need the ability to run the activity_attempts_submitted index before migration to prevent extended downtime and CI disconnect to make the deployment go smoother.

This PR adds the `IF NOT EXISTS` specification to the index creation during migration to prevent the migration from throwing an error when the index was created ahead of time.


simonchoxx (approved):
LGTM!

PR #5290, opened 2024-12-06
[BUG FIX] [MER-4099] time limit should always be present on prologue page

Francisco-Castro:
Ticket: [MER-4099](https://eliterate.atlassian.net/browse/MER-4099)

This PR adds logic to display a page term when a time limit is set, but no due date is specified. The message is as follows:

> You have **X minutes** to complete the assessment. If you exceed this time, it will be marked as late.

It also pluralizes the message when more than 1 minute is specified.

Additionally, it conjugates the verb correctly based on whether the due date has passed.

https://github.com/user-attachments/assets/98782b81-8a93-4433-93bb-3bf654a2cc18

[MER-4099]: https://eliterate.atlassian.net/browse/MER-4099?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli_web/live/delivery/student/utils.ex
+++ b/lib/oli_web/live/delivery/student/utils.ex
@@ -210,34 +210,46 @@ defmodule OliWeb.Delivery.Student.Utils do
   attr :effective_settings, Oli.Delivery.Settings.Combined
   attr :ctx, SessionContext
 
-  defp page_due_term(%{effective_settings: %{end_date: end_date}} = assigns)
-       when not is_nil(end_date) do
+  defp page_due_term(%{effective_settings: %{end_date: nil, time_limit: time_limit}} = assigns)
+       when time_limit > 0 do
+    minute_label =
+      case time_limit do
+        1 -> "minute"
+        _ -> "minutes"
+      end
+
+    assigns = assign(assigns, minute_label: minute_label)
+
     ~H"""
-    <div>
-      <span>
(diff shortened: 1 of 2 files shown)


lib/oli_web/live/delivery/student/utils.ex
```diff
@@ -210,34 +210,46 @@ defmodule OliWeb.Delivery.Student.Utils do
...
-    </div>
+    This assignment is <b>not yet scheduled.</b>
+    """
+  end
+
+  defp page_due_term(%{effective_settings: %{end_date: end_date}} = assigns) do
+    verb_form =
+      case DateTime.compare(DateTime.utc_now(), end_date) do
+        :gt -> "was"
+        :lt -> "is"
```
gastonabella, line 239: Super edge case: if the dates are equal (very unlikely) this will fail. You could do `_ -> "is"` instead :)

gastonabella (approved):
LGTM!

lib/oli_web/live/delivery/student/utils.ex
Francisco-Castro: Good point! I considered that while writing the code but forgot to include it as a catch-all! However, it's very unlikely for this situation to occur since we're checking the time at the millisecond level.

PR #5307, opened 2024-12-17
[BUG FIX] [MER-4117] Infinite loading and errors thrown on adaptive pages

dtiwarATS:
Hey @bsparks, could you please look at the PR. Thanks.

Ticket#->  [MER-4117](https://eliterate.atlassian.net/browse/MER-4117)

**Root Cause:**

It appears that the server is now preventing the client from creating more attempts than the maximum allowed, as defined in the activity configuration, which is the correct behavior. For example, if an activity has a maximum attempt limit of 5, the server returns an error when the client tries to create a 6th attempt. This results in an undefined value being assigned to the attempt variable in the triggercheck.tsx page, causing the student to become stuck on the screen.

**Note:** I believe this issue has always existed, but it only became apparent now because the server has started enforcing the restriction.

**Use Case:**
Activity A has max attempt set to 3
There is a trap state that will get triggerd if `attemptNumber >3`. This trap state does only set the correct answers and shows feedback.
when this trap state gets triggered, since it's a wrong trap state (e.g. too many attempt), the client is trying to create a new attempt however since only 3 attempt was allowed for this screen, we created 3 attempts with when user has 3 wrong attempts so the server does not allow the client to create attempt because limit is exhausted.

**Solution:**
The client now verifies if there are any remaining attempts before requesting the server to create a new attempt.

Additionally, changes were made to handle the 'session.attemptNumber' in resume mode.
Previously, if a student had already made 3 attempts and refreshed the screen, `'session.attemptNumber'` was being reset to 1.
This behavior has been corrected. Now, `'session.attemptNumber'` is restored to the correct value (e.g., 3) from the resource attempt state, instead of being reset to the default value of 1.

[TRIAGE-1240]: https://eliterate.atlassian.net/browse/TRIAGE-1240?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[MER-4117]: https://eliterate.atlassian.net/browse/MER-4117?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/assets/src/apps/delivery/store/features/adaptivity/actions/triggerCheck.ts
+++ b/assets/src/apps/delivery/store/features/adaptivity/actions/triggerCheck.ts
@@ -364,12 +364,16 @@ export const triggerCheck = createAsyncThunk(
     }
     //Even If the check result contains a wrong trap state and has a navigation to different screen, we should not create a new attempt for that screen because
     // the student will be navigated to different screen so it does not make sense to create a new attempt for the current screen
-    if (!isCorrect && !doesCheckResultContainsNavigationToDifferentScreen) {
+    if (
+      !isCorrect &&
+      !doesCheckResultContainsNavigationToDifferentScreen &&
+      attempt?.hasMoreAttempts
+    ) {
       /* console.log('Incorrect, time for new attempt'); */
       const { payload: newAttempt } = await dispatch(
         createActivityAttempt({ sectionSlug, attemptGuid: currentActivityAttemptGuid }),
       );
-      attempt = newAttempt;
+      if (attempt) attempt = newAttempt;
       const updateSessionAttempt: ApplyStateOperation[] = [
         {
           target: 'session.attemptNumber',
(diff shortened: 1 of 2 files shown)


bsparks (approved):

darrensiegel (changes requested):
This should not target the `prerelease-v0.29.0` branch as this fix will need to actually be a part of a hotfix (29.1).   That hotfix branch will be created momentarily

dtiwarATS:
> This should not target the `prerelease-v0.29.0` branch as this fix will need to actually be a part of a hotfix (29.1). That hotfix branch will be created momentarily

Done.

dtiwarATS:
Changed the target version of PR to hotfix-v0.29.1

darrensiegel (approved):

PR #5320, opened 2024-12-20
[BUG FIX] [MER-4120] One at a time setting can prevent submissions

simonchoxx:
[MER-4120](https://eliterate.atlassian.net/browse/MER-4120)

This PR fixes the bug where the auto submit of a page was not working in case the page has a due date and the presentation was `one at a time`. 

With these changes, even if the page has the presentation as `one at a time` if it has a stipulated due date, it will auto submit when the time limit is reached, as it already happens with pages that have a traditional presentation.

https://github.com/user-attachments/assets/90949bb7-0f0c-4523-b5ce-ca4edfeec997

[MER-4120]: https://eliterate.atlassian.net/browse/MER-4120?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Francisco-Castro (approved):
Good job! 🏅 

I tested locally and it's working as expected.

darrensiegel (approved):

PR #5339, opened 2025-01-08
 [FEATURE] [MER-3794] Granted Certificates PDF generation via AWS Lambda

manelli:
Wrapper function that calls the AWS Lambda function that given an HTML generates a PDF and stores it in S3 so later it can be viewed by anyone that wants to check if a given certification has been granted.


gastonabella (approved):
Great work! 👏

PR #5340, opened 2025-01-08
[FEATURE] [MER-3782] no payment feature

rgachuhi:
This PR addresses an issue where users registered into a course that requires payment are still redirected to a payment provider even when the price is zero. The implementation adds a new paywall bypass for cases where the amount listed in the course is zero.


darrensiegel (changes requested):
I'm concerned that we are using the cost being 0 as a paywall bypass mechanism.
  
This will preclude use cases such as a paywall guarding a section with access codes and a $0 cost

If what we are really intending to build (according to the ticket) is a mechanism for an institution to bypass completely the default paywall (since it is being offered for free at the institution), we should be modeling that explicitly by adding a new "Bypass paywall" or "Offer this for free" in the Discount impl ( Oli.Delivery.Paywall.Discount)

rgachuhi:
> I'm concerned that we are using the cost being 0 as a paywall bypass mechanism.
> 
> This will preclude use cases such as a paywall guarding a section with access codes and a $0 cost
> 
> If what we are really intending to build (according to the ticket) is a mechanism for an institution to bypass completely the default paywall (since it is being offered for free at the institution), we should be modeling that explicitly by adding a new "Bypass paywall" or "Offer this for free" in the Discount impl ( Oli.Delivery.Paywall.Discount)

I added the price enforcement we discussed to the Section "amount" field. This implementation ensures that if payment is required, the minimum amount cannot be zero or negative.

darrensiegel (approved):

PR #5342, opened 2025-01-09
[BUG FIX] [MER-4131] Dig into chem labs simulations

dtiwarATS:
Hey @bsparks, could you please look at the PR? Thanks.

The state of these SIMs is saved within the SIM itself and is linked to the repository: `git@github.com:SmartSparrow/componentLibrary-Redux.git#~2`.

During my investigation, I discovered that the state is maintained at the lesson ID level. This means that when a student restarts a lesson, the SIM retrieves the state for that lesson rather than resetting itself.

I also found that appending a random number to the lesson ID causes the SIM to reset, as it perceives it as a new lesson.

To address this, we introduced an iframe property called `IFRAME_randomizeLessonId`, similar to `IFRAME_frameVisible`. In authoring, if this property is set to `true`, we append a random number to the lesson ID. This ensures that when a new attempt is started, the SIM resets itself. 

In the initial state, we set the property to `true` and then use a mutate state action to set its value to `false`.

### DIFF
--- a/assets/src/components/parts/janus-capi-iframe/ExternalActivity.tsx
+++ b/assets/src/components/parts/janus-capi-iframe/ExternalActivity.tsx
@@ -29,6 +29,7 @@ const ExternalActivity: React.FC<PartComponentProps<CapiIframeModel>> = (props) 
   const [scriptEnv, setScriptEnv] = useState<any>();
   // model items, note that we use default values now because
   // the delay from parsing the json means we can't set them from the model immediately
+  const [randomizeLessonId, setrandomizeLessonId] = useState<boolean>(false);
   const [frameX, setFrameX] = useState<number>(0);
   const [frameY, setFrameY] = useState<number>(0);
   const [frameZ, setFrameZ] = useState<number>(0);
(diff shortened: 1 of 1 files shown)


assets/src/components/parts/janus-capi-iframe/ExternalActivity.tsx
```diff
@@ -657,7 +667,9 @@ const ExternalActivity: React.FC<PartComponentProps<CapiIframeModel>> = (props)
     // taken from simcapi.js TODO move somewhere, use from settings
     simLife.handshake.config = {
       context: context,
-      lessonId,
+      lessonId: randomizeLessonId
+        ? `${lessonId}_${Math.floor(Math.random() * 100)}_${Math.floor(Math.random() * 100)}`
```
bsparks, line 671: using a random number seems like it has the potential for creating a needle in a haystack bug. I presume that while the chance is very small, it still exists that the random number will choose the same number as before. so you will have a really small chance of this thing failing and it will be very hard to know why. so I suggest what about just increasing it? or maybe putting attempt number or guid on the end? so like `${lessonId}_${previousIndex + 1}` or something?

bsparks: also in general I don't really like the "IFRAME" so much because all of those other variables are something that affects the actual iframe so the purpose is clear... this is something different so it could potentially be a different key since it really just has to be something that the sim itself wouldn't set naturally. so maybe even `LESSON_ID_UNIQUE` ?

dtiwarATS: I have updated the code base to use the` 'guid'` as well as changed the variable to `LESSON_ID_UNIQUE`

bsparks (approved):

PR #5344, opened 2025-01-09
[FEATURE] [MER-4068] Reimplement user's invite - part 2

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-4068) to the ticket

As [commented on MER-3893](https://eliterate.atlassian.net/browse/MER-3893?focusedCommentId=25179), I forgot to use an absolute URL instead of a relative URL for the email invitation link. This PR fixes that error


manelli (approved):
LGTM 🚀

PR #5347, opened 2025-01-13
[FEATURE] [MER-3345] Dataset integration

darrensiegel:
This PR integrates the AWS EMR serverless "dataset creation" infrastructure into Torus, as an author facing feature. 

This work allows authors (and, of course, admins) to create and monitor cloud based jobs for dataset generation.  Users initiate a job by selecting the job type and the data source (some number of course sections).  

Torus will periodically poll the EMR infrastructure to obtain status information regarding these jobs.  Once a job transitions to a terminal state (success, failure or cancelled) the user that initiated it is notified via an email.  The UI then allows the user to click through and access the list of resultant dataset files (and by clicking the link, download them locally).


config/config.exs
```diff
@@ -65,6 +65,24 @@ config :oli,
     String.to_integer(System.get_env("SCREEN_IDLE_TIMEOUT_IN_SECONDS", "1800")),
   log_incomplete_requests: true
 
+config :oli, :dataset_generation,
```
eliknebel, line 68: Unless these configs are strictly required at compile time or when the release is being built, they can probably solely live in `runtime.exs` to simplify the config and avoid duplication.

lib/oli/analytics/datasets/setttings.ex
```diff
@@ -0,0 +1,37 @@
+defmodule Oli.Analytics.Datasets.Settings do
```
eliknebel, line 1: It looks like there is a typo in the file name of this module

eliknebel (approved):

PR #5349, opened 2025-01-13
[BUG FIX] [MER-4154] Ensure current_user set on superactivity support API calls

andersweinstein:
Since recent change, superactivity support api calls were failing due to unset `current_user` assign. This changes the route to use the `:api` pipeline which will set that, and seems appropriate since these are api calls.


Francisco-Castro (approved):
LGTM! I tested it locally, and it works without crashing the page.

PR #5358, opened 2025-01-17
[BUG FIX] [MER-4131] | Dig into chem labs simulations

dtiwarATS:
Hey @bsparks, could you please look at this PR? Thanks

The reason for this change is because of the changes made in https://github.com/Simon-Initiative/oli-torus/pull/5282. The new `PartId` variable only needs to be used in preview mode. It should never get sent to server. The server will throw an error because of the format in which it expects the part responses.


dtiwarATS:
closing this PR. Will create another one.

PR #5361, opened 2025-01-17
[FEATURE] [MER-3777] Cannot see comments from manually graded adaptive assignments

simonchoxx:
[MER-3777](https://eliterate.atlassian.net/browse/MER-3777)

This PR aims to show the feedback that an instructor gives to a student if the activity that the student solves is set to be manually evaluated. 

This implementation works for both basic and adaptive pages.

The feedback that the instructor can give will be displayed in two places:

- For the student, in the prologue view of the attempts the student made for a given page.

- For the instructor, in the view showing the attempts a student made on a given page.

https://github.com/user-attachments/assets/655c8a31-a5dc-43bd-a45a-1af9f1ed69f9

[MER-3777]: https://eliterate.atlassian.net/browse/MER-3777?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli/delivery/attempts/core.ex
+++ b/lib/oli/delivery/attempts/core.ex
@@ -590,10 +590,7 @@ defmodule Oli.Delivery.Attempts.Core do
               where: ra.resource_access_id == ^id,
               order_by: ra.attempt_number,
               select: ra,
-              preload: [
-                :revision,
-                activity_attempts: [:part_attempts]
-              ]
+              preload: [:revision]
             )
           )
 
@@ -1024,5 +1021,16 @@ defmodule Oli.Delivery.Attempts.Core do
       select: count(ra.id)
     )
     |> Repo.one()
+  end
+
+  @doc """
+  Preloads `activity_attempts` and their associated `part_attempts` for a given resource attempt or list of attempts.
+
+  This allows fetching detailed attempt data only when needed, avoiding unnecessary preloading in other queries.
+  """
+  def preload_activity_part_attempts(resource_attempts) do
+    Repo.preload(resource_attempts,
+      activity_attempts: [:part_attempts]
+    )
   end
 end
--- a/lib/oli/delivery/attempts/core.ex
+++ b/lib/oli/delivery/attempts/core.ex
@@ -590,7 +590,10 @@ defmodule Oli.Delivery.Attempts.Core do
               where: ra.resource_access_id == ^id,
               order_by: ra.attempt_number,
               select: ra,
-              preload: [:revision]
+              preload: [
+                :revision,
+                activity_attempts: [:part_attempts]
+              ]
             )
           )
 
--- a/lib/oli_web/live/delivery/student/prologue_live.ex
+++ b/lib/oli_web/live/delivery/student/prologue_live.ex
@@ -11,7 +11,6 @@ defmodule OliWeb.Delivery.Student.PrologueLive do
   alias Oli.Delivery.Sections
   alias Oli.Delivery.Settings
   alias Oli.Publishing.DeliveryResolver, as: Resolver
-  alias Oli.Repo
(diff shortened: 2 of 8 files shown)


lib/oli_web/live/delivery/student/prologue_live.ex
```diff
@@ -11,9 +11,10 @@ defmodule OliWeb.Delivery.Student.PrologueLive do
   alias Oli.Delivery.Sections
   alias Oli.Delivery.Settings
   alias Oli.Publishing.DeliveryResolver, as: Resolver
-  alias OliWeb.Common.FormatDateTime
+  alias Oli.Repo
```
eliknebel, line 14: We should be careful to not import Oli.Repo into the oli_web domains and liveviews if it's avoidable. Instead a function in the oli domain (`Oli.Delivery.Attempts` perhaps?) would be cleaner

lib/oli_web/live/delivery/student/prologue_live.ex
```diff
@@ -205,6 +206,12 @@ defmodule OliWeb.Delivery.Student.PrologueLive do
   attr :request_path, :string
 
   defp attempt_summary(assigns) do
+    activity_attempts =
+      Repo.preload(assigns.attempt, activity_attempts: [:part_attempts]).activity_attempts
```
eliknebel, line 210: This will make a database call for every instance of attempt summary rendered. We should instead be
preloading the part attempt data up front with whatever database call is being made for the
attempt or perhaps with an async call during connected mount

eliknebel (changes requested):

lib/oli_web/live/delivery/student/prologue_live.ex
simonchoxx: It makes sense. I'll fix it.

lib/oli/delivery/attempts/core.ex
```diff
@@ -590,7 +590,10 @@ defmodule Oli.Delivery.Attempts.Core do
               where: ra.resource_access_id == ^id,
               order_by: ra.attempt_number,
               select: ra,
-              preload: [:revision]
+              preload: [
```
eliknebel, line 593: My concern with this approach is that we are now preloading all of the nested activity attempts and part
attempts for the other existing callers to this function. Do we have an idea of what the performance impact
of this could be on these other callers?

A safer approach here would be to create another function in this module called
`preload_activity_part_attempts` that takes a list of resource attempts and does he preloading of
part attempt data. That way we can execute it only where it is required to load all of this data.

lib/oli_web/live/progress/student_resource_view.ex
```diff
@@ -106,7 +106,7 @@ defmodule OliWeb.Progress.StudentResourceView do
         nil
 
       ra ->
-        Oli.Repo.preload(ra, :resource_attempts)
+        Oli.Repo.preload(ra, resource_attempts: [activity_attempts: [:part_attempts]])
```
eliknebel, line 109: If we create that new `preload_activity_part_attempts` core function, we could use it here as well
instead of further polluting this live view with calls directly to `Oli.Repo`

eliknebel (changes requested):

lib/oli_web/live/progress/student_resource_view.ex
simonchoxx: I have left this as it was before. I have removed the preloads

lib/oli/delivery/attempts/core.ex
simonchoxx: I agree with your concern. I have already made the changes. Thanks!

eliknebel (approved):
Nice work

PR #5364, opened 2025-01-22
[FEATURE] [MER-4151] Certificate PDF Generation using AWS Lambda

manelli:
Add the code for the lambda function that generates PDF certificates and stores them in S3.

See: https://eliterate.atlassian.net/browse/MER-4151


manelli:
Converting to draft because there has been a change regarding the signatures in the generated certificate. See: https://eliterate.atlassian.net/browse/MER-4087?focusedCommentId=25408

I need to check if there is a way for WeasyPrint to support these cursive fonts.

rgachuhi (approved):

PR #5367, opened 2025-01-23
[BUG FIX] [MER-4185] Reverts the survey tool report to match legacy display 

rgachuhi:
This PR reverts the Likert report back to how it looked in the previous PR, which closely matches what legacy displays. See the recording in the PR [https://github.com/Simon-Initiative/oli-torus/pull/4862](https://github.com/Simon-Initiative/oli-torus/pull/4862) for details
Screenshot 2025-01-24 at 8 42 36 AM

### DIFF
--- a/lib/oli/activities/reports/providers/OliLikert.ex
+++ b/lib/oli/activities/reports/providers/OliLikert.ex
@@ -232,7 +232,8 @@ defmodule Oli.Activities.Reports.Providers.OliLikert do
                   "y": {
                       "field": "choice",
                       "type": "nominal",
-                      "sort": "descending"
+                      "sort": "descending",
+                      "title": "Choice"
                   }
               }
           }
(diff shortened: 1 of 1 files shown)


darrensiegel:
Hold on reviewing this - I'm not convinced that we want to make this change.

darrensiegel:
Reverting back to the Legacy implementation is the wrong thing to do here. Bar charts are for aggregated data, and not for raw, ordinal data.  

Visualizing a single student’s ratings as a bar chart suggests there’s some quantitative comparison happening, which isn’t true.  This misuse of the chart confuses viewers into thinking they’re looking at something aggregated or meaningful when it’s not. (it did for me when I saw this in Legacy)

Another PR (https://github.com/Simon-Initiative/oli-torus/pull/5368) has been opened which makes, instead, a targeted improvement to the current visualization - while correctly adhering to basic visualization practices.

andersweinstein:
Note this work resulted because instructors using this hated the more 'correct" implementation foisted upon them, finding it unacceptable and demanding it be changed back to the legacy style, which they liked. It's a user-driven demand.

PR #5376, opened 2025-01-28
[BUG FIX] [MER-4075] Cant access video settings (Part 2)

eliknebel:
https://eliterate.atlassian.net/browse/MER-4075

This is a follow-up to fix issues uncovered as a result of https://github.com/Simon-Initiative/oli-torus/pull/5374.

The approach here is to revert all changes made in the previous PR and instead take a more limited approach to the fix. This fix takes a similar approach to other fixes that have been made in the past to resolve modal and missing store issues. Instead of trying to resolve the root cause of the component not being under the umbrella of a parent provider, and new provider is simply used by instantiating a new store.

While this works to resolve the immediate issue, it is just a bandaid on the larger issue that we are using global references to dispatch actions and calling `configureStore` in multiple different components (50+ different places) when it should only really be called once at the top level per "App". This could result in strange issues where multiple stores are now not in sync with each other and mostly defeats the purpose of using redux to begin with. Trying to address the root cause here is too risky for a hotfix. I filed a separate ticket to track this larger issue https://eliterate.atlassian.net/browse/MER-4210.

### DIFF
--- a/assets/src/components/editing/elements/audio/AudioSettings.tsx
+++ b/assets/src/components/editing/elements/audio/AudioSettings.tsx
@@ -1,9 +1,10 @@
 import React, { useState } from 'react';
-import { useDispatch } from 'react-redux';
+import { Provider } from 'react-redux';
 import { Tooltip } from 'components/common/Tooltip';
 import { CommandContext } from 'components/editing/elements/commands/interfaces';
 import { Modal, ModalSize } from 'components/modal/Modal';
 import * as ContentModel from 'data/content/model/elements/types';
+import { configureStore } from 'state/store';
 import { modalActions } from '../../../../actions/modal';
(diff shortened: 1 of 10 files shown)


assets/src/components/editing/elements/audio/AudioSettings.tsx
```diff
@@ -96,60 +97,62 @@ const AudioSettingsModal = (props: AudioSettingsProps) => {
...
-                      {isPlaying ? (
-                        <i className="fa-solid fa-circle-stop"></i>
-                      ) : (
-                        <i className="fa-solid fa-circle-play"></i>
-                      )}
-                    </span>
-                  </button>
-                </Tooltip>
-              )}
+    <Provider store={store}>
```
eliknebel, line 100: The only change here is to wrap this component in a Provider

PR #5390, opened 2025-01-31
[BUG FIX] [MER-4191] Exploration pages dont display in explorations tab

simonchoxx:
[MER-4191](https://eliterate.atlassian.net/browse/MER-4191)

This PR addresses a bug where exploration purpose pages that are not contained within a Unit or Module type container (orphaned pages) were not displayed in the Explorations tab.

**Root Cause**

The issue was caused by incorrectly retrieving the hierarchy data to display exploration pages. Previously, we were using the function `Hierarchy.flatten`, which only considers pages inside Unit or Module type containers. This approach overlooked pages that are directly attached to the root container, leading to them not being displayed in the Explorations tab.

**Fix**

To ensure that all exploration purpose pages are correctly included, we replaced `Hierarchy.flatten` with `Hierarchy.flatten_hierarchy`. Unlike the previous function, `flatten_hierarchy` properly retrieves all elements in the hierarchy, including those directly attached to the root container. This fix ensures that both basic and adaptive exploration pages are now correctly displayed in the Explorations tab.

Additionally, the method used to retrieve the hierarchy was updated based on an existing To Do in the code, which aimed to optimize the way the hierarchy is fetched. Previously, `DeliveryResolver.full_hierarchy(section_slug)` was used, but it has now been replaced with `MinimalHierarchy.full_hierarchy(section_slug)`. 

- Before
explorations-tab-before

- After
explorations-tab-after

[MER-4191]: https://eliterate.atlassian.net/browse/MER-4191?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli/delivery/sections.ex
+++ b/lib/oli/delivery/sections.ex
@@ -22,7 +22,8 @@ defmodule Oli.Delivery.Sections do
     Enrollment,
     EnrollmentBrowseOptions,
     EnrollmentContextRole,
-    Scheduling
+    Scheduling,
+    MinimalHierarchy
   }
 
   alias Lti_1p3.Tool.ContextRole
@@ -1803,12 +1804,11 @@ defmodule Oli.Delivery.Sections do
   def fetch_ordered_container_labels(section_slug, opts \\ []) do
     short_label = opts[:short_label] || false
 
-    # TODO: OPTIMIZATION replace this with a minimal hierarchy query after v26.2 is merged
     %Section{customizations: customizations} = get_section_by_slug(section_slug)
-    full_hierarchy = DeliveryResolver.full_hierarchy(section_slug)
+    full_hierarchy = MinimalHierarchy.full_hierarchy(section_slug)
 
     full_hierarchy
-    |> Hierarchy.flatten()
+    |> Hierarchy.flatten_hierarchy()
     |> Enum.filter(fn node ->
       node.revision.resource_type_id == ResourceType.get_id_by_type("container")
     end)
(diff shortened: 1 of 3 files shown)


eliknebel:
@simonchoxx Could you update the description here to include a detailed explanation of the diagnosed root cause? We also want to include an explanation of the fix being made here (According to the new code review checklist https://github.com/Simon-Initiative/oli-torus/wiki/Code-Review-Checklist)

Thanks!

rgachuhi (changes requested):
I have tested this PR locally, and as far as I can tell, the problem described in the ticket persists. Any exploration pages created outside the curriculum do not show up in the list of exploration on delivery. Only those exploration pages attached to the curriculum are listed. 

https://github.com/user-attachments/assets/81e34a23-0801-4a9b-97bc-e144a40d2f40


Uploading Screen Recording 2025-02-04 at 1.38.22 PM.mov…


Uploading Screen Recording 2025-02-04 at 1.43.41 PM.mov…

rgachuhi:
sdf

rgachuhi:
Uploading Screen Recording 2025-02-04 at 1.43.41 PM.mov…

simonchoxx:
Thanks for your review @rgachuhi.  I can't see the second recording you attached. 

I tried the same thing you did in my local environment. 

Check if the page you create and associate to a module is visible to the student after publishing the changes and if it is also visible in the curriculum on the author side. 

Because we may be in the presence of a totally different issue.

rgachuhi (approved):

PR #5398, opened 2025-02-05
[BUG FIX] [MER-4219] OAuth google login may result in infinite confirm email step

eliknebel:
https://eliterate.atlassian.net/browse/MER-4219

Fixes an issue where OAuth social login may result in infinite confirmation step loop.

**Cause**
The issue identified was that if an existing OAuth social login account was never confirmed, then subsequent logins would never update this field and the user would be stuck in an infinite "confirm email" loop.

**Fix**
The fix here is to actually update user details on every OAuth flow login so that the email will be automatically confirmed if `email_verified: true` is passed along. This has the additional benefit of keeping other user details up to date such as name and profile picture. A step was also added here to properly handle the case where `email_verified: false` is passed in, in which case a confirmation email is automatically sent, prompting the user to confirm their email. While this should never be true for email based OAuth integrations such as google, it is good to have in place for Github or other OAuth integration types.

### DIFF
--- a/lib/oli/assent_auth/user_assent_auth.ex
+++ b/lib/oli/assent_auth/user_assent_auth.ex
@@ -14,6 +14,7 @@ defmodule Oli.AssentAuth.UserAssentAuth do
   """
   def authentication_providers() do
     Application.get_env(:oli, :user_auth_providers)
+    |> Enum.filter(fn {_, config} -> config != nil end)
   end
 
   @doc """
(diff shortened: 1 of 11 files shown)


lib/oli/assent_auth/user_assent_auth.ex
```diff
@@ -14,6 +14,7 @@ defmodule Oli.AssentAuth.UserAssentAuth do
   """
   def authentication_providers() do
     Application.get_env(:oli, :user_auth_providers)
+    |> Enum.filter(fn {_, config} -> config != nil end)
```
eliknebel, line 17: Only show OAuth providers if they are configured

rgachuhi (approved):

PR #5399, opened 2025-02-06
[FEATURE] [MER-4068] Reimplement user and author invite functionality - part 4

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-4068?focusedCommentId=25417) to the comment in the ticket

### Admin invites an existing author
The following flash message was added to handle that case (and no email invitation is sent)

### Instructor adds enrollments

This PR allows an instructor to re-send an invitation to any student with an enrollment with any of these statuses:
- rejected
- suspended
- pending confirmation

All previous invitations for that user-course are invalidated before the new invitation is sent. So if the user enters an old invitation through the email invitation link this view will be prompted:

On the other hand, the `step 2` of the "Add Enrollments" wizard was extended to warn about:

- users that do not yet exist in the system (this warning already existed)
- users that already have an "inactive" enrollment -pending confirmation, rejected or suspended- (new warning)
- users that already have an active enrollment in the course (new warning)

Finally, the `step 3` of the same wizard was extended to exclude the existing enrollments from the invitation count (since no invitations are sent to already enrolled users)

### DIFF
--- a/lib/oli_web/components/delivery/students/students.ex
+++ b/lib/oli_web/components/delivery/students/students.ex
@@ -138,8 +138,9 @@ defmodule OliWeb.Components.Delivery.Students do
          pending_confirmation: [],
          rejected: [],
          non_existing_users: [],
-         existing_users_without_an_enrollment: []
+         not_enrolled_users: []
        },
+       add_enrollments_effective_count: 0,
        inviter: if(is_nil(ctx.author), do: "user", else: "author"),
        current_user: ctx.user,
        current_author: ctx.author,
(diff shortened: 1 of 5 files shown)


lib/oli_web/components/delivery/students/students.ex
```diff
@@ -723,33 +742,171 @@ defmodule OliWeb.Components.Delivery.Students do
...
+        </p>
+        <div>
+          <li class="list-none mt-4 max-h-80 overflow-y-scroll">
+            <%= for email <- @add_enrollments_grouped_by_status[:non_existing_users] do %>
+              <ul class="odd:bg-gray-200 dark:odd:bg-neutral-600 even:bg-gray-100 dark:even:bg-neutral-500 p-2 first:rounded-t last:rounded-b">
+                <div class="flex items-center justify-between">
+                  <p><%= email %></p>
+                  <button
+                    phx-click={
+                      JS.push("add_enrollments_remove_from_list",
```
gastonabella, line 762: Just a thought, but I believe that we could avoid using the `JS` library when no JS is being used after the event is triggered. As a dev I'd expect to see some javascript code running but it is just Elixir/LiveView.

rgachuhi (approved):

PR #5400, opened 2025-02-06
[FEATURE] [MER-4087] Certificates Design and Preview Tab

manelli:
See: https://eliterate.atlassian.net/browse/MER-4087 and https://eliterate.atlassian.net/browse/MER-3793

Implement the Design tab where the author will input the values needed for generating certificates and preview them.

There are 2 differences with the Figma design:
- There are simple buttons to navigate the different certificate previews instead of chevrons and a carousel style (This will be implemented in another PR)
- In order to be able to have access to the uploaded images, the form had to be submitted, that is why there is only one button "Save Design" instead of two ("Preview" and "Save Design")

https://github.com/user-attachments/assets/778d7933-c300-47e8-a93e-dc4601c5a20e

### DIFF
--- a/lib/oli/delivery/certificates/certificate_renderer.ex
+++ b/lib/oli/delivery/certificates/certificate_renderer.ex
@@ -6,38 +6,63 @@ defmodule Oli.Delivery.Certificates.CertificateRenderer do
     @page {
       size: landscape;
     }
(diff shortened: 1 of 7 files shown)


gastonabella (approved):
LGTM!

PR #5401, opened 2025-02-07
[BUG FIX] [MER-4175] Redirect instructors to the correct workspace after social login

rgachuhi:
[https://eliterate.atlassian.net/jira/software/c/projects/MER/boards/2?selectedIssue=MER-4175](https://eliterate.atlassian.net/jira/software/c/projects/MER/boards/2?selectedIssue=MER-4175)

This PR adds support for social login to the issue where anyone who logs in via the Instructor interface is redirected to the Student workspace instead of the Instructor workspace


lib/oli_web/controllers/user_authorization_controller.ex
```diff
@@ -23,6 +23,20 @@ defmodule OliWeb.UserAuthorizationController do
   def new(conn, %{"provider" => provider}) do
     config = conn.assigns.assent_auth_config
 
+    conn =
```
eliknebel, line 26: Would it be better to use a user's platform roles or `can_create_sections` field to determine where to redirect to?

eliknebel (changes requested):

lib/oli_web/controllers/user_authorization_controller.ex
rgachuhi: The first time an "instructor" signs up via social login, the `can_create_sections ` field is not automatically set. So, instructor signup will always initially land the user `/workspaces/student` until an admin sets that field. An option would be automatically setting the field, which is not wise.

eliknebel (approved):

lib/oli_web/controllers/user_authorization_controller.ex
eliknebel: Ah, makes sense

PR #5404, opened 2025-02-11
[FEATURE] [MER-3809] Certificate Verification

manelli:
Implement page for verification of granted certificates

Changes:
- Add new endpoint `GET /certificates/` for searching a certificate by its ID (guid)
- Add new endpoint `POST /certificates/verify` for veryfing if the certificate is earned and if so, display it on the browser
- **cloud/generate-pdf-certificate/lambda_function.py**:
  - Store certificate in S3 so it is served with `Content-Type: 'application/pdf'`
- **lib/oli/delivery/certificates/certificate_renderer.ex**: 
  - `render/1 `now allows to pass a `GrantedCertificate` struct.
  - Use `DejaVu Serif` if available, if not `Times New Roman` for the title
 - **lib/oli/delivery/granted_certificates.ex**:
   - Add `get_granted_certificate_by_guid/1` function
   - Rely on `CertificateRenderer.render/1` for generating the certificate html in `generate_pdf/1`
 - The URL in a `GrantedCertificate` is now derived from the `CERTIFICATES_S3_PDF_URL` env variable
 - The lambda function used to generate the pdf for a certificate given a html can be changed via the `CERTIFICATES_GENERATE_PDF_LAMBDA` env variable

https://github.com/user-attachments/assets/998e2d42-d3c3-4df4-b31a-230bcf21a204

See: https://eliterate.atlassian.net/browse/MER-3809

### DIFF
--- a/lib/oli_web/controllers/certificate_controller.ex
+++ b/lib/oli_web/controllers/certificate_controller.ex
@@ -2,16 +2,20 @@ defmodule OliWeb.CertificateController do
   use OliWeb, :controller
 
   alias Oli.Delivery.GrantedCertificates
-  alias Oli.Repo
 
-  def show(conn, %{"guid" => guid}) do
-    case GrantedCertificates.get_granted_certificate_by_guid(guid) do
-      nil ->
-        send_resp(conn, 404, "Not Found")
+  def index(conn, _params), do: render(conn, "index.html")
 
-      gc ->
-        granted_certificate = Repo.preload(gc, [:certificate, :user])
-        render(conn, "show.html", certificate: granted_certificate)
+  def verify(conn, params) do
+    if recaptcha_verified?(params) do
+      certificate = GrantedCertificates.get_granted_certificate_by_guid(params["guid"]["value"])
+      render(conn, "show.html", certificate: certificate)
+    else
+      render(conn, "index.html", recaptcha_error: "ReCaptcha failed, please try again")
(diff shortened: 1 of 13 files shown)


lib/oli_web/controllers/certificate_controller.ex
```diff
@@ -0,0 +1,31 @@
...
+    end
+  end
+
+  defp recaptcha_verified?(params) do
+    recaptcha_response = Map.get(params, "g-recaptcha-response", "")
+    Oli.Recaptcha.verify(recaptcha_response) == {:success, true}
+  end
+
+  defp s3_certificate_url(%{guid: guid}),
+    do: "https://torus-pdf-certificates.s3.amazonaws.com/certificates/#{guid}.pdf"
```
darrensiegel, line 28: This full URL, which presumably is driven from an S3 bucket, must be able to be varied for different Torus instances.  So this must be driven from an application variable from `runtime.exs`

darrensiegel (changes requested):

lib/oli_web/controllers/certificate_controller.ex
manelli: 👍 Fixed in dac52eeef65a7fe896ff0f1caaca7211d3507555
Let me know what you think

gastonabella (approved):
Looks good!

PR #5406, opened 2025-02-12
[BUG FIX] [MER-4218] Advanced Authoring: Unable to de-select "refresh warning pop up"

dtiwarATS:


assets/src/apps/authoring/components/PropertyEditor/schemas/lesson.ts
```diff
@@ -319,7 +319,7 @@ export const transformModelToSchema = (model: any) => {
       title: model.title,
       customCSS: model.customCss,
       enableHistory: model.custom.allowNavigation || model.custom.enableHistory || false,
-      displayRefreshWarningPopup: model.custom.displayRefreshWarningPopup || true,
```
dtiwarATS, line 322: By default, `displayRefreshWarningPopup` property is true. When user was unchecking it, the condition here was getting set as 'false || true' and due to that the check box was getting checked.

bsparks (approved):

PR #5411, opened 2025-02-13
[BUG FIX] [MER-3000] Lost Quiz Answers by Working in a Second Tab of an Already-Submitted Attempt

dtiwarATS:
Hey @darrensiegel  / @eliknebe, this is the PR that contains the changes done in the original [PR](https://github.com/Simon-Initiative/oli-torus/pull/4765).

We already removed the change from the test as part of this [PR](https://github.com/Simon-Initiative/oli-torus/pull/5091) because we didn't need that hence I didn't included that.

This is just for reference. You can merge the original PR or this.


eliknebel:
Addressed by https://github.com/Simon-Initiative/oli-torus/pull/5414

PR #5418, opened 2025-02-18
[BUG FIX] [MER-4145] hints not working tokamak

eliknebel:
https://eliterate.atlassian.net/browse/MER-4145

The issue stems from the fact that `datashop_session_id` was not always being set for a session in all cases, in particular LTI and Guest users (which do not use the `UserAuth.log_in_user` method). This means that code which pulls the `datashop_session_id` from the session was getting nil values. Even though there is a changeset requirement for this value in the code,  there is no database level constraint and there are queries that bulk create part attempts which do not enforce this, these records were being initialized with nil values.

For example, the following flow:
- Endpoint `PageDeliveryController.start_attempt` calls `do_start_attempt`
- `do_start_attempt` will pull a possibly `nil` value from the session, in cases described above (LTI, Guest users)
- `query_driven_part_attempt_creation` called with `datashop_session_id: nil` and runs a hard coded bulk creation query, bypassing the `PartAttempt.changeset` allowing this value to be set to nil

Once these part attempts are created, they cannot be updated when a hint is requested due to the `required` constraint in the PartAttempt schema, manifesting in the hints error (and possibly other errors) we were seeing.

At some point we might want to add a database-level non-null constraint to prevent this scenario from happening and get earlier feedback when this issue is occurring (in the form of an error upon part attempt creation).

The fix in this PR is simpler, which is to remove the application-level changeset requirement on required `datashop_session_id` for a part attempt. This aligns better with the now real world state that existing records may, in fact, be null.

This PR also includes several other necessary fixes for issues related to the datashop_session_id that I identified, including:
- Better ensure that we always have a stable datashop identifier in the session by guarding delivery_protected endpoints with a new `EnsureDatashopId` plug
- Ensure that we are properly clearing the datashop identifier when a session ends and minting a new one on subsequent session creation (undoing https://github.com/Simon-Initiative/oli-torus/pull/5356). This keeps the `datashop_session_id` identifier scoped to a particular session, per original requirements.
- Remove "on-the-fly" datashop id creation during datashop export, as this could result in non-deterministic and misleading data. It is better that a researcher can identify the cases where a datashop id may be missing and filter those out of their dataset then to mint a random id on the fly


lib/oli/analytics/datashop.ex
```diff
@@ -144,10 +142,7 @@ defmodule Oli.Analytics.Datashop do
         date: part_attempt.date_submitted,
         email: email,
         sub: sub,
-        # datashop_session_id will be set to the latest part attempt datashop_session_id
-        # unless it is nil, then it will be set to a generated UUID. This will handle any
-        # part attempts that existed before this value was tracked in the part attempt record
-        datashop_session_id: value_or(part_attempt.datashop_session_id, UUID.uuid4()),
```
eliknebel, line 150: We should not be generating a datashop id on the fly if one doesn't exist. This could lead to non-deterministic and misleading data

lib/oli_web/user_auth.ex
```diff
@@ -97,8 +96,7 @@ defmodule OliWeb.UserAuth do
         "browser_timezone",
         "author_token",
         "author_live_socket_id",
-        "current_author_id",
-        "datashop_session_id"
```
eliknebel, line 101: Here we do not want to retain the datashop_session_id when a session is ended. Otherwise, we would simply be tracking a user forever and not just their session

nicocirio (approved):
Nice work! Thanks for the detailed explanation of the root cause

PR #5428, opened 2025-02-21
[CHORE] [MER-0000] Remove dead analytics code

darrensiegel:
This PR removes all of the code related to the Raw Analytics and Datashop generation and download, since all of this can now be done with the Dataset infrastructure.


nicocirio (approved):
🚀

PR #5429, opened 2025-02-22
[ENHANCEMENT] [MER-3809] Certificate Verification: placeholder text white

manelli:
So it follows the Figma Design: 
image

### DIFF
--- a/lib/oli_web/templates/certificate/index.html.eex
+++ b/lib/oli_web/templates/certificate/index.html.eex
@@ -7,7 +7,7 @@
       <div class="flex w-full">
         <%= text_input f, :value,
           placeholder: "Enter certificate ID...",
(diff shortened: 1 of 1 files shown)


nicocirio (commented):
Do we need to adjust [the header](https://github.com/Simon-Initiative/oli-torus/pull/5429/files#diff-de296ef06eedeb25a593cf519df7ba1c98f7bc986aba29e6b07f32dd09e6c3c3R2) contrast as well?

manelli:
> Do we need to adjust [the header](https://github.com/Simon-Initiative/oli-torus/pull/5429/files#diff-de296ef06eedeb25a593cf519df7ba1c98f7bc986aba29e6b07f32dd09e6c3c3R2) contrast as well?

That is already with the correct color:
image

nicocirio (approved):

PR #5433, opened 2025-02-24
[BUG FIX] [MER-4300] Error thrown and can't see inspector on some screens - Advanced Author

dtiwarATS:


assets/src/apps/delivery/components/preview-tools/Inspector.tsx
```diff
@@ -120,7 +120,10 @@ const Inspector: React.FC<InspectorProps> = ({ reviewMode, isInstructor }) => {
       const allState = getEnvState(defaultGlobalEnv);
       const variableTypes: Array<any> = getVariablesTypeState(allState);
       const globalStateAsVars = Object.keys(allState).reduce((collect: any, key) => {
-        const result = variableTypes?.filter((part) => key.includes(part.key));
```
dtiwarATS, line 123: `key.includes` was giving results for all the matching keys and from the result we were always taking the first value hence even for a Boolean variable type, Enum was coming-up hence throwing the error.

PR #5441, opened 2025-02-25
[BUG FIX] [MER-4262] 500 internal server error when accessing assessment settings

eliknebel:
https://eliterate.atlassian.net/browse/MER-4262

Fixes an issue where when logged in as an admin, the assessment settings page throws a 500 error.

The issue was caused by the result user of `Mount.for` in `SettingsLive` overwriting the `current_user` assigns with an Author (when logged in admin account). This was causing some of the UserAuth logic to break.

This was originally being worked on as part of MER-4262, but since there is a separate ticket targeted for this specific issue I am submitting this fix for MER-4262 and any additional work done for MER-4262 will be tracked by that Jira ticket and completed with another PR.

### DIFF
--- a/lib/oli_web/live/sections/assessment_settings/settings_live.ex
+++ b/lib/oli_web/live/sections/assessment_settings/settings_live.ex
@@ -14,7 +14,7 @@ defmodule OliWeb.Sections.AssessmentSettings.SettingsLive do
       {:error, error} ->
         {:ok, redirect(socket, to: Routes.static_page_path(OliWeb.Endpoint, error))}
 
-      {_user_type, _user, section} ->
+      {_user_type, user, section} ->
         section =
           section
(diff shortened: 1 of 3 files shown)


simonchoxx (commented):
I have tested the changes locally. I can enter the view correctly as an Admin. 

However, I find that there are some issues when trying to modify some settings as Admin (the liveview crashes). 

I think the problem is happening when saving the changes. 

I'm not sure if it should be fixed in this PR or in some other PR (maybe it would be good to fix it in this PR as it should not be too expensive).


https://github.com/user-attachments/assets/f9ff5c6a-1320-47bb-8169-cac2748c19e9

eliknebel:
Good catch, there was an issue related to this view expecting an Author to be loaded into session context. I fixed the issue.

simonchoxx (approved):
Nice work 🚀

PR #5447, opened 2025-02-27
[FEATURE] [MER-3418] Add a hub and spoke component to simple author/flowchart

dtiwarATS:

https://github.com/user-attachments/assets/866d6818-8ece-49b0-a947-972e2dd6de04


bsparks (approved):
lgtm, might want additional eyes since I've been away so long now...

assets/src/components/parts/janus-hub-spoke/HubSpoke.tsx
```diff
@@ -0,0 +1,378 @@
...
+  }
+  const progressPercentage = options?.length ? (completedSpokeCount / options?.length) * 100 : 0;
+
+  return (
+    <React.Fragment>
+      {ready ? (
+        <>
+          <style>
+            {`
+
```
eliknebel, line 277: I'm no expert in the advanced author space, but directly embedding styles like this seems like an anti-pattern. Should we not be defining these styles in a css file?

Also, using `!important` is a pretty heavy hammer that should really only be used when necessary. Most of the time when it is used, the same result can be achieved by properly following CSS specificity and precedence https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_cascade/Specificity

assets/src/components/parts/janus-hub-spoke/HubSpoke.tsx
```diff
@@ -0,0 +1,378 @@
...
+  );
+};
+
+interface SpokeOptionModel extends Item {
+  index: number;
+  value: number;
+}
+
+const HubSpoke: React.FC<PartComponentProps<hubSpokeModel>> = (props) => {
+  const [model, setModel] = useState<any>(Array.isArray(props.model) ? props.model : {});
```
eliknebel, line 83: There are a lot of different `useState` defined here, maybe it makes sense to encapsulate all of this state mutation logic in a reducer and `useReducer` instead?

PR #5452, opened 2025-02-28
[BUG FIX] [MER-4339] handle scenario where an invited user is registering an account from invitation email

eliknebel:
https://eliterate.atlassian.net/browse/MER-4339

This PR addresses an issue where users who are invited via bulk invitation were not able to complete the registration process. The issue here was the Pow auth custom `create` logic was not taking into account the scenario where a user may already have an existing User record from the bulk invite process and when attempting to complete the registration process the system would assume that a user was trying to register an account that already exists.

The fix here includes a check for the invitation token, and if present along with a nil invitation_accepted_at, then it will process the registration as completing the invitation flow.

### DIFF
--- a/lib/oli_web/controllers/invite_controller.ex
+++ b/lib/oli_web/controllers/invite_controller.ex
@@ -54,7 +54,7 @@ defmodule OliWeb.InviteController do
           case user.status do
             :new_user ->
               {"Join now",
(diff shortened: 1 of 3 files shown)


eliknebel:
Closing for now. We can re-open if we decide we want to take this as part of a hotfix

PR #5459, opened 2025-03-03
[BUG FIX] [MER-4350]Simplify, and use the current author for is_admin check

darrensiegel:
This PR fixes the problem that an "Account" only admin was not seeing the Admin menu.  The fix here was to directly use the `Author` struct that is in the context for determining if the user is an Admin

### DIFF
--- a/lib/oli_web/components/delivery/user_account.ex
+++ b/lib/oli_web/components/delivery/user_account.ex
@@ -39,29 +39,6 @@ defmodule OliWeb.Components.Delivery.UserAccount do
   attr(:class, :string, default: "")
   attr(:dropdown_class, :string, default: "")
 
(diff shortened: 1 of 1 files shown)


lib/oli_web/components/delivery/user_account.ex
```diff
@@ -39,6 +39,29 @@ defmodule OliWeb.Components.Delivery.UserAccount do
   attr(:class, :string, default: "")
   attr(:dropdown_class, :string, default: "")
 
+  def workspace_menu(%{ctx: %{author: %{system_role_id: system_role_id}}} = assigns) when system_role_id in [2, 3, 4] do
```
eliknebel, line 42: The `is_admin` assign should be set if *any* admin roles are present, so it is a bit strange that it is not picking that up here before this change

PR #5461, opened 2025-03-04
[FEATURE] [MER-4360] update dark mode styles

darrensiegel:
This PR adjusts the dark mode styles for the trigger editing UIs, so that there is better contrast for the sample dropdown and no longer 'white text on white background' for the prompt editing.

### DIFF
--- a/assets/src/apps/page-editor/PageTriggerEditor.tsx
+++ b/assets/src/apps/page-editor/PageTriggerEditor.tsx
@@ -8,9 +8,9 @@ export const PageTriggerEditor: React.FC<{
   const [promptsExpanded, setPromptsExpanded] = useState<boolean>(false);
 
   const ExpandablePromptHelp = () => (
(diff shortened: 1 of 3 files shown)


andersweinstein:
There is reasonable dark-mode behavior in the TriggerAuthoring code.  Would be appropriate to make the dropdown prompt help  into a reuseable component to get consistent appearance from a single source.  (If prompt list is different in different uses it could just be passed as a prop. And for the textarea used for the trigger prompt, the only issue is that you have to set `bg-inherit` to make it inherit background colors -- this happens by default for regular text input controls, but not for textareas.)

Note however: the initial implementation I did in TriggerAuthoring did NOT use the colors from the Figma designs. Those were blueish to match a similar control used in course-progress liveview, I believe the view_example_container. So if we are fixing up the colors, we could also consider trying to change them to match the original design.

andersweinstein (approved):

PR #5462, opened 2025-03-04
[BUG FIX] [MER-4357] Fix issue preventing threshold checks from triggering on completed jobs

Francisco-Castro:
Ticket: [MER-4357](https://eliterate.atlassian.net/browse/MER-4357)

When a student submits a graded page, our system kicks off an Oban job. This job’s job (pun intended!) is to figure out if the student has hit the required thresholds. It’s supposed to run every time a student submits a graded page (with some smart optimizations in the mix). But right now, it only runs once, and any follow-up submissions get skipped. This is a problem because the student’s page thinks the thresholds have been reached, but that’s not actually the case since we’re not running all the necessary checks.

The Issue: Once an Oban job runs, it’s marked as "completed." Because of a unique constraint, we can’t run the same job again for the same user and section.

The Fix We’re Proposing: Before inserting a new Oban job, we’ll clear out any existing jobs for the same user_id and section_id combo. Then, we’ll add a fresh Oban job to make sure all the checks run properly.

[MER-4357]: https://eliterate.atlassian.net/browse/MER-4357?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli/delivery/sections/certificates/workers/check_certification.ex
+++ b/lib/oli/delivery/sections/certificates/workers/check_certification.ex
@@ -5,11 +5,30 @@ defmodule Oli.Delivery.Sections.Certificates.Workers.CheckCertification do
     max_attempts: 1
 
   alias Oli.Delivery.GrantedCertificates
+  alias Oli.Delivery.Sections.Certificates.Workers.CheckCertification
 
   @impl Oban.Worker
   def perform(%Oban.Job{args: %{"user_id" => user_id, "section_id" => section_id}}) do
     GrantedCertificates.has_qualified(user_id, section_id)
 
     :ok
+  end
+
+  def restart_certificate_check(user_id, section_id) do
+    Ecto.Adapters.SQL.query!(
+      Oli.Repo,
+      """
+      DELETE FROM oban_jobs
+      WHERE id = (
+        SELECT id FROM oban_jobs
+        WHERE queue = 'certificate_eligibility'
+        AND args @> $1
+        LIMIT 1
+      )
+      """,
+      [%{"user_id" => user_id, "section_id" => section_id}]
+    )
+
+    %{user_id: user_id, section_id: section_id} |> CheckCertification.new() |> Oban.insert()
   end
 end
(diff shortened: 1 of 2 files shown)


lib/oli/delivery/sections/certificates/workers/check_certification.ex
```diff
@@ -5,11 +5,30 @@ defmodule Oli.Delivery.Sections.Certificates.Workers.CheckCertification do
...
 
   @impl Oban.Worker
   def perform(%Oban.Job{args: %{"user_id" => user_id, "section_id" => section_id}}) do
     GrantedCertificates.has_qualified(user_id, section_id)
 
     :ok
   end
+
+  def restart_certificate_check(user_id, section_id) do
+    Ecto.Adapters.SQL.query!(
```
eliknebel, line 18: Should we wrap this in a database transaction to ensure we guard against the (unlikely) case where a job is inserted between the check and the insert?

eliknebel (changes requested):

darrensiegel:
Why don't we just adjust which fields and states we use for uniqueness consideration?

https://hexdocs.pm/oban/unique_jobs.html

nicocirio:
> Why don't we just adjust which fields and states we use for uniqueness consideration?

Yes, I was searching for how to do that same configuration. I found this elixir forum post -> https://elixirforum.com/t/unique-job-handling/57844/2

manelli:
Agree with previous comments:

I think that with the following configuration for the `CheckCertificate` worker would suffice.
```elixir
keys: [:user_id, :section_id]],
states: [:executing, :available, :scheduled, :retryable]
```

this way we can ensure that we have a single job for a `{user_id, section_id}` that is executing / about to execute / scheduled to execute / retrying a previous execution

darrensiegel:
I do believe that we want to include `:executing` in the uniqueness check.  That could lead to this situation:

1. Student takes an action which doesn't lead to earning the certificate.  The Oban queue for checking is queued
2. The Oban job starts to execute (maybe it only gets past the step in GrantedCertificates.has_qualified where it counts posts)
3. Student takes another action which does lead to earning the certificate.  Since there is a job that is already executing, this job will not be queued and the fact that the student has earned the certificate may not be correctly calculated.

Francisco-Castro:
Thanks for all your comments! 🏅 I followed the suggestion to use `GradeUpdateWorker` as a model, and it worked as expected.

nicocirio (approved):

PR #5469, opened 2025-03-06
[BUG FIX] [MER-4243] allow non-confirmed user to access student workspace

eliknebel:

Fixes an issue uncovered in QA where an unconfirmed student can get lost and have no way to return to their section.

The fix here allows unconfirmed students to access the student workspace when they are enrolled in any course that does not require email confirmation.

### DIFF
--- a/lib/oli/delivery/sections.ex
+++ b/lib/oli/delivery/sections.ex
@@ -5395,7 +5395,7 @@ defmodule Oli.Delivery.Sections do
     from(e in Enrollment,
       join: s in assoc(e, :section),
       where: e.user_id == ^user.id,
-      where: s.skip_email_verification == true
+      where: e.status == :enrolled and s.skip_email_verification == true
     )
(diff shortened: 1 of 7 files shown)


lib/oli/delivery/sections.ex
```diff
@@ -5387,4 +5387,20 @@ defmodule Oli.Delivery.Sections do
...
+    from(e in Enrollment,
+      join: s in assoc(e, :section),
+      where: e.user_id == ^user.id,
+      where: s.skip_email_verification == true
+    )
+    |> Repo.all()
+    |> case do
+      [] -> false
+      _ -> true
+    end
```
nicocirio, line 5404: NIT: we could use [Repo.exists?()](https://hexdocs.pm/ecto/Ecto.Repo.html#c:exists?/2) instead

nicocirio (approved):

PR #5470, opened 2025-03-07
[BUG FIX] [MER-4361] Fix product credentials issued tab

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-4361) to the ticket

### Granted certificates not being listed

The main issue was that the query that fetches the granted certificates for the current product did not include all granted certificates of courses created based on that product.
The fix was to extend `browse_granted_certificates/4` to include those certificates.

That same query was not filtering by state -> now [we only consider GC with state = :earned](https://eliterate.atlassian.net/browse/MER-4361?focusedCommentId=25919)

Tests for all modified/extended context functions are included in the PR.

#### Before
Screenshot 2025-03-07 at 10 57 19 AM

#### After
Screenshot 2025-03-07 at 10 50 13 AM

### Download CSV returned a blank spreadsheet

The `download CSV` link returned a blank spreadsheet since the query used to fetch the GCs had the same issue as the other one (it was not considering GC from courses created based on that product and was not filtering by earned state).
Now it returns the same GCs shown in the table.

A test was added to the `OliWeb.GrantedCertificatesController` module (the one that handles the CSV download)

#### Before
Screenshot 2025-03-07 at 10 52 40 AM

#### After
Screenshot 2025-03-07 at 10 51 14 AM

### Filtering GC by text was broken.
The CertificatesIssuedTab component is rendered both in the authoring and in the instructor UI.
The `delivery` path case was missing on the `select_path/4` helper function (the one used to patch the URL while filtering by text).

#### Before

https://github.com/user-attachments/assets/52508318-3b49-4de4-8c48-84d528564409

#### After

https://github.com/user-attachments/assets/fab5c4bc-aab6-4ed0-b209-24cb1f019410

### DIFF
--- a/lib/oli/delivery/certificates.ex
+++ b/lib/oli/delivery/certificates.ex
@@ -124,6 +124,7 @@ defmodule Oli.Delivery.Certificates do
         on: a.id == gc.issued_by and gc.issued_by_type == :author
       )
       |> where(^filter_by_section_or_blueprint)
+      |> where([gc, c, s, u, u1, a], gc.state == :earned)
       |> offset(^offset)
       |> limit(^limit)
       |> select(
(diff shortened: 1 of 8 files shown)


Francisco-Castro (approved):
Nice work! 🏅 Thanks for adding more testing.

PR #5475, opened 2025-03-08
[BUG FIX] [MER-4375] Do not create duplicate certificates in thresholds tab

manelli:
Assign certificate to socket assigns after inserting it so next time an update happens instead a new insert.
See: https://eliterate.atlassian.net/browse/MER-4375


Francisco-Castro (approved):
LGTM!

PR #5477, opened 2025-03-10
[BUG FIX] [MER-4377] mobile hamburger menu is broken in student workspace view

eliknebel:
https://eliterate.atlassian.net/browse/MER-4377

Fixes an issue with the mobile nav header menu. In some circumstances the mobile menu was rendering but would not respond to and actions. This was because of an id mismatch, however there were a number of other issues as well that needed fixed here such as:

- In some layouts the user menu renders off screen and cannot be accessed
- When a student shouldn't access other workspaces, the mobile menu was showing those workspaces
- User and author account settings were not mobile responsive

Before
Screenshot 2025-03-10 at 11 40 26 AM

After
Screenshot 2025-03-10 at 11 45 15 AM


Francisco-Castro:
While testing this PR locally, I found that for a **student**, the route `/sections/__SOME_COURSE_SECTION__/welcome` shows the hamburger menu, but it doesn’t do anything when clicked.

Additionally, for an **instructor**, the course section routes `/sections/coursesectionexample/manage` are not mobile-friendly. This is because the navbar contains different tabs, such as "Overview," "Insight," "Manage," etc.

eliknebel:
> While testing this PR locally, I found that for a **student**, the route `/sections/__SOME_COURSE_SECTION__/welcome` shows the hamburger menu, but it doesn’t do anything when clicked.
> 
> Additionally, for an **instructor**, the course section routes `/sections/coursesectionexample/manage` are not mobile-friendly. This is because the navbar contains different tabs, such as "Overview," "Insight," "Manage," etc.

Good catch, I will fix the instructor hamburger menu.

As for the manage sections view, these were out of scope for the short term mobile fixes as they are not as critical as the student views and we will address them in a more comprehensive mobile improvements task in the future.

lib/oli_web/components/delivery/layouts.ex
```diff
@@ -379,7 +382,7 @@ defmodule OliWeb.Components.Delivery.Layouts do
         </div>
       </nav>
       <nav
-        id="mobile-workspace-nav-menu"
+        id="mobile-nav-menu"
```
eliknebel, line 385: This was the root cause of the issue. However, more broadly because the menu button and this element and the menu button that triggers it are rendered separately, there was no guarantee that this element actually exists on the page and therefore would be broken on certain pages.

Refactoring the header to accept a `sidebar_enabled` attribute that is more in sync with the logic that renders the sidebar will help guarantee that the header hamburger menu button is only rendered when the supporting sidebar nav item is rendered as well.

lib/oli_web/components/delivery/user_account.ex
```diff
@@ -141,6 +142,7 @@ defmodule OliWeb.Components.Delivery.UserAccount do
   attr(:is_admin, :boolean, required: true)
   attr(:class, :string, default: "")
   attr(:dropdown_class, :string, default: "")
+  attr(:show_support_link, :boolean, default: false)
```
eliknebel, line 145: Added some logic to show a support menu item in the account menu when one is not available from the sidebar nav so that a support button is always available

lib/oli_web/components/layouts/workspace.html.heex
```diff
@@ -1,25 +1,13 @@
...
-      </div>
-    </div>
+    <Components.Delivery.Layouts.header
+      ctx={@ctx}
+      is_admin={@is_admin}
+      preview_mode={@preview_mode}
+      sidebar_expanded={@sidebar_expanded}
+      sidebar_enabled={!@disable_sidebar?}
+      include_logo
+    />
```
eliknebel, line 10: Removed duplicated markup in favor of a single implementation

rgachuhi (approved):

lib/oli_web/components/layouts/workspace.html.heex
Francisco-Castro: @eliknebel I was analyzing this part and found that it still works as expected even if we don't pass `sidebar_enabled`. The sidebar's show/hide behavior is controlled by `sidebar_expanded`. You did a great job simplifying this part. Thanks!

PR #5481, opened 2025-03-10
[BUG FIX] [MER-4243] Rework checkbox for omit student email verification not being honored

eliknebel:

Reworks an issue with omit email verification where there are more routes that need to bypass email verification, specifically attempt API routes.

The rework refactors the logic to determine whether to skip email verification or not into plugs instead of directly in the UserAuth module.


lib/oli_web/plugs/require_section.ex
```diff
@@ -28,21 +27,25 @@ defmodule Oli.Plugs.RequireSection do
...
     section = Repo.preload(section, [:brand, lti_1p3_deployment: [institution: [:default_brand]]])
 
     conn
     |> assign(:section, section)
+    |> assign(:skip_email_verification, skip_email_verification?(section))
     |> assign(:brand, Oli.Branding.get_section_brand(section))
   end
 
+  defp skip_email_verification?(%Section{skip_email_verification: true}), do: true
+  defp skip_email_verification?(_section), do: true
```
nicocirio, line 47: Shouldn't this return `false`?

eliknebel: Yes, good catch

rgachuhi: Quick question. Does this new plug based implementation account for cases where a student is registered in two sections, on that requires email verification and the other does not?

nicocirio (approved):

PR #5483, opened 2025-03-11
[BUG FIX] [MER-4378] Show error message when exceeding the number of entries

simonchoxx:
[MER-4378](https://eliterate.atlassian.net/browse/MER-4378)

This PR improve user feedback for Image upload errors in `Certificate Design Tab`.

Root Cause

Previously, when users attempted to upload more than the allowed three (3) images in the Design tab of the certificate settings, there was no clear indication that the upload limit had been exceeded. Additionally, errors related to file uploads, such as invalid file types or oversized files, were not consistently displayed in an organized manner.

Fix

- Added validation to detect when the user exceeds the maximum of three images and display a clear error message.

- Ensured that this error appears only once and is removed when the user deletes excess images

- Modified the layout to display file upload errors below the row of uploaded images, ensuring a clearer and more structured presentation of validation messages.

https://github.com/user-attachments/assets/c5111d8f-879a-402a-89f6-73190fc58d43


lib/oli_web/live/certificates/components/design_tab.ex
```diff
@@ -276,10 +280,20 @@ defmodule OliWeb.Certificates.Components.DesignTab do
...
   def handle_event("cancel", %{"ref" => ref}, socket) do
-    {:noreply,
-     socket
-     |> cancel_upload(:logo, ref)
-     |> assign(logo_upload_errors: formatted_upload_errors(socket.assigns.uploads.logo, ref))}
+    socket = cancel_upload(socket, :logo, ref)
+
+    uploads = socket.assigns.uploads.logo
+
+    updated_uploads = %{
```
manelli, line 287: I think we can remove this part altogether and in the `formatted_upload_errors/2` function at the end just do:

```elixir 
if uploads.entries > uploads.max_entries, do:
["You can upload up to #{uploads.max_entries} files only." | errors],
else: errors
```

manelli (commented):
Left minor comment about the logic of adding the errors. The UI is 💯

lib/oli_web/live/certificates/components/design_tab.ex
simonchoxx: Nice catch!

manelli (approved):
LGTM 🚀

PR #5488, opened 2025-03-12
[BUG FIX] [MER-4395] Ensures that a section resource page query with zero page ids returns empty array

rgachuhi:
This PR addresses an issue where a course without surveys lists all section resources when one views the `Surveys` table under the course `Insights` tab.


nicocirio (changes requested):
LGTM. Can we please add a test? thanks!

rgachuhi:
> LGTM. Can we please add a test? thanks!

I changed the solution after talking to Darren. Also, I added a test case

PR #5492, opened 2025-03-14
[ENHANCEMENT] [MER-4405] Allow erlang download to failover to secondary download location

Charles-Jones-Unicon:
The erlang.org site primary downloads become unavailable when the site is under maintenance. This will allow getting the source from the secondary location if the first wget fails.


Francisco-Castro (approved):
LGTM!

PR #5496, opened 2025-03-17
[BUG FIX] [MER-4243] checkbox for omit student email verification not being honored

eliknebel:

As seen by the multiple reworks of this issue, it is tricky to identify all routes that might need this check in place and as it expands, it is starting to impact more users and clutters the router pipelines.

This PR takes a more measured approach to restricting access to non-confirmed users. Instead of trying to "whack-a-mole" all of the routes that may or may not be required for non-confirmed user access, this PR simply adds the check for non-confirmed user to the delivery_protected plug and only runs the check for users who are not confirmed.

So for **all** delivery protected routes, if a user is non-confirmed then a check will be performed to see if the student is enrolled in any section that omits student email verification, and if so will allow access for the user. This database query check will only run if a user is non-confirmed, saving all other users the performance hit of an extra database query on every request.

Certain routes which do not utilize the `:delivery_protected` plug pipeline, such as account settings, remain as to require email confirmation before allowing access.

### DIFF
--- a/lib/oli_web/live/user_login_live.ex
+++ b/lib/oli_web/live/user_login_live.ex
@@ -69,7 +69,7 @@ defmodule OliWeb.UserLoginLive do
           <Components.Auth.login_form
             title="Sign In"
             form={@form}
-            action={~p"/users/log_in?#{[request_path: ~p"/workspaces/instructor"]}"}
+            action={~p"/users/log_in"}
             registration_link={~p"/users/register"}
             reset_password_link={~p"/users/reset_password"}
             authentication_providers={@authentication_providers}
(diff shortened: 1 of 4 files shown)


lib/oli_web/live/user_login_live.ex
```diff
@@ -69,7 +69,7 @@ defmodule OliWeb.UserLoginLive do
           <Components.Auth.login_form
             title="Sign In"
             form={@form}
-            action={~p"/users/log_in?#{[request_path: ~p"/workspaces/instructor"]}"}
+            action={~p"/users/log_in"}
```
eliknebel, line 72: This was a somewhat-related issue, where users were incorrectly being sent to the instructor workspace if they were logging in with the generic sign in page. This method will instead relegate that determination to the login redirection logic instead of incorrectly specifying it here

rgachuhi (approved):

PR #5322, opened 2024-12-20
[FEATURE] [MER-4068] Reimplement user invite functionality

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-4068) to the ticket

As discussed in the ticket, this PR also introduced the `:pending_confirmation` and `:rejected` statuses for and enrollment, to allow the user to decide what to do with an invitation (in the previous version the user was enrolled and the email invitation ended up being an information email)

So, the student's overview page (in the instructor dashboard) was expanded to allow the instructor to filter by these two new values.
On the other hand, now the student workspace won't show the course until the student ends up accepting the invitation (in a future iteration we might want to show a `pending confirmation` course as "grayed" with a link to the UserInvite liveview)

### A non-existing user invitation flow

https://github.com/user-attachments/assets/99d3c9a4-8ac6-4c83-b622-3d8a9335396f

### An existing user invitation flow (not logged in -> password must be provided) 

https://github.com/user-attachments/assets/e17f246b-4f01-40f6-8a1b-87ed7c274067

### An existing user invitation flow (logged in with the same account as the invitation)
Since the user is already logged in, the user invitation is confirmed and the user gets redirected to the course as soon as it clicks the "Accept" button

https://github.com/user-attachments/assets/d17ba6e9-9006-4a8b-bb6c-060cf35646dd

### An existing user invitation flow (logged in with a different account than the invitation)
The user sees a warning and will be logged in with the account corresponding to the invitation

https://github.com/user-attachments/assets/a44c9fcb-3491-4d56-90cf-37d5c9ab9929

### Rejecting an invitation flow

https://github.com/user-attachments/assets/54fc8960-f07a-4b5f-9f82-84f6be96266e

### An already rejected invitation
image

### An already accepted invitation
image

### A non-existing invitation
image

#### Student invitation
image

#### Instructor invitation
image

PR #5499, opened 2025-03-18
[BUG FIX] [MER-4401] User login is not being constrained to independent users

simonchoxx:
[MER-4401](https://eliterate.atlassian.net/browse/MER-4401)

Fixed a 500 error when logging in with an email associated with both an LTI user and an independent learner.

Root Cause

The issue was caused by a query that did not properly filter independent users, resulting in multiple records for the same email. Since LTI users and independent learners can share the same email, the query returned multiple results, leading to an Ecto.MultipleResultsError.

Fix

An existing function that properly filters users based on the independent_learner flag was used, ensuring that any login attempt by email only considers independent learners and preventing the multiple results error.

[MER-4401]: https://eliterate.atlassian.net/browse/MER-4401?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli_web/controllers/user_session_controller.ex
+++ b/lib/oli_web/controllers/user_session_controller.ex
@@ -29,7 +29,7 @@ defmodule OliWeb.UserSessionController do
         user_params
       end
 
-    if user = Accounts.get_user_by_email_and_password(email, password) do
+    if user = Accounts.get_independent_user_by_email_and_password(email, password) do
       conn
       |> maybe_add_flash_message(opts[:flash_message])
       |> UserAuth.log_in_user(user, user_params)
--- a/test/oli_web/controllers/user_session_controller_test.exs
+++ b/test/oli_web/controllers/user_session_controller_test.exs
@@ -1,6 +1,8 @@
 defmodule OliWeb.UserSessionControllerTest do
   use OliWeb.ConnCase, async: true
 
+  import Oli.Factory
+
   setup do
     %{user: user_fixture()}
   end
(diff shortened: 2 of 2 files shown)


lib/oli_web/controllers/user_session_controller.ex
```diff
@@ -29,7 +29,7 @@ defmodule OliWeb.UserSessionController do
         user_params
       end
 
-    if user = Accounts.get_user_by_email_and_password(email, password) do
+    if user = Accounts.get_independent_user_by_email_and_password(email, password) do
```
nicocirio, line 32: I am wondering why I may have [changed this](https://github.com/Simon-Initiative/oli-torus/pull/5322/files#diff-2d21dcf6d1de8fc14bffffc64b0100679aca60d63abe84cee70f06ce1d0f9764R24) two months ago. Probably I made a mistake.
I think we should use the `get_independent_user_by_email_and_password` but also extend the [user invitate changeset](https://github.com/Simon-Initiative/oli-torus/pull/5322/files#diff-92715fea158bdca1d531d6c60b6a121a0fa824308c00d0bad55398818e11b6d3R503-R507) to set the independent_learner flag to true when inviting users into a course (`put_change(:independent_learner, true)`)

@eliknebel does that make sense?

eliknebel: I think so, but it looks like this field defaults to true in the user schema
```
field :independent_learner, :boolean, default: true
```
which means we might not have to do anything to set it

nicocirio (approved):
Nice work Santi!

PR #5500, opened 2025-03-18
[PERFORMANCE] [MER-4412] Truncate in memory representation of log messages

darrensiegel:
This PR introduces and installs a configurable `Oli.LoggerTruncator`, which truncates items being logged.  

This addresses a shortcoming of the currently used `:logger` `truncate` (see `runtime.exs`) which only truncates what is actually written to the logs.  The new `Oli.LoggerTruncator` truncates the in-memory representation, which is important as that is what is passed around between things like the Logger and downstream backends like AppSignal.  

We have been seeing huge memory spikes on Proton which look like this:

Screenshot 2025-03-18 at 7 34 30 AM

This PR will eliminate this problem by ensure that the seemingly HUGE message being stored in memory gets truncated down to max size in bytes (default `5_000`)

This is a low risk addition because we can completely turn this off by setting `LOGGER_TRUNCATION_ENABLED` equal to `false`

## TO VERIFY DURING PR REVIEW

Set both `LOGGER_TRUNCATION_ENABLED` to `true` and `LOGGER_TRUNCATION_LENGTH` to something observably low like `10`.  Then start the app and click around, you should see the logging messages being appropriately truncated.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -35,7 +35,14 @@ get_env_as_boolean = fn key, default ->
   end
 end
 
+get_env_as_integer = fn key, default ->
+  System.get_env(key, default)
+  |> String.to_integer()
+end
+
 config :oli,
+  logger_truncation_enabled: get_env_as_boolean.("LOGGER_TRUNCATION_ENABLED", "true"),
+  logger_truncation_length: get_env_as_integer.("LOGGER_TRUNCATION_LENGTH", "5000"),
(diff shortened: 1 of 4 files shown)


eliknebel (approved):

nicocirio (approved):
LGTM!

We could also extend `oli.example.env`:
```
LOGGER_TRUNCATION_ENABLED=false
LOGGER_TRUNCATION_LENGTH=5000
```

In this way a dev can decide to apply or not the truncate feature (probably in dev we do not want to truncate, to be able to debug)

PR #5502, opened 2025-03-19
[BUG FIX] [MER-4411] using social login with existing lti account email adds identity provider to lti user

eliknebel:
https://eliterate.atlassian.net/browse/MER-4411

After an extensive investigation, I was unable to reproduce this issue but verified the situation that occurred was due to existing production data from an issue with a previous release. I confirmed that the user_identities record used to link the user account with the google provider was actually created back in August 2024 so it could not have been caused by an issue in the latest version.

The PR confirms this by introducing a unit test to exercise this case and is passing with the expected behavior. I also noticed and removed a duplicate log statement as part of this work.

### DIFF
--- a/lib/oli_web/common/assent_auth_web.ex
+++ b/lib/oli_web/common/assent_auth_web.ex
@@ -107,8 +107,6 @@ defmodule OliWeb.Common.AssentAuthWeb do
   end
 
   defp split_user_identity_params(params, _provider) do
-    Logger.error("No sub found in user params: #{inspect(params)}")
-
     {:error, {:invalid_user_identity_params, {:missing_param, "sub", params}}}
   end
 
--- a/test/oli_web/common/assent_auth_web_test.exs
+++ b/test/oli_web/common/assent_auth_web_test.exs
@@ -4,12 +4,11 @@ defmodule OliWeb.Common.AssentAuthWebTest do
   import Oli.Factory
 
   alias Oli.Accounts
-  alias Oli.AssentAuth.{AuthorAssentAuth, AuthorIdentity}
-  alias OliWeb.AuthorAuth
+  alias Oli.AssentAuth.{AuthorAssentAuth, AuthorIdentity, UserAssentAuth}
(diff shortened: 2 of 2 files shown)


test/oli_web/common/assent_auth_web_test.exs
```diff
@@ -130,5 +129,109 @@ defmodule OliWeb.Common.AssentAuthWebTest do
...
+               {:invalid_user_identity_params,
+                {:missing_param, "sub",
+                 %{
+                   "email" => author_email,
+                   "name" => author_name
+                 }}}
+    end
+  end
+
+  defp user_test_config(),
```
Francisco-Castro, line 166: Could you please move this function to the bottom?

lib/oli_web/common/assent_auth_web.ex
```diff
@@ -107,8 +107,6 @@ defmodule OliWeb.Common.AssentAuthWeb do
   end
 
   defp split_user_identity_params(params, _provider) do
-    Logger.error("No sub found in user params: #{inspect(params)}")
-
     {:error, {:invalid_user_identity_params, {:missing_param, "sub", params}}}
   end
```
Francisco-Castro, line 111: Could you please remove `require Logger` at the beginning of the module since we're no longer logging from here?

Francisco-Castro (changes requested):
Thanks for adding more tests to `handle_authorization_success/4`.

It seems that `handle_authorization_success/4` can return `{:error, error, conn}`, but Dialyzer isn’t recognizing that output correctly.

Francisco-Castro (commented):
Thanks for making the changes.

Do you think it's safe to delete the following condition from `handle_authorization_success/4`? I checked every function in the `with` statement, and it looks like we cannot reach that part.

```
{:error, error, conn} ->
        {:error, error, conn}
```

Francisco-Castro (approved):
Thank you for addressing all the comments.

PR #5505, opened 2025-03-20
[BUG FIX] [MER-4426] Attempts auto submit after suggested date has passed

simonchoxx:
[MER-4426](https://eliterate.atlassian.net/browse/MER-4426)

This PR ensures that when an assessment has a past suggested date with the late policy set to “Disallow late start and late submit” and the scheduling type is :read_by, the attempt is not auto-submitted upon entering the page.

Root Cause

Previously, the logic determining whether an attempt should be auto-submitted did not account for assessments with :read_by scheduling type. As a result, even when late start and late submit were disallowed, and the suggested date had passed, the attempt was incorrectly auto-submitted upon page entry.

Fix

The logic now correctly excludes :read_by assessments from the auto-submit condition, ensuring that students can enter and view their attempt without it being immediately submitted.

https://github.com/user-attachments/assets/770a3a60-bdcb-494f-b132-eda5abe1ff5d

[MER-4426]: https://eliterate.atlassian.net/browse/MER-4426?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli_web/live/delivery/student/lesson_live.ex
+++ b/lib/oli_web/live/delivery/student/lesson_live.ex
@@ -135,7 +135,15 @@ defmodule OliWeb.Delivery.Student.LessonLive do
       now = DateTime.utc_now() |> to_epoch
 
       attempt_expired_auto_submit =
-        now > effective_end_time and auto_submit and !page_context.review_mode
+        with true <- now > effective_end_time,
+             true <- auto_submit,
+             false <- page_context.review_mode,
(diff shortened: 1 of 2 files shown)


nicocirio (approved):
LGTM! 🚀

PR #5506, opened 2025-03-20
[BUG FIX] [MER-3510] Cannot create adaptive page from all pages

simonchoxx:
[MER-3510](https://eliterate.atlassian.net/browse/MER-3510)

This PR fixes the issue where attempting to create an adaptive page from the `All Pages` view resulted in a page reload without actually creating a new adaptive page.

Root Cause

The root cause of the issue was that incorrect parameters were being passed to the function responsible for creating both adaptive and basic pages.

Fix

The fix was based on fixing the parameters passed to the function responsible for creating pages, ensuring the page type and grading status are properly defined.

https://github.com/user-attachments/assets/b0e723b9-7502-4e95-913e-3f735bcda3e4

[MER-3510]: https://eliterate.atlassian.net/browse/MER-3510?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli_web/live/workspaces/course_author/pages_live.ex
+++ b/lib/oli_web/live/workspaces/course_author/pages_live.ex
@@ -398,7 +398,7 @@ defmodule OliWeb.Workspaces.CourseAuthor.PagesLive do
     {:noreply, assign(socket, modal_assigns: modal_assigns)}
   end
 
-  def handle_event("create_page", %{"type" => type}, socket) do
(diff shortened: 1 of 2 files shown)


Francisco-Castro (approved):
LGTM.

Thanks for adding tests for all three cases. 🏅

PR #5511, opened 2025-03-21
[BUG FIX] [MER-4429] Instructor Dashboard search: ignore nil instructor names

manelli:
If for any given reason the instructors of a section end up in a inconsistent state and they do not have a name associated, when filtering the sections list (specifically when downcasing the instructor names, see: [Appsignal Incident](https://appsignal.com/open-learning-initiative/sites/618539ab2cf81d7e3cd051ca/exceptions/incidents/62/samples/618539ab2cf81d7e3cd051ca-477208838546314286517424913201)) it would fail returning a 500.

I will add a log to detect these cases in a separate PR, related to this same bug, since the other ticket also involves work in the "Manage Source Materials" section.

See: https://eliterate.atlassian.net/jira/software/c/projects/MER/issues/MER-4429


Francisco-Castro (approved):
Tested in an IEx session, and it works correctly.

Thanks for considering adding a log to detect the source of the issue.

PR #5512, opened 2025-03-21
[BUG FIX] [MER-4430] Fixes email cleared on any action in the create dataset ui

rgachuhi:
This PR addresses an issue where the email list in the "Enter additional emails" field on the "create dataset job" tab gets cleared every time a user interacts with the courses table


Francisco-Castro (approved):
LGTM

I left a recommendation.

PR #5515, opened 2025-03-25
[ENHANCEMENT] [MER-3577] Prevent Adding Question Components to the End Screen in Simple Author

dtiwarATS:

### DIFF
--- a/assets/src/apps/authoring/components/Flowchart/screens/screen-factories.ts
+++ b/assets/src/apps/authoring/components/Flowchart/screens/screen-factories.ts
@@ -16,6 +16,19 @@ export const screenTypes = [
   'dropdown',
 ];
 
+export const screenType = {
+  BLANK_SCREEN: 'blank_screen',
+  WELCOME_SCREEN: 'welcome_screen',
+  MULTIPLE_CHOICE_SCREEN: 'multiple_choice',
+  MULTILINE_SCREEN: 'multiline_text',
+  SLIDER_SCREEN: 'slider',
+  HUB_SPOKE_SCREEN: 'hub_spoke',
+  END_SCREEN: 'end_screen',
+  NUMBER_INPUT_SCREEN: 'number_input',
+  TEXT_INPUT_SCREEN: 'text_input',
+  DROPDOWN_SCREEN: 'dropdown',
+};
+
 export const screenTypeToTitle: Record<string, string> = {
   blank_screen: 'Instructional Screen',
   welcome_screen: 'Welcome Screen',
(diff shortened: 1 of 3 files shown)


assets/src/apps/authoring/components/Flowchart/screens/screen-factories.ts
```diff
@@ -16,6 +16,19 @@ export const screenTypes = [
   'dropdown',
 ];
 
+export const screenType = {
```
bsparks, line 19: I think if you are going to create an enum for the screenType then it should be an actual typescript string enum, and also you should use it again in the other items here, like the above screenTypes would then be like:
```
export const screenTypes = [
  screenType.BLANK_SCREEN,
  screenType.WELCOME_SCREEN,
  // ... etc
  ```

File: assets/src/apps/authoring/components/Flowchart/toolbar/FlowchartHeaderNav.tsx [hunk 2]
```diff
@@ -143,7 +150,8 @@ export const FlowchartHeaderNav: React.FC<HeaderNavProps> = () => {
 
   const questionType = getScreenQuestionType(currentActivity);
   const hasQuestion = questionType !== 'none';
-
+  const isLessonEndScreen =
```
bsparks, line 153: besides the enum idea, I notice there is also this: 
https://github.com/Simon-Initiative/oli-torus/blob/master/assets/src/apps/authoring/components/Flowchart/screens/screen-utils.ts#L58

```
const isEndScreen = (screen: IActivity): boolean =>
  screen.authoring?.flowchart?.screenType === 'end_screen';
```

so you could just potentially use the utils instead of doing the enum

assets/src/apps/authoring/components/Flowchart/toolbar/FlowchartHeaderNav.tsx
```diff
@@ -363,11 +372,14 @@ export const FlowchartHeaderNav: React.FC<HeaderNavProps> = () => {
                   <Tooltip id="button-tooltip" style={{ fontSize: '12px' }}>
                     <strong>Paste Component</strong>
                     {hasQuestion && <div>Only one question component per screen is allowed</div>}
+                    {isLessonEndScreen && (
```
bsparks, line 375: also doing this makes me feel funny because it's a bit specific. maybe I'd prefer having some kind of `canAddQuestions` variable instead which right now would be `!isLessonEndScreen` but could change later to have other criteria;

separately from that however, I notice there is some sort of screen validation thing that already has some rules and reasons in it, can that be enhanced instead?

dtiwarATS: I thought of renaming the variable but the `'Question/interaction components cannot be added to the last screen.'` validation message was too specific to the end of lesson use case.

bsparks (approved):

PR #5516, opened 2025-03-25
[BUG FIX] [MER-4437] Progress for practice page not counting on homepage

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-4437) to the ticket

The issue was that we were calculating the progress in a way that it could never reach 1.0, since the denominator was always slightly larger than the numerator (because we were adding 0.0001 to the denominator):

### Before

```
completed_count / (total_count * ((COALESCE(rev.full_progress_pct, 100) / 100) + 0.0001))
```

### After

```
completed_count / (total_count * ((COALESCE(rev.full_progress_pct, 100) / 100.0)))
```

### Before

https://github.com/user-attachments/assets/236ac93d-eca6-4024-9d33-14d4193dcac7

### After

https://github.com/user-attachments/assets/f5805475-01e2-4cd4-bd7a-2df2bcccc301

### DIFF
--- a/lib/oli/delivery/metrics.ex
+++ b/lib/oli/delivery/metrics.ex
@@ -1211,7 +1211,7 @@ defmodule Oli.Delivery.Metrics do
             GREATEST(
               (
                 SELECT
-                  completed_count / (total_count * ((COALESCE(rev.full_progress_pct, 100) / 100) + 0.0001))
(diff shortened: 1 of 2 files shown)


simonchoxx (approved):
Nice work 👏

PR #5517, opened 2025-03-26
[BUG FIX] [MER-4446] assessments are being marked late when submitting after suggested date

Francisco-Castro:
Ticket: [MER-4446](https://eliterate.atlassian.net/browse/MER-4446)

This PR fixes an issue where an instructor sets a graded page with a suggested "read by" date, but it is not honored.

The fix can be found in commit e4508cce38e6815f549f66d6f8ae992b3dba5da3 , where the `Oli.Delivery.Settings.was_late?/3` function was not considering the case when `%Combined{scheduling_type: :read_by}`. This case is used to determine that the grade page is not late.

[MER-4446]: https://eliterate.atlassian.net/browse/MER-4446?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli/delivery/settings.ex
+++ b/lib/oli/delivery/settings.ex
@@ -238,7 +238,7 @@ defmodule Oli.Delivery.Settings do
     |> Repo.update()
   end
 
-  def was_late?(_, %Combined{late_submit: :disallow}, _now), do: false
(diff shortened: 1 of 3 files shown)


nicocirio (approved):
LGTM 🚀

PR #5519, opened 2025-03-26
[BUG FIX] [MER-4454] Keep original granted certificate distinction when approving/reapproving

manelli:
Also fix MER-4452, that involved a duplicated module name in the tests

See:
- https://eliterate.atlassian.net/browse/MER-4454
- https://eliterate.atlassian.net/browse/MER-4452


lib/oli_web/components/delivery/students/certificates/state_approval_component.ex
```diff
@@ -204,7 +207,7 @@ defmodule OliWeb.Components.Delivery.Students.Certificates.StateApprovalComponen
              url: nil,
              student_email_sent: false,
              guid: UUID.uuid4(),
-             with_distinction: false
+             with_distinction: current_gc.with_distinction
```
nicocirio, line 210: Instead of fetching the current_gc and then assigning its with_distinction value in the update, I believe we could directly remove the `with_distinction` key from the attrs passed to the update_granted_certificate function. The value of `with_distinction` will remain the same, and we will avoid running a query (the one that fetches current_gc).

nicocirio (changes requested):
I think we could avoid fetching the current_gc. Please let me know if it makes sense.

lib/oli_web/components/delivery/students/certificates/state_approval_component.ex
manelli: 🤦 i forgot it was an update 😆 
Thanks for the catch, fixed in 64e3f4aba8dda3aea01036822dee991251a7ea0a

nicocirio (approved):

PR #5527, opened 2025-03-27
[ENHANCEMENT] [MER-3983] Flaky Tests (Part 1)

manelli:
Now flaky tests are tagged as such, and tests that _always_ fail are marked as skipped (we need to get into those and try to fix them and remove the ones that are related to legacy features)

The Github action that runs the tests, now reruns the initial failed tests once more in order to avoid having to retrigger the CI because of flaky tests.

See: https://eliterate.atlassian.net/browse/MER-3983

----

## Skipped tests

| File                                                                                             | Count |
|--------------------------------------------------------------------------------------------------------|----------------|
| test/oli_web/controllers/page_delivery_controller_test.exs                                             | 38             |
| test/oli_web/live/new_course/course_details_test.exs                                                  | 13             |
| test/oli_web/live/delivery/instructor_dashboard/insights/learning_objectives_tab_test.exs             | 3              |
| test/oli_web/live/delivery/student/learn_live_test.exs                                                | 3              |
| test/oli_web/live/delivery/student_dashboard/components/learning_objectives_tab_test.exs              | 3              |
| test/oli_web/live/ingest_live_test.exs                                                                | 3              |
| test/oli_web/live/objectives_live_test.exs                                                            | 3              |
| test/oli_web/controllers/legacy_superactivity_controller_test.exs                                     | 1              |
| test/oli_web/live/sections/gating_and_scheduling_test.exs                                             | 1              |
| test/oli_web/live/workspaces/course_author/curriculum_live_test.exs                                   | 1              |

## Flaky tests
| Test File                                                                                     | Count |
|-----------------------------------------------------------------------------------------------|----------------|
| test/oli/analytics/datasets_test.exs                                                          | 2              |
| test/oli_web/live/objectives_live_test.exs                                                    | 2              |
| test/oli_web/live/delivery/student/learn_live_test.exs                                        | 2              |
| test/oli/delivery/section_cache_test.exs                                                      | 1              |
| test/oli/delivery/sections/blueprint_test.exs                                                 | 1              |
| test/oli/publishing/authoring_resolver_test.exs                                               | 1              |
| test/oli_web/live/certificates/components/certificates_issued_tab_test.exs                    | 1              |
| test/oli_web/live/delivery/student/discussions_live_test.exs                                  | 1              |
| test/oli_web/live/delivery/student/lesson_live_test.exs                                       | 1              |
| test/oli_web/live/grades_live_test.exs                                                        | 1              |
| test/oli_web/live/products_test.exs                                                           | 1              |
| test/oli_web/live/sections/gating_and_scheduling_test.exs                                     | 1              |
| test/oli_web/plugs/header_size_logger_test.exs                                                | 1              |

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -72,7 +72,10 @@ jobs:
         run: mix format --check-formatted
 
       - name: ▶️ Run unit tests
-        run: set -a;source oli.env && MIX_ENV=test mix ecto.reset && mix test
+        run: |
+          set -a;source oli.env
+          MIX_ENV=test mix ecto.reset
+          mix test || if [[ $? = 2  ]]; then mix test --failed; else false; fi # https://angelika.me/2024/01/08/do-not-run-mix-test-failed/
 
       - name: 📦 Install node_module dependencies
         run: yarn --cwd assets --frozen-lockfile
--- a/test/oli_web/live/delivery/student/learn_live_test.exs
+++ b/test/oli_web/live/delivery/student/learn_live_test.exs
@@ -1276,8 +1276,8 @@ defmodule OliWeb.Delivery.Student.ContentLiveTest do
              )
     end
 
-    @tag :skip
     # This feature was disabled in ticket NG-201 but will be reactivated with NG23-199
+    @tag :skip
     test "can see module learning objectives (if any) in the tooltip", %{
       conn: conn,
       section: section
@@ -1427,7 +1427,6 @@ defmodule OliWeb.Delivery.Student.ContentLiveTest do
     end
 
     # TODO: finish this test when the handle event for the "Let's discuss" button is implemented
-    @tag :skip
     test "can click on let's discuss button to open DOT AI Bot interface", %{
       conn: conn,
       section: section
@@ -1549,7 +1548,6 @@ defmodule OliWeb.Delivery.Student.ContentLiveTest do
     end
 
     @tag :flaky
-    @tag :skip
     test "can see units, modules and page (at module level) progresses", %{
       conn: conn,
       user: user,
@@ -2213,7 +2211,6 @@ defmodule OliWeb.Delivery.Student.ContentLiveTest do
     end
 
     @tag :flaky
-    @tag :skip
     test "can see checked square icon and score details for attempted graded pages in the module index details",
          %{
            conn: conn,
--- a/test/oli_web/live/delivery/student/learn_live_test.exs
+++ b/test/oli_web/live/delivery/student/learn_live_test.exs
@@ -2210,55 +2210,6 @@ defmodule OliWeb.Delivery.Student.ContentLiveTest do
              )
     end
 
-    @tag :flaky
-    test "can see checked square icon and score details for attempted graded pages in the module index details",
-         %{
-           conn: conn,
-           user: user,
-           section: section,
-           mcq_1: mcq_1,
-           page_4: page_4_revision,
-           project: project,
-           publication: publication
-         } do
-      set_progress(section.id, page_4_revision.resource_id, user.id, 1.0, page_4_revision)
-
-      set_activity_attempt(
-        page_4_revision,
-        mcq_1,
-        user,
-        section,
-        project.id,
-        publication.id,
-        "id_for_option_a",
-        true
-      )
-
-      set_activity_attempt(
-        page_4_revision,
-        mcq_1,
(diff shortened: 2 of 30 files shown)


test/oli_web/live/delivery/student/learn_live_test.exs
```diff
@@ -2209,54 +2210,6 @@ defmodule OliWeb.Delivery.Student.ContentLiveTest do
              )
     end
 
-    test "can see checked square icon and score details for attempted graded pages in the module index details",
```
manelli, line 2212: This test was removed because it was duplicated in this same file

manelli:
Also maybe we can remove the `auto-format` (Auto format and lint) job, I don't think is really necessary since if we have a format error it would be detected on the `elixir-build-test` job which does a `mix format --check-formatted`

WDYT?

manelli:
Aaaalso, is this warning message necessary?  Its polluting the output when running tests
```
 [warning] The section association [:brand] has not been preloaded for branding. The association will be loaded now but may result in performance issues
 ```
 
 It's from `Branding.ensure_preloaded/2` private function and emits this warning when a section does not have the `brand` or `lti_1p3_deployment->institution->default_brand` association preloaded.

File: .github/workflows/build.yml [hunk 2]
```diff
@@ -72,7 +72,10 @@ jobs:
         run: mix format --check-formatted
 
       - name: ▶️ Run unit tests
-        run: set -a;source oli.env && MIX_ENV=test mix ecto.reset && mix test
+        run: |
+          set -a;source oli.env
+          MIX_ENV=test mix ecto.reset
+          mix test || if [[ $? = 2  ]]; then mix test --failed; else false; fi # https://angelika.me/2024/01/08/do-not-run-mix-test-failed/
```
gastonabella, line 78: I agree this would save us some time and headaches when pushing code, but it also was the way to identify and track flaky tests.

gastonabella (approved):

eliknebel (changes requested):
We do not want to turn off the mix test enforcement

.github/workflows/build.yml
eliknebel: We do not want to turn off the mix test enforcement.

It seems like this perhaps just reprints the failed tests? If that is the case and the CI still fails then this is fine and a nice addition. But we need to make sure no PR can land with failing tests, flaky or not

eliknebel:
> Also maybe we can remove the `auto-format` (Auto format and lint) job, I don't think is really necessary since if we have a format error it would be detected on the `elixir-build-test` job which does a `mix format --check-formatted` and it takes 6 minutes to run
> 
> If we end up removing the `auto-format` job I think we can remove the `check-simon-bot-access-token` one WDYT?

This auto format should remain. It will do a best attempt to format the PR automatically for a developer and commit the changes to kick off a new build so that a developer is not required to fix every little trivial formatting issue themselves

eliknebel:
> Aaaalso, is this warning message necessary? Its polluting the output when running tests
> 
> ```
>  [warning] The section association [:brand] has not been preloaded for branding. The association will be loaded now but may result in performance issues
> ```
> 
> It's from `Branding.ensure_preloaded/2` private function and emits this warning when a section does not have the `brand` or `lti_1p3_deployment->institution->default_brand` association preloaded.

I agree that this is noisy in the logs, but it is actually indicative of an issue that when this function is called, the brand has not been preloaded and it should have been. We should find where this is happening and make sure there is a proper plug in place to preload the brand for the section

eliknebel (changes requested):
We should track down the branding preloading issue and replace the warning log message with a crash so that tests cannot pass when this condition occurs. This will ensure that branding is always preloaded for delivery routes that require it

eliknebel:
What is the difference between the :skip and :flaky tags here? It looks like there were not many tests that were resolved, but more are being skipped. Didn't we want to either fix or remove the flaky tests as part of this work?

manelli:
> What is the difference between the :skip and :flaky tags here? It looks like there were not many tests that were resolved, but more are being skipped. Didn't we want to either fix or remove the flaky tests as part of this work?

Skipped tests are tests that always fail (with exception of 2 that are flaky but fail 90% of the time)
Flaky tests are marked for informational purposes (they run on mix test).

I have a separate branch (part-2) with some test fixes but in order to not pollute this PR.

eliknebel (approved):

PR #5530, opened 2025-04-01
[FEATURE] [MER-4097] Student outline view search and expand features

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-4097) to the ticket

https://github.com/user-attachments/assets/0f14a17e-866b-4180-a93a-bffc51b58bb1

### DIFF
--- a/assets/css/app.css
+++ b/assets/css/app.css
@@ -136,7 +136,7 @@ https://tailwindcss.com/docs/utility-first
 }
 
 .search-result em {
-  @apply not-italic bg-yellow-50;
+  @apply not-italic bg-yellow-100 dark:bg-yellow-800;
 }
 
 .progress-options {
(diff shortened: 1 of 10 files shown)


Francisco-Castro (approved):
Awesome work! ⭐ 

I really like the part about defining anonymous functions—pretty cool! The highlight feature is very clever. I could go on and on...

One thing that came to my mind is that we are using streams, and I’m wondering why we’re not using normal lists instead. We could discuss this outside.

nicocirio:
Thanks Fran! Regarding the use of streams, I think it is a good use case because we do not want to hold the hierarchy list in the server memory. 
Streams allow us to modify an element on the client's list (insert/delete or even change its position) without having to store the full list in the server's memory. In this case, we are just taking advantage of the memory benefit of streams (and not the insert/delete/positioning advantage). 
Another way to achieve the "free-server-memory" advantage could have been by combining a `common` list with temporary_assigns.

PR #5531, opened 2025-04-01
[FEATURE] [MER-3452] Bottom bar improvement lesson pages

simonchoxx:
[MER-3452](https://eliterate.atlassian.net/browse/MER-3452)

This PR introduces changes and improvements to the navigation bar displayed at the bottom of the `Lesson Pages`.

It improves the way the navigation bar is displayed, leaving it hidden by default and showing it when the user hovers on it. 
Also, when the user navigates to the bottom of the page, the bar will be fixed at the bottom. 

On the other hand, icons are added next to the title of the previous page and the next page, indicating what type of page they are and if they are completed or not. 

For the icons, the same logic used in the `Learn View` is maintained, i.e.: 

- Scored pages completed ---> Square check icon
- Scored pages not completed ---> Orange flag icon
- Practice pages completed ---> Check icon
- Practice pages not completed ---> No icon

Additionally the text of the previous and next pages is truncated, adding ellipses if the text is too long. In this case a tooltip is also added over the title to show it completely. 

https://github.com/user-attachments/assets/9bed1615-c3b2-4fd1-b4c5-6e7be92f8632

[MER-3452]: https://eliterate.atlassian.net/browse/MER-3452?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli/delivery/metrics.ex
+++ b/lib/oli/delivery/metrics.ex
@@ -214,6 +214,26 @@ defmodule Oli.Delivery.Metrics do
   end
 
   @doc """
+  Calculate the progress for a given student for a list of pages.
+  """
+
+  def progress_for_pages(section_id, user_id, page_ids) do
+    from(ra in ResourceAccess,
+      where:
+        ra.resource_id in ^page_ids and
+          ra.section_id == ^section_id and
+          ra.user_id == ^user_id,
+      group_by: ra.resource_id,
+      select: {
+        ra.resource_id,
+        fragment("COALESCE(SUM(?), 0)", ra.progress)
+      }
+    )
+    |> Repo.all()
+    |> Enum.into(%{})
+  end
+
+  @doc """
   Calculate the percentage of students that have completed a container or page
   """
   def completion_for(section_id, container_id) do
--- a/lib/oli_web/components/delivery/layouts.ex
+++ b/lib/oli_web/components/delivery/layouts.ex
@@ -850,6 +850,7 @@ defmodule OliWeb.Components.Delivery.Layouts do
   attr(:previous_page, :map)
   attr(:next_page, :map)
   attr(:section_slug, :string)
+  attr(:pages_progress, :map)
   attr(:request_path, :string)
   attr(:selected_view, :string, doc: "The selected view for the Learn page (gallery or outline)")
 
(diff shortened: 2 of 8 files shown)


lib/oli/delivery/metrics.ex
```diff
@@ -213,6 +213,26 @@ defmodule Oli.Delivery.Metrics do
...
+          ra.user_id == ^user_id,
+      group_by: ra.resource_id,
+      select: {
+        ra.resource_id,
+        fragment("COALESCE(SUM(?), 0)", ra.progress)
+      }
+    )
+    |> Repo.all()
+    |> Enum.into(%{})
+  end
```
nicocirio, line 234: Does it make sense to add a test for this function?

lib/oli_web/components/delivery/layouts.ex
```diff
@@ -1122,4 +1139,16 @@ defmodule OliWeb.Components.Delivery.Layouts do
...
   end
+
+  defp maybe_add_icon(page, pages_progress) do
+    page_id = String.to_integer(page["id"])
+    progress = Map.get(pages_progress, page_id, nil)
+
+    case {progress, page["graded"]} do
+      {1.0, "false"} -> apply(OliWeb.Icons, :check, [%{}])
+      {1.0, "true"} -> apply(OliWeb.Icons, :square_checked, [%{}])
+      {progress, "true"} when progress < 1.0 -> apply(OliWeb.Icons, :flag, [%{}])
```
nicocirio, line 1150: NIT: I believe the progress cannot be bigger than 1.0, so the `progress < 1.0` guard is not strictly necessary

nicocirio (changes requested):
Nice work Santi! I loved the transition animations.
I left some NITs and requested changes about some code comments that we might want to delete or translate into English

lib/oli_web/components/delivery/layouts.ex
simonchoxx: I had a conversation with @Francisco-Castro  via slack a few days ago about this case. 
This could be written as it is already written, that is: 

`{progress, “true”} when progress < 1.0 -> apply(OliWeb.Icons, :flag, [%{}]) `

or this other way: 

`{_, “true”} -> apply(OliWeb.Icons, :flag, [%{}])`

Both are correct and functional. 

I decided to leave it as it is now so that the case is explicit that if I receive any progress less than 1.0 I will enter there.

simonchoxx: mmmm but on second thought, now I like the simpler line.
I'll change it. Thanks!

nicocirio: I have no strong opinion about it, both approaches are fine for me, it was just a NIT comment 😄

nicocirio (approved):
🚀

PR #5539, opened 2025-04-10
[BUG FIX] [MER-4483] Improve Certificates signatures font stack

manelli:
Add cursive paragraph style with cross-platform font stack because in certain operating systems/browsers where the font was not available signatures were rendered in Comic Sans.

See: https://eliterate.atlassian.net/browse/MER-4483

### DIFF
--- a/lib/oli/delivery/certificates/certificate_renderer.ex
+++ b/lib/oli/delivery/certificates/certificate_renderer.ex
@@ -44,6 +44,7 @@ defmodule Oli.Delivery.Certificates.CertificateRenderer do
     .signature {
       margin: 0;
       font-family: 'Alex Brush', 'Segoe Script', 'Brush Script MT', 'Lucida Handwriting', cursive;
+      font-weight: bold;
       font-size: 20px;
     }
   </style>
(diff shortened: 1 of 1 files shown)


lib/oli/delivery/certificates/certificate_renderer.ex
```diff
@@ -56,7 +62,7 @@ defmodule Oli.Delivery.Certificates.CertificateRenderer do
             <div style="display: flex; justify-content: space-around;">
               <%= for {admin_name, admin_description} <- @administrators do %>
                 <div>
-                  <p style="margin: 0; font-weight: bold; font-family: 'Alex Brush', cursive; font-size: 20px;">
+                  <p class="signature">
```
nicocirio, line 65: We missed the `font-weight: bold` in the signature class definition.

nicocirio (changes requested):

lib/oli/delivery/certificates/certificate_renderer.ex
manelli: Nice catch! thanks 👍 
Fixed in 73f674186a87f111ae7dc822dd9e77ebbecb0b63

nicocirio (approved):
🚀

PR #5546, opened 2025-04-11
[BUG FIX] [MER-4446] Re revert assessments are being marked late when submitting after suggested date

nicocirio:
According to the [Schedule and Assessment Settings Logic](https://docs.google.com/drawings/d/1LDzW7DH6fX7BW6oFBYgPR9H2ZkqHn-YP5JuvcfSq1xw/edit) doc, we do not want to mark as late a submission of a page that has a suggested by date

image


manelli (approved):
LGTM 🚀

PR #5555, opened 2025-04-17
[ENHANCEMENT] [MER-4449] LTI external tools groundwork

eliknebel:
https://eliterate.atlassian.net/browse/MER-4449

This PR contains some of the LTI external tools groundwork refined from prototyping. With these changes and using https://github.com/Cvmcosta/ltijs-demo-server, one can create a new platform instance (and, in turn, a new LTI activity registration) and can launch the LTI tool as an admin, author and user (instructor/student).

Platform instances can be created and browsed with the URL http://localhost/admin/platform_instances


assets/src/components/activities/creation.ts
```diff
@@ -20,7 +20,7 @@ export function registerCreationFunc(manifest: Manifest, fn: creationFn) {
     window.oliCreationFuncs = {};
   }
 
-  window.oliCreationFuncs[manifest.id] = fn;
+  window.oliCreationFuncs[manifest.authoring.element] = fn;
```
eliknebel, line 23: Change the creation function key from the id to the authoring element to support multiple different LTI activities which will all have different IDs but all have the same authoring element. After all, this authoring element is actually what this creation functions map is used for.

assets/src/data/content/editors.ts
```diff
@@ -1,9 +1,7 @@
-import React from 'react';
-
 export type EditorDesc = {
   slug: string;
-  deliveryElement: string | React.FunctionComponent;
-  authoringElement: string | React.FunctionComponent;
+  deliveryElement: string;
```
eliknebel, line 3: I couldn't find any places where this was actually using this as a react component, so this change seemed to make sense and simplified things a bit

config/dev.exs
```diff
@@ -66,50 +66,31 @@ config :oli, :cashnet_provider,
...
-    protocol_options: [
-      max_header_name_length: http_max_header_name_length,
-      max_header_value_length: http_max_header_value_length,
-      max_headers: http_max_headers
-    ]
+    port: String.to_integer(System.get_env("HTTP_PORT", "80"))
   ],
   url: [
-    scheme: System.get_env("SCHEME", "https"),
+    scheme: System.get_env("SCHEME", "http"),
```
eliknebel, line 80: These are som config changes/cleanup to help facilitate development mode using HTTP, which is required for the demo tool since it fails the handshake with an invalid SSL cert

lib/oli/activities/activity_registration.ex
```diff
@@ -18,6 +18,11 @@ defmodule Oli.Activities.ActivityRegistration do
     field :variables, {:array, :string}, default: []
     field :generates_report, :boolean, default: false
 
+    # Optionally, this activity registration can be associated with an LTI deployment.
+    # If an LTI deployment is associated, the activity is considered an LTI activity.
+    has_one :lti_external_tool_activity_deployment,
```
eliknebel, line 23: Instead of adding an explicit field to identify the activity registration "subtype", an activity registration is simply identified as an LTI activity type by the presence of an `lti_external_tool_activity_deployment`

config/config.exs
```diff
@@ -186,15 +186,15 @@ config :oli, :upgrade_experiment_provider,
...
   url: [host: "localhost"],
   secret_key_base: "GE9cpXBwVXNaplyUCYbIWqERmC/OlcR5iVMwLX9/W7gzQRxkD1ETjda9E0jW/BW1",
   render_errors: [
     accepts: ~w(html json),
     root_layout: {OliWeb.LayoutView, :error},
     view: OliWeb.ErrorView
   ],
-  pubsub_server: Oli.PubSub
+  pubsub_server: Oli.PubSub,
+  live_view: [signing_salt: "zsLMAebo"]
```
gastonabella, line 197: Shouldn't we keep this field as a secret in an env var?

gastonabella (approved):
Looks great! I just left a small suggestion to improve readability. Nothing blocking the PR from being merged.

mix.exs
```diff
@@ -167,8 +167,11 @@ defmodule Oli.MixProject do
       {:jason, "~> 1.3"},
       {:joken, "~> 2.2.0"},
       {:jose, "~> 1.10"},
-      {:lti_1p3, "~> 0.6.0"},
-      {:lti_1p3_ecto_provider, "~> 0.6.0"},
+      {:lti_1p3,
```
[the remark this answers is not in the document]
eliknebel: Good idea https://eliterate.atlassian.net/browse/MER-4550

PR #5559, opened 2025-04-21
[FEATURE] [MER-4472] Incorrect Total Points sent for Basic Graded Assessments to LMS

rgachuhi:
This PR addresses an issue where, when publishing basic assessments from Torus and using them in an LMS, the total points value sent is incorrect. Instead of sending the sum of the point values assigned to each question, Torus sends a total equal to the number of questions in the assessment.

The PR fixes this issue by updating the `Oli.Grading` module's `determine_page_out_of` function for basic assessments to walk through each activity on a graded page and for each activity, determine the maximum score value from the activity parts and responses.

### DIFF
--- a/lib/oli/authoring/editing/page_editor.ex
+++ b/lib/oli/authoring/editing/page_editor.ex
@@ -95,11 +95,77 @@ defmodule Oli.Authoring.Editing.PageEditor do
           Broadcaster.broadcast_revision(r, project_slug)
         end)
 
+        act_refs = determine_page_out_of(project_slug, revision)
+
+        IO.inspect(
+          act_refs,
+          label: "activity references"
+        )
+
+        # IO.inspect(determine_page_out_of(project_slug, revision), label: "page out of")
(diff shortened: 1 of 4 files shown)


darrensiegel (changes requested):
The proposed changes would *serially* resolve each activity revision, as it is encountered during the flat filter.   We have to have an upper bound on the number of DB queries that this runs - so this must instead collect the activity ids during the flat filter, and then afterwards issue a single DeliveryResolver.from_resource_id call (giving it an array of resource ids)

rgachuhi:
> The proposed changes would _serially_ resolve each activity revision, as it is encountered during the flat filter. We have to have an upper bound on the number of DB queries that this runs - so this must instead collect the activity ids during the flat filter, and then afterwards issue a single DeliveryResolver.from_resource_id call (giving it an array of resource ids)

Thanks. Change applied

darrensiegel (approved):

PR #5560, opened 2025-04-21
[BUG FIX] [MER-4490] Fix date comparison and submit page term - part 2

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-4490?focusedCommentId=27024) to the comment in the ticket

When I opened the previous PR, I understood that none of the logic of the [Schedule and Assessment Settings](https://docs.google.com/drawings/d/1LDzW7DH6fX7BW6oFBYgPR9H2ZkqHn-YP5JuvcfSq1xw/edit) document had to be applied when the page had a scheduling type of `:read_by`. The note that made me confuse was: 
image

Amanda [explained](https://eliterate.atlassian.net/browse/MER-4490?focusedCommentId=27030) to me that `“Suggested by” date has no impact on the logic, meaning one should follow the logical path on the flow chart that exists when there is no due date”`.

That being clarified, then we should mark an attempt as late when the page has read_by + allow_late_submit + time_limit and the student submits after the countdown has finished.

### DIFF
--- a/lib/oli/delivery/settings.ex
+++ b/lib/oli/delivery/settings.ex
@@ -240,7 +240,22 @@ defmodule Oli.Delivery.Settings do
 
   def was_late?(%ResourceAttempt{}, %Combined{late_submit: :disallow}, _now), do: false
 
-  def was_late?(%ResourceAttempt{}, %Combined{scheduling_type: :read_by}, _now), do: false
+  def was_late?(
+        %ResourceAttempt{} = resource_attempt,
+        %Combined{scheduling_type: :read_by, late_submit: :allow, time_limit: time_limit} =
+          effective_settings,
+        now
+      ) do
+    if time_limit in [0, nil, ""] do
+      false
+    else
+      effective_deadline =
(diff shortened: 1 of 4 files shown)


lib/oli/delivery/settings.ex
```diff
@@ -240,7 +240,22 @@ defmodule Oli.Delivery.Settings do
...
+    if time_limit in [0, nil, ""] do
+      false
+    else
+      effective_deadline =
+        DateTime.add(resource_attempt.inserted_at, time_limit, :minute)
+        |> DateTime.add(effective_settings.grace_period, :minute)
+
+      DateTime.compare(now, effective_deadline) == :gt
+    end
+  end
```
Francisco-Castro, line 258: @nicocirio I haven't tried it yet, but maybe if we get rid of this function and add a catch-all one at the end that returns false, it would do the same.

Francisco-Castro: Thanks @nicocirio for the explanation.

Francisco-Castro (approved):
LGTM

PR #5562, opened 2025-04-22
[BUG FIX] [MER-4507] Branding isnt working on course pages as expected

simonchoxx:
[MER-4507](https://eliterate.atlassian.net/browse/MER-4507)

This PR updates the `required_keys_per_assign` configuration to include the `:open_and_free` key in the section assign. This makes sure that the open_and_free value stays in the LiveView socket after calling slim_assigns/1, which is used to reduce memory usage.

Before this change, open_and_free was not included in the slimmed version of section. Because of that, some logic depending on this flag—like the branding logic that decides which logo to show—did not work correctly.

As a result, even when a section had a custom brand with its own logo, the system was always showing the default logo, since section.open_and_free was missing and treated as false.

With this fix, the open_and_free value is kept, and the branding logic can now use the correct value from the section data.

[MER-4507]: https://eliterate.atlassian.net/browse/MER-4507?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli_web/live/delivery/student/assignments_live.ex
+++ b/lib/oli_web/live/delivery/student/assignments_live.ex
@@ -22,7 +22,8 @@ defmodule OliWeb.Delivery.Student.AssignmentsLive do
          :lti_1p3_deployment,
          :contains_discussions,
          :contains_explorations,
-         :contains_deliberate_practice
+         :contains_deliberate_practice,
+         :open_and_free
        ], %Section{}},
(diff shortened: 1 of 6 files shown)


nicocirio (approved):
🚀

PR #5563, opened 2025-04-23
[BUG FIX] [MER-4417] Switching from Flowchart view to Screen Panel view causes the canvas to be off centered

dtiwarATS:

### DIFF
--- a/assets/styles/authoring/flowchart.scss
+++ b/assets/styles/authoring/flowchart.scss
@@ -33,8 +33,6 @@ $hover: #dce7f9;
   display: flex;
   flex-direction: row;
 
-
-
   &.storybook {
     position: relative !important;
     bottom: unset;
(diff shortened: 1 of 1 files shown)


assets/styles/authoring/flowchart.scss
```diff
@@ -675,16 +673,16 @@ $hover: #dce7f9;
...
+    left: 180px;
     right: 270px;
     bottom: 0;
     overflow: scroll;
     background-color: #f3f5f8;
 
     .aa-stage-pan {
-      width: 2000px;
-      height: 1000px;
+      width: auto;
```
dtiwarATS, line 684: Setting the height and width to auto because it was causing a vertical and horizontal scroll bar even though there were very less parts on the screen

bsparks (approved):

PR #5565, opened 2025-04-23
[BUG FIX] [MER-4536] make instructor sign-in page use same page title as student and author sign-in

andersweinstein:
Instructor sign-in page was defaulting to hard-coded page title of "Sign in", which was inconsistent with other sign-in pages which have page title taken from brand name, coming out as "OLI Torus" under default branding. This inconsistency was interfering with automating QA testing. This PR changes to use the ~~name from branding~~ short name from vendor properties consistently for page title in all cases.

Note "page title" here means the title shown in the browser tab, obtained from meta information in the HTML head element. It is not part of the page body content.

### DIFF
--- a/lib/oli_web/live/user_login_live.ex
+++ b/lib/oli_web/live/user_login_live.ex
@@ -85,7 +85,7 @@ defmodule OliWeb.UserLoginLive do
 
   @impl Phoenix.LiveView
   def mount(_, session, socket) do
-    title = session["title"] || brand_name()
+    title = session["title"] || product_short_name()
 
     email = Phoenix.Flash.get(socket.assigns.flash, :email)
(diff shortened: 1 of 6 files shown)


darrensiegel (changes requested):
I don't think we want the brand name here, I think we want the short name of the system from Vendor Properties

andersweinstein:
If you search for for `brand_name()` in the codebase you will find several other places where it is set as the page title. Are these incorrect also? 

The QA automation requirement is that the page title be consistent across sign-in pages. Is `product_short_name` currently used in the student and author sign-in pages? I am not immediately finding how these pages set their page title.

darrensiegel (approved):

PR #5569, opened 2025-04-24
[FEATURE] [MER-4261] student home page when no schedule set

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-4261) to the ticket

The ticket requested to adjust the home layout based on the status of the agenda (on/off), and how the agenda component is shown based on the course schedule status (no schedule or schedule defined). 

Originally, the course schedule status was checked with the `has_scheduled_resources?` variable, boosted by a query on the mount function in the Student's index liveview.
Since the `has_scheduled_resources?` value will only change when the instructor changes the course schedule (not too often), it made sense for me to replace the Ecto.Repo query implementation with a SectionResourceDepot implementation.
To do so, in the SectionResourceDepot, I needed to check for any scheduled resources in the section ETS table. That is why I created the `Depot.exists?` function (works as `Ecto.Repo.exists?` but for ETS). Probably this same function may be reused in other X-Depot functions in the future.

Finally, I moved the `had_scheduled_resources?` evaluation/assign from the student's home LiveView into a live_session plug, since that value will be needed by other tickets of the `Schedule` epic.

At this point of the task, it seemed quite straightforward to implement the hide/show of the Schedule sidebar link based on the course schedule state, which is why this PR also addresses [MER-4264](https://eliterate.atlassian.net/browse/MER-4264)

**##### When this PR gets approved, both tickets should be moved to QA ####**

#### Adjust the layout based on the agenda status (on/off)

https://github.com/user-attachments/assets/990499b4-10ee-4e65-b15e-12b5e3a473ed

#### Change the agenda component and show/hide the Schedule link based on the Schedule Status

https://github.com/user-attachments/assets/82dcdb1a-9a75-4358-8d9f-e31d57441b1c

#### When there is no schedule, show the first 6 groups until they are completed
When they are all completed, the next 6 (or all remaining if they are less than 6) will be shown.

https://github.com/user-attachments/assets/77007d39-f891-4cb2-9bdb-bf15a622a9ff

[MER-4264]: https://eliterate.atlassian.net/browse/MER-4264?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/oli/delivery/depot.ex
```diff
@@ -146,4 +146,55 @@ defmodule Oli.Delivery.Depot do
...
+    Example:
+      exists?(desc, table_id, [
+        [start_date: {:!=, nil}, hidden: false],
+        [end_date:   {:!=, nil}]
+      ])
+
+    Equivalent to the expression:
+      (start_date != nil and hidden == false) or (end_date != nil)
+  """
+  def exists?(%DepotDesc{} = depot_desc, table_id, filters) when is_list(filters) do
```
gastonabella, line 170: Something to keep in mind is that this function is not flexible enough to express a logic operation such as:

```
(a or b) and (c or d)
```

where the main logic operator is an `and`, and the inner expressions are composed by `or`s. It's probably enough for our current purposes, but we might need to extend it to achieve something like the example above.

lib/oli/delivery/depot.ex
```diff
@@ -146,4 +146,55 @@ defmodule Oli.Delivery.Depot do
...
+    groups =
+      if Keyword.keyword?(filters) do
```
gastonabella, line 172: Nit: we could avoid one level of indentation by using a `cond` here :)

lib/oli/delivery/depot.ex
```diff
@@ -146,4 +146,55 @@ defmodule Oli.Delivery.Depot do
...
+        else
+          raise ArgumentError,
+                "exists?/3 expects a keyword list or a list of keyword lists, got: #{inspect(filters)}"
+        end
+      end
+
+    # Translate each group into an ETS match spec
+    specs =
+      Enum.map(groups, fn conds ->
+        conditions = Enum.map(conds, fn {field, value} -> {field, value} end)
```
gastonabella, line 188: Is this line making any difference?

lib/oli_web/components/layouts/page.html.heex
```diff
@@ -7,15 +7,6 @@
   />
 
   <main role="main" class="flex-1 flex flex-col relative md:flex-row">
-    <Components.Delivery.Layouts.sidebar_nav
```
gastonabella, line 10: Why are we removing the sidebar nav from this layout?

gastonabella (commented):
Looks really good!
I love seeing the `Depot` being reused and extended 🚀

lib/oli_web/components/layouts/page.html.heex
nicocirio: I'm 99% sure this page template is the old one we used before migrating to our liveview NG approach. But to honor that 1% of the possibility, I will put it back and assign a default value to the required has_scheduled_resources? assign in case it is not in the socket. And I will add this research to our "Friday's tech debt", as we may end up removing this entire template and it's corresponding "old" controller.

lib/oli/delivery/depot.ex
nicocirio: It seems to be doing nothing. To be honest, I made some iterations with Cursor to build this `exists?` function, and the Enum made sense in one of those iterations. Thanks for the comment!

lib/oli/delivery/depot.ex
nicocirio: I agree, nice catch! I will try to see if extending it is easy to implement in this PR; if not, I will prefer to leave that optimization to a future iteration

gastonabella (approved):
LGTM!

PR #5570, opened 2025-04-24
[CHORE] [MER-0000] Adjust pr title regex

darrensiegel:


andersweinstein:
Note you can match between 1 and 4 digits in a regex concisely with `\d{1,4}`

andersweinstein (approved):

PR #5575, opened 2025-04-25
[BUG FIX] [MER-4547] More robust trigger type determination for ResponseMult

darrensiegel:
The ResponseMult activity type can break trigger type detection code - as when parts are linked there are then less parts than there are inputs. 

The change here is to catch the exception that `findTargetedResponses` throws in this case, and omit this activity instance from including targeted triggers.


andersweinstein:
The bug arises because the default incorrect (catchall) response for responseMulti's has rule of the form
`input_ref_136853124 like {.*} && (input_ref_1282200811 like {.*})`
but the  generic getIncorrectResponse routine in responses.ts is not prepared for this structure. It looks only for a catchall of form `... like {.*}` assuming one argument only and so fails to find responseMulti catchAlls. 

The generic getIncorrectResponse is NOT normally used within ResponseMulti authoring. That uses a more general ruleIsCatchAll routine that also works on conjunctive rules. 

But the AI trigger authoring uses the generic routine while getting targeted feedback, going through a routine that expects to partition all responses into [correct, incorrect, targeted]. 

A more durable fix would be to correct getIncorrectResponse to work on the sort of rules that are used by responseMulti's.

andersweinstein (approved):

PR #5580, opened 2025-04-28
[BUG FIX] [MER-4510] Fix Bypass Paywall logic

gastonabella:
[MER-4510](https://eliterate.atlassian.net/browse/MER-4510)

This PR fixes the case when an admin or instructor wants to give students a 100% discount (meaning $0 cost) for a given product or institution. A new field is added to the `discounts` table and the "Create Discounts" form.

**Before:** 

https://github.com/user-attachments/assets/b3b55908-a838-4024-979f-f885d57a4d8c

**After:**

https://github.com/user-attachments/assets/e7045505-7370-4cfe-be6b-b17a25737db7

NOTE:
I added to our tech debt list that we need to add tests for the Paywall and Discounts integrated logic (there are no test files for this flow) and for the section creation logic (create from product and project). Since this is a bug that needs to be addressed ASAP, I didn't add those test cases in this PR.

[MER-4510]: https://eliterate.atlassian.net/browse/MER-4510?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli_web/live/products/payments/discounts/form.ex
+++ b/lib/oli_web/live/products/payments/discounts/form.ex
@@ -30,36 +30,46 @@ defmodule OliWeb.Products.Payments.Discounts.Form do
         </div>
       <% end %>
 
+      <% bypass_paywall? = @form[:bypass_paywall].value in [true, "true"] %>
+
       <div class="form-group">
         <.input
           type="select"
           field={@form[:type]}
           label="Type"
-          class="form-control"
+          class={"form-control #{if(bypass_paywall?, do: "cursor-not-allowed", else: "")}"}
           options={[Percentage: "percentage", "Fixed price": "fixed_amount"]}
         />
       </div>
 
       <div class="form-group">
+        <% price_disabled? =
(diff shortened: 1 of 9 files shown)


lib/oli_web/live/products/payments/discounts/form.ex
```diff
@@ -30,38 +30,48 @@ defmodule OliWeb.Products.Payments.Discounts.Form do
...
       <div class="form-group">
         <.input
           type="select"
           field={@form[:type]}
           label="Type"
-          class="form-control"
+          class={"form-control #{if(bypass_paywall?, do: "cursor-not-allowed", else: "")}"}
           options={[Percentage: "percentage", "Fixed price": "fixed_amount"]}
         />
       </div>
```
nicocirio, line 43: Should we disable this input when `bypass_paywall` is true?

PR #5583, opened 2025-04-30
[ENHANCEMENT] [MER-4506] Updates aggregates mode calculation to include all students

rgachuhi:
This PR updates the insights in the content and objectives table to account for users who have not yet interacted with the content. It also ensures that, where there is more than one mode, the displayed value should be the highest value according to this hierarchy:

1. Low
2. Medium
3. High
4. Not enough information

### DIFF
--- a/lib/oli/delivery/metrics.ex
+++ b/lib/oli/delivery/metrics.ex
@@ -10,6 +10,8 @@ defmodule Oli.Delivery.Metrics do
   alias Oli.Delivery.Attempts.Core
   alias Oli.Resources.Revision
 
+  alias Oli.Delivery.Sections
+
   alias Oli.Delivery.Sections.{
     ContainedPage,
     Enrollment,
@@ -862,7 +864,7 @@ defmodule Oli.Delivery.Metrics do
     }
   """
   def proficiency_per_container(
-        %Section{id: section_id, analytics_version: _both},
+        %Section{id: section_id, slug: slug, analytics_version: _both},
         contained_pages
       ) do
     page_type_id = Oli.Resources.ResourceType.id_for_page()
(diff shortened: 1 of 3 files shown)


lib/oli/delivery/metrics.ex
```diff
@@ -1324,7 +1326,12 @@ defmodule Oli.Delivery.Metrics do
   # the level of proficiency for that container.
   # The proficiency is calculated by finding the mode where the highest number of
   # users fall into a proficiency range for that container.
-  defp bucket_into_container_mode(page_data, contained_pages) do
+  defp bucket_into_container_mode(page_data, contained_pages, section_slug) do
+    student_ids =
```
darrensiegel, line 1330: If all that this implementation needs is the set of user ids of enrolled students, it should execute a query *exactly* for that - as opposed to fetching all of the user records and filtering and mapping.   User records can be arbitrarily large, thanks to the `state` field, which is used by extrinsic storage.

lib/oli/delivery/sections.ex
```diff
@@ -4875,16 +4875,40 @@ defmodule Oli.Delivery.Sections do
           )
       end
 
+    student_ids =
+      Sections.enrolled_students(section_slug)
```
darrensiegel, line 4879: Same comment as before

darrensiegel (changes requested):

lib/oli/delivery/metrics.ex
rgachuhi: Good point. Change applied

lib/oli/delivery/sections.ex
rgachuhi: Change applied

darrensiegel (approved):

PR #5585, opened 2025-04-30
[FEATURE] [MER-4293] LTI tools table information

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-4293) to the ticket

https://github.com/user-attachments/assets/792ce5e1-2f93-4eb6-a33a-f81874d7e5e6

Note that the `User Count` and the `Actions` column have placeholders that will be addressed in future tickets.
The 'User Count` has random numbers as placeholders.


simonchoxx (commented):
Great job Nico! 👏

Leave some comments to understand whether or not we should make some modifications to the code!

lib/oli_web/live/admin/external_tools/external_tools_view.ex
```diff
@@ -20,14 +34,110 @@ defmodule OliWeb.Admin.ExternalToolsView do
...
+      )
+
+    total_count = SortableTableModel.determine_total(tools)
+
+    {:ok, table_model} = TableModel.new(tools, socket.assigns.ctx)
+
     {:ok,
      assign(socket,
-       breadcrumbs: set_breadcrumbs()
+       title: "Manage LTI 1.3 External Tools",
```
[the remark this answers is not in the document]
nicocirio: This assign is used in the layout, and determines the title of the browser tab:

image

lib/oli_web/live/admin/external_tools/table_model.ex
```diff
@@ -0,0 +1,96 @@
...
+        }
+      ],
+      event_suffix: "",
+      id_field: [:id],
+      data: %{ctx: ctx}
+    )
+  end
+
+  def sort_tool_name_column(sort_order, _sort_spec),
+    do: {fn r -> String.downcase(r.name) end, sort_order}
```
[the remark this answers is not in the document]
nicocirio: Nice catch! I used it in a previous iteration, but then I ended up sorting in the `browse_platform_instances/3` function

lib/oli/lti/platform_instances.ex
```diff
@@ -3,6 +3,78 @@ defmodule Oli.Lti.PlatformInstances do
...
+        where: ^filter_by_text,
+        where: ^filter_by_status,
+        limit: ^limit,
+        offset: ^offset,
+        select: %{
+          id: p.id,
+          name: p.name,
+          description: p.description,
+          inserted_at: p.inserted_at,
+          usage_count: fragment("floor(random() * 101)::integer AS usage_count"),
```
gastonabella, line 59: I know this won't be used until we release the whole feature, but isn't it less confusing to not define this field for now (leave it `nil`) and add the column when the value has a meaning? I don't see a point in showing a random number.

lib/oli/lti/platform_instances.ex
```diff
@@ -3,6 +3,78 @@ defmodule Oli.Lti.PlatformInstances do
 
   alias Oli.Repo
   alias Lti_1p3.DataProviders.EctoProvider.PlatformInstance
+  alias Oli.Lti.PlatformExternalTools.LtiExternalToolActivityDeployment
+  alias Oli.Lti.PlatformInstances.BrowseOptions
+  alias Oli.Repo.{Paging, Sorting}
+
+  @doc """
```
gastonabella, line 10: We’re using two different contexts to manage similar (if not the same) parts of the app. This current one is older, and a recent PR introduced a new context, `Oli.Lti.PlatformExternalTools`, to handle external tools information. I think we should either consolidate and stick to one of them, or keep both but clearly define their boundaries. That way, if we need to add new functions related to LTI tools in the future, it’ll be clear where they belong.

simonchoxx (approved):
LGTM 🚀

lib/oli/lti/platform_instances.ex
nicocirio: I assigned a random number just to check that the column's sorting was working as expected. But I agree I can leave a nil value.

lib/oli/lti/platform_instances.ex
nicocirio: Oh, I believe that new context was not in the master branch by the time I worked on the Browse function. I think I should move that function to this new `Oli.Lti.PlatformExternalTools`, wdyt?

lib/oli_web/live/admin/external_tools/external_tools_view.ex
```diff
@@ -1,9 +1,23 @@
 defmodule OliWeb.Admin.ExternalToolsView do
```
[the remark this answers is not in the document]
nicocirio: I agree! I will adjust that and the corresponding test file

gastonabella (approved):
LGTM!

PR #5586, opened 2025-05-01
[FEATURE] [MER-4348] Score as you go

darrensiegel:
This PR introduces the "Score as you go" scoring policy.  For a graded page operating in this mode, as a student answers questions their overall page score is updated, even to the point of syncing to the LMS. 

Demo videos: 

“Score as you go” feature demo 1: Authoring, student, instructor flows
https://drive.google.com/file/d/1is9Wj80NWZTMvpdlqqUyqW4A0aijdRi7/view?usp=drive_link

“Score as you go” feature demo 2: Handling publication updates
https://drive.google.com/file/d/15VaItxf0nfT4drVj5WEoZRkQO66v0y9s/view?usp=drive_link

The following are the major components of the technical design and implementation:

## Authoring control over score as you go

Two new settings are now available to the author, the scoring policy and replacement policy.  The scoring policy is "Score at the end" or the new "Score as you go".  Replacement policy governs how the system handles subsequent attempts of activities on a page (both graded, ungraded): does the system generate a new question or keep the original from the previous attempt. 

## `Oli.Delivery.Attempts.ActivityLifecycle.RollUp` refactor

Multiple places in the system were finalizing activity evaluations using a few different implementations.  Each of these endpoints had to "roll up" part to activity to page attempt results.  To ensure consistency of implementation, this PR refactors these locations (legacy super activity, manual grading, client side eval, server side eval) down to `Evaluate.evaluate_from_input` and `Evaluate.apply_client_evaluations`.  Both of these impls call a now singularized roll up implementation from module `Oli.Delivery.Attempts.ActivityLifecycle.RollUp`.  

## Aggregate "tracking" score and out of

Two new fields are added to `ActivityAttempt`:  `:aggregate_score` and `:aggregate_out_of`.  These fields track the current "aggregate" score for a series of attempts for a single activity in a score as you go page.  As an example, think about a page with two activities where the activities are allowed to be answered 10 times and the scoring strategy is "best".  After the student answers the first question and receives 1 / 2 points, they then reset to try again, but this time they receive 0 / 1.   The system needs a way to store that, overall, for this activity their score is 1 /2 (the best one), but the most recent attempt was 0 / 1.   This "best" score is captured in these these fields.  

## Activity bank "points per activity"

Score as you go impl requires that upfront, when a page attempt is started that we know the total "out of" points (so that we can render the overall score to the student).  Activity bank selections complicate this, because total points can differ across student attempts due to random fulfillment.  Consider just the case where student A gets a multi input question with 4 parts and student B gets an MCQ with 1 part.  The former is worth 4 points while the latter is worth 1.  The new "points per activity" attribute of selections now normalizes each of the fulfilled activities from a bank to be the specified point value.   If a bank selection is set to 2 points per activity and it selects an MCQ with custom scoring of 10 for a correct answer, 5 for partial and 0 for incorrect - that 10 will be scaled down to 2, the 5 to 1, etc.  

With "points per activity" in place, now we can deterministically calculate the "out of" for a page give the max "out of" for each embedded activity and the "points per activity" * "count" for each bank selected activity. 

## One at a time mode changes

The One at a time question mode had to be reworked significantly to allow client driven submission and reset.  A new notification system is present which powers question submission notification for the server. 

## Publication application

Some adjustments are made to what happens when a student visits a graded page with an active attempt.  Previously, the system did nothing:  students had to explicitly finish that active attempt and start a new one to pick up the new page revision.  Score as you go, however, introduces implicit page attempt management for the student - so we cannot continue with this "do nothing" approach.  Instead (and only for Score as you go mode) these active page attempts will automatically move forward, by way of creating a new page attempt, to pick up the latest page revision.  This involves pulling forward all current state and scores from exist activity attempts.


lib/oli/delivery/attempts/activity_lifecycle/roll_up.ex
```diff
@@ -0,0 +1,542 @@
...
+    end
+  end
+
+  @doc """
+  From the update to a collection of part attempts, return a function that will properly roll
+  up the score to the activity attempt, and possibly the resource attempt and resource access.
+
+  The returned function takes zero arguments, and returns either :ok or :error.
+
+  This function makes at most two read queries to the database.
```
eliknebel, line 108: Nit: Function docs say this function makes at most 2 queries, but it looks if it is score as you go it will
make 3

lib/oli/delivery/attempts/page_lifecycle/graded.ex
```diff
@@ -59,17 +62,46 @@ defmodule Oli.Delivery.Attempts.PageLifecycle.Graded do
           resource_attempts: graded_attempts
         }}}
     else
-      # Unlike ungraded pages, for graded pages we do not throw away attempts and create anew in the case
-      # where the resource revision has changed.  Instead we return back the existing attempt tree and force
-      # the page renderer to resolve this discrepancy by indicating the "revised" state.
+      # Se if we are in score as you go mode, and the page revision has changed since the start of this
```
eliknebel, line 65: Nit typo: See

lib/oli_web/templates/layout/delivery.html.heex
```diff
@@ -160,6 +160,34 @@
 
     <%= Map.get(assigns, :inner_layout) || @inner_content %>
 
+    <div
```
eliknebel, line 163: I think we already have an existing modal implementation using
`window.oliDispatch(modalActions.display(...`. Granted this one is much simpler, but maybe we should
use the existing one or replace that one to standardize on a single approach so we dont have two
different approaches to displaying modals?

eliknebel (changes requested):
Everything here looks great. I just had a few small comments and found a leftover IO.inspect

eliknebel (approved):

andersweinstein:
Including the ScoreAsYouGoHeader has the effect of showing "Question #null" over every question on a practice page. Logged as https://eliterate.atlassian.net/browse/TRIAGE-1585  I believe the ScoreAsYouGoHeader should not render anything when on a practice page. Or it could render an empty div, though I am not sure why this would be needed.

PR #5587, opened 2025-05-02
[FEATURE] [MER-4292] Admin LTI tools add tool

simonchoxx:
[MER-4292](https://eliterate.atlassian.net/browse/MER-4292)

This PR introduces a new view that allows admin users to create LTI 1.3 external tools at the system level.

https://github.com/user-attachments/assets/cfc7e5f0-11eb-42e8-83d6-8ef4ef68acc2

[MER-4292]: https://eliterate.atlassian.net/browse/MER-4292?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/oli_web/live/admin/external_tools/new_external_tool_view.ex
```diff
@@ -0,0 +1,256 @@
...
+     )}
+  end
+
+  def render(assigns) do
+    ~H"""
+    <div class="flex flex-col justify-end mx-12 mt-4">
+      <%= render_custom_flash(@custom_flash) %>
+      <div class="w-[1247px] inline-flex flex-col justify-start items-start gap-3">
+        <div class="self-stretch flex flex-col justify-start items-start">
+          <div class="justify-center text-color-blue-24 text-2xl font-normal font-['Open_Sans'] leading-9">
```
gastonabella, line 37: Same here about the font

lib/oli_web/live/admin/external_tools/new_external_tool_view.ex
```diff
@@ -0,0 +1,256 @@
...
+        {:error, %Ecto.Changeset{} = changeset} ->
+          {:noreply,
+           socket
+           |> assign(form: to_form(changeset, as: :tool_form))
+           |> assign(:custom_flash, %{
+             type: :error,
+             message: "One or more of the required fields is missing; please check your input."
+           })}
+      end
+    rescue
```
gastonabella, line 192: Couldn't we leverage changeset validations and avoid using a `try rescue` block here? This works, but does not seem to be the standard way of handling database errors such as unique constraints. Since this is a schema from a library, one way could be to extend or redefine the existing `PlatformInstance` changeset to have a `unique_constraint` over that field. Another option could be to define the `register_lti_external_tool_activity` function as an `Ecto.Multi` instead of a `Repo.transaction` and validate there the uniqueness in the pipeline.

gastonabella (approved):
Looks great!

PR #5588, opened 2025-05-05
[FEATURE] [MER-4564] Admins can access all instructor and student routes

darrensiegel:
This PR allows admin users - when logged in only as an admin user - to be able to access any and all delivery focused routes. 

This is done via a new "hidden" instructor account that will get created and enrolled into a course section, just in time, when an admin user visits a delivery route for a section.

Given that this "auto enroll" feature ONLY kicks in when the Admin is NOT signed in with a delivery account, a new UI is added to the "Instructor Workspace" for Admins.  It shows the details of the current delivery account and a link to log out.  This log out is necessary to facilitate a workflow where the Admin wants to "bounce" between many different course sections (log out, visit section and get auto enrolled, log out, etc)

Screenshot 2025-05-05 at 11 00 04 AM

### DIFF
--- a/lib/oli_web/user_auth.ex
+++ b/lib/oli_web/user_auth.ex
@@ -321,11 +321,14 @@ defmodule OliWeb.UserAuth do
   end
 
   def auto_enroll_admin(conn, _opts) do
+
+    # If we only have an admin user logged in, and the section is not nil, we will
+    # auto-enroll a new (hidden) instructor account into this section.
     if conn.assigns[:current_user] == nil and conn.assigns[:is_admin] and conn.assigns[:section] != nil do
 
       case Oli.Delivery.Sections.fetch_hidden_instructor(conn.assigns[:section].id) do
 
-        {:ok, user, token} ->
+        {:ok, {user, token}} ->
 
           # reset the user and the SessionContext, push the token into the session
           conn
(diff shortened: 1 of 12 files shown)


lib/oli_web/user_auth.ex
```diff
@@ -320,6 +320,27 @@ defmodule OliWeb.UserAuth do
     end
   end
 
+  def auto_enroll_admin(conn, _opts) do
+    # If we only have an admin user logged in, and the section is not nil, we will
+    # auto-enroll a new (hidden) instructor account into this section.
+    if conn.assigns[:current_user] == nil and conn.assigns[:is_admin] and
```
nicocirio, line 326: If I, as an admin, access one course page and these conditions are met, a new "Admin User" will be created on the fly, enrolled, and I will be able to visit that page.
But if then I visit a page from another course without logging out that new user, the app will crash since that "Admin User" will still be in the session but not enrolled in that second course.

nicocirio: I also noticed that, after signing out, a new "Admin User" is created when I visit another page. Does it make sense to reuse the same "Admin User" and enroll it "on the fly" to the visited courses to avoid populating the DB with many "repeated" records?
image

eliknebel (approved):

PR #5595, opened 2025-05-07
[FEATURE] [MER-4400] Admin Edit LTI Tool Details

gastonabella:
[MER-4400](https://eliterate.atlassian.net/browse/MER-4400)

Add new Details page for LTI External Tools.

https://github.com/user-attachments/assets/31961d93-6368-48ce-8a2d-fac5e4a30012

[MER-4400]: https://eliterate.atlassian.net/browse/MER-4400?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli_web/live/admin/external_tools/table_model.ex
+++ b/lib/oli_web/live/admin/external_tools/table_model.ex
@@ -81,9 +81,11 @@ defmodule OliWeb.Admin.ExternalTools.TableModel do
   end
 
   def render_actions_column(assigns, row, _) do
+    assigns = Map.merge(assigns, %{platform_instance_id: row.id})
+
     ~H"""
     <.link
-      navigate={~p"/admin/external_tools/#{row.id}/details"}
+      navigate={~p"/admin/external_tools/#{@platform_instance_id}/details"}
       class="w-20 text-center text-sm font-semibold leading-none h-7 rounded-lg shadow-[0px_2px_4px_0px_rgba(0,52,99,0.10)] outline outline-1 outline-offset-[-1px] inline-flex justify-center items-center"
     >
       Details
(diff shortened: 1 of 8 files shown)


lib/oli_web/live/admin/external_tools/table_model.ex
```diff
@@ -79,11 +80,12 @@ defmodule OliWeb.Admin.ExternalTools.TableModel do
...
 
-  def render_actions_column(assigns, _row, _) do
-    # TODO: link this to the details page (https://eliterate.atlassian.net/browse/MER-4400)
+  def render_actions_column(assigns, row, _) do
+    assigns = Map.merge(assigns, %{platform_instance_id: row.id})
+
     ~H"""
     <.link
-      navigate="#"
+      navigate={~p"/admin/external_tools/#{@platform_instance_id}/details"}
```
nicocirio, line 88: I realized that by using `navigate` we mount a new LiveView while keeping the current layout, so the breadcrumbs do not get updated. We should somehow force to update the breadcrumbs on the layout or use `href` instead.

nicocirio (changes requested):
Nice work! I liked how you reused the tool form.
While testing it locally I found that the 'navigate' is not compatible with the way we are handling the breadcrumbs on the layout (they do not get updated). One approach can be simply changing that link to the `href` flavour. But if you agree, it would be nice to explore if we can adapt the way we render our breadcrumbs on `lib/oli_web/templates/layout/_workspace_header.html.heex` to keep the live navigation. WDYT?

lib/oli_web/live/admin/external_tools/table_model.ex
gastonabella: Good catch! I agree we should somehow leverage live navigation here, but I believe this is a general issue in Torus. Since breadcrumbs are rendered in a LiveView inside an app layout, I think it is possible to make them reload accordingly. Something to explore in our tech debt sessions :)
I ended up using `href` here.

nicocirio (approved):
🚀

PR #5600, opened 2025-05-12
[FEATURE] [MER-4332] Admin LTI Tool Usage Count

manelli:


lib/oli/lti/platform_external_tools.ex
```diff
@@ -260,4 +272,96 @@ defmodule Oli.Lti.PlatformExternalTools do
...
+        join: d in assoc(ar, :lti_external_tool_activity_deployment),
+        where: d.platform_instance_id == ^platform_instance_id,
+        select: ar.id
+      )
+      |> Repo.all()
+
+    # Step 2: Get IDs of all section resources that are LTI activities
+    lti_activity_ids =
+      from(sr in SectionResource,
+        join: r in Revision,
```
gastonabella, line 345: Are we using this join here? I think we could only use the `revisions` table here.

lib/oli/lti/platform_external_tools.ex
```diff
@@ -260,4 +272,96 @@ defmodule Oli.Lti.PlatformExternalTools do
...
+      Enum.reduce(referenced_lti_activities, acc, fn lti_id, inner_acc ->
+        existing = Map.get(inner_acc, lti_id, [])
+        Map.put(inner_acc, lti_id, [section_resource | existing])
+      end)
+    end)
+  end
+
+  @doc """
+  Given a section, return a map where the keys are LTI activity resource IDs and the values are lists of section resources that reference those activities.
+  """
```
gastonabella, line 331: I believe this @doc does not describe what this function does.

lib/oli/lti/platform_external_tools.ex
```diff
@@ -178,11 +180,21 @@ defmodule Oli.Lti.PlatformExternalTools do
         dynamic([p, lad], lad.status == :enabled)
       end
 
-    # TODO: calculate usage_count (we need https://eliterate.atlassian.net/browse/MER-4469)
+    usage_counts_query =
```
gastonabella, line 183: I might be missing something, but here we are counting the number of activity registrations associated with a platform instance. Does this number represent the usage count?

gastonabella (commented):
It's looking great! I left some comments/questions

lib/oli/lti/platform_external_tools.ex
eliknebel: In testing this out, I noticed the same issue. The number being displayed does not represent the actual usage counts displayed in the table it links to

manelli: Thanks for pointing that out. Fixed in b4ee9b81540f4e48307fcb0db81b0ce0ee45da1c

lib/oli/lti/platform_external_tools.ex
manelli: Fixed in b4ee9b81540f4e48307fcb0db81b0ce0ee45da1c

lib/oli/lti/platform_external_tools.ex
manelli: Updated @doc and @spec in bbdc967d9f19396e44c57d7afadfbe3dd5f28c1c

nicocirio:
> @manelli I tested it locally and found that the tools aren't counted until a student enters a page that contains those tools. I mean, on the authoring side, I created some tools, added to a page in a project, and published the changes; as an instructor, I applied the changes to my course. No tools `usage count` was detected. As soon as I visited that page as a student, the `usage count` was updated and matched the expected result. Is that how we expect it to work? I expected to see the count updated as soon as I publish and apply the changes to a course, but I might be wrong.

While adding tests in another PR I was bitten by the same issue... the problem root cause is that some fields of the SectionResource aren't populated when we create the section resources for a given section. Those values are populated when [a SectionResourceDepot table is first triggered](https://github.com/Simon-Initiative/oli-torus/pull/5185/files#diff-9cd70d1d862781e5ecd2a468b858c410ad804e68e0bb40ea8d86491a192649d9R124-R130), for example, when the first student visits a section page. 
So, if we build queries that fetch the revision by joining the section resource on the revision_id field, it might be the case that that field has not yet been populated. I believe that was what happened in the case I described above... as an instructor I applied the changes that were just published, and the Usage Count was not updated [as it's query relies on joining the revision on the revision_id field](https://github.com/Simon-Initiative/oli-torus/pull/5600/files#diff-69e1e0ec963c1956627c2f537c67f4e47d2d813551e250a2122b000ef9f8f3f3R309-R316) (not yet populated)

PR #5609, opened 2025-05-15
[FEATURE] [MER-4399] Admin soft delete external tool

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-4399) to the ticket

https://github.com/user-attachments/assets/168cec6d-7a9f-490b-aca6-149fc7f72b22


test/oli/lti/platform_external_tools_test.exs
```diff
@@ -177,6 +177,29 @@ defmodule Oli.Lti.PlatformExternalToolsTest do
...
+        deployment_2,
+        %{"status" => :deleted}
+      )
+
+      results =
+        PlatformExternalTools.browse_platform_external_tools(
+          %Paging{offset: 0, limit: 10},
+          %Sorting{field: :name, direction: :asc},
+          %BrowseOptions{include_disabled: true}
+        )
```
Francisco-Castro, line 197: NIT: I experimented with adding default values to `browse_platform_external_tools/3` to achieve something like:

```elixir
  def browse_platform_external_tools(
        paging \\ %Paging{limit: 10, offset: 0},
        sorting \\ %Sorting{field: :name, direction: :asc},
        options
      )
```

This way, we can simplify testing (and other usages) to just:

```elixir
PlatformExternalTools.browse_platform_external_tools(%BrowseOptions{include_disabled: true})
```

This allows us to focus on the core part: %BrowseOptions{include_disabled: true}.

Francisco-Castro (approved):
It looks good to me. I left a few NITs, but the core functionality is working as expected.

PR #5611, opened 2025-05-16
[FEATURE] [MER-4294] Enable/Disable a tool at system level - rework

simonchoxx:
[MER-4294](https://eliterate.atlassian.net/browse/MER-4294)

This PR excludes LTI tools that were disabled or removed at the system level, showing only LTI tools enabled to be added to a curriculum page. 

https://github.com/user-attachments/assets/cc92e9d9-9b1d-4583-9097-e0e4032e51ea

[MER-4294]: https://eliterate.atlassian.net/browse/MER-4294?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli/activities.ex
+++ b/lib/oli/activities.ex
@@ -175,6 +175,9 @@ defmodule Oli.Activities do
         end)
 
       list_activity_registrations()
+      |> Enum.filter(fn a ->
+        is_nil(a.status) or a.status == :enabled
+      end)
       |> Enum.map(&ActivityMapEntry.from_registration/1)
       |> Enum.reduce(%{}, fn e, m ->
         e =


manelli (approved):
LGTM 🚀

lib/oli/activities.ex
```diff
@@ -175,6 +175,9 @@ defmodule Oli.Activities do
         end)
 
       list_activity_registrations()
+      |> Enum.filter(fn a ->
```
Francisco-Castro, line 178: A late NIT: `|> Enum.reject(&(&1.status == :disabled))`

Francisco-Castro: A future improvement would be to make `list_activity_registrations/0` more flexible. I was looking at the code, and most of the time when we call that function, we end up doing some kind of post-filtering.

PR #5614, opened 2025-05-19
[BUG FIX] [MER-4034] Add logging on failure

Francisco-Castro:
Ticket: [MER-4034](https://eliterate.atlassian.net/browse/MER-4034)

This PR adds logging for failed help requests, which is important for diagnosing the cause of the failures. I set up logging on both the front-end and back-end, but we'll likely remove the front-end one.

[MER-4034]: https://eliterate.atlassian.net/browse/MER-4034?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5617, opened 2025-05-20
[BUG FIX] [MER-4034] improvements to the torus support modal rework 2

Francisco-Castro:
Ticket: [MER-4034](https://eliterate.atlassian.net/browse/MER-4034)

This PR adds the captcha directly to the TechSupport LiveView, as AppSignal logs indicate the error originates there:

```
05-19-2025 10:15:41.950[application][ERROR][tokamak.oli.cmu.edu][phoenix]Error when processing help message "reCaptcha failure"
```

Additionally, this PR includes a fix for the flash message rendering twice. A custom flash implementation was added, inspired by PR: https://github.com/Simon-Initiative/oli-torus/pull/5595.

Finally, this PR updates the `knowledgebase_url` to open in a new tab (see commit: 01b6d68c748a7cf32c737049db283984ca02475a) and removes the `console.log` added in PR: https://github.com/Simon-Initiative/oli-torus/pull/5614, as it was not providing useful information. (see commit: a62bd51dc6cc8f839d4f03648d078efa53131757)

Note: The last commit adds a debounce to the textarea input.

[MER-4034]: https://eliterate.atlassian.net/browse/MER-4034?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli_web/components/tech_support_live.ex
+++ b/lib/oli_web/components/tech_support_live.ex
@@ -142,9 +142,20 @@ defmodule OliWeb.TechSupportLive do
               &times;
             </a>
           </div>
-
           <div class="w-full flex flex-col lg:flex-row gap-2 md:justify-between">
-            <.render_recaptcha recaptcha_error={@recaptcha_error} class="md:m-0 m-auto" />
+            <%!-- Start Captcha --%>
+            <div class="w-80 mx-auto">
+              <div
+                id="recaptcha"
+                phx-hook="Recaptcha"
+                data-sitekey={Application.fetch_env!(:oli, :recaptcha)[:site_key]}
+                data-theme="dark"
+                phx-update="ignore"
+              >
+              </div>
+
+              <.error :if={@recaptcha_error}><%= @recaptcha_error %></.error>
+            </div>
 
             <div class="flex w-full justify-around lg:justify-end items-center">
               <.button type="link" variant={:link} phx-click={Modal.hide_modal(@modal_id)}>
--- a/lib/oli_web/components/tech_support_live.ex
+++ b/lib/oli_web/components/tech_support_live.ex
@@ -3,11 +3,10 @@ defmodule OliWeb.TechSupportLive do
   alias Oli.Help.HelpContent
   alias Oli.Help.HelpRequest
   alias OliWeb.Components.Modal
+  alias OliWeb.Icons
 
   @modal_id "tech-support-modal"
   @base_url Oli.Utils.get_base_url()
-
-  require Logger
 
   @impl true
   def mount(_params, session, socket) do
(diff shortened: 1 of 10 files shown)


eliknebel (changes requested):
I don't think we should be using a new custom flash message. In fact, even though imo it looks better than the existing flash message, I think we should revert the custom flash added in the referenced PR, as it does not match the other flash messages throughout the system and requires additional custom code in each LiveView to support it.

We should be using the existing Flash infrastructure via `socket = put_flash(socket, :info, info)` and if we desire to update the look and feel of that, we can do it in a way that is consistent throughout the system (in a separate ticket)

lib/oli_web/components/tech_support_live.ex
```diff
@@ -28,16 +27,15 @@ defmodule OliWeb.TechSupportLive do
...
     {:ok, socket, layout: false}
   end
 
   @impl true
   def render(assigns) do
     ~H"""
-    <div class="mx-auto z-50 top-0 absolute right-0">
-      <.flash_group flash={@flash} />
-    </div>
+    <%= render_custom_flash(@custom_flash) %>
```
eliknebel, line 38: We should just remove this altogether and still use the existing flash infrastructure e.g. `put_flash(socket, :info, info)`

Francisco-Castro: @eliknebel Nico suggested a way to show the flash message without making a custom one. You'll see that approach in commits: [ecc4709](https://github.com/Simon-Initiative/oli-torus/pull/5617/commits/ecc4709b8b8d079f38267d89781f65c09cdb550a) and [9d5d704](https://github.com/Simon-Initiative/oli-torus/pull/5617/commits/9d5d704d956fc0dec71c392b6df7995836b6960e).

### New outcome:

https://github.com/user-attachments/assets/afd6bc52-1a9f-4429-8f6c-a8a5f71af372


### When getting an error:

Screenshot 2025-05-21 at 4 50 49 PM

Francisco-Castro:
@eliknebel This is how the flow of the request for help looks now:

https://github.com/user-attachments/assets/5eefdeda-6306-4e6e-a5eb-b3a418806c5b

I swapped how we retrieve the user in commit https://github.com/Simon-Initiative/oli-torus/pull/5617/commits/8573e723854dda222ee291b89a4bfc4030545e93 because, in  [PR-5588](https://github.com/Simon-Initiative/oli-torus/pull/5588), a user without an email is created when an admin navigates to a section under specific conditions. While the conditions aren't relevant to this PR, the result—having a user account without an email—occurs because we were previously retrieving the email by checking the user first, which may not always contain one.

Reference:
https://github.com/Simon-Initiative/oli-torus/blob/c46f490d17fba0c85a1e40e8e0048f8e329b3e8a/lib/oli/delivery/sections.ex#L93

PR #5628, opened 2025-05-27
[FEATURE] [MER-4399] Rework: Admin soft delete external tool

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-4399?focusedCommentId=28408) to the comment in the ticket

https://github.com/user-attachments/assets/f319c563-7b51-49fe-b48a-ff3afed1d52c

### DIFF
--- a/lib/oli/lti/platform_external_tools.ex
+++ b/lib/oli/lti/platform_external_tools.ex
@@ -200,7 +200,7 @@ defmodule Oli.Lti.PlatformExternalTools do
       [enabled: true, disabled: options.include_disabled, deleted: options.include_deleted]
       |> Enum.reduce([], fn
         {status, true}, acc -> [status | acc]
-        {_status, false}, acc -> acc
+        {_status, _}, acc -> acc
       end)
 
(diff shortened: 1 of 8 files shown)


lib/oli/lti/platform_external_tools.ex
```diff
@@ -197,7 +197,9 @@ defmodule Oli.Lti.PlatformExternalTools do
       end
 
     allowed_statuses =
-      [:enabled] ++ if options.include_disabled, do: [:disabled], else: []
+      [:enabled] ++
+        if(options.include_disabled, do: [:disabled], else: []) ++
+        if options.include_deleted, do: [:deleted], else: []
```
simonchoxx, line 202: NIT: Just a small thing. Instead of chaining multiple ++ operations, we could consider using `Enum.reduce` for a slightly more expressive approach like this

```
allowed_statuses =
  Enum.reduce([enabled: true, disabled: options.include_disabled, deleted: options.include_deleted], [], fn
    {status, true}, acc -> [status | acc]
    {_status, false}, acc -> acc
  end)
```

PR #5633, opened 2025-05-28
[BUG FIX] [MER-4614] Fix issue where checkbox label affects other LTI activity on page

eliknebel:
https://eliterate.atlassian.net/browse/MER-4614

This PR fixes an issue in authoring where changing the value of the "Launch tool in new window"  by clicking the label would non-deterministically affect other LTI activities on the page.

The issue was the htmlFor prop for the label was a static string and therefore pointed to all activities of this type. The fix here is to append a unique identifier to the id so that all checkboxes are unique on the page.


nicocirio (approved):
I tested it locally and works as expected 🚀

PR #5640, opened 2025-05-29
[ENHANCEMENT] [MER-4620] Pass in new results bucket name arg for Dataset job invocation

darrensiegel:
This PR adds an additional argument for invoking EMR serverless jobs.  This allows the job to vary which S3 bucket to write results to, which is necessary for independent execution on Tokamak vs Proton.


Francisco-Castro (approved):
LGTM!

PR #5643, opened 2025-05-29
[BUG FIX] [MER-3114] Fix license selection resetting other project attributes in form

andersweinstein:
Selecting a license type in the Project Attributes section of the Project Overview page was having the effect of resetting any other unsaved attribute values in that section (Language and, for admins, Calculate Embeddings on Publish). This issue posed a nuisance for automated QA testing.

The bug was caused by an error in the state management in the event handler for selections from the license type dropdown, which is needed to maintain state to show/hide the input box for custom license details. 

This recodes to use a simple `custom_license` flag in the assigns to show/hide the details, setting on mount and updating on changes to license type. This avoids the problem of trying to do this by manipulating the involved changeset structure.


Francisco-Castro (approved):
Nice work! Clean solution.

PR #5645, opened 2025-05-30
[FEATURE] [MER-4270] Adds support for save and preview of schedule and agenda

rgachuhi:
[MER-4270](https://eliterate.atlassian.net/jira/software/c/projects/MER/boards/2?selectedIssue=MER-4270)

This PR introduces a new **“View” button** to the scheduling toolbar, enabling instructors to preview how students will see the **Agenda** and **Schedule** based on the current scheduling configuration.

### Key Features

- Added **“View” button** between “Expand all” and “Clear” on the schedule tab toolbar.
- Clicking the button opens a **dropdown with two options**:
  - **View Agenda**
  - **View Schedule**
- Selecting an option opens the corresponding **student-facing view in a new tab** using the last saved schedule.
- If unsaved changes exist, a **confirmation modal** is shown:
  - **“Keep editing”** brings the user back to the scheduling tool.
  - **“View without saving”** proceeds with the last saved data.

[MER-4270]: https://eliterate.atlassian.net/browse/MER-4270?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/assets/src/apps/scheduler/ScheduleEditor.tsx
+++ b/assets/src/apps/scheduler/ScheduleEditor.tsx
@@ -139,7 +139,7 @@ export const ScheduleEditor: React.FC<SchedulerProps> = ({
 
   const { Modal: unsavedModal, showModal: showUnsavedModal } = useBackdropModal(
     <div>
-      <p>Viewing your schedule requires that you save all changes first.</p>
+      <p>Please save your changes before viewing your schedule.</p>
     </div>,
     () => {},
     () => {
--- a/assets/src/apps/scheduler/ScheduleEditor.tsx
+++ b/assets/src/apps/scheduler/ScheduleEditor.tsx
@@ -140,7 +140,7 @@ export const ScheduleEditor: React.FC<SchedulerProps> = ({
   const { Modal: unsavedModal, showModal: showUnsavedModal } = useBackdropModal(
     <div>
       <p>
-        If you view without saving, you’ll see the last saved changes—not your most recent edits.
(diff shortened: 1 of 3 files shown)


assets/src/apps/scheduler/ScheduleEditor.tsx
```diff
@@ -110,6 +137,22 @@ export const ScheduleEditor: React.FC<SchedulerProps> = ({
     'Cancel',
   );
 
+  const { Modal: unsavedModal, showModal: showUnsavedModal } = useBackdropModal(
+    <div>
+      <p>
+        If you view without saving, you’ll see the last saved changes—not your most recent edits.
```
darrensiegel: Is this the correct message?  The message seems to imply that I have a choice here:  that if I view without saving that I will see my last saved changes, not my most recent edits - but the implementation here forces a save?   Maybe the message should read something more along the lines of "Previewing your schedule requires that you save all changes first".

darrensiegel (changes requested):

assets/src/apps/scheduler/ScheduleEditor.tsx
rgachuhi: Urgh. I copy pasted from figma. Will update the text

rgachuhi: Text updated

PR #5648, opened 2025-06-02
[BUG FIX] [MER-4572] Cannot paste static components onto screens with question components

dtiwarATS:


dtiwarATS:
Hey @bsparks, this one also needs your review :)

dtiwarATS:
oh, Got it. Looks the build was failing due to some unchanged file showing up in the PR so just made a comment and checked-in the code so that the jobs re-runs

PR #5649, opened 2025-06-02
[ENHANCEMENT] [MER-4621] Refine approach for determining student ids to ignore

darrensiegel:
This PR adjusts the approach that Datasets infrastructure uses to determine which user ids to ignore.  The intent here is to ignore all "non students" and all students who opt out of research consent.  The previous impl was not robust against students having multiple roles (`:context_learner` and `:context_member`), which certain LMSes were assigning. 

The new approach handles this edge case.


Francisco-Castro (approved):
LGTM!

I left a NIT to make a small optimization in `determine_ignored_student_ids/1`

PR #5654, opened 2025-06-04
[FEATURE] [MER-4607] Update downloadable student report

Francisco-Castro:
Ticket: [MER-4607](https://eliterate.atlassian.net/browse/MER-4607)

This PR adapts our current `download_students_progress/2` function to meet the requirements outlined in the ticket. It also attempts to abstract as much code as possible for reuse in upcoming reports.

To avoid over-sorting, only the first three columns — status, name, and email — are taken into account.

[MER-4607]: https://eliterate.atlassian.net/browse/MER-4607?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli_web/controllers/delivery_controller.ex
+++ b/lib/oli_web/controllers/delivery_controller.ex
@@ -28,8 +28,6 @@ defmodule OliWeb.DeliveryController do
     ContextRoles.get_role(:context_instructor)
   ]
 
-  # %{"Enrolled" => 0, "Rejected invitation" => 1, ...}
-
   @doc """
   This is the default entry point for delivery users. It will redirect to the appropriate page based
   on whether the user is an independent learner or an LTI user.


lib/oli_web/controllers/delivery_controller.ex
```diff
@@ -507,4 +517,25 @@ defmodule OliWeb.DeliveryController do
       student_ids
     )
   end
+
+  defp sort_data(results) do
+    Enum.sort_by(results, &{@status_rank[&1.status], &1.name, &1.email})
```
nicocirio: What do you think about doing something like 
`Enum.sort_by(results, &{&1.status, &1.name, &1.email})` and avoid defining `@status_rank`?

nicocirio (dismissed):
We are in a good path Fran! I tested it locally and non-students are included in the report  and student names are not being sorted asc by Name. Could we please recheck that?
On the other hand, it is confusing for me if we need to label "Pending Confirmation" or "Waiting Confirmation". The ticket, the UI and the specs talk about "Pending Confirmation" except on the spreadsheet example

Screenshot 2025-06-10 at 12 00 32 PM

lib/oli_web/controllers/delivery_controller.ex
Francisco-Castro: It makes sense now that we are using the Pending Confirmation instead of the Waiting on.

Francisco-Castro:
Thanks, @nicocirio, for the review. I’ve fixed the filtering issue—could you please take a look when you have a moment?

lib/oli_web/controllers/delivery_controller.ex
```diff
@@ -27,6 +28,8 @@ defmodule OliWeb.DeliveryController do
     ContextRoles.get_role(:context_instructor)
   ]
 
+  # %{"Enrolled" => 0, "Rejected invitation" => 1, ...}
+
```
nicocirio: We forgot to remove this comment

nicocirio (approved):
Nice work! I tested it locally and it worked as expected:

image


I removed an old comment before merging it.

lib/oli_web/controllers/delivery_controller.ex
Francisco-Castro: Nice catch! I'll remove it in my current ticket --[MER-4610](https://eliterate.atlassian.net/browse/MER-4610). Thanks!

nicocirio: I've already removed it [here](https://github.com/Simon-Initiative/oli-torus/pull/5654/commits/e6d352c41020ca2d28e8744deecc82e628d86633) 😄

PR #5658, opened 2025-06-06
[ENHANCEMENT] [TQA-6] Added related test to the ticket

lperezCES:


assets/automation/src/systems/torus/pom/workspace/administrator/AdminUserDetailsPO.ts
```diff
@@ -0,0 +1,42 @@
...
+    this.utils = new Utils(this.page);
+    this.editButton = this.page.getByRole('button', { name: 'Edit' });
+    this.checkBox = this.page.getByRole('checkbox', {
+      name: 'Can Create Sections',
+    });
+    this.saveButton = this.page.getByRole('button', { name: 'Save' });
+  }
+
+  async goToUserEditPage(userId: string) {
+<<<<<<< HEAD
```
darrensiegel: Looks to be a left over merge conflict

darrensiegel (changes requested):
I saw an instance where a local merge conflict was still in the codebase. 

Also, I'm not sure what is going on with this PR, but many (or maybe most) of the files in this PR are already in our codebase - but this PR suggests that the entirety of the contents of their files have changed.  For instance, `assets/automation/src/systems/torus/pom/component/AddResourceCO.ts` is already on `master` but it appears in this PR as having every single line changed - even though the true diff is smaller.  We need to understand why this is happening - we want traditional "diffs" to appear in the PRs as opposed to seeing all files appear as if everything has changed.

PR #5660, opened 2025-06-09
[FEATURE] [MER-4482] Prevent LTI external tool creation in the activity bank

simonchoxx:
[MER-4482](https://eliterate.atlassian.net/browse/MER-4482)

This PR updates the activity bank authoring interface to exclude LTI external tool activities, ensuring only supported (non-LTI) activities appear as options for authors.

- Before

before

- After

after

[MER-4482]: https://eliterate.atlassian.net/browse/MER-4482?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli/authoring/editing/bank_editor.ex
+++ b/lib/oli/authoring/editing/bank_editor.ex
@@ -44,7 +44,7 @@ defmodule Oli.Authoring.Editing.BankEditor do
            ) do
       editor_map =
         Activities.create_registered_activity_map(project_slug)
-        |> Enum.filter(fn {_key, entry} -> entry.isLtiActivity == false end)
+        |> Enum.reject(fn {_key, entry} -> entry.isLtiActivity end)
         |> Enum.into(%{})
 
       project = Oli.Authoring.Course.get_project_by_slug(project_slug)
--- a/test/oli_web/live/workspaces/course_author/activity_bank_live_test.exs
+++ b/test/oli_web/live/workspaces/course_author/activity_bank_live_test.exs
@@ -84,14 +84,7 @@ defmodule OliWeb.Workspaces.CourseAuthor.ActivityBankLiveTest do
       {:ok, props} = Jason.decode(props_json)
       editor_map = props["editorMap"]
 
-      # 1. Ensure that LTI activities exist in the database
-      lti_activities = Oli.Activities.list_activity_registrations()
-
-      assert Enum.all?([lti_activity1.slug, lti_activity2.slug], fn slug ->
-               Enum.any?(lti_activities, &(&1.slug == slug))
-             end)
-
-      # 2. Ensure that they are NOT present in the editorMap (React props)
+      # Ensure that they are NOT present in the editorMap (React props)
       editor_map_slugs = Map.keys(editor_map)
 
       assert Enum.all?([lti_activity1.slug, lti_activity2.slug], fn slug ->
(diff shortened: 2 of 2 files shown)


lib/oli/authoring/editing/bank_editor.ex
```diff
@@ -42,7 +42,11 @@ defmodule Oli.Authoring.Editing.BankEditor do
              },
              %Paging{limit: 1, offset: 0}
            ) do
-      editor_map = Activities.create_registered_activity_map(project_slug)
+      editor_map =
+        Activities.create_registered_activity_map(project_slug)
+        |> Enum.filter(fn {_key, entry} -> entry.isLtiActivity == false end)
```
Francisco-Castro: This could also be written like

```
|> Enum.reject(fn {_key, entry} -> entry.isLtiActivity end)
```

test/oli_web/live/workspaces/course_author/activity_bank_live_test.exs
```diff
@@ -67,13 +67,71 @@ defmodule OliWeb.Workspaces.CourseAuthor.ActivityBankLiveTest do
...
+    lti_activity2 =
+      insert(:activity_registration, %{
+        slug: "lti_activity_2",
+        title: "LTI Activity 2"
+      })
+
+    insert(:lti_external_tool_activity_deployment, %{
+      activity_registration: lti_activity1,
+      deployment_id: Ecto.UUID.generate(),
+      status: :enabled
```
Francisco-Castro: By default those values are provided by the factory.

test/oli_web/live/workspaces/course_author/activity_bank_live_test.exs
```diff
@@ -67,13 +67,71 @@ defmodule OliWeb.Workspaces.CourseAuthor.ActivityBankLiveTest do
...
+      props_json = Floki.attribute(activity_bank_html, "data-live-react-props") |> List.first()
+      {:ok, props} = Jason.decode(props_json)
+      editor_map = props["editorMap"]
+
+      # 1. Ensure that LTI activities exist in the database
+      lti_activities = Oli.Activities.list_activity_registrations()
+
+      assert Enum.all?([lti_activity1.slug, lti_activity2.slug], fn slug ->
+               Enum.any?(lti_activities, &(&1.slug == slug))
+             end)
```
Francisco-Castro: I'm not sure why this check is here, since you explicitly define those inserts in you HELPER FUNCTIONS block.

Francisco-Castro (commented):
The functionality is working as expected, but there are a few parts that could be improved for better readability.

lib/oli/authoring/editing/bank_editor.ex
simonchoxx: While there’s no efficiency difference between `Enum.filter` and `Enum.reject`, I agree that using `Enum.reject` is clearer in this case, as you suggest. I’ll make the change—thanks!

test/oli_web/live/workspaces/course_author/activity_bank_live_test.exs
simonchoxx: I added this check to explicitly verify that the LTI activities exist before asserting they’re filtered out. It helps clarify test intent and avoids misleading failures if the setup changes.

Francisco-Castro: @simonchoxx There's no need to check them explicitly — they're already inserted in your helper function. It’s essentially like testing whether `insert/3` works, which isn’t necessary.

PR #5664, opened 2025-06-10
[FEATURE] [MER-4647] Text blob storage replacement for main DB extrinsic storage

darrensiegel:
This PR introduces a (configurable, optional) replacement for the database backed storage for global and page attempt level.   Prior to this PR, the `users` and `resource_attempts` each used a `state` field to store arbitrary JSON blobs.  This feature is used by the Adaptive Player to store EverApp and Simulation data, but the large of amount of data being stored has proved to be a performance and scalability issue.   We solve these performance issues by 1) Storing these large JSON blobs in S3 storage and 2) Eliminating all of the server JSON encoding and decoding by providing a pure `plain/text` API and distributing the JSON encoding/decoding to the client browser.

To enable this new "text blob storage", we set a new run time environment variable `BLOB_STORAGE_USE_DEPRECATED_API=false`.  The default here is `true` which allows a server to continue to operate with the current DB backed extrinsic service.     **WE MUST SET THIS FOR PROTON, but NOT SET THIS FOR ETX PROD.** 

A change had to be made in this client facing API for pure text as we now must read and write individual keys instead of getting and upsertting a JSON object with potentially many keys. 

Also, this only affects the Adaptive Page use of extrinsic storage.  The basic page has a couple of uses of user and attempt level extrinsic storage (pagination, alternatives selection) that will continue to be stored in the DB record.  These usages do not at all contribute to the scaling and performance issues and can continue to operate this way.

### DIFF
--- a/assets/src/data/persistence/blob.ts
+++ b/assets/src/data/persistence/blob.ts
@@ -18,9 +18,7 @@ const lastSet = () => {
   return arr[arr.length - 1];
 };
 
-
-export const read = async(key: string) => {
-
+export const read = async (key: string) => {
   const params = {
     method: 'GET',
     url: '/blob/' + key,
@@ -40,10 +38,9 @@ export const read = async(key: string) => {
   });
 
   return { result };
-}
+};
 
 export const write = async (key: string, state: any) => {
-
   const body = JSON.stringify(state);
 
   const params = {
@@ -64,12 +61,9 @@ export const write = async (key: string, state: any) => {
   });
 
   return { result };
-}
-
-
+};
 
 function readGlobal(keys: string[] | null = null) {
-
   const promises = keys?.map((key) => {
     const params = {
       method: 'GET',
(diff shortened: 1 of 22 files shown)


assets/src/data/persistence/blob.ts
```diff
@@ -0,0 +1,196 @@
...
+    }
+  }
+  return result;
+};
+
+export const internalUpdateGlobalUserState = async (
+  updates: { [topKey: string]: { [key: string]: any } },
+  useLocalStorage = false,
+) => {
+  /* console.log('updateGlobalUserState', updates); */
```
gastonabella, line 130: Nit: remove console log comment

assets/src/data/persistence/blob.ts
```diff
@@ -0,0 +1,196 @@
...
+const updateInterval = 300;
+const [batchedUpdate] = batchedBuffer(internalUpdateGlobalUserState, updateInterval);
+
+export const updateGlobalUserState = async (
+  updates: { [topKey: string]: { [key: string]: any } },
+  useLocalStorage = false,
+) => {
+  /*console.log('updateGlobalUserState called', { updates, useLocalStorage });*/
+  const result = await batchedUpdate(updates, useLocalStorage);
+  /* console.log('updateGlobalUserState result', { result, updates }); */
```
gastonabella, line 167: Same here

PR #5677, opened 2025-06-16
[ENHANCEMENT] [MER-4609] Update Downloadable Gradebook Report

gastonabella:
[MER-4609](https://eliterate.atlassian.net/browse/MER-4609)

Update Gradebook report according to new requirements:
- Display new columns (such as `Enrollment Status` `LMS ID`).
- Add a `Percentage` column for each assessment, as well as the `Points Possible` and `Points Earned`.
- Order rows alphabetically and assessments hierarchically.

https://github.com/user-attachments/assets/9dbea814-db02-4070-97e4-b819f9ffcf44

[MER-4609]: https://eliterate.atlassian.net/browse/MER-4609?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli/grading.ex
+++ b/lib/oli/grading.ex
@@ -115,21 +115,20 @@ defmodule Oli.Grading do
   Returns a Stream which can be written to a file or other IO
   """
   def export_csv(%Section{} = section) do
-    {gradebook, assessments_column_labels, points_possible} =
-      generate_gradebook_for_section(section)
+    {gradebook, assessments_column_labels} = generate_gradebook_for_section(section)
 
     column_labels =
       [status: "Status", name: "Name", email: "Email", lms_id: "LMS ID"]
       |> Keyword.merge(assessments_column_labels)
 
-    Enum.map(gradebook, &build_gradebook_row(&1, points_possible))
+    Enum.map(gradebook, &build_gradebook_row/1)
     |> sort_data()
     |> DataTable.new()
     |> DataTable.headers(column_labels)
     |> DataTable.to_csv_content()
   end
 
-  defp build_gradebook_row(%{user: user} = row, points_possible) do
+  defp build_gradebook_row(%{user: user} = row) do
     main_attrs = %{
       status: StudentUtils.parse_enrollment_status(user.enrollment_status),
       name: Utils.name(user),
@@ -143,10 +142,10 @@ defmodule Oli.Grading do
       percentage_column_label = :"#{score.label} - Percentage"
 
       acc
-      |> Map.put(points_earned_column_label, score.score)
+      |> Map.put(points_earned_column_label, format_score(score.score))
       |> Map.put(
         points_possible_column_label,
-        score.out_of || Map.get(points_possible, score.resource_id)
+        format_score(score.out_of)
       )
       |> Map.put(
         percentage_column_label,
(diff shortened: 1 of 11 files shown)


lib/oli/grading.ex
```diff
@@ -104,112 +106,138 @@ defmodule Oli.Grading do
...
-    |> CSV.encode()
+  defp sort_data(results) do
+    Enum.sort_by(results, &{&1.status, &1.name, &1.email})
   end
 
+  defp parse_enrollment_status(:enrolled), do: "Enrolled"
+  defp parse_enrollment_status(:suspended), do: "Suspended"
+  defp parse_enrollment_status(:pending_confirmation), do: "Pending confirmation"
+  defp parse_enrollment_status(:rejected), do: "Rejected invitation"
+  defp parse_enrollment_status(_status), do: "Unknown"
```
nicocirio: Does it make sense to move this to a common module and use it here and in the `delivery_controller.ex`?

lib/oli/grading.ex
```diff
@@ -104,112 +106,138 @@ defmodule Oli.Grading do
...
+  Returns a tuple containing a list of GradebookRow for every enrolled user, an ordered list of column labels and a map of points possible for each resource id.
 
-  `{[%GradebookRow{user: %User{}, scores: [%GradebookScore{}, ...]}, ...], ["Quiz 1", "Quiz 2"]}`
+  `{[%GradebookRow{user: %User{}, scores: [%GradebookScore{}, ...]}, ...], ["Quiz 1 - Points Earned", "Quiz 1 - Points Possible", "Quiz 1 - Percentage"], %{1 => 20, 2 => 10}}`
   """
   def generate_gradebook_for_section(%Section{} = section) do
-    # get publication page resources, filtered by graded: true
-    graded_pages = Sections.fetch_scored_pages(section.slug)
+    # get publication page resources, filtered by graded: true and ordered by numbering index
+    graded_pages = Sections.fetch_scored_pages(section.slug, :numbering_index)
```
nicocirio: There is a `Oli.Delivery.Sections.SectionResourceDepot.graded_pages` function that could be the perfect match for this use case.

lib/oli/grading.ex
Francisco-Castro: Maybe in `OliWeb.Delivery.Student.Utils`

lib/oli/utils.ex
```diff
@@ -515,6 +515,28 @@ defmodule Oli.Utils do
...
+  ## Examples
+
+      iex> StringUtils.title_case("hello world")
+      "Hello World"
+  """
+  @spec title_case(String.t()) :: String.t()
+  def title_case(string) do
+    string
+    |> String.split()
+    |> Enum.map(&capitalize_word/1)
```
Francisco-Castro: I think this could also be written as `String.capitalize/1`

lib/oli/grading.ex
gastonabella: Good idea! Just moved it to Student Utils.

lib/oli/grading.ex
gastonabella: Although this query returns revisions and the depot returns section resources, it contains all the fields I need, so I made the improvement. Thanks for the suggestion!

lib/oli/utils.ex
gastonabella: I'm just a guy who wanted to play around with binaries 😂 .

gastonabella:
> We are in a good path! I found that we are only showing `:enrolled` students instead of all students (regardless of status). I also found that we are showing more than 2 digits after the decimal point:
> 
> image

@nicocirio good catch! Looks like the old version of the report was using a function that only retrieved enrollments with that specific status. It's updated now!
Regarding the decimal points, I was waiting for confirmation on the format of the numbers, which is why the PR was still a draft. Two decimal points is the rule for all.

@Francisco-Castro is ready for review again

lib/oli_web/live/delivery/student/utils.ex
```diff
@@ -808,6 +808,24 @@ defmodule OliWeb.Delivery.Student.Utils do
...
 
+  @doc """
+  Calculates and formats the percentage score based on the given score and out_of value.
+
+  Returns `nil` if the score is `nil`.
+
+  ## Examples
+
+      iex> parse_percentage(45, 50)
+      "90%"
```
Francisco-Castro, line 819: NIT: Remove the % symbol

lib/oli/utils/db_seeder.ex
```diff
@@ -1397,13 +1409,10 @@ defmodule Oli.Seeder do
 
     # Enroll users
     user_tags
-    |> Enum.each(fn user_tag ->
-      Sections.enroll(map[user_tag].id, map[section_tag].id, [
-        Lti_1p3.Roles.ContextRoles.get_role(:context_learner)
-      ])
+    |> Enum.reduce(map, fn user_tag, acc ->
+      tag = :"#{user_tag}_enrollment"
```
Francisco-Castro, line 1413: NIT: Replace this with a string.

PR #5223, opened 2024-10-29
[BUG FIX] [MER-3931] students can submit after time limit

rgachuhi:
[MER-3931](https://eliterate.atlassian.net/browse/MER-3931)

This PR addresses an issue where, if a student starts a graded attempt, then navigates away from the quiz and then returns to the quiz after the timeout is expired, the quiz erroneously allows the student to answer and submit responses. 

[MER-3931]: https://eliterate.atlassian.net/browse/MER-3931?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5682, opened 2025-06-18
[BUG FIX] [MER-4648] Autosubmission not working correctly

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-4648) to the ticket

### How does auto-submission work?
  - When a student is in the browser and the time limit is reached, a js script auto-triggers a click on the submit button.
  - When the student closes the browser and the time limit is reached, an oban job auto-submits the attempt (one minute after the time limit is reached)

So, there was a one-minute window that allowed the student to re-enter the assessment and make changes to the responses. That bug was fixed on [MER-3931](https://github.com/Simon-Initiative/oli-torus/pull/5223)

### Issue
The logic for determining the deadline was considering a soft end date as part of the logic. According to the [Schedule and Assessment Settings Logic](https://docs.google.com/drawings/d/1LDzW7DH6fX7BW6oFBYgPR9H2ZkqHn-YP5JuvcfSq1xw/edit) only hard end dates (scheduling type = :due_by) should be considered in the logic `EARLIEST(due date, start time + time limit) + grace period`

### Fix
This PR changes the `determine_effective_deadline` function to consider only hard end dates.

### Before

https://github.com/user-attachments/assets/3c85b3a8-b00d-418f-b1f8-0752b175846e

### After

https://github.com/user-attachments/assets/5f65ccba-20dc-486e-8483-aee79e107d03

[MER-3931]: https://eliterate.atlassian.net/browse/MER-3931?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli/delivery/settings.ex
+++ b/lib/oli/delivery/settings.ex
@@ -325,30 +325,52 @@ defmodule Oli.Delivery.Settings do
     end
   end
 
+  @doc """
+  Determine the effective deadline for a resource attempt.
+  The effective deadline is the deadline that is used to determine if the attempt is late
+  or if the attempt should be auto-submitted.
+
+  - When end date is a hard one (scheduling type is due_by):
+    effective_deadline = EARLIEST(due date, start time + time limit) + grace period
+
+  - When end date is a soft one (scheduling type is read_by):
+    - If there is a time limit:
+      effective_deadline = start time + time limit + grace period
+    - If there is no time limit:
+      effective_deadline = nil
+  """
+
   def determine_effective_deadline(nil, _), do: nil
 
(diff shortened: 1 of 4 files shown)


lib/oli/delivery/settings.ex
```diff
@@ -325,30 +325,52 @@ defmodule Oli.Delivery.Settings do
     end
   end
 
+  @doc """
```
manelli, line 328: Thank you for the docs! 🙌

PR #5684, opened 2025-06-19
[BUG FIX] [MER-4662] Opening a student's attempt as an admin returns them to first screen

dtiwarATS:

### DIFF
--- a/assets/src/apps/delivery/layouts/deck/DeckLayoutView.tsx
+++ b/assets/src/apps/delivery/layouts/deck/DeckLayoutView.tsx
@@ -447,6 +447,7 @@ const DeckLayoutView: React.FC<LayoutProps> = ({ pageTitle, pageContent, preview
       partAttemptGuid: string,
       response: StudentResponse,
     ) => {
+      if (reviewMode) return;
       // save before submitting
       const { result, snapshot } = await handleActivitySavePart(
         activityId,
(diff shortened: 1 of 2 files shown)


assets/src/apps/delivery/store/features/groups/actions/deck.ts
```diff
@@ -198,6 +197,7 @@ export const initializeActivity = createAsyncThunk(
     };
     if (!isReviewMode && !isHistoryMode) {
       sessionOps.push(targetVisitTimeStampOp);
+      sessionOps.push(resumeTarget);
```
dtiwarATS, line 200: If the board is opened in Review or History mode, the session.resume value should not be updated.

The issue occurred because when an author opened an in-progress attempt, the resumeId was being set to the screen they accessed. As a result, when the student tried to resume their lesson, it would start from the screen the admin last viewed, instead of where the student left off.

assets/src/apps/delivery/layouts/deck/DeckLayoutView.tsx
```diff
@@ -447,6 +447,7 @@ const DeckLayoutView: React.FC<LayoutProps> = ({ pageTitle, pageContent, preview
       partAttemptGuid: string,
       response: StudentResponse,
     ) => {
+      if (reviewMode) return;
```
dtiwarATS, line 450: I noticed that when a lesson is opened in Review mode, the Janus navigation button gets automatically triggered because its selected property is set to true. This was causing a check event to fire, even in Review mode—which should not happen.

For example, if an admin opens a student’s attempt in Review mode and the first screen includes a navigation button to start the lesson, the button appears disabled. However, the part still calls prop.onSubmit() due to the selected property being true.

To fix this, I added the following code to prevent this behavior in Review mode.

bsparks (approved):

PR #5691, opened 2025-06-23
[BUG FIX] [MER-4537] Allow triggers to display when Explanation fires

darrensiegel:
This PR makes an adjustment to the logic that checks to see if an explanation trigger exists, when an explanation has fired.  Previously, it simply took the result of the `Triggers. check_for_explanation_trigger` function and overwrote the `trigger` attribute of the feedback action.  When there is no explanation trigger (that is, when this function returns `nil`) this effectively overwrites a non-explanation trigger that would have already been set. 

The change here is to preserve the previously set trigger, and only replace it with the explanation trigger if there is actually an explanation trigger.  This gives us the original intent that Explanation triggers have higher priority than non explanation triggers, when an explanation has been tripped.


Francisco-Castro (approved):
LGTM!

PR #5696, opened 2025-06-25
[FEATURE] [MER-4319] LTI Data Passback

eliknebel:
https://eliterate.atlassian.net/browse/MER-4319

This PR implements grade pass back functionality for LTI activities.

Due to the attempts model and activity constraints, torus AGS is restricted to only sending a score to a single lineitem per activity. This service endpoint is provided via the 'lineitem' key within the AGS claim, specified in the LTI AGS spec https://www.imsglobal.org/spec/lti-ags/v2p0#assignment-and-grade-service-claim.

A tool can issue multiple grade updates for that lineitem and they will be registered in torus as subsequent activity attempts, for both unscored and scored pages (resource attempts). This is functionally equivalent to pressing the "Reset" button in an activity on an unscored page, however does not directly have an equivalent existing function for scored pages. I tested this out for scored pages however, and it works as one would expect - The latest activity attempt/LTI posted score is processed as the final score for the activity when the scored page attempt is submitted.

### Steps to test:
You should be able to use any LTI 1.3 tool you are familiar with to test this. However, due to the single lineitem limiation described above, it might be difficult to find a tool that only supports single lineitem AGS. I tested with both the lti-example-tool as well as the more commonly used [lti.js library](https://github.com/Cvmcosta/ltijs) using a [simplified setup repo here](https://github.com/Simon-Initiative/ltijs-demo-server).

1. Clone the [lti-example-tool](https://github.com/Simon-Initiative/lti-example-tool) repository and set up according to the [Development](https://github.com/Simon-Initiative/lti-example-tool?tab=readme-ov-file#development) instructions in the README
2. In torus, create a new external tool named "LTI Example Tool" as you would under http://localhost/admin/external_tools
3. In the lti-example-tool, create a registration for your local torus dev environment. This can also be done in the UI under /registrations - OR - copy seeds.example.yml to seeds.yml, entering the registration entry details below and running `gleam run -m lti_example_tool/database/migrate_and_seed seed`:
```
  - name: "Local Torus"
    issuer: "http://localhost"
    client_id: "EXAMPLE_CLIENT_ID"
    auth_endpoint: "http://localhost/lti/authorize_redirect"
    access_token_endpoint: "http://localhost/lti/auth/token"
    keyset_url: "http://localhost/.well-known/jwks.json"
    // Replace deployment_id  with the Deployment ID from your local torus instance
    // This can be found in any project Overview under the tool's name
    deployment_id: "7ad727c9-3f14-4ffc-91e5-8e1f0df27359"

```

4. In torus, go into any project and enable the LTI Example Tool previously registered
5. Go into both an unscored and score page and add the tool
6. Publish the project and create a new section
7. As a student, access both unscored and scored pages and launch the tool. You should see all the claims details (including AGS claim "https://purl.imsglobal.org/spec/lti-ags/claim/endpoint") as well as an Assignment and Grade Services section with the message "AGS Service is available". Here you can set a score, maximum and comment and send the score back to torus.  NOTICE: score/maximum values must include a decimal point. e.g. '1.0'.
8. As a student when you submit a graded page and review the results, you should see your posted LTI AGS grade above the LTI activity
9. As an instructor you can explore the dashboard for score as you would for any other activity. e.g. Overview > Students > [Student Name] > Quiz Scores, Insights > Scored Activities, ...
10. Notice if you launch into the LTI activity as an instructor or admin, the AGS service is not available

### DIFF
--- a/lib/oli_web/controllers/api/lti_controller.ex
+++ b/lib/oli_web/controllers/api/lti_controller.ex
@@ -241,18 +241,36 @@ defmodule OliWeb.Api.LtiController do
   end
 
   defp validate_client_assertion(client_assertion) do
-    with {:ok, %JOSE.JWT{fields: %{"iss" => client_id, "sub" => client_id_sub} = jwt}} <-
+    with {:ok, %JOSE.JWT{fields: %{"iss" => client_id, "sub" => client_id}}} <-
            peek_jwt(client_assertion),
-         true <- client_id == client_id_sub,
-         platform_instance when not is_nil(platform_instance) <-
-           PlatformInstances.get_platform_instance_by_client_id(client_id),
-         {:ok, jwk} <- get_jwk_for_assertion(client_id, jwt["kid"]),
+         %JOSE.JWS{fields: %{"kid" => kid}} <- JOSE.JWT.peek_protected(client_assertion),
+         {:ok, _iss, _sub} <- validate_matching_iss_sub(client_id, client_id),
+         {:ok, platform_instance} <- get_platform_instance_by_client_id(client_id),
+         {:ok, jwk} <- get_jwk_for_assertion(platform_instance.keyset_url, kid),
          {true, _jwt, _jws} <- JOSE.JWT.verify(jwk, client_assertion) do
       {:ok, %{sub: client_id}}
     else
       e ->
         Logger.error("Failed to validate client assertion: #{inspect(e)}")
         {:error, :invalid_client_assertion}
+    end
+  end
+
+  defp validate_matching_iss_sub(iss, sub) do
+    if iss == sub do
+      {:ok, iss, sub}
+    else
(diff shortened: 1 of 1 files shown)


test/oli_web/controllers/lti_ags_controller_test.ex
```diff
@@ -0,0 +1,938 @@
...
+      params = %{
+        "userId" => student_1.sub,
+        "scoreGiven" => 8,
+        "scoreMaximum" => 10,
+        "activityProgress" => "Completed",
+        "gradingProgress" => "FullyGraded",
+        "timestamp" => DateTime.utc_now() |> DateTime.to_iso8601()
+      }
+
+      conn =
```
darrensiegel, line 721: Nice work with these unit tests

darrensiegel (approved):

PR #5700, opened 2025-06-26
[BUG FIX] [MER-4591] Ingest failing on some courses

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-4591) to the ticket

The function `rewire_activity_references/2` was reverted to return the mapper (like the other `rewire_*` functions of the `Oli.Interop.Ingest.Processing.Rewiring` module), restoring the expected contract and fixing the import process. 
A test was added to ensure this contract is enforced in the future.

### Before

https://github.com/user-attachments/assets/9970d6e4-2f23-4ac6-a9f0-e1ac2a13f88a

### After

https://github.com/user-attachments/assets/5f574867-436f-4b8c-809e-790cae90cfca

### DIFF
--- a/lib/oli/interop/ingest/processor/rewiring.ex
+++ b/lib/oli/interop/ingest/processor/rewiring.ex
@@ -1,4 +1,10 @@
 defmodule Oli.Interop.Ingest.Processing.Rewiring do
+  @moduledoc """
+  Rewiring functions for the ingest processor.
+  This module is responsible for rewiring (updating or remapping) references inside content data structures during an ingest (import) process.
+  When content is imported, IDs for activities, tags, bibliographic references, etc., may change. This module updates those references in the content to point to the new, correct IDs.
+  """
+
   alias Oli.Resources.PageContent
 
   defp retrieve(map, key) do
--- a/test/oli/interop/ingest/processor/rewiring_test.exs
+++ b/test/oli/interop/ingest/processor/rewiring_test.exs
@@ -1,0 +1,131 @@
+defmodule Oli.Interop.Ingest.Processing.RewiringTest do
+  use ExUnit.Case, async: true
+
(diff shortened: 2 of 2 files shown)


lib/oli/interop/ingest/processor/products.ex
```diff
@@ -105,7 +105,7 @@ defmodule Oli.Interop.Ingest.Processor.Products do
 
     {certificate_params, new_product_attrs} =
       case Map.get(product, "certificate") do
-        nil ->
+        cert when cert in [nil, "null"] ->
```
Francisco-Castro, line 108: Just curious—why did we add another type to check?

lib/oli/interop/ingest/processor/rewiring.ex
```diff
@@ -13,28 +19,24 @@ defmodule Oli.Interop.Ingest.Processing.Rewiring do
 
   @spec rewire_activity_references(map(), any) :: map()
```
Francisco-Castro, line 20: Interesting that Dialyzer didn't catch the mismatch! 🤔

Francisco-Castro (approved):
Great job! 🏅 Thanks for adding tests to the Rewiring module.

I tested locally with a course that was previously failing, and it was successfully ingested now.

lib/oli/interop/ingest/processor/products.ex
nicocirio: Oh, I forgot to comment on that in the description; while testing with various .zip files, one exported from Proton failed. After inspecting the JSON, I realized it was because the `certificates` key had a "null" value.

image

Francisco-Castro: Good catch! Thanks for the explanation.

PR #5703, opened 2025-06-27
[BUG FIX] [MER-4686] Can leave names empty on account editing

simonchoxx:
[MER-4686](https://eliterate.atlassian.net/browse/MER-4686)

This PR fixes the bug  view that allowed leaving the First Name and Last Name fields empty and saving the changes in the Account Settings. Now, the First Name cannot be empty and the Last Name must have at least two non-whitespace characters.

Additionally, the Full Name is automatically updated on the screen as the First Name or Last Name fields are edited. If the entered values do not meet the validation requirements, error messages are shown on the corresponding fields.

Both the First Name and Last Name are saved without leading or trailing spaces, as was already the case. This applies both when a user edits their own account and when an administrator edits another user’s account.

Author Account Settings

https://github.com/user-attachments/assets/9ac30013-a681-48af-a608-8279bf14be60

User Account Settings

https://github.com/user-attachments/assets/6619d7ca-e048-4471-a823-6a48265bd8cd

[MER-4686]: https://eliterate.atlassian.net/browse/MER-4686?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli/utils.ex
+++ b/lib/oli/utils.ex
@@ -533,4 +533,34 @@ defmodule Oli.Utils do
 
   defp compare(f, t, true), do: f <= t
   defp compare(f, t, false), do: f < t
+
+  @doc """
+  Validates the given_name and family_name fields.
+
+  ## Examples
+
+      iex> common_name_validations(changeset)
+      %Ecto.Changeset{data: %Author{}}
+
+      iex> common_name_validations(changeset)
+      %Ecto.Changeset{data: %User{}}
+
+  """
+  @spec common_name_validations(Ecto.Changeset.t()) :: Ecto.Changeset.t()
+  def common_name_validations(changeset) do
+    changeset
+    |> update_change(:given_name, &maybe_trim/1)
+    |> update_change(:family_name, &maybe_trim/1)
+    |> validate_length(:given_name, min: 1, message: "Please enter a First Name.")
+    |> validate_length(:family_name,
+      min: 2,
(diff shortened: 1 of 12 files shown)


lib/oli/utils.ex
```diff
@@ -533,4 +533,34 @@ defmodule Oli.Utils do
...
+      iex> common_name_validations(changeset)
+      %Ecto.Changeset{data: %User{}}
+
+  """
+  @spec common_name_validations(Ecto.Changeset.t()) :: Ecto.Changeset.t()
+  def common_name_validations(changeset) do
+    changeset
+    |> update_change(:given_name, &maybe_trim/1)
+    |> update_change(:family_name, &maybe_trim/1)
+    |> validate_length(:given_name, min: 1, message: "Please enter a First Name.")
```
nicocirio, line 554: As per the ticket requirements, the first name is not a required field.

nicocirio (changes requested):
Nice work!
We need to make a small adjustment to the required fields to match the ticket's description before merging the PR

lib/oli/utils.ex
simonchoxx: Thanks for the review! According to the expected behavior in the ticket the `first name` should be a required field!

```
Expected Behavior

The user cannot save the changes if the First and Last name inputs are empty.
```

PR #5708, opened 2025-06-27
[BUG FIX] [MER-4737] Fix LTI launches for H5P activities

eliknebel:
https://eliterate.atlassian.net/browse/MER-4737

- [x] NOTE: This PR is branched from `MER-4319-instructor-lti-data-passback`. [The Data Passback PR](https://github.com/Simon-Initiative/oli-torus/pull/5696) should be reviewed and landed before this one.

Fixes an issue where the LTI launches were missing some critical claims in order to function properly, specifically the 'context' claim in the delivery views. I also refactored this logic to clean up the implementation details and removed the now deprecated Platform Instances controller and views superseded by the new LTI External Tools admin interface.

This likely also fixes (at least some of) the launch issues seen in other LTI integrations as well.

With this fix in place, I was able to successfully launch into H2P and access the configuration view. From my testing, it seems that H2P doesn't respond to the roles provided from the launch, instead the H2P-specific role (e.g. administrator, author, learner) must be set in the H2P interface itself. Although more testing is warranted here to fully understand how this works. From Torus perspective, the appropriate roles are being passed along in the different contexts and this can be seen by testing with the [lti-example-tool](https://lti-example-tool.oli.cmu.edu), configured in tokamak as OLI Example Tool. It seems like either deep linking or custom parameter support is required to actually specify which activity to display in H5P.


lib/oli_web/controllers/lti_controller.ex
```diff
@@ -261,6 +171,104 @@ defmodule OliWeb.LtiController do
...
+
+    roles =
+      [
+        Lti_1p3.Roles.ContextRoles.get_role(:context_content_developer)
+      ] ++
+        if Oli.Accounts.is_admin?(author),
+          do: [
+            Lti_1p3.Roles.PlatformRoles.get_role(:system_administrator),
+            Lti_1p3.Roles.PlatformRoles.get_role(:institution_administrator),
+            Lti_1p3.Roles.ContextRoles.get_role(:context_content_developer)
```
gastonabella: Is it necessary to put this role twice?

gastonabella (commented):
Looks much cleaner now!
I wonder if we should remove `lib/oli_web/views/api/platform_instance_view.ex` and `lib/oli_web/views/platform_instance_view.ex` files as well.

lib/oli_web/controllers/lti_controller.ex
eliknebel: Ah, no good catch

PR #5710, opened 2025-07-01
[ENHANCEMENT] [TQA-9] Test scripts are added

lperezCES:


eliknebel:
For future PR's, it would be great if we had a link to the ticket in the description. Thanks!

PR #5713, opened 2025-07-02
[BUG FIX] [MER-4754] Cannot Set Incorrect Value Points in Activities

gastonabella:
[MER-4757](https://eliterate.atlassian.net/browse/MER-4757)

### Problem
Previously, the code assumed that the "incorrect" response for an activity part always had a score of `0`. This caused errors when using custom scoring, where an "incorrect" response could have a different score (e.g., `1`). As a result, the system failed to identify the catch-all incorrect response and threw an exception.

### Solution
The logic for finding the "incorrect" response was updated to match only the catch-all rule (e.g., `input like {.*}`), regardless of its score. This makes the code compatible with custom scoring scenarios and prevents runtime errors when the score is not zero.

https://github.com/user-attachments/assets/66b99505-983e-48ff-b314-b8d3bc3db3b5

[MER-4757]: https://eliterate.atlassian.net/browse/MER-4757?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/assets/src/data/activities/model/responses.ts
+++ b/assets/src/data/activities/model/responses.ts
@@ -80,10 +80,9 @@ export const getIncorrectResponse = (model: HasParts, partId: string) => {
   return Maybe.maybe(
     getResponsesByPartId(model, partId).find((r) => {
       return (
-        r.score === 0 &&
-        (r.rule === matchRule('.*') ||
-          // Allow for special rule form used by ResponseMulti
-          (r.rule.startsWith('input_ref') && MultiRule.ruleIsCatchAll(r.rule)))
+        r.rule === matchRule('.*') ||
+        // Allow for special rule form used by ResponseMulti
+        (r.rule.startsWith('input_ref') && MultiRule.ruleIsCatchAll(r.rule))
       );
     }),
   ).valueOrThrow(new Error('Could not find incorrect response'));


andersweinstein:
The reason for verifying that score === 0 is that an author can set a _correct_ answer regex rule of .* (perhaps on a targeted feedback), and then this action can interfere with finding the incorrect response, breaking the activity. It is better if authors never do this, but nothing prevents this. So I think it would be more robust to change the
  `r.score === 0`
to 
 ` r.score === (getIncorrectPoints(model, partId) || 0)`
  
(it would also be smoother if `getIncorrectPoints` returned 0 in case no custom incorrect score is defined, but this works)

andersweinstein:
[Gaston] _Regarding your suggestion, I tried to do that exactly: compare the response score with the incorrectPoints from the model. It does not work because when you update the incorrect score in the UI, the incorrectScore field within the part gets updated, but the corresponding score within the responses field doesn't, so there's a mismatch between these two values._

Perhaps it is not so important to defend against author setting correct answer pattern of .* this way (though I believe this has happened) and we should just keep your simple fix.

gastonabella:
Closing this PR to allow for a more thoughtful and robust solution.

PR #5715, opened 2025-07-03
[BUG FIX] [MER-4755] Uploading image does not override image source URL

dtiwarATS:

### DIFF
--- a/assets/src/components/parts/janus-image/Image.tsx
+++ b/assets/src/components/parts/janus-image/Image.tsx
@@ -98,7 +98,7 @@ const Image: React.FC<PartComponentProps<ImageModel>> = (props) => {
     props.onReady({ id, responses: [] });
   }, [ready]);
 
-  const { width, height, src, imageSrc,defaultSrc, alt } = model;
+  const { width, height, src, imageSrc, defaultSrc, alt } = model;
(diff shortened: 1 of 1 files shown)


assets/src/components/parts/janus-image/Image.tsx
```diff
@@ -118,7 +118,7 @@ const Image: React.FC<PartComponentProps<ImageModel>> = (props) => {
   }, [width, height]);
   useEffect(() => {
     //Image Source will take precedence ( if there is an image link present in it). If Image Sorce is blank then it will display image link from src.
-    const imageSource = imageSrc?.length ? imageSrc : src;
+    const imageSource = imageSrc?.length && imageSrc != defaultSrc ? imageSrc : src;
```
dtiwarATS, line 121: These changes were already implemented in the Image Authoring component but were missing in the Delivery component.

Previously, `imageSrc` was not being validated against the `defaultSrc`, so even when a valid `src` was provided, the component would still render the default image in Delivery mode. This update ensures that `imageSrc` takes precedence only if it's non-empty and different from the `defaultSrc`; otherwise, it falls back to `src`.

nicocirio (approved):

PR #5716, opened 2025-07-03
[BUG FIX] [MER-4760] Issue with incorrect attempt in Hub and Spoke component

dtiwarATS:

### DIFF
--- a/assets/src/apps/authoring/components/Flowchart/rules/create-hub-spoke-rules.ts
+++ b/assets/src/apps/authoring/components/Flowchart/rules/create-hub-spoke-rules.ts
@@ -107,7 +107,7 @@ export const createSpokeCorrectCondition = (
   const correctconditions = correctScreens
     .filter((screen: string) => screen?.length)
     .map((correctScreen: any) => {
-      return createCondition(`session.visits.${correctScreen}`, '1', 'equal');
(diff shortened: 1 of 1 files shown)


assets/src/apps/authoring/components/Flowchart/rules/create-hub-spoke-rules.ts
```diff
@@ -107,7 +107,7 @@ export const createSpokeCorrectCondition = (
   const correctconditions = correctScreens
     .filter((screen: string) => screen?.length)
     .map((correctScreen: any) => {
-      return createCondition(`session.visits.${correctScreen}`, '1', 'equal');
+      return createCondition(`session.visits.${correctScreen}`, '1', 'greaterThanInclusive');
```
dtiwarATS, line 110: **Updated condition from `equal `to `greaterThanInclusive` for `session.visits.${destinationScreenId}`**

Previously, the condition used equal (value = 1) to ensure each spoke screen was visited once before marking the hub as complete. However, when a student submitted an incorrect answer on a spoke screen, the s`ession.visits.${destinationScreenId} `counter gets incremented with each attempt.

As a result, the hub completion check would fail, even if all spoke screens were genuinely visited. This update changes the condition to `greaterThanInclusive`, allowing the logic to pass as long as the student has visited the screen at least once.

nicocirio (approved):

PR #5719, opened 2025-07-04
[BUG FIX] [MER-4762] Fix broken redirect when canceling section creation

Francisco-Castro:
Ticket: [MER-4762](https://eliterate.atlassian.net/browse/MER-4762)

This PR fixes the redirect bug when canceling the creation of a section. The issue was caused by a route that no longer exists.

https://github.com/user-attachments/assets/b2a108cf-9331-4503-beea-fefe0bfe7ae0


nicocirio (approved):
🚀

PR #5721, opened 2025-07-08
[BUG FIX] [MER-4166] Instructors should be able to review answers when "View Answers" is disallowed

manelli:
https://github.com/user-attachments/assets/32d9b530-fb63-47f2-889b-457bff352690

See: https://eliterate.atlassian.net/browse/MER-4166


nicocirio (approved):
🚀

PR #5723, opened 2025-07-09
[FEATURE] [MER-4689] Support modal routing by publisher

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-4689) to the ticket

For demo purposes, in my local environment, I have set Wikipedia as the default knowledge base link and `support@example.com` as the default support email.

### No knowledge base link or support email set (default ones are used)

https://github.com/user-attachments/assets/f8c7ec38-b4f7-48ad-a72a-9e1e2a6e6abf

### Set knowledge base and support email.

https://github.com/user-attachments/assets/26a49ef6-0d8a-4b39-8a09-311a16999a22

### DIFF
--- a/lib/oli/inventories.ex
+++ b/lib/oli/inventories.ex
@@ -201,6 +201,24 @@ defmodule Oli.Inventories do
     Publisher.changeset(publisher, attrs)
   end
 
+  @doc """
+  Determines the publisher for a given context map.
+
+  Accepts a map with either a `"section"` or `"project"` key, each pointing to their respective structs.
+  Returns the associated publisher for the section or project, or the default publisher if not found.
+
+  ## Examples
+
+      iex> get_publisher_for_context(%{"section" => %Oli.Delivery.Sections.Section{id: 1}})
+      %Publisher{} # the publisher for the section
+
+      iex> get_publisher_for_context(%{"project" => %Oli.Authoring.Course.Project{id: 1}})
+      %Publisher{} # the publisher for the project
+
+      iex> get_publisher_for_context(%{})
+      %Publisher{} # the default publisher
+  """
+  @spec get_publisher_for_context(map()) :: Publisher.t()
   def get_publisher_for_context(%{"section" => %Oli.Delivery.Sections.Section{id: section_id}}) do
     from(s in Oli.Delivery.Sections.Section,
       join: p in Oli.Authoring.Course.Project,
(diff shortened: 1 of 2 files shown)


lib/oli/inventories.ex
```diff
@@ -200,4 +200,60 @@ defmodule Oli.Inventories do
   def change_publisher(%Publisher{} = publisher, attrs \\ %{}) do
     Publisher.changeset(publisher, attrs)
   end
+
+  def get_publisher_for_context(%{"section" => %Oli.Delivery.Sections.Section{id: section_id}}) do
```
gastonabella, line 222: What about adding a `@spec` and `@doc` here as well?

lib/oli/inventories/publisher.ex
```diff
@@ -3,6 +3,15 @@ defmodule Oli.Inventories.Publisher do
   import Ecto.Changeset
   import Oli.Utils
 
+  @moduledoc """
+  The Publisher schema represents a content publisher in the system.
+
+  ## Fields
```
gastonabella, line 9: Nit: it'd be great to mention here all the fields on this table and not only the new ones.

lib/oli/inventories/publisher.ex
```diff
@@ -50,4 +64,16 @@ defmodule Oli.Inventories.Publisher do
         false
     end
   end
+
+  defp maybe_validate_support_email(changeset) do
```
gastonabella, line 68: The regex here accepts an invalid email as valid (it can not include more than one `@` symbol):
`test@@example.com`

Something like `~r/^[^\s@]+@[^\s@]+\.[^\s@]+$/` would do the trick here.

Anyway, since we have several places where we accept emails as user input, it'd be better if we unify the logic behind all these places and have a shared module that takes over email validation.
Maybe as a future tech debt, we could delegate this to an existing library such as [email_checker](https://github.com/maennchen/email_checker)

lib/oli/inventories/publisher.ex
nicocirio: Yes, I agree. TBH, I was not sure of the meaning of all the fields, so I went to read the ticket when this schema was created, but unfortunately it was not documented there. So I finally opted to document the meaning of the fields I was 100% sure.

gastonabella (approved):
LGTM! Thanks for the changes 🚀

PR #5725, opened 2025-07-11
[FEATURE] [MER-4683] Critical information missing from my courses page

simonchoxx:
[MER-4683](https://eliterate.atlassian.net/browse/MER-4683)

This PR adds the display of the course start and end dates, as well as instructor names, on the “My Courses” page for both students and instructors. 

This information is now visible directly on each course card.

**Instructor view**

instructor-view

**Student view**

student-view

[MER-4683]: https://eliterate.atlassian.net/browse/MER-4683?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli_web/live/common/utils.ex
+++ b/lib/oli_web/live/common/utils.ex
@@ -3,7 +3,6 @@ defmodule OliWeb.Common.Utils do
 
   alias Oli.Accounts.{User, Author}
   alias OliWeb.Common.SessionContext
-  alias OliWeb.Common.FormatDateTime
 
   require Logger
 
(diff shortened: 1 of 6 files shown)


lib/oli_web/live/common/utils.ex
```diff
@@ -27,6 +28,19 @@ defmodule OliWeb.Common.Utils do
...
+  @doc """
+  Formats a date using the given format.
+
+  ## Examples
+
+      iex> format_date(~U[2023-09-06 13:28:00Z], %{SessionContext.init() | local_tz: "America/Montevideo"}, "{0M}/{0D}/{YY}")
+      "09/06/23"
+  """
+  @spec format_date(DateTime.t(), map(), String.t()) :: String.t()
+  def format_date(date, context, format) do
```
gastonabella: Why do we need to define a function that does the same as `FormatDateTime.to_formatted_datetime/3`? Couldn't we just use that function instead?

simonchoxx: I agree. I will remove that function. Thanks!

PR #5728, opened 2025-07-14
[FEATURE] [MER-4583] Student Timezone Detection

gastonabella:
[MER-4583](https://eliterate.atlassian.net/browse/MER-4583)

Display current timezone information on the following student pages:
- My Assignments page
- Schedule page
- Learn page (both gallery and outline views)

**Note**: The bug mentioned in the Jira ticket, where students see due dates changing unexpectedly, was not addressed in this PR.

https://github.com/user-attachments/assets/b873ca6e-0aac-4df0-9062-d9e9afb1ad3d

[MER-4583]: https://eliterate.atlassian.net/browse/MER-4583?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli_web/components/utils.ex
+++ b/lib/oli_web/components/utils.ex
@@ -72,14 +72,10 @@ defmodule OliWeb.Components.Utils do
   end
 
   attr :timezone, :string, required: true
-  attr :class, :string, default: nil
 
   def timezone_info(assigns) do
     ~H"""
-    <div
-      id="timezone_info"
-      class={["flex items-center gap-2 text-[#757682] dark:text-[#bab8bf] mb-1", @class]}
-    >
+    <div id="timezone_info" class="flex items-center gap-2 text-[#757682] dark:text-[#bab8bf] mb-1">
       <div class="w-5 h-5 flex items-center justify-center">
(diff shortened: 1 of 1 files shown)


lib/oli_web/components/utils.ex
```diff
@@ -69,4 +70,23 @@ defmodule OliWeb.Components.Utils do
         false
     end
   end
+
+  attr :timezone, :string, required: true
+  attr :class, :string, default: nil
```
simonchoxx: I see that the `:class` attribute is defined with **default: nil**, but it’s never passed when the component is used. Would it make sense to remove it for now, since it doesn’t seem necessary and can always be added later if needed?

simonchoxx (approved):
Great work 🚀

I left a small comment! I’d like to hear your thoughts on it!

lib/oli_web/components/utils.ex
gastonabella: Makes sense!

PR #5729, opened 2025-07-14
[FEATURE] [MER-4684] Add available dates to student assignment UI - rework

simonchoxx:
[MER-4684](https://eliterate.atlassian.net/browse/MER-4684)

This PR introduces changes to ensure that due and available dates are always displayed, regardless of whether the course resources have a schedule set or not.

If the resources don’t have a schedule set, it will display:

`Available: Now         Due by: None`

- Gallery view

gallery-view-dates

- Outline view

outline-view-dates

[MER-4684]: https://eliterate.atlassian.net/browse/MER-4684?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli_web/live/delivery/student/utils.ex
+++ b/lib/oli_web/live/delivery/student/utils.ex
@@ -388,7 +388,7 @@ defmodule OliWeb.Delivery.Student.Utils do
   def container_label_for_scheduling_type([:read_by]), do: "Read by: "
   def container_label_for_scheduling_type(_), do: "Due by: "
 
-  def label_for_scheduling_type(:due_by), do: "Due by: "
+  def label_for_scheduling_type(type) when type in [:due_by, "Not yet scheduled"], do: "Due by: "
   def label_for_scheduling_type(:read_by), do: "Read by: "
   def label_for_scheduling_type(:inclass_activity), do: "In-class activity by: "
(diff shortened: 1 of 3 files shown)


lib/oli_web/live/delivery/student/utils.ex
```diff
@@ -388,7 +388,7 @@ defmodule OliWeb.Delivery.Student.Utils do
   def container_label_for_scheduling_type([:read_by]), do: "Read by: "
   def container_label_for_scheduling_type(_), do: "Due by: "
 
-  def label_for_scheduling_type(:due_by), do: "Due by: "
+  def label_for_scheduling_type(type) when type in [:due_by, "Not yet scheduled"], do: "Due by: "
```
gastonabella: I find it a bit odd that this expects a string like 'Not yet scheduled', given that (based on the function name) it should expect a scheduling type, which is an atom.

simonchoxx: I plan to change this so that instead of waiting for a string, it will wait for a nil which would mean that there is no schedule for that resource. WDYT?

gastonabella (approved):
LGTM!

PR #5745, opened 2025-07-22
[BUG FIX] [MER-4790] Wrong container numbering in "Add Materials"

manelli:
Use AuthoringResolver for displaying hierarchy when adding materials when remixing a section so it displays the correct numbering.

https://github.com/user-attachments/assets/db2c4db8-1619-4699-afd6-875310260dd8

See: https://eliterate.atlassian.net/issues/MER-4790


gastonabella (approved):
Looks good!

PR #5746, opened 2025-07-22
[ENHANCEMENT] [MER-3090] Implements email validation in compliance with RFC 3696

rgachuhi:
This PR enforces RFC 3696 Section 3-compliant email validation across the system to ensure consistency and robustness.

## Scope

- Validate emails against RFC 3696 Section 3
- Enforce validation universally across the system: new user creation, user profile updates, and admin edits
- Apply validation on both frontend and backend

## Technical Notes

- Centralized the validation logic for consistent enforcement.
- Implemented validation on both frontend and backend to prevent bypass.
- Standardized error messaging across all interfaces.


lib/oli/accounts/schemas/author.ex
```diff
@@ -116,8 +116,13 @@ defmodule Oli.Accounts.Author do
...
     |> validate_required([:email])
-    |> validate_format(:email, ~r/^[^\s]+@[^\s]+$/, message: "must have the @ sign and no spaces")
-    |> validate_length(:email, max: 160)
+    |> validate_change(:email, fn :email, email ->
+      if Oli.Utils.validate_email(email) do
+        []
+      else
+        [email: "must be a valid email address"]
+      end
+    end)
```
gastonabella: Nit, but you could define this as a helper function in `Oli.Utils` or `Oli.Accounts` (something like `validate_email`) and reuse twice in this module and in the user's schema module.

lib/oli/utils.ex
```diff
@@ -562,4 +562,151 @@ defmodule Oli.Utils do
...
+    # Split email into local and domain parts
+    case String.split(email, "@") do
+      [local_part, domain] ->
+        with true <- String.length(email) < 255,
+             true <- validate_local_part(local_part),
+             true <- validate_domain(domain) do
+          true
+        else
+          _ -> false
+        end
```
gastonabella: What if we do instead:
```elixir
String.length(email) < 255 and validate_local_part(local_part) and validate_domain(domain)
```

I'd use `with` for more complex validations, and here we could use just `and`s :)

lib/oli/utils.ex
```diff
@@ -562,4 +562,151 @@ defmodule Oli.Utils do
...
+
+      _ ->
+        false
+    end
+  end
+
+  def validate_email(_), do: false
+
+  # Validates the local part of the email (before @)
+  defp validate_local_part(local) do
```
gastonabella: This works great for almost all real-world cases! 👍
Just a heads up: there are a few valid RFC examples (like `"Abc@def"@example.com`, `Abc\@def@example.com`, or `"Fred Bloggs"@example.com`) that get rejected by this logic, even though they’re technically allowed. Not a blocker, but worth noting if strict RFC compliance is needed.

lib/oli/utils.ex
```diff
@@ -562,4 +562,151 @@ defmodule Oli.Utils do
...
+    # - Must only contain valid characters
+    String.length(local) < 65 and
+      not String.starts_with?(local, ".") and
+      not String.ends_with?(local, ".") and
+      not String.contains?(local, "..") and
+      String.match?(local, ~r/^[a-zA-Z0-9!#$%&'*+\-\/=?^_`{|}~.]+$/)
+  end
+
+  # Validates the domain part of the email (after @)
+  defp validate_domain(domain) do
```
gastonabella, line 649: I tested this validation and it works perfectly! I just have a suggestion that could improve readability. It might be a good opportunity to use a `with` clause here, which would also help avoid so many nested ifs and cases that make the logic a bit harder to follow :)

```elixir
defp validate_domain(domain) do
  with true <- String.length(domain) <= 255, # Domain must not exceed 255 characters
       parts when length(parts) > 1 <- String.split(domain, "."), # Must have at least one dot and a valid TLD
       true <- valid_tld?(List.last(parts)), # TLD must be at least 2 characters
       ascii_domain <- try_idna_encode(domain), # Convert to punycode for IDN support
       true <- valid_labels?(String.split(ascii_domain, ".")) # Each part must be 63 characters or less, valid chars, no leading/trailing hyphens
  do
    true
  else
    _ -> false
  end
end

defp valid_tld?(tld), do: String.length(tld) >= 2

defp valid_labels?(labels) do
  Enum.all?(labels, fn label ->
    # Each part must be 63 characters or less
    String.length(label) <= 63 and
      # Can't start or end with hyphen
      not String.starts_with?(label, "-") and
      not String.ends_with?(label, "-") and
      # Must only contain letters, numbers, and hyphens
      String.match?(label, ~r/^[a-zA-Z0-9\-]+$/)
  end)
end
```

gastonabella (commented):
Amazing work! I just left a few comments and suggestions. Let me know what you think 🚀

rgachuhi:
> Amazing work! I just left a few comments and suggestions. Let me know what you think 🚀

All good suggestions. I will make the updates.

lib/oli/accounts/schemas/author.ex
rgachuhi: Function added to Oli.Accounts

lib/oli/utils.ex
rgachuhi: updated

lib/oli/utils.ex
rgachuhi: Support added except for Abc\@def@example.com which is not valid without quotations around "Abc\@def"

PR #5749, opened 2025-07-23
[ENHANCEMENT] [MER-4078] Designer or TA Role in Canvas Does Not Work

gastonabella:
[MER-4078](https://eliterate.atlassian.net/browse/MER-4078)

### Summary:

This PR updates Torus to treat users with the Canvas "Designer" role (represented as ContentDeveloper in LTI) as instructors throughout the platform. Previously, Designers were incorrectly mapped to students, preventing them from configuring or managing course sections. With this change, both Instructors and Designers are recognized as instructors in the system.

- The LTI role ContentDeveloper is now included wherever instructor roles are checked (e.g., enrollments, permissions, filters, and UI helpers).
- All queries and helper functions that previously only considered context_instructor now also include context_content_developer.

💡 **Note:** The TA role was already mapped to instructor in Torus, so no changes were required for TAs.

https://github.com/user-attachments/assets/e10ca7f1-caf5-4ca1-8c78-34357c021a60

[MER-4078]: https://eliterate.atlassian.net/browse/MER-4078?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli/delivery/sections.ex
+++ b/lib/oli/delivery/sections.ex
@@ -64,6 +64,12 @@ defmodule Oli.Delivery.Sections do
   @instructor_context_role_id ContextRoles.get_role(:context_instructor).id
   @student_role_id ContextRoles.get_role(:context_learner).id
 
+  @instructor_roles [
+    ContextRoles.get_role(:context_instructor),
+    ContextRoles.get_role(:context_content_developer)
+  ]
+
+  @instructor_role_ids Enum.map(@instructor_roles, & &1.id)
   @doc """
   Fetches the hidden instructor for a given section. If no hidden instructor exists,
   it creates one.  The "hidden" instructor is a special user account that is used to
@@ -189,8 +195,6 @@ defmodule Oli.Delivery.Sections do
         %Sorting{field: field, direction: direction},
         %EnrollmentBrowseOptions{} = options
       ) do
-    instructor_role_id = ContextRoles.get_role(:context_instructor).id
-
     filter_by_role =
       case options do
         %EnrollmentBrowseOptions{is_instructor: true, is_student: true} ->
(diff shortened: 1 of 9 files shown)


lib/oli/delivery/sections.ex
```diff
@@ -5587,4 +5586,27 @@ defmodule Oli.Delivery.Sections do
...
+  @doc """
+  Returns true if the list of roles contains any instructor role
+  """
+  @spec contains_instructor_role?(list(%Lti_1p3.DataProviders.EctoProvider.ContextRole{})) ::
+          boolean()
+  def contains_instructor_role?(roles) do
+    Enum.any?(@instructor_roles, fn r ->
+      ContextRoles.contains_role?(roles, r)
+    end)
+  end
```
simonchoxx, line 5611: I have a question here. Why do we create this new function instead of using `ContextRoles.has_roles?/4`? Do we avoid any database query or it has some other purpose?

test/oli/sections_test.exs
```diff
@@ -2145,4 +2154,65 @@ defmodule Oli.SectionsTest do
...
+      Sections.enroll(user.id, section.id, [ContextRoles.get_role(:context_content_developer)])
+
+      result = Sections.instructors_per_section([section.id])
+      assert Map.has_key?(result, section.id)
+      assert Enum.any?(result[section.id], &(&1 == "Content Dev"))
+    end
+  end
+
+  describe "contains_instructor_role?/1" do
+    alias Lti_1p3.Roles.ContextRoles
```
simonchoxx: I think we should remove this line as this alias is already defined along with the other aliases.

gastonabella: Good catch! It's removed now

lib/oli/delivery/sections.ex
gastonabella: Good question! Exactly, this function is for cases where we already have the roles in memory, so we avoid an extra, unnecessary query.

simonchoxx (approved):
Great work 🚀🚀

PR #5751, opened 2025-07-25
[FEATURE] [MER-4563] Advanced activities and lti tool selector

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-4563) to the ticket

## Overview
Implements a new modal-based interface for managing advanced activities and LTI tools in project settings, replacing the previous approach where all tools were listed directly in the overview page.
Now the user can search for tools through the modal, select the ones that wants to add to the project. Then, after being added, the user may disable the tool.

https://github.com/user-attachments/assets/d49bd574-4c5f-48b3-9a0f-ac0f160cdbdd

### Architecture
- Removed ProjectActivityController - changes (enable/disable) are now handled directly from the LiveView, re-rendering changes immediately instead of having to fully reload the page.
- Added AddActivitiesAndToolsModal LiveComponent for activity/tool selection
### Database
- Added status column to activity_registration_projects table. This new field tracks if the tool, once selected and added, is enabled or disabled.

### Features
- Updated project overview page with "Add Activities and Tools" button (instead of listing all tools)
- Modal with "Advanced Activities" and "External Tools" tabs
- Bulk add/remove operations with search functionality
- Visual status indicators and change summaries -> While not originally requested in the ticket, visual status indicators and change summaries were added to improve user clarity. This addresses scenarios where users select activities from one tab, switch to another tab to select tools, and need clear visibility into what changes will be applied when clicking "Apply". Without these indicators, it was unclear which selections from different tabs would be included in the final operation.


simonchoxx (approved):
Fantastic work 🚀🚀

I tested it locally and everything works as expected!

Great to see so many tests included!

PR #5753, opened 2025-07-25
[FEATURE] [MER-0000] New Generative AI infrastructure

darrensiegel:
Pending design description.


priv/repo/migrations/20250721171203_genai_infra.exs
```diff
@@ -0,0 +1,102 @@
...
+    if anthropic_api_key do
+      execute("""
+      INSERT INTO registered_models
+        (name, provider, model, url_template, api_key,
+        secondary_api_key, inserted_at, updated_at)
+      VALUES
+        ('claude', 'claude', 'claude-3-haiku-20240307', 'https://api.anthropic.com/v1/messages',
+        '#{anthropic_api_key}', NULL, NOW(), NOW());
+      """)
+    end
```
nicocirio: NIT: use parametrized queries to avoid sql injection (this is almost impossible since we are the ones that will provide those keys...)

```suggestion
    if anthropic_api_key do
      execute(fn ->
        repo().query!(
          """
          INSERT INTO registered_models
            (name, provider, model, url_template, api_key,
            secondary_api_key, inserted_at, updated_at)
          VALUES
            ('claude', 'claude', 'claude-3-haiku-20240307', 'https://api.anthropic.com/v1/messages',
            $1, NULL, NOW(), NOW());
          """,
          [anthropic_api_key]
        )
      end)
    end
```

priv/repo/migrations/20250721171203_genai_infra.exs
```diff
@@ -0,0 +1,102 @@
...
+    if open_ai_key do
+      execute("""
+      INSERT INTO registered_models
+        (name, provider, model, url_template, api_key,
+        secondary_api_key, inserted_at, updated_at)
+      VALUES
+        ('openai-gpt4', 'open_ai', 'gpt-4-1106-preview', 'https://api.openai.com',
+        '#{open_ai_key}', '#{open_ai_org_key}', NOW(), NOW());
+      """)
+    end
```
nicocirio: NIT: use parametrized queries to avoid SQL injection (this is almost impossible since we are the ones that will provide those keys...)
```suggestion
    if open_ai_key do
      execute(fn ->
        repo().query!(
          """
          INSERT INTO registered_models
            (name, provider, model, url_template, api_key,
            secondary_api_key, inserted_at, updated_at)
          VALUES
            ('openai-gpt4', 'open_ai', 'gpt-4-1106-preview', 'https://api.openai.com',
            $1, $2, NOW(), NOW());
          """,
          [open_ai_key, open_ai_org_key]
        )
      end)
    end
```

priv/repo/migrations/20250721171203_genai_infra.exs
```diff
@@ -0,0 +1,102 @@
...
+      add :section_id, references(:sections, on_delete: :nothing)
+
+      timestamps(type: :timestamptz)
+    end
+
+    create index(:gen_ai_feature_configs, [:section_id, :feature], unique: true)
+
+    open_ai_key = System.get_env("OPENAI_API_KEY")
+    open_ai_org_key = System.get_env("OPENAI_ORG_KEY")
+    anthropic_api_key = System.get_env("ANTHROPIC_API_KEY")
```
nicocirio: Shouldn't we encrypt these api keys before storing them in the DB?

lib/oli/gen_ai/completions/function.ex
```diff
@@ -0,0 +1,75 @@
...
+        if Code.ensure_loaded?(module) do
+          {:ok, module, name}
+        else
+          {:error, :invalid_function_name}
+        end
+
+      _ ->
+        {:error, :invalid_function_name}
+    end
+  end
```
nicocirio: The following early return will do nothing without the `case` within the `else` statement: 

```
  if full_name == nil do
      {:error, :invalid_function_name}
  end
```

We could instead use a `with` statement:

```suggestion
  defp verify_module_and_function(available_functions, name) do
    as_map =
      Enum.reduce(available_functions, %{}, fn function, acc ->
        Map.put(acc, function.name, function)
      end)

    with {:ok, full_name} <- get_full_name(as_map, name),
         {:ok, parts} <- split_function_name(full_name),
         {:ok, module, function_name} <- parse_module_and_function(parts),
         {:ok, _} <- ensure_module_loaded(module) do
      {:ok, module, function_name}
    end
  end

  defp get_full_name(as_map, name) do
    case Map.get(as_map, name) do
      %{full_name: full_name} when not is_nil(full_name) -> {:ok, full_name}
      _ -> {:error, :invalid_function_name}
    end
  end

  defp split_function_name(full_name) do
    case String.split(full_name, ".") do
      parts when is_list(parts) and length(parts) > 1 -> {:ok, parts}
      _ -> {:error, :invalid_function_name}
    end
  end

  defp parse_module_and_function(parts) do
    module_parts = Enum.take(parts, Enum.count(parts) - 1)
    function_name = Enum.at(parts, -1) |> String.to_existing_atom()

    module =
      Enum.join(module_parts, ".")
      |> String.to_existing_atom()

    {:ok, module, function_name}
  rescue
    # handles the case where String.to_existing_atom() fails
    ArgumentError -> {:error, :invalid_function_name}
  end

  defp ensure_module_loaded(module) do
    if Code.ensure_loaded?(module) do
      {:ok, module}
    else
      {:error, :invalid_function_name}
    end
  end
```

lib/oli_web/live/dialogue/window_live.ex
```diff
@@ -87,26 +92,32 @@ defmodule OliWeb.Dialogue.WindowLive do
...
+             allow_submission?: true,
+             trigger_queue: [],
+             active_message: nil,
+             title: "Dot",
+             current_user: Oli.Accounts.get_user!(current_user_id),
+             height: 500,
+             width: 400,
+             section: section,
+             resource_id: session["resource_id"],
+             is_page: session["is_page"] == true || false
```
nicocirio: NIT
```suggestion
             is_page: session["is_page"] == true
```

lib/oli/gen_ai/completions.ex
```diff
@@ -0,0 +1,30 @@
...
+  Synchronous and asynchronous chat completion are supported via the
+  `generate` and `stream` functions.
+  """
+
+  alias Oli.GenAI.Completions.RegisteredModel
+
+  def generate(messages, functions, %RegisteredModel{} = registered_model) do
+    get_provider(registered_model)
+    |> apply(:generate, [messages, functions, registered_model])
+  end
```
nicocirio, line 16: Just curious, what will be our use case for this `generate/3` sync function? I understand DOT uses the `stream/4` async flavour

lib/oli/gen_ai/completions/open_ai_compliant_provider.ex
```diff
@@ -0,0 +1,267 @@
...
+      [%{"delta" => %{"content" => content}}] ->
+        {:tokens_received, content}
+
+      _ ->
+        {:error}
+    end
+  end
+
+  def encode_messages(messages) do
+    # Delete the id and input from messages and open ai does not require these
```
nicocirio: NIT
```suggestion
    # Delete the id and input from messages as open ai does not require these
```

lib/oli/gen_ai/completions/open_ai_compliant_provider.ex
```diff
@@ -0,0 +1,267 @@
...
+        body
+    end
+  end
+
+  def handle_response(httpoison_response) do
+    case httpoison_response do
+      {:ok, %HTTPoison.Response{status_code: 200, body: {:ok, body}}} ->
+        res =
+          body
+          |> Enum.map(fn {k, v} -> {String.to_atom(k), v} end)
```
nicocirio, line 136: Using `String.to_atom/1` in `handle_response/1` may lead to atom-table overflow and creates a security vulnerability. In general, we use string keys for external data, so we should avoid this conversion and let the caller (who knows the expected keys) handle the transformation if needed.

**Note**: This `handle_response` function is only used for synchronous requests via `Oli.GenAI.Completions.OpenAICompliantProvider.generate/3` → `api_post/3` → `handle_response/1`, and I understand this flow is not currently being used in the codebase.

priv/repo/migrations/20250721171203_genai_infra.exs
```diff
@@ -0,0 +1,102 @@
+defmodule Oli.Repo.Migrations.GenaiInfra do
+  use Ecto.Migration
+
+  def change do
```
nicocirio: This migration can't be rolled back (`mix ecto.rollback`). To do so, we can change `def change do` here to be `def up do`, and then define:

```
 def down do
    drop table(:gen_ai_feature_configs)
    drop table(:completions_service_configs)
    drop table(:registered_models)
 end
  ```

**Review by nicocirio — COMMENTED** [review]
Excellent feature! I tested it locally with an OpenAI key and with no provider, and it worked as expected (I do not have a Claude key to test it).
There are some security aspects we might want to improve (for example, encrypting API keys in the database, or avoiding using String.to_atom).

File: lib/oli/gen_ai/completions.ex (hunk 6 above)
Comment (darrensiegel) [reply lib/oli/gen_ai/completions.ex:16]: Any future "agent" feature will use the synchronous generate.

File: lib/oli/vault.ex [hunk 10]
```diff
@@ -0,0 +1,19 @@
...
+      )
+
+    {:ok, config}
+  end
+
+  defp decode_env!(var) do
+    var
+    |> System.get_env()
+    |> Base.decode64!()
+  end
```
nicocirio, line 18: We should take into account that if we forget to set the `CLOAK_VAULT_KEY` var the app will crash when we pass `nil` to `Base.decode64!`.

PR #5757, opened 2025-07-28
[ENHANCEMENT] [MER-4745] Allow soft deletions of LTI platform instances

manelli:
Now the 'Delete Tool' button on `http://localhost/admin/external_tools/:tool_id/details` soft deletes the activity deployment and the platform instance.

See: https://eliterate.atlassian.net/browse/MER-4745

### DIFF
--- a/test/oli/lti/platform_instances_test.exs
+++ b/test/oli/lti/platform_instances_test.exs
@@ -153,20 +153,22 @@ defmodule Oli.Lti.PlatformInstancesTest do
       active_instance =
         platform_instance_fixture(%{client_id: "same_client_id", status: :active})
 
-      # Create second platform instance with deleted status and same client_id - should succeed
-      deleted_instance =
+      # Create other platform instances with deleted status and same client_id - should succeed
+      deleted_instance1 =
         platform_instance_fixture(%{client_id: "same_client_id", status: :deleted})
 
-      # Verify both instances exist
-      assert active_instance.id != deleted_instance.id
-      assert active_instance.client_id == deleted_instance.client_id
-      assert active_instance.status == :active
-      assert deleted_instance.status == :deleted
+      deleted_instance2 =
+        platform_instance_fixture(%{client_id: "same_client_id", status: :deleted})
 
-      # Try to create another active instance with same client_id - should fail due to unique constraint
-      assert_raise Ecto.ConstraintError, fn ->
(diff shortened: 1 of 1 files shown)


test/oli/lti/platform_instances_test.exs
```diff
@@ -143,5 +147,129 @@ defmodule Oli.Lti.PlatformInstancesTest do
...
+               )
+
+      assert reenabled_instance.status == :active
+      assert reenabled_instance.client_id == "test_client_id"
+      assert reenabled_instance.name == "Test Platform"
+
+      # Verify the re-enabled instance is properly active
+      final_instance = PlatformInstances.get_platform_instance!(reenabled_instance.id)
+      assert final_instance.status == :active
+      assert final_instance.id == platform_instance.id
```
nicocirio: Although we can re-enable a PI in the test, I believe we do not have a way to do this through the UI. Actually, when we delete a PI we warn the user that this action can not be undone. So I would recommend removing this part of the test to avoid confusion on how the interface should work, wdyt?

test/oli/lti/platform_instances_test.exs
```diff
@@ -143,5 +147,129 @@ defmodule Oli.Lti.PlatformInstancesTest do
...
+      assert active_instance.id != deleted_instance.id
+      assert active_instance.client_id == deleted_instance.client_id
+      assert active_instance.status == :active
+      assert deleted_instance.status == :deleted
+
+      # Try to create another active instance with same client_id - should fail due to unique constraint
+      assert_raise Ecto.ConstraintError, fn ->
+        platform_instance_fixture(%{client_id: "same_client_id", status: :active})
+      end
+    end
```
nicocirio: NIT: We are not testing what the test description says. I believe we should add a third PI with a `deleted` status and then verify all exist to honor the "allows multiple deleted" part of the description

test/oli/lti/platform_instances_test.exs
manelli: 👍 Fixed in 55ce0af3b60b009c84295d50b1b6693505993388

test/oli/lti/platform_instances_test.exs
manelli: 👍 Fixed in 55ce0af3b60b009c84295d50b1b6693505993388

PR #5758, opened 2025-07-29
[BUG FIX] [MER-4789] Resources to add in "Customize Content" are out of order

rgachuhi:
This PR ensures that all curriculum content displayed during the course content remixing process (Manage → Customize Content → Add Materials) is ordered consistently with the authoring view.

**Changes Made**

- Updated the ordering logic for containers (units, modules) and pages in the "Add Materials" flow
- Ensured that hierarchical navigation (unit → module → page) preserves authoring order
- Fixed the "All Pages" view to match the default ordering from the authoring interface


lib/oli_web/live/delivery/remix_section.ex
```diff
@@ -1047,6 +1047,8 @@ defmodule OliWeb.Delivery.RemixSection do
 
   defp transform_section_pages(section_pages) do
     section_pages
+    # Sort by title to ensure it matches the default order in the "all pages" authoring view
+    |> Enum.sort_by(fn rev -> rev.title end, :asc)
```
simonchoxx: I believe this would not be necessary and would even break with the logic of the sorting, since the database query we perform to obtain the data already sorts it there, and this `Enum.sort_by` would override the sorting we already have.
I believe it is not necessary here and we should remove it.

test/oli_web/live/remix_section_test.exs
```diff
@@ -1249,13 +1224,13 @@ defmodule OliWeb.RemixSectionLiveTest do
       assert view
              |> has_element?(
                ".remix_materials_table tbody tr:first-of-type td:nth-of-type(2)",
-               "Elixir Page"
+               "Another orph. Page"
```
simonchoxx, line 1228: With the suggested changes I think this should go back to how it was before
This should be the `last-of-type`.

test/oli_web/live/remix_section_test.exs
```diff
@@ -1249,13 +1224,13 @@ defmodule OliWeb.RemixSectionLiveTest do
...
              )
 
       assert view
              |> has_element?(
                ".remix_materials_table tr:last-of-type td:nth-of-type(2)",
-               "Another orph. Page"
+               "Elixir Page"
```
simonchoxx, line 1234: With the suggested changes I think this should go back to how it was before
This should be the `first-of-type`.

simonchoxx (changes requested):
I'll leave you a couple of comments to review and let me know what you think. 

We should also perhaps add these two keys to the params map we built in the init_state function:
```

sort_by: :title,
sort_order: :asc

```
Try it out if you want and let me know what you think and if you think these changes meet the acceptance criteria.

lib/oli_web/live/delivery/remix_section.ex
rgachuhi: Makes sense, since the "all pages" table allows user to change the sorting order. I will remove this. I don't think it is necessary for the all pages to match what is listed in the remix

rgachuhi: I have updated the param map adding  sort_by: :title and sort_order: :asc. Also removed the sort_by function

test/oli_web/live/remix_section_test.exs
rgachuhi: Updated accordingly

test/oli_web/live/remix_section_test.exs
rgachuhi: Updated accordingly

simonchoxx (approved):
LGTM 🚀 

Nice work!

PR #5763, opened 2025-07-30
[ENHANCEMENT] [MER-4597] tech debt singularize course section creation code

eliknebel:
https://eliterate.atlassian.net/browse/MER-4597

This PR singularizes the course section implementation for all source types (e.g. project, product) and delivery types (e.g. direct, lti). It does this by adding a new SectionSpecification struct and maintaining a changeset-driven approach to section creation in order to facilitate incremental changes by the course builder wizard.

It also handles the case where an LTI section could be accidentally configured in parallel by two different users by only proceeding with section creation if an LTI section does not already exist for a given context.


lib/oli_web/router.ex
```diff
@@ -1531,11 +1546,9 @@ defmodule OliWeb.Router do
...
   end
 
-  scope "/course", OliWeb do
-    pipe_through([:browser, :delivery_protected])
-
-    live("/select_project", Delivery.NewCourse, :lms_instructor, as: :select_source)
-  end
+  ###
+  # Delivery
+  ###
```
nicocirio, line 1551: NIT: we may want to remove this comments

lib/oli_web/delivery_web.ex
```diff
@@ -0,0 +1,80 @@
...
+  @allow_configure_section_roles [
+    PlatformRoles.get_role(:system_administrator),
+    PlatformRoles.get_role(:institution_administrator),
+    ContextRoles.get_role(:context_administrator),
+    ContextRoles.get_role(:context_instructor)
+  ]
+
+  @roles_claims "https://purl.imsglobal.org/spec/lti/claim/roles"
+
+  def redirect_user(conn, opts \\ []) do
```
nicocirio, line 22: It would be great to test this function, wdyt?

nicocirio (approved):
Awesome work!

PR #5767, opened 2025-07-31
[BUG FIX] [MER-4820] Handle old "submit-and-compare" type question models w/o catchall

andersweinstein:
Some existing short-answer text questions aiming to approximate legacy "submit and compare" type behavior (no wrong answer, student just submits answer for purpose of reflectively comparing to a correct answer ) wound up including only a correct response with regex rule .* but no catchall response. However, torus authoring UI expects to find both and a recent change* now causes these questions to crash on the interface, rendering them uneditable, due to exception from getIncorrectResponse.

*(The change was to include a check for score of 0 when finding incorrect response, rather than just looking for pattern of .* The change was put in to ALLOW for authors to set a correct response of .* which previously caused problems. )

This fixes by using authoring's just-in-time model migration facility to ensure that a catchAll exists on short Answer questions, adding one if it is not found. If a correct response is defined to be .* it will always pre-empt the catch all on evaluation because correct response rules are always tried first (they are supposed to be first in the response list). But authoring will be satisfied and it is possible to change the correct response rule (for example, to .+ to require a non-empty response) while keeping the catchall.

### DIFF
--- a/assets/src/components/activities/short_answer/ShortAnswerAuthoring.tsx
+++ b/assets/src/components/activities/short_answer/ShortAnswerAuthoring.tsx
@@ -14,16 +14,23 @@ import { InputEntry } from 'components/activities/short_answer/sections/InputEnt
 import { getTargetedResponses, shortAnswerOptions } from 'components/activities/short_answer/utils';
 import {
   GradingApproach,
+  HasParts,
   Manifest,
   Response,
   RichText,
   makeResponse,
 } from 'components/activities/types';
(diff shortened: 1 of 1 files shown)


eliknebel:
Tested with this recent export of real chem and was able to reproduce the issue on the hotfix branch. I then switched to this branch and verified it fixes the authoring issue. Finally, I committed the migration changes by editing the activities to commit the migrated content, published and tested the submit and compares as a student and everything worked as expected. Changes look good, nice work.

eliknebel (approved):

PR #5770, opened 2025-07-31
[ENHANCEMENT] [MER 4767] Add missing components tests

manelli:
Part of https://eliterate.atlassian.net/browse/MER-4767


gastonabella (approved):
Great work! 🚀 
Our codebase now has `70.8%` test coverage  📈
I've added a comment to the Jira ticket as well, but since you have better insight into which tests are missing, would it make sense to create a follow-up ticket to continue this effort? We could detail which other files would be good candidates for additional test coverage.

PR #5775, opened 2025-08-04
[ENHANCEMENT] [MER-4732] Update to latest version of Phoenix and LiveView - part 1

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-4732) to the ticket

Phoenix was not updated since we are already using the latest stable version (1.7.21). There are some v1.8, but they are currently release candidates, so we should wait for the stable version before upgrading.
##### 08/07/25 update: [v1.8 was just released](https://x.com/chris_mccord/status/1952839142993252736), so the update to that version is handled in [this PR](https://github.com/Simon-Initiative/oli-torus/pull/5783). ####

This first PR updates LiveView to v1.0. 
Last week, while I was updating LiveView to v1.0, the v1.1 was released. So, the update to v1.1 is handled in [this PR](https://github.com/Simon-Initiative/oli-torus/pull/5785)

## ⚠️ Important ⚠️ 
This PR can be reviewed and approved, but please **do not merge** until the pending tasks are merged into the current branch.

Pending tasks
- [x] Fix all duplicate-id warnings while running tests -> done in [this PR](https://github.com/Simon-Initiative/oli-torus/pull/5781) to avoid making the current PR's size even bigger. 
- [x] Run `mix format` -> this was done on [this PR](https://github.com/Simon-Initiative/oli-torus/pull/5776) since all template interpolations `<%= %>` will be updated to the latest interpolation `{ }`, and about 350 files will be changed. To avoid having a huge PR that will be difficult to review, I split those changes into a new PR (that PR re-enables auto format [here](https://github.com/Simon-Initiative/oli-torus/pull/5775/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R124-R126) and [here](https://github.com/Simon-Initiative/oli-torus/pull/5775/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R125-R127))
- [x] Merge part 5

### DIFF
--- a/lib/oli_web/live/delivery/student/utils.ex
+++ b/lib/oli_web/live/delivery/student/utils.ex
@@ -1035,8 +1035,7 @@ defmodule OliWeb.Delivery.Student.Utils do
           </p>
 
           <p class="mt-4">
-            If you think this is an error or would like more information, please
-            <OliWeb.Components.Common.tech_support_link
+            If you think this is an error or would like more information, please <OliWeb.Components.Common.tech_support_link
               id="tech_support_lti_error"
               class="text-[#006CD9] hover:text-[#1B67B2] dark:text-[#4CA6FF] dark:hover:text-[#99CCFF] hover:underline text-base font-['Open Sans'] tracking-tight cursor-pointer"
             >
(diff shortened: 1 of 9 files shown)


lib/oli_web/live/user_settings_live.ex
```diff
@@ -8,6 +8,11 @@ defmodule OliWeb.UserSettingsLive do
   alias Oli.AssentAuth.UserAssentAuth
   alias OliWeb.Common.Properties.{Groups, Group}
 
+  ### TODO lunes:
```
gastonabella: Looks like these comments should be removed 😅

PR #5779, opened 2025-08-05
[ENHANCEMENT] [TQA-10] [TQA-11] [TQA-12] Test scripts are added

lperezCES:
https://eliterate.atlassian.net/browse/TQA-10
https://eliterate.atlassian.net/browse/TQA-11
https://eliterate.atlassian.net/browse/TQA-12


lperezCES:
https://eliterate.atlassian.net/browse/TQA-13

assets/automation/README.md
```diff
@@ -61,3 +61,57 @@ Withd this plugin we can:
...
+## Pre-conditions for `user-accounts` suite
+
+To successfully run the `user-accounts` test suite, make sure the following conditions are met:
+
+### ✅ General requirements for `user-accounts`
+
+- All required environment variables listed in the `.env` file must be set (see [Required environment variables](#required-environment-variables)).
+- Playwright must be installed and configured.
+- The `loginData` used in this suite must be correctly set up with valid credentials for each role.
+
```
andersweinstein: I don't think it is necessary to repeat that the environment variables are required. This has already been said; there is no need to repeat that. 

It is also not necessary to say again that Playwright must be installed and configured. 

And the loginData gets setup automatically from the environment variables, right?

So I don't think we don't need these last two sections (both Pre-conditions for user accounts suite and General requirements for user accounts) at all. These are not giving new steps someone has to perform to run the tests.

assets/automation/README.md
```diff
@@ -61,3 +61,57 @@ Withd this plugin we can:
...
+## Pre-conditions for `course-authoring` suite
+
+To successfully run the `course-authoring` test suite, make sure the following conditions are met:
+
+### ✅ General requirements for `course-authoring`
+
+- All required environment variables listed in the `.env` file must be set (see [Required environment variables](#required-environment-variables)).
+- The test data used in this suite assumes that specific projects already exist on the platform with fixed IDs
+
```
andersweinstein: Same comments as above, I don't think we need these last two sections

assets/automation/README.md
```diff
@@ -61,3 +61,57 @@ Withd this plugin we can:
   |--> resources
   |--> torus
 ```
+
```
andersweinstein: I think instructions for how to setup the system for a test should come first. It should mention which accounts and projects one has to create.

assets/automation/README.md
```diff
@@ -61,3 +61,57 @@ Withd this plugin we can:
...
+### 🧪 Test data setup
+
+Before running this suite, verify that:
+
+- The **author** user has access to projects with the following IDs:
+  - `tqa10automation`
+  - `tqa11automation`
+  - `tqa12automation` must have `img-mock-05-16-2025.jpg`
+  - `tqa13automation` must have `img-mock-05-16-2025.jpg` `audio-test-01.mp3` `video-test-01.mp4`
```
andersweinstein, line 79: I think it would be clearer to say `must have an image file named 'img-mock-05-16-2025.jpg' uploaded into its media library. It does not matter what image file is used or where it exists in the course` Similarly for the other examples, though you don't need to repeat the second sentence.

[Note:  each project has its own "media library", although all images in all torus courses are stored within the underlying Amazon S3 storage. However, only the items in a project's media library are available as choices in the Media Insertion modal dialog for that project.

assets/automation/README.md
```diff
@@ -61,3 +61,57 @@ Withd this plugin we can:
...
+
+To successfully run the `course-authoring` test suite, make sure the following conditions are met:
+
+### ✅ General requirements for `course-authoring`
+
+- All required environment variables listed in the `.env` file must be set (see [Required environment variables](#required-environment-variables)).
+- The test data used in this suite assumes that specific projects already exist on the platform with fixed IDs
+
+### 🧪 Test data setup
+
```
andersweinstein, line 72: As mentioned above, I think this should come first because it is step 1, the first thing you have to do (after installing dependencies) in order to run the tests.

andersweinstein (approved):

PR #5786, opened 2025-08-08
[FEATURE] [MER-4765] Implement color tokens

simonchoxx:
[MER-4765](https://eliterate.atlassian.net/browse/MER-4765)

This PR introduces the initial implementation of design tokens to define color values based on the Figma design system.

The tokens are defined in the `tailwind.tokens.js` file, where each token includes both its light and dark mode variants. If you need to add or update a token, this file should be modified directly.

In this PR, I started replacing hardcoded color values in tailwind classes with the newly defined semantic tokens, as part of the effort to move toward a more consistent and scalable design system.

Additionally, I added an internal page at [/dev/tokens](http://localhost/dev/tokens), which provides an easy way to browse available tokens and copy their names or hexadecimal values.

[MER-4765]: https://eliterate.atlassian.net/browse/MER-4765?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/assets/css/table.css
+++ b/assets/css/table.css
@@ -3,7 +3,7 @@
 }
 
 .instructor_dashboard_table th {
-  @apply h-[50px] border-r-0 border-l-0 bg-white dark:bg-[#000000] text-[#353740] dark:text-[#EEEBF5] border-b border-[#CED1D9] dark:border-[#3B3740] text-left font-normal text-xs sm:text-sm relative;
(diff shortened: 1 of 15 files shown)


nicocirio (approved):
This is awesome! 

🚀 🚀 🚀 
image

PR #5787, opened 2025-08-08
[ENHANCEMENT] [MER-4816] handle use of v1 ingest by AutomationSetup facility

andersweinstein:
[Ticket](https://eliterate.atlassian.net/browse/MER-4816) We have been removing the deprecated v1 ingestion implementation in favor of the newer v2 (also called "ScalableIngest" at some point)implementation, see https://github.com/Simon-Initiative/oli-torus/pull/5755, but complete removal of v1 ingest code was blocked by its use in module Oli.AutomationSetup, the need for which was not understood. This module provides implementation for seeding torus with accounts/section/project for the purpose of automated end to end testing, then clearing them out when done. 

On investigation it appears this testing project pursued at Argos was never completed, so in principle this module could be removed without loss. HOWEVER: this facility could be useful for the newer QA Automation testing under development. So I decided to keep it for now.

This PR preserves existing uses of v1 ingest APIs (also used in some tests, e.g. export_test) by providing a (nearly) compatible implementation using the newer v2 ingest implementation, so client code does not change, removing the v1 implementation and tests that were specific to it. (*v1 had specific error checks and codes not preserved in v2). 

Search for `Ingest.process( `and `Ingest.ingest(` to see uses. Tests using this are
test/oli/interop/export_test.exs
test/oli/interop/ingest_test.exs
test/oli_web/controllers/api/automation/automation_setup_controller_test.exs

### DIFF
--- a/lib/oli/interop/ingest.ex
+++ b/lib/oli/interop/ingest.ex
@@ -1,9 +1,13 @@
 defmodule Oli.Interop.Ingest do
+  @moduledoc """
+    v1 ingest apis for non-interactive use. Now implemented using v2 ingest methods
+  """
   alias Oli.Interop.Ingest
 
-  # v1 ingest apis in use now implemented using v2 ingest methods
-
-  # ingest project from file for author
+  @doc """
+  ingest project from file for author
+  Returns {:ok, project} on success and {:error, first_error_string} on failure
+  """
   def ingest(file, as_author) do
     %{
       Ingest.State.new()
@@ -14,7 +18,10 @@ defmodule Oli.Interop.Ingest do
     |> do_ingest()
   end
 
-  # ingest project from unzipped entry list
+  @doc """
+  ingest project from unzipped entry list for author
+  Returns {:ok, project} on success and {:error, first_error_string} on failure
+  """
   def process(entries, author) do
     %{
       Ingest.State.new()
@@ -31,7 +38,7 @@ defmodule Oli.Interop.Ingest do
       |> Ingest.Preprocessor.preprocess()
       |> Ingest.Processor.process()
 
-    # v1 ingest returned {:ok, project} on success
+    # convert to v1 return
     case result do
       {:ok, %Ingest.State{project: project}} -> {:ok, project}
       {:error, %Ingest.State{errors: [err1 | _rest]}} -> {:error, err1}


lib/oli/interop/ingest.ex
```diff
@@ -1,1143 +1,41 @@
...
-  A course digest archive is a zip file containing a flat list of JSON files.
-  There are three required files in a course archive:
-  _project.json: a document containing top-level project meta data
-  _hierarchy.json: a document that specifies the course project container hierarchy
-  _media-manifest.json: a manifest listing all of the media items referenced by the digest
-  Any number of other JSON files corresponding to pages and activities can exist
-  in the digest archive.
-  Returns {:ok, project} on success and {:error, error} on failure
-  """
+  # ingest project from file for author
```
nicocirio: NIT: If we want to comment on a function's purpose, we could use `@doc`. And we could even add some `@spec`

lib/oli/interop/ingest.ex
```diff
@@ -1,1143 +1,41 @@
...
-  alias Oli.Resources.PageContent
-  alias Oli.Utils.SchemaResolver
-  alias Oli.Resources.ContentMigrator
-  alias Oli.Authoring.Course
+  alias Oli.Interop.Ingest
 
-  @project_key "_project"
-  @hierarchy_key "_hierarchy"
-  @media_key "_media-manifest"
+  # v1 ingest apis in use now implemented using v2 ingest methods
```
nicocirio: NIT: This comment could be wrapped in `@module_doc`

lib/oli/interop/ingest/processor.ex
```diff
@@ -23,6 +23,9 @@ defmodule Oli.Interop.Ingest.Processor do
     InternalActivityRefs
   }
 
+  # for prior error such as bad zip file
```
nicocirio, line 26: NIT: `@doc` can be used here

nicocirio (approved):
Nice work Anders! I agree that it is good to keep the ingest functionality for QA automation testing purposes.
I left some minor NIT comments

lib/oli/interop/ingest/processor.ex
andersweinstein: Changed all to doc comments, thanks.

PR #5790, opened 2025-08-11
[ENHANCEMENT] [MER-4835] Change the tool tip info contents of the CAPI Iframe component in advanced author

dtiwarATS:

@gastonabella, as part of this PR, I had to execute a update statement on DB ->` part_component_registrations` table. I haven't done that but looking at the migration script, I thought this is the right place to put the update script. Let me know your thoughts.

### DIFF
--- a/priv/repo/migrations/20250811145546_updateCapiPartRegistrations.exs
+++ b/priv/repo/migrations/20250811145546_updateCapiPartRegistrations.exs
@@ -1,0 +1,11 @@
+defmodule Oli.Repo.Migrations.UpdateCapiPartRegistrations do
+  use Ecto.Migration
+
+  def change do
+    execute "update part_component_registrations
+    set title='Iframe',description='A webpage displayed in an iframe, or a simulation,
+    virtual tour, or component that uses CAPI.'
+    where slug='janus_capi_iframe' and delivery_element='janus-capi-iframe'"
+    flush()
+  end
+end
--- a/priv/repo/migrations/20250811145546_updateCapiPartRegistrations.exs
+++ b/priv/repo/migrations/20250811145546_updateCapiPartRegistrations.exs
@@ -1,11 +1,24 @@
 defmodule Oli.Repo.Migrations.UpdateCapiPartRegistrations do
   use Ecto.Migration
 
-  def change do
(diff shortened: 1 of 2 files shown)


priv/repo/migrations/20250811145546_updateCapiPartRegistrations.exs
```diff
@@ -0,0 +1,11 @@
...
+  use Ecto.Migration
+
+  def change do
+    execute "update part_component_registrations
+    set title='Iframe',description='A webpage displayed in an iframe, or a simulation,
+    virtual tour, or component that uses CAPI.'
+    where slug='janus_capi_iframe' and delivery_element='janus-capi-iframe'"
+    flush()
+  end
+end
```
gastonabella: This migration is not reversible as written, meaning that if we attempt to rollback, it won't fail but it also won't restore the table to its original state. While this might not cause immediate issues, it's always a good practice to define safe migrations by adding separate `up` (executed when running the migration) and `down` (executed when rolling back) functions, especially if we know the original values.
```elixir
defmodule Oli.Repo.Migrations.UpdateCapiPartRegistrations do
  use Ecto.Migration

  def up do
    execute """
    UPDATE part_component_registrations
    SET title = 'Iframe',
        description = 'A webpage displayed in an iframe, or a simulation, virtual tour, or component that uses CAPI.'
    WHERE slug = 'janus_capi_iframe' 
      AND delivery_element = 'janus-capi-iframe'
    """
  end

  def down do
    # Replace with actual original values when known
    execute """
    UPDATE part_component_registrations
    SET title = 'Original Title',
        description = 'Original description'
    WHERE slug = 'janus_capi_iframe' 
      AND delivery_element = 'janus-capi-iframe'
    """
  end
end
```
The `down` function should be updated with the actual original title and description values once they're identified.
If we don't know the original values, I'm fine with leaving this as is.

dtiwarATS: Thanks @gastonabella. I have made the changes and updated the PR. Please look at it and if everything looks good then we can proceed with merging the PR. Thanks

gastonabella (approved):
LGTM!

PR #5792, opened 2025-08-11
[FEATURE] [MER-4813] GenAI infra Admin UI

darrensiegel:
This PR adds the minimal UI necessary for a System Admin to be able to manage the new GenAI infra.  Three new LiveViews are added that:

1. Allow creation and management of "Registered Models"
2. Allow creation and management of "Service Configs" - these are the definitions of primary and backup registered models that can then be referenced from:
3. Feature Configs - The most important UI which allows service configs to be specified for use in "Default" feature configurations or in "Course Section" specific configurations.  

With this UI, Admins will now be able to do the following things (example use cases)
1. Add support in Torus for a new model and start to use it without code change.  When GPT-6 is released, as long as it remains compliant to the OpenAI API spec, an Admin would be able to create a new "Registered Model" and then wire it into existing Service Configs.
2. Allow one or two course sections to use an "experimental model".  A course section feature config can be created which uses a service config to point to that experimental model.
3. If a managed LLM provider such as OpenAI has an extended outage, an Admin can go in and change the service configs to swap it out as the Primary Registered Model - thus avoiding the runtime penalty for a "Just in Time" fallback.


lib/oli/gen_ai.ex
```diff
@@ -0,0 +1,109 @@
+defmodule Oli.GenAI do
+  alias Oli.GenAI.Completions.{RegisteredModel, ServiceConfig}
+  alias Oli.GenAI.FeatureConfig
+  alias Oli.Repo
+
+  import Ecto.Query, warn: false
+
+  @doc """
+  Returns a list of all registered GenAI models.
+  """
```
gastonabella, line 15: What do you think about adding a `@doc` to every public function under this context?

lib/oli_web/live/gen_ai/registered_models_view.ex
```diff
@@ -0,0 +1,369 @@
...
+      Breadcrumb.new(%{
+        link: ~p"/admin/gen_ai/registered_models",
+        full_title: "Registered LLM Models"
+      })
+    ]
+
+  # Returns all registered models sorted by ID, which provides a stable sorting order
+  def all do
+    GenAI.registered_models()
+  end
```
gastonabella, line 368: nit: do we need to add this function if it’s not providing any additional value?

darrensiegel:
@gastonabella Take another look - I added those docs and fixed the issues with the FeatureConfigs UI.

darrensiegel:
@gastonabella I address those two additional comments.  Flash message displays and the defaults is handled correctly

PR #5793, opened 2025-08-12
[FEATURE] [MER-4800] Add timezone field to course sections

darrensiegel:
## Summary
• Implements timezone field for course sections as requested in acceptance criteria
• Adds timezone to course creation UI as required field
• Updates course management pages to display and edit timezone
• Implements timezone fallback order: Student timezone → Section timezone → Browser timezone → UTC

## Changes Made
• **Database**: Added timezone field to sections table with UTC default
• **Schema**: Updated Section model to include timezone field with validation
• **Course Creation**: Added timezone dropdown to course creation step 3
• **Course Management**: Added timezone field to section editing interface
• **Timezone Logic**: Updated SessionContext and FormatDateTime to support section timezone fallback
• **LTI Integration**: Framework ready for LTI timezone parameter handling

## Test plan
- [x] Course creation includes timezone field as required
- [x] Timezone displays on manage course page
- [x] Instructors can edit timezone in section settings
- [x] Compilation passes without errors
- [ ] Manual testing of timezone fallback logic
- [x] Verification that existing sections default to UTC
- [ ] End-to-end testing of scheduling with section timezone

🤖 Generated with [Claude Code](https://claude.ai/code)

Co-Authored-By: Claude <noreply@anthropic.com>


.claude/commands/open-feature-pr.md
```diff
@@ -0,0 +1,19 @@
+---
```
eliknebel: Do we want to include these types of files in the final PR or should we ignore the .claude directory in .gitignore?

priv/repo/migrations/20250811220833_add_timezone_to_sections.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Oli.Repo.Migrations.AddTimezoneToSections do
+  use Ecto.Migration
+
+  def change do
+    alter table(:sections) do
+      add :timezone, :string, default: "Etc/UTC"
```
eliknebel: Perhaps the default here should be nil, so that the fallback logic works as intended? Otherwise there is no good way to know whether someone explicitly set this timezone for a section or if it was just the default. I could imagine a feature where some torus deployments may want to set a system-level default timezone

.claude/commands/open-feature-pr.md
darrensiegel: We do want a set of Claude Code files to be checked in, so that we have a standard set of tooling that folks can use.  This file, though, isn't one of them. I need to delete it.

priv/repo/migrations/20250811220833_add_timezone_to_sections.exs
darrensiegel: Yes, that makes sense

darrensiegel:
@eliknebel Take another look, I've addressed your review comments

eliknebel (approved):

PR #5796, opened 2025-08-13
[BUG FIX] [MER-4809] Slider Text Component does not delete

dtiwarATS:

### DIFF
--- a/assets/src/components/parts/janus-text-slider/SliderTextAuthor.tsx
+++ b/assets/src/components/parts/janus-text-slider/SliderTextAuthor.tsx
@@ -1,5 +1,4 @@
 import React, { CSSProperties, useEffect, useRef, useState } from 'react';
-import isEqual from 'lodash/isEqual';
 import { AuthorPartComponentProps } from 'components/parts/types/parts';
 import { clone } from 'utils/common';
 import './Slider-Text.scss';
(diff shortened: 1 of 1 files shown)


assets/src/components/parts/janus-text-slider/SliderTextAuthor.tsx
```diff
@@ -29,15 +28,19 @@ const SliderTextAuthor: React.FC<AuthorPartComponentProps<SliderTextModel>> = (p
...
-    if (!isEqual(prevSliderOptionLabelsRef.current?.length, sliderOptionLabels?.length)) {
+    if (prevSliderOptionLabelsRef.current === null) {
+      prevSliderOptionLabelsRef.current = sliderOptionLabels?.length ?? 0;
+      return;
+    }
+    if (prevSliderOptionLabelsRef.current !== sliderOptionLabels?.length) {
       const modelClone = clone(model);
       //The max range will be the total text items - 1.
       modelClone.maximum = sliderOptionLabels?.length - 1;
       onSaveConfigure({ id, snapshot: modelClone });
```
dtiwarATS, line 42: The  `onSaveConfigure()` was getting called on render every time. Due to that, even though the part was deleted, when this  `onSaveConfigure()` was getting called, it use to save the part again for that screen. Made changes in code to make sure this functiion only gets called when the length of items are changed so that the correct values gets set in `Maximum `variable

PR #5799, opened 2025-08-15
[FEATURE] [MER-4857] Activity Bank Instructor Blacklisting

darrensiegel:
## Summary

  This PR introduces the ability for instructors to exclude (blacklist) specific activities from activity bank selections within course sections. This feature gives instructors granular control over which activities students can access from randomized activity pools, improving course customization and content curation capabilities.

## Key Features

  - Instructor Review Interface: New LiveView at /sections/:section_slug/page/:page_id/selection/:selection_id/review for reviewing and managing activity bank selections
  - Per-Selection Blacklisting: Activities are blacklisted at the selection level, allowing different exclusions for different activity banks within the same course
  - Smart Constraints: Prevents instructors from disabling too many activities (maintains minimum viable pool based on selection count)
  - Pagination Support: Handles large activity banks efficiently with built-in pagination

## Database Schema

  New Table: blacklisted_activities

  - section_id (bigint, FK to sections, NOT NULL)
  - activity_id (bigint, NOT NULL)
  - selection_id (string, NOT NULL)
  - timestamps

## How `ReviewActivitiesLive` Works

  The ReviewActivitiesLive LiveView (lib/oli_web/live/instructor/review_activities_live.ex) is the core UI component:

  1. Initialization Flow

  - Loads the page revision and finds the specific selection element within the content
  - Queries all matching activities from the bank using the selection's criteria
  - Fetches current blacklisted activities for the section/selection combination
  - Initializes pagination (10 activities per page by default)
  - Allows user to select and view details of activity and to toggle blacklist state
  
## Integration with Activity Provider

  The ActivityProvider module integrates blacklisting during activity realization:

  1. Blacklist Loading: When processing a page with selections, blacklisted activities are loaded from the database
  2. Selection Fulfillment: During Selection.fulfill_from_bank, blacklisted activity IDs are passed to exclude them from random selection
  3. Cascading Application: Blacklists are maintained through nested content structures (groups, surveys)
Screenshot 2025-08-15 at 11 56 52 AM

### DIFF
--- a/test/oli_web/live/instructor/review_activities_live_test.exs
+++ b/test/oli_web/live/instructor/review_activities_live_test.exs
@@ -19,7 +19,11 @@ defmodule OliWeb.Instructor.ReviewActivitiesLiveTest do
       Sections.enroll(user_id, section.id, [ContextRoles.get_role(:context_instructor)])
 
       # Test that the route exists and handles invalid page gracefully
-      result = live(conn, ~p"/sections/#{section.slug}/preview/page/invalid_page/selection/invalid_selection")
+      result =
+        live(
+          conn,
+          ~p"/sections/#{section.slug}/preview/page/invalid_page/selection/invalid_selection"
+        )
 
       # Should redirect with error flash
       assert {:error, {:redirect, %{flash: %{"error" => error_message}}}} = result
--- a/lib/oli/resources.ex
+++ b/lib/oli/resources.ex
@@ -493,7 +493,7 @@ defmodule Oli.Resources do
 
   @doc """
   Gets bank entries for a given publication ID.
-  
+
   Returns a list of maps containing resource_id, tags, objectives, and activity_type_id
   for all banked activities in the given publication.
   """


lib/oli/delivery/activity_provider.ex
```diff
@@ -237,6 +265,22 @@ defmodule Oli.Delivery.ActivityProvider do
         {:ok, %Selection{id: id} = selection} =
           decrement_for_prototypes(result, fulfillment_state.prototypes_by_selection)
 
+        # Exclude any section specific blacklisted activities
+        blacklisted_activity_ids =
+          Enum.filter(blacklisted_activities, fn ba -> ba.selection_id == id end)
+          |> Enum.map(fn ba -> ba.activity_id end)
```
nicocirio: To avoid traversing the list twice, we could use a list comprehension instead

```suggestion
         blacklisted_activity_ids =
          for ba <- blacklisted_activities, ba.selection_id == id, do: ba.activity_id
```

lib/oli_web/live/instructor/review_activities_live.ex
```diff
@@ -0,0 +1,445 @@
...
+    else
+      text
+    end
+  end
+
+  defp set_breadcrumbs(section, revision) do
+    [
+      Breadcrumb.new(%{
+        full_title: revision.title,
+        link: Routes.page_delivery_path(OliWeb.Endpoint, :page, section.slug, revision.slug)
```
nicocirio: This link should point to the preview of the page
```suggestion
        link: ~p"/sections/#{section.slug}/preview/page/#{revision.slug}"
```

lib/oli_web/live/instructor/review_activities_live.ex
```diff
@@ -0,0 +1,445 @@
...
+                    where: r.resource_type_id == ^activity_type_id,
+                    where: r.scope == :banked,
+                    select: %{
+                      resource_id: pr.resource_id,
+                      tags: r.tags,
+                      objectives: r.objectives,
+                      activity_type_id: r.activity_type_id
+                    }
+                  )
+                  |> Repo.all()
```
nicocirio: NIT: This query should be in a context module

lib/oli_web/live/instructor/review_activities_live.ex
```diff
@@ -0,0 +1,445 @@
+defmodule OliWeb.Instructor.ReviewActivitiesLive do
```
nicocirio, line 1: Since we have already tested the `Oli.Delivery.BlacklistedActivities` module, I think we could add some basic tests to this liveview (for example, "page loads correctly", "header lists activity bank selection count correctly", etc.)

nicocirio (changes requested):
Nice work! I tested it locally and it works as expected.
The only thing we should change before merging it is the link in the breadcrumb (that points to the page instead of the page preview)

test/oli_web/live/instructor/review_activities_live_test.exs
```diff
@@ -0,0 +1,39 @@
...
+      # Should redirect with error flash
+      assert {:error, {:redirect, %{flash: %{"error" => error_message}}}} = result
+      assert error_message =~ "Failed to load selection"
+    end
+
+    test "LiveView module exists and has mount function", %{conn: _conn} do
+      # Basic module test
+      assert function_exported?(OliWeb.Instructor.ReviewActivitiesLive, :mount, 3)
+      assert function_exported?(OliWeb.Instructor.ReviewActivitiesLive, :render, 1)
+    end
```
nicocirio, line 37: I think this test does not add much value. Instead, I would suggest a test that checks the liveview mounts correctly when a valid page and selection are provided (asserting on some content of the page, like the number of activities)

darrensiegel:
Closed, we are clearly not taking this work

PR #5804, opened 2025-08-19
[BUG FIX] [MER-4805] Fix student lesson view to use ResourceAccess.out_of for consistent "points possible" display

manelli:
When page revisions change (e.g., author adds questions to assessment), ResourceAttempt.out_of gets updated immediately with the new calculated value from the updated page content, while ResourceAccess.out_of is only updated during activity roll-up or finalization.

This caused a discrepancy where the student interface was showing outdated "points possible" values from ResourceAttempt.out_of, while the instructor interface was showing current values from ResourceAccess.out_of. The "points possible" shown in the student lesson view also didn't match the out-of score displayed in the assignments page where students can see all their assignments.

The fix changes lesson_live.ex to retrieve current_out_of from ResourceAccess.out_of instead of ResourceAttempt.out_of, ensuring both student and instructor interfaces display consistent "points possible" values that reflect the most up-to-date assessment content. This also ensures the "points possible" shown in the student lesson view now matches the out-of score displayed in the assignments page where students can see all their assignments.

See: https://eliterate.atlassian.net/browse/MER-4805

### DIFF
--- a/test/oli_web/live/delivery/student/lesson_live_test.exs
+++ b/test/oli_web/live/delivery/student/lesson_live_test.exs
@@ -1850,6 +1850,74 @@ defmodule OliWeb.Delivery.Student.LessonLiveTest do
       assert updated_attempt.lifecycle_state == :evaluated
       refute updated_attempt.was_late
     end
+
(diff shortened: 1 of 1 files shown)


nicocirio (approved):
🚀

PR #5809, opened 2025-08-19
[FEATURE] [MER-4617] [MER-4618] Audit log feature

darrensiegel:
Implements the Audit Log feature (core infrastructure, Admin UI, initial usages for account deletion and course project publication)


lib/oli/auditing/log_event.ex
```diff
@@ -0,0 +1,182 @@
...
+    :institution_created,
+    :institution_updated,
+    :institution_deleted,
+    :role_changed,
+    :permission_granted,
+    :permission_revoked,
+    :content_updated,
+    :grade_updated,
+    :system_setting_changed
+  ]
```
nicocirio: We defined many event types that we are not yet using. We are only capturing the `project_published` and `section_created`.
This might be on purpose (a future PR will wire the rest of the the captures) or maybe it was intended to be part of this PR

lib/oli_web/live/admin/audit_log_live.ex
```diff
@@ -0,0 +1,297 @@
...
+        Breadcrumb.new(%{
+          full_title: "Audit Log",
+          link: Routes.live_path(OliWeb.Endpoint, __MODULE__)
+        })
+      ]
+  end
+
+  def mount(_, _session, socket) do
+    # Check if user has system admin role
+    if !Oli.Accounts.has_admin_role?(socket.assigns.current_author, :system_admin) do
```
nicocirio: Instead of validating the admin role here, we should add the route under this scope:

```
    scope "/" do
      pipe_through([:require_authenticated_system_admin])
```

We already have that scope in the router file.

lib/oli_web/live/admin/audit_log_live.ex
```diff
@@ -0,0 +1,297 @@
...
+              id="actor_type_filter"
+              name="actor_type"
+              phx-hook="SelectListener"
+              phx-change="filter_actor_type"
+              class="mt-1 block w-full pl-3 pr-10 py-2 text-base border-gray-300 focus:outline-none focus:ring-indigo-500 focus:border-indigo-500 sm:text-sm rounded-md"
+            >
+              <option value="">All Actors</option>
+              <option value="user" selected={@options.actor_type == :user}>Users</option>
+              <option value="author" selected={@options.actor_type == :author}>Authors</option>
+            </select>
```
nicocirio, line 153: We should also allow filtering log events by `:system` (user_id = nil and author_id = nil)

lib/oli/auditing.ex
```diff
@@ -0,0 +1,340 @@
...
+        do: dynamic([e], e.event_type == ^options.event_type),
+        else: true
+
+    # Actor type filter
+    filter_by_actor_type =
+      case options.actor_type do
+        :user -> dynamic([e], not is_nil(e.user_id))
+        :author -> dynamic([e], not is_nil(e.author_id))
+        _ -> true
+      end
```
nicocirio, line 247: We should extend this dynamic filter to allow filtering by `system`

nicocirio (changes requested):
Nice work! I tested it locally and it works as expected. 
On the liveview we do not have the option to filter by "system" author. I believe that would be a nice-to-have filter.
On the other hand, I noticed we added many event_types that we are not yet auditing.

lib/oli/auditing.ex
```diff
@@ -0,0 +1,340 @@
...
+    # Resource filters
+    filter_by_project =
+      if options.project_id,
+        do: dynamic([e], e.project_id == ^options.project_id),
+        else: true
+
+    filter_by_section =
+      if options.section_id,
+        do: dynamic([e], e.section_id == ^options.section_id),
+        else: true
```
nicocirio: These 4 filters will always be true since we do not have ways to set these values in the UI.
On the other hand, we should carefully handle the `date_from` and `date_to` options to be able to compare them with the `inserted_at` field (we should probably use `fragment`)

lib/oli/auditing/log_event.ex
darrensiegel: Yes this was on purpose

darrensiegel: Actually i am going to remove them for now.  And in the future as we add support for new ones, we add them back one by one

nicocirio: makes sense 😄

lib/oli/auditing.ex
darrensiegel: Removing these as they are unused.

darrensiegel:
@nicocirio I've addressed your review comments:
- Fixed it to rely on the router scoping
- Removed all of the "not yet implemented" event types
- Remove the "not yet used" dynamic filters

I didn't implement the "filter by system" idea yet - as we do not have any events yet that are strictly "system" driven. But that is a good idea and I captured it as a comment next to the event_types definition.

nicocirio:
> @nicocirio I've addressed your review comments:
> 
> * Fixed it to rely on the router scoping
> * Removed all of the "not yet implemented" event types
> * Remove the "not yet used" dynamic filters
> 
> I didn't implement the "filter by system" idea yet - as we do not have any events yet that are strictly "system" driven. But that is a good idea and I captured it as a comment next to the event_types definition.

Great, thanks! One last thing, we can now remove the author system role check on the LiveView mount callback, since now that is checked in the router.

darrensiegel:
@nicocirio  remove that system role check as it is redundant, good catch!

PR #5811, opened 2025-08-20
[FEATURE] [MER-4677] All Projects CSV Download AND Claude Code infrastructure

darrensiegel:
Implements the All Projects CSV Download feature, driven from a new Claude Code infrastructure. 

The CC infrastructure includes:
- Updated `CLAUDE.md` file, which guides CC but balances token size
- `torus-architect`: command to research and produce a feature design document
- `torus-build-plan`: command to build an implementation plan from an FDD
- `torus-develop`: command to implement a plan
- `elixir-code-reviewer` subagent that reviews elixir code


CLAUDE.md
```diff
@@ -0,0 +1,119 @@
+## Torus Overview
+Learning engineering platform for authoring, delivering and improving online courses. Traditional monolithic Elixir/Phoenix app with
+React frontend.
+
+## Essential Commands
+```bash
+# Backend
+mix deps.get && mix phx.server # install & run
+mix test # test
+mix test test/path/to/file.exs # run one test
```
nicocirio, line 11: in our case we should add `set -a && source oli.env`

CLAUDE.md
```diff
@@ -0,0 +1,119 @@
...
+## Critical Performance Rules
+
+- Use SectionResourceDepot cache, not SectionResource queries
+- Never use DeliveryResolver.full_hierarchy directly
+- Avoid cross-table queries in attempt hierarchy tables
+
+## Dev Guidelines
+
+- Use `IO.inspect` for Elixir debugging
+- All new UIs should use LiveView not React
```
nicocirio, line 106: LiveView 🫶

nicocirio (changes requested):
Nice work! 
I tested the "Download CSV" feature locally and it worked as expected.
Regarding the Claude Code infrastructure, those definitions are awesome. We'll likely refine and improve those `.md` files as we gain experience with Claude.
I requested changes because we need to make a minor addition to our *Essential Commands* in the `CLAUDE.md` file

darrensiegel:
@nicocirio  Good catch, i made that addition.  And you are absolutely correct, these claude files will change and improve over time.  Devs will also have the ability to define their own "personal" claude commands and subagents (which live in `~/.claude/` and thus aren't checked in)

nicocirio (approved):

PR #5817, opened 2025-08-21
[BUGFIX] [MER-4867] fix test for divide-by-zero in proficiency bucketing calculation

andersweinstein:
Existing was blocking divide-by-zero by testing for zero values with `x in [0.0, -0.0]` This form was considered unnecessary, but harmless, so preserved.  However, further testing revealed routine was still tripping ArithmeticError exceptions. Tracing revealed that at some point, the relevant counts came in to the calculation as integers, rather than floats. This was infrequent, but it happened; reasons are unclear.

Apparently the "in" operator relies on strict equality so fails to detect an integer 0 in this case:
```
ex(10)> 0 in [0.0]
false
iex(11)> 0 == 0.0
true
```
This PR fixes by using the simpler `x == 0.0` which would have been preferable all along.


eliknebel (approved):
Nice job figuring this one out

PR #5819, opened 2025-08-21
[BUG FIX] [MER-4834] Adaptive page global variables not recalled - part 2

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-4834?focusedCommentId=32567) to the comment in the ticket

This PR moves the response parsing into the `readGlobal` function.
The function used to return an array of maps, and now it returns a map (as adaptive pages expect).

### After
Added observations are kept between different adaptive pages

https://github.com/user-attachments/assets/df7cab87-f3bd-4c0f-b9f3-1e6c1b538c13


simonchoxx (approved):
LGTM

PR #5822, opened 2025-08-22
[FEATURE] [MER-1879] Advanced Author - Click and Drag Screens

dtiwarATS:

This PR introduces **drag & drop support for the Sequence Editor using two different approaches**, allowing us to evaluate and decide which implementation best fits our use case.

**Key Changes**

1. Added the following dependencies for drag & drop functionality:

```
@dnd-kit/core
@dnd-kit/sortable
@dnd-kit/utilities
dnd-kit-sortable-tree
```
2. Implemented **two approaches** for drag & drop:
    2.1 **Sortable Tree**
             1. Drag & drop is always enabled.
             2. Uses dnd-kit-sortable-tree.
             3. Provides a nested sortable tree structure where items can be reordered freely.

    2.2. **Hierarchy View**
            1. Drag & drop can be toggled on/off.
            2. Uses @dnd-kit/core + custom sortable logic.
            3. Offers a more controlled drag mode, closer to our current hierarchy interactions.

3. Added a toggle in the UI to switch between **Sortable Tree and Hierarchy View** (see screenshot).

**Next Steps: **

1. Once this PR is merged, **Mia will test both approaches** and provide feedback on usability, performance, and maintainability.
2. Based on the feedback:
    2.1 We will **remove one approach** along with its related code/dependencies.
    2.2 Keep and refine the chosen solution.

**Notes:**

1. This PR intentionally includes **both implementations** for side-by-side comparison.
2. Some code duplication exists temporarily but will be removed after we finalize the approach.

https://github.com/user-attachments/assets/951273b9-cc71-4d06-a6d2-15cf9ef56e2f


assets/yarn.lock
```diff
@@ -3259,6 +3259,37 @@
   resolved "https://registry.npmjs.org/@discoveryjs/json-ext/-/json-ext-0.5.7.tgz"
   integrity sha512-dBVuXR082gk3jsFp7Rd/JI4kytwGHecnCoTtXFb7DB6CNHp4rg5k1bhg0nWdLGLnOV71lmDzGQaLMy8iPLY0pw==
```
dtiwarATS, line 3261: Hey @gastonabella, I added some packages for Drag & Drop and saw this check failing:

`error Your lockfile needs to be updated, but yarn was run with --frozen-lockfile.`


I went ahead and committed the updated yarn.lock file. Just wanted to confirm — is that the right thing to do when adding new packages?

PR #5825, opened 2025-08-25
[BUG FIX] [MER-4837] Cannot move canvas if no components present

dtiwarATS:

https://github.com/user-attachments/assets/c2730962-424c-4f87-beff-735808d16d34


assets/src/components/activities/adaptive/components/authoring/LayoutEditor.tsx
```diff
@@ -96,6 +96,18 @@ const LayoutEditor: React.FC<LayoutEditorProps> = (props) => {
     setParts(props.parts);
   }, [props.parts]);
 
+  useEffect(() => {
+    if (!parts?.length) {
+      setParts([
```
dtiwarATS, line 101: If no parts are displayed on the screen, a dummy part will be added dynamically. This dummy part won’t be saved in the database—it’s only there to let the user drag the blank activity.

PR #5826, opened 2025-08-26
[ENHANCEMENT] [MER-4802] [MER-4711] "Instructor Preview" mode, student response visualizations

darrensiegel:
This PR adds a new `mode` attribute to the activity authoring elements, with values of `authoring` (default) and `instructor_preview`.  The latter mode instructs the UI to remove some UI affordances (the X buttons to remove choices, the "Enable Dynamic Variables" button, etc).  

This PR also makes the visualization changes in MER-4711, though there is some inconsistencies in those specifications and this impl had to deviate slightly.   

 Most notably, this PR makes no changes the MultiInput activity, though MER-4711 calls out that the new visualization approach in the "Student Responses" tab should apply to MultiInput.  The problem, though, is that this design does not take into account Text and Numeric parts of Multi-Input - it only considers "Dropdown" parts.  Many multi-input questions actually contain a mixture of all three part types.  So, for now, this PR leaves MultiInput untouched. 

The new "Student Responses" tab is now present in Instructor Preview mode for MCQs and CATAs - each with the requested visualization.  One deviation exists in the visualization per the screenshot from legacy - a labelled X-axis with counts was used instead of the "X students" label.  

Screenshot 2025-08-26 at 11 50 46 AM

### DIFF
--- a/assets/src/components/activities/common/choices/delivery/LabelledChoices.tsx
+++ b/assets/src/components/activities/common/choices/delivery/LabelledChoices.tsx
@@ -2,6 +2,7 @@ import React from 'react';
 import { useAuthoringElementContext } from 'components/activities/AuthoringElementProvider';
 import { Choice } from 'components/activities/types';
 import { SlateOrMarkdownEditor } from 'components/editing/SlateOrMarkdownEditor';
+import { DEFAULT_EDITOR } from 'data/content/resource';
 
 interface Props {
   choices: Choice[];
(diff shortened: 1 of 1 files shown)


nicocirio (commented):
I reviewed the Elixir files (`.ex` and `.exs`) and they all look good to me. 
I do not approve the PR since I would like someone with more knowledge about React to review those files.

eliknebel (approved):

PR #5828, opened 2025-08-27
[BUG FIX] [MER-4880] Fix for the activity reset race condition

darrensiegel:
Here's a detailed description of the three-pronged approach to address the race condition issues which are affecting (but not caused by) Score as you Go.  I believe this issue has been in the codebase for quite some time, but we never saw reports about it because up until this point in time, there had never been a way to "reset" an activity in a high stakes context.  

### The Problem

  The original implementation had a critical race condition where multiple concurrent reset requests could:
  - Read the same activity attempt simultaneously
  - Both calculate the same "next attempt number"
  - Create duplicate attempts with identical attempt numbers
  - Corrupt the scoring calculations that depend on sequential attempt numbering 

### The Solution: Database-Level Concurrency Control

  Serializable Isolation Level: The transaction runs with isolation: :serializable, the highest isolation level
  that prevents:
  - Dirty reads
  - Non-repeatable reads
  - Phantom reads
  - Serialization anomalies

  Row-Level Locking: Uses FOR UPDATE to lock the parent resource_attempt row.

  Concurrent and Serial Reset Detection: After acquiring the lock, validates that the activity attempt being reset is still the latest:
```
  {max_attempt_number} =
    Repo.one(
      from(a in ActivityAttempt,
        where: a.resource_attempt_id == ^resource_attempt.id and
               a.resource_id == ^activity_attempt.resource_id,
        select: {max(a.attempt_number)}
      )
    )
  if activity_attempt.attempt_number != max_attempt_number do
    Repo.rollback({:already_reset})
  end
```

### Why This Approach Works

  1. Mutual Exclusion: Only one transaction can hold the FOR UPDATE lock at a time
  2. Atomic Validation: The "latest attempt" check happens within the same transaction as the creation
  3. Clean Failure Mode: Subsequent requests fail fast with {:error, {:already_reset}} rather than creating
  corrupt data
  4. Performance: Minimal overhead - just one additional query and lock per reset operation

### Frontend Safeguards

Button Debouncing: Implement client-side debouncing to prevent rapid-fire clicks from sending multiple network requests to reset. 

Visual Feedback: Provide immediate UI feedback to discourage multiple submissions:
  - Disable the reset button during processing
  - Show loading spinners or progress indicators
  - Display "Processing..." messages

## 3. Inline Record Cleanup with fix_bad_records Function

While the locking prevents new race conditions, existing corrupted data needed immediate remediation for score-as-you-go assessments.

Cleanup Strategy

  1. Prefer Evaluated Records: If duplicates exist, choose evaluated over non-evaluated
  2. Recency Priority: Among evaluated records, pick the most recently evaluated
  3. Deterministic Fallback: If no evaluated records exist, consistently pick the first record
  4. Preserve Good Data: Non-duplicate records pass through unchanged
  5. Maintain Order: Original ordering is preserved where possible

### DIFF
--- a/lib/oli/delivery/attempts/activity_lifecycle/roll_up.ex
+++ b/lib/oli/delivery/attempts/activity_lifecycle/roll_up.ex
@@ -342,7 +342,6 @@ defmodule Oli.Delivery.Attempts.ActivityLifecycle.RollUp do
   Account for bad records by filtering out multiple records for the same attempt number.
   """
   def fix_bad_records(attempts) do
-
     # Group the records by activity id
     by_activity_id = Enum.group_by(attempts, & &1.resource_id)
 
@@ -350,11 +349,9 @@ defmodule Oli.Delivery.Attempts.ActivityLifecycle.RollUp do
     Enum.reduce(by_activity_id, [], fn {_activity_id, records}, acc ->
       acc ++ fix_bad_records_for_activity_id(records)
     end)
-
   end
 
   defp fix_bad_records_for_activity_id(attempts) do
-
     grouped = Enum.group_by(attempts, & &1.attempt_number)
 
     # For all groups where there is more than one entry, that
@@ -362,15 +359,12 @@ defmodule Oli.Delivery.Attempts.ActivityLifecycle.RollUp do
     # the same attempt number
     Enum.reduce(grouped, [], fn {_attempt_number, records}, acc ->
       if length(records) > 1 do
-
         # Filter down to only those that are evaluated
-        evaluated = Enum.filter(records, & &1.lifecycle_state == :evaluated)
+        evaluated = Enum.filter(records, &(&1.lifecycle_state == :evaluated))
 
         case Enum.count(evaluated) do
-
           # None were evaluated, so it is safe to use any one of them
-          0 -> acc ++ [(records |> hd())]
-
+          0 -> acc ++ [records |> hd()]
           # Sort by date evaluated descending, return the most recently evaluated
           _n -> acc ++ [Enum.sort_by(evaluated, & &1.date_evaluated, :desc) |> List.first()]
         end
--- a/lib/oli/delivery/attempts/activity_lifecycle/roll_up.ex
+++ b/lib/oli/delivery/attempts/activity_lifecycle/roll_up.ex
@@ -148,6 +148,8 @@ defmodule Oli.Delivery.Attempts.ActivityLifecycle.RollUp do
             a.resource_id == activity_id and a.attempt_guid != activity_attempt_guid
           end)
 
+        other_attempts_for_this_activity = fix_bad_records(other_attempts_for_this_activity)
+
         all_attempts = [
           %{score: score, out_of: out_of, date_evaluated: now} | other_attempts_for_this_activity
         ]
@@ -164,6 +166,7 @@ defmodule Oli.Delivery.Attempts.ActivityLifecycle.RollUp do
           end)
           |> Enum.group_by(fn a -> a.resource_id end)
           |> Enum.map(fn {_, attempts} ->
+            attempts = fix_bad_records(attempts)
             Enum.max_by(attempts, fn a -> a.attempt_number end)
           end)
           |> List.flatten()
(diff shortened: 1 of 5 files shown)


test/oli/delivery/attempts/activity_lifecycle/roll_up_fix_bad_records_test.exs
nicocirio, line 1: To keep consistency with other test files and with Elixir best practices, this file should be named `roll_up_test.exs`

lib/oli/delivery/attempts/activity_lifecycle/roll_up.ex
```diff
@@ -335,6 +338,42 @@ defmodule Oli.Delivery.Attempts.ActivityLifecycle.RollUp do
...
+  @doc """
+  Account for bad records by filtering out multiple records for the same attempt number.
+  """
+  def fix_bad_records(attempts) do
+    # Group the records by activity id
+    by_activity_id = Enum.group_by(attempts, & &1.resource_id)
+
+    # Then disambiguate the records for each activity id
+    Enum.reduce(by_activity_id, [], fn {_activity_id, records}, acc ->
+      acc ++ fix_bad_records_for_activity_id(records)
```
nicocirio, line 350:
```suggestion
      [fix_bad_records_for_activity_id(records) | acc]
```

lib/oli/delivery/attempts/activity_lifecycle/roll_up.ex
```diff
@@ -335,6 +338,42 @@ defmodule Oli.Delivery.Attempts.ActivityLifecycle.RollUp do
...
+    # means that we somehow ended up with multiple records for
+    # the same attempt number
+    Enum.reduce(grouped, [], fn {_attempt_number, records}, acc ->
+      if length(records) > 1 do
+        # Filter down to only those that are evaluated
+        evaluated = Enum.filter(records, &(&1.lifecycle_state == :evaluated))
+
+        case Enum.count(evaluated) do
+          # None were evaluated, so it is safe to use any one of them
+          0 -> acc ++ [records |> hd()]
```
nicocirio, line 367:
```suggestion
          0 -> [hd(records) | acc]
```

lib/oli/delivery/attempts/activity_lifecycle/roll_up.ex
```diff
@@ -335,6 +338,42 @@ defmodule Oli.Delivery.Attempts.ActivityLifecycle.RollUp do
...
+          # Sort by date evaluated descending, return the most recently evaluated
+          _n -> acc ++ [Enum.sort_by(evaluated, & &1.date_evaluated, :desc) |> List.first()]
+        end
+      else
+        acc ++ records
```
nicocirio, line 372:
```suggestion
        [records | acc]
```

lib/oli/delivery/attempts/activity_lifecycle/roll_up.ex
```diff
@@ -335,6 +338,42 @@ defmodule Oli.Delivery.Attempts.ActivityLifecycle.RollUp do
...
+    Enum.reduce(grouped, [], fn {_attempt_number, records}, acc ->
+      if length(records) > 1 do
+        # Filter down to only those that are evaluated
+        evaluated = Enum.filter(records, &(&1.lifecycle_state == :evaluated))
+
+        case Enum.count(evaluated) do
+          # None were evaluated, so it is safe to use any one of them
+          0 -> acc ++ [records |> hd()]
+          # Sort by date evaluated descending, return the most recently evaluated
+          _n -> acc ++ [Enum.sort_by(evaluated, & &1.date_evaluated, :desc) |> List.first()]
```
nicocirio, line 369:
```suggestion
          _n -> [(Enum.sort_by(evaluated, & &1.date_evaluated, :desc) |> List.first()) | acc]
```

nicocirio (commented):
Eli will be reviewing this PR since it’s a hotfix, but I left a few comments as well. I hope they’re helpful!

eliknebel (approved):
Changes look good. I agree with Nico on using the more efficient operation to append to list, however given the relatively small number of records this will be processing it probably won't have a huge impact on performance.

PR #5829, opened 2025-08-27
[BUG FIX] [MER-4882] Fix major updates causing remix content removal

darrensiegel:
The new unit tests demonstrates the bug and was written (and failed) prior to changing any production code. That unit test shows a remixed course section getting a major publication pushed to it.  During application of that update, the course section has a container that was added from the "remix" project completely dropped from the hieararchy. 

The root cause of the error is in a (somewhat recently added) step that "cleans" each container after the updates are applied.  This cleaning step is given a map of "all known resource ids in the course section" and was instructed to remove any from the current section resource `:children` list that either do not exist (`nil`) or are deleted.  

The issue here was the the clean function was given ONLY a mapping of the resources in the project being updated - and NOT the complete mapping of all remixed projects.  These changes now pass the unit test.

### DIFF
--- a/lib/oli/delivery/sections/updates.ex
+++ b/lib/oli/delivery/sections/updates.ex
@@ -454,6 +454,18 @@ defmodule Oli.Delivery.Sections.Updates do
     end
   end
 
+  defp all_resource_mappings(section) do
+    # Query the section project publications to get all publication ids, then create the
+    # published resource map for each and MERGE them all into one map
+    from(spp in Sections.SectionsProjectsPublications,
+      where: spp.section_id == ^section.id,
+      select: spp.publication_id
+    )
+    |> Repo.all()
+    |> Enum.map(fn publication_id -> MinimalHierarchy.published_resources_map(publication_id) end)
+    |> Enum.reduce(%{}, fn m, merged -> Map.merge(merged, m) end)
+  end
+
   defp update_container_children(section, prev_publication, new_publication) do
     container = ResourceType.id_for_container()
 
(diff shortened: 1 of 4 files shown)


lib/oli/delivery/sections/updates.ex
```diff
@@ -454,6 +454,18 @@ defmodule Oli.Delivery.Sections.Updates do
...
 
+  defp all_resource_mappings(section) do
+    # Query the section project publications to get all publication ids, then create the
+    # published resource map for each and MERGE them all into one map
+    from(spp in Sections.SectionsProjectsPublications,
+      where: spp.section_id == ^section.id,
+      select: spp.publication_id
+    )
+    |> Repo.all()
+    |> Enum.map(fn publication_id -> MinimalHierarchy.published_resources_map(publication_id) end)
```
eliknebel, line 465: I noticed this map is making multiple DB requests for each iteration, but I suspect this is okay given the complexity of any alternative solution and since this code executes in an update background job

eliknebel: I noticed this map is making multiple DB requests for each iteration, but I suspect this is okay given the complexity of any alternative solution and since this code executes in an update background job.

I see there is a version of `MinimalHierarchy.published_resources_map` that takes multiple publication_ids and executes a single query. Maybe that would be more efficient?

eliknebel (changes requested):

darrensiegel:
@eliknebel  Take another look.  I used that multi publication id version (didn't realize it existed)

PR #5833, opened 2025-08-27
[CHORE] [MER-4885] Devops dynamic deployments

eliknebel:
https://eliterate.atlassian.net/browse/MER-4885

This PR adds preview auto deployment for new PRs. When a PR is opened or updated the branch will be deployed to our self-hosted plasma server for testing/demos. Simon-bot will comment on the PR with a link to the preview deployment, as seen in this PR below.

This work also tweaks the way the ecto cloak vault key gets loaded in order to fix some issues I was running into with running seeds/migrations. making the env variable not required (falling back to a default value) in dev or test environments.

There is some more fine-tuning and followup work to do here but this represents the core of this functionality and is probably a good checkpoint to get this landed.

NOTE: We are currently working to get a wildcard SSL cert in place but until then, links will have an SSL warning. You can click through the warning in order to proceed for now.


darrensiegel:
This is fantastic and opens up several different use cases from demos to isolated, automated testing. 

A few notes for some potential enhancements (if this doesn't already account for them):
1. It would be great to be able to specify somehow that the PR instance should start from a copy of Tokamak DB. 
2. It will be necessary to bootstrap GEN AI provider API Keys, so that features that use GenAI will be available.

darrensiegel (approved):

PR #5837, opened 2025-08-28
[BUG FIX] [MER-4871] [MER-4754] fix code syntax highlighting & add line numbering

andersweinstein:
This addresses two issues with code block styling in delivery:

[MER-4871 ](https://eliterate.atlassian.net/browse/MER-4871) Even though the `highlightjs` syntax highlighting script was being loaded by the page, syntax highlighting was not getting applied in delivery. This happened because the highlightjs processing of code elements on the page was not getting kicked off at the right time for a LiveView page. Added `HighlightCode` hook for this purpose. Wrapped together with `EvaluateMathJaxExpressions` in new `PageContentHooks` composite hook to execute both over the same element scope to find and adjust content elements on page. 

[MER-4754](https://eliterate.atlassian.net/browse/MER-4754) Code line numbering as in legacy was not being done. This implements via the `highlightjs-line-numbers` add-on to highlightjs, defining custom styles for light and dark mode. 

Note: discovered pages were loading BOTH light and dark mode stylesheets for a certain theme from highlightjs, with the latter link controlled by a media query for user's preferred colored scheme. The media query apparently did not work to prevent loading the darkmode stylesheet (possibly a webpack artifact), but in any case is not sensitive torus dark mode, which is implemented as an app-specific mode not related to user's browser-wide or OS-wide preference. With our style customizations adjusting background colors for dark mode, it appears the highlightjs dark mode stylesheet is unnecessary, so just removed it.

Current:  

Screenshot 2025-08-28 at 6 46 27 AM

New:  

Screenshot 2025-08-28 at 6 26 32 AM

Screenshot 2025-08-28 at 6 26 57 AM

<br><br>Note 2: now that we are seeing the highlightjs syntax coloring for the first time, we should examine it carefully to see if we like the theme it is using or need to customize some of the colors. For example, this theme ('atom-one") colors types, variables and numeric constants the same, which is different than the highlighting done by other highlighters including the component used in our authoring editor. But this can be a separate issue.

Java in highlightjs with atom-one theme:  

Screenshot 2025-08-28 at 7 02 58 AM

[MER-4754]: https://eliterate.atlassian.net/browse/MER-4754?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Same in authoring editor component:  

Screenshot 2025-08-28 at 10 23 33 AM

The CS111 course content includes screenshots from a third-party web-based Java IDE used in the course. This also uses different highlighting conventions; they look more similar to those of preview component:

java-ide

### DIFF
--- a/assets/src/hooks/evaluate_mathjax_expressions.ts
+++ b/assets/src/hooks/evaluate_mathjax_expressions.ts
@@ -3,7 +3,6 @@
 
 export const EvaluateMathJaxExpressions = {
   mounted() {
-    console.log('EvaluateMathJax');
     const elements = document.querySelectorAll('.formula, .formula-inline, .chat-message');
 
     const getGlobalLastPromise = () => {
--- a/assets/src/hooks/evaluate_mathjax_expressions.ts
+++ b/assets/src/hooks/evaluate_mathjax_expressions.ts
@@ -3,6 +3,7 @@
 
 export const EvaluateMathJaxExpressions = {
   mounted() {
+    console.log('EvaluateMathJax');
     const elements = document.querySelectorAll('.formula, .formula-inline, .chat-message');
 
     const getGlobalLastPromise = () => {
--- a/assets/src/hooks/highlight_code.ts
+++ b/assets/src/hooks/highlight_code.ts
@@ -1,0 +1,12 @@
+export const HighlightCode = {
+  mounted() {
+    const hljs = (window as any).hljs;
+
+    hljs.configure({
+      cssSelector: 'pre code.torus-code',
+    });
+
+    hljs.highlightAll();
+    hljs.initLineNumbersOnLoad();
+  },
+};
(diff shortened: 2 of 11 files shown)


assets/src/hooks/evaluate_mathjax_expressions.ts
```diff
@@ -3,6 +3,7 @@
 
 export const EvaluateMathJaxExpressions = {
   mounted() {
+    console.log('EvaluateMathJax');
```
darrensiegel: Remove log message

darrensiegel (changes requested):

assets/src/hooks/evaluate_mathjax_expressions.ts
andersweinstein: Oopsie. Done.

PR #5839, opened 2025-08-28
[BUG FIX] [MER-4854] Course with scored pages has no assignments

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-4854) to the ticket

### Before

https://github.com/user-attachments/assets/a28bf790-c65f-404d-9302-d87c3db96330

### After

https://github.com/user-attachments/assets/3e70d2c0-c2d0-453d-a757-fb26587ba447

### Background
When we improved the app's performance by implementing the SectionResourceDepot, we extended the `section_resources` table to include/replicate some fields from the revision (such as the `graded` field). This optimization eliminated the need for four joins to retrieve revision data for a specific section resource, instead providing all necessary data directly in the Depot's section resource record.

### Issue
When an author made minor changes to a page (e.g., changing "PAGE_A" from `graded: false` to `graded: true`) and published those changes, the updates were not reflected in the corresponding `section_resource` record. This caused the SectionResourceDepot —which powers most of our LiveViews, including the Assignments page— to serve stale data. Consequently, when students accessed the assignments page, "PAGE_A" was excluded because it still appeared as `graded: false`.

### Solution
We now trigger section resource migration during the minor-changes publishing flow, ensuring that `section_resources` values remain synchronized with their corresponding revision values.

**Why didn't this issue occur in the major-changes flow?** Because major changes trigger a complete rebuild of section resources from scratch, migrating all current revision data into the section resource records.

### Technical Implementation
- Created a `migrate_specific_resources` function for the minor-changes flow to avoid updating all section resources unnecessarily
- Abstracted the common query logic from the existing `migrate` function into a shared implementation
- Refactored the original raw SQL query into Ecto for improved readability and maintainability

### DIFF
--- a/lib/oli/delivery/sections/section_resource_migration.ex
+++ b/lib/oli/delivery/sections/section_resource_migration.ex
@@ -2,7 +2,16 @@ defmodule Oli.Delivery.Sections.SectionResourceMigration do
   alias Oli.Repo
   import Ecto.Query
   alias Oli.Delivery.Sections.SectionResource
+  alias Oli.Delivery.Sections.SectionsProjectsPublications
+  alias Oli.Publishing.Publications.Publication
+  alias Oli.Publishing.PublishedResource
+  alias Oli.Resources.Revision
+  alias Oli.Authoring.Course.Project
 
+  @doc """
+  Returns true if at least one SectionResource record requires migration, false otherwise.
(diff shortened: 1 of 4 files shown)


darrensiegel:
Great catch, good solution!

PR #5841, opened 2025-08-28
[FEATURE] [MER-4697] Instructor | Insights > Content | Send email to students

manelli:
This PR adds functionality for instructors to email students from the Insights > Content page. When one or more students are selected, an Email button appears in the toolbar. The button provides two options:

Copy email addresses: copies the selected students’ addresses to the clipboard

Send email: opens an email modal

The email modal shows the recipient information based on whether a single student or multiple students are selected. Instructors can type a message or insert an existing template. Templates are read-only for instructors and can only be managed by admins.

Emails are sent individually, not as a group, and use the OLI-standard format and subject line. The system prevents sending blank messages.

---
See: https://eliterate.atlassian.net/browse/MER-4697

---

## TODO:
- Success confirmation should not be missing after a message is sent
- Cannot scroll after sending the email


assets/src/hooks/copy_to_clipboard_event.ts
```diff
@@ -0,0 +1,18 @@
+export const CopyToClipboardEvent = {
+  mounted() {
+    this.handleEvent('copy_to_clipboard', ({ text }: { text: string }) => {
+      navigator.clipboard
+        .writeText(text)
+        .then(() => {
+          console.log('Text copied to clipboard:', text);
+        })
+        .catch((err) => {
+          console.error('Failed to copy text to clipboard:', err);
```
nicocirio: We may want to remove these 2 `console.log`

assets/src/hooks/copy_to_clipboard_event.ts
```diff
@@ -0,0 +1,18 @@
...
+        });
+    });
+
+    this.handleEvent('remove-overflow-hidden', () => {
+      document.body.classList.remove('overflow-hidden');
+    });
```
nicocirio: Is this event handler duplicated with the one we defined on `app.ts`?

assets/src/hooks/copy_to_clipboard_event.ts
```diff
@@ -0,0 +1,18 @@
+export const CopyToClipboardEvent = {
```
nicocirio: There is an existing CopyToClipboard hook we might re-use instead of creating this new one

nicocirio (commented):
Awesome work! 
I left some minor comments regarding the CopyToClipboard hook.
And while testing it locally I found a minor bug; the all-checkbox does not keep state when clicking on it:

https://github.com/user-attachments/assets/68cb3b2c-0477-434f-be7e-42235be84eb2

nicocirio:
Thanks for addressing those review comments! I tested it locally and found an edge case where I am not sure how the feature is supposed to work: 
If we have a paginated table and click to select all, only the students from the current page are actually selected (only those emails are copied - emails are only sent to those). Should the checkbox select all students (including ones from other pages)?

https://github.com/user-attachments/assets/7dc1f939-18bd-45f3-970f-a806aa146328

Before recording the video, I tweaked `page_sizes` in `lib/oli_web/live/common/paging.ex` to include the value 1. In this way, I avoided having to create and enroll up to 10 students to test this edge case

nicocirio (approved):
🚀

PR #5845, opened 2025-08-30
[BUG FIX] [MER-4889] Submitted (not evaluted) part attempts trigger failure in snapshot worker

darrensiegel:
Snapshot worker for XAPI upload didn't handle case that part attempt is submitted, not evaluated.  The failure here triggers a retry, which fails again, another retry, which also fails. 

So explicitly handling this saves us 3 failed oban jobs. 

https://appsignal.com/open-learning-initiative/sites/618539ab2cf81d7e3cd051ca/exceptions/incidents/825/samples/618539ab2cf81d7e3cd051ca-1426492702470160536017565166202

** (FunctionClauseError) no function clause matching in Oli.Analytics.XAPI.StatementFactory.to_statements/1**

lib/oli/analytics/xapi/statement_factory.ex:10 Oli.Analytics.XAPI.StatementFactory.to_statements(nil)
lib/oli/delivery/snapshots/worker.ex:73 Oli.Delivery.Snapshots.Worker.perform_now/3
lib/oban/queue/executor.ex:129 Oban.Queue.Executor.perform/1
lib/oban/queue/executor.ex:74 Oban.Queue.Executor.call/1
lib/task/supervised.ex:101 Task.Supervised.invoke_mfa/2
lib/task/supervised.ex:36 Task.Supervised.reply/4

### DIFF
--- a/lib/oli/delivery/snapshots/worker.ex
+++ b/lib/oli/delivery/snapshots/worker.ex
@@ -68,6 +68,10 @@ defmodule Oli.Delivery.Snapshots.Worker do
       end
 
     case Oli.Analytics.Summary.execute_analytics_pipeline(results, project_id, host_name()) do
+      {:ok, %Pipeline{data: nil}} ->
+        # No evaluated part attempts found - job completes successfully without sending bundle
+        :ok
+
       {:ok, %Pipeline{data: attempt_group}} ->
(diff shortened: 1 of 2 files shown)


manelli (approved):
LGTM 👍

PR #5850, opened 2025-08-30
[BUG FIX] [MER-4893] AppSignal: Avoid nil comparison in query

darrensiegel:
https://appsignal.com/open-learning-initiative/sites/618539ab2cf81d7e3cd051ca/exceptions/incidents/67/samples/618539ab2cf81d7e3cd051ca-733585056698300711117565334801

```
** (ArgumentError) comparison with nil is forbidden as it is unsafe. If you want to check if a value is nil, use is_nil/1 instead
```

```
lib/ecto/query/builder.ex:1069 Ecto.Query.Builder.not_nil!/1
lib/oli/conversation/triggers.ex:70 Oli.Conversation.Triggers.verify_access/2
lib/oli/conversation/triggers.ex:99 Oli.Conversation.Triggers.possibly_invoke_trigger/3
lib/task/supervised.ex:101 Task.Supervised.invoke_mfa/2
```

### DIFF
--- a/lib/oli/conversation/triggers.ex
+++ b/lib/oli/conversation/triggers.ex
@@ -58,19 +58,23 @@ defmodule Oli.Conversation.Triggers do
   with the AI agent and triggers enabled.
   """
   def verify_access(section_slug, user_id) do
-    case Oli.Accounts.User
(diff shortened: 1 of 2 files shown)


manelli (approved):
Looks good! 👍

PR #5857, opened 2025-09-01
[FEATURE] [MER-4901] Unit test scenarios

darrensiegel:
# Oli.Scenarios

**YAML-driven integration testing at the Elixir unit test level**

## Overview

`Oli.Scenarios` enables you to write sophisticated integration tests as simple unit tests without writing any Elixir code. By describing test scenarios in YAML files, you can rapidly script complex workflows that would normally require hundreds of lines of setup code.

### Key Benefits

- **Zero-code test creation**: Define entire test scenarios in readable YAML files - no Elixir code required
- **Integration tests as unit tests**: Test complex multi-step workflows (project creation → publishing → section creation → updates → verification) with the speed and isolation of unit tests
- **Rapid iteration**: Add new test cases by creating YAML files, not writing code
- **Self-documenting**: YAML scenarios serve as both tests and documentation of system behavior
- **Real infrastructure testing**: Scenarios are powered by real infrastructure and **NOT** fixtures or other antiquated approaches. (e.g. `DBSeeder`) That means that things like simulating project hierarchy changes directly uses and tests the actual `ContainerEditor` module, the same code paths used when authors manipulate their curriculum.

### Example

Instead of writing hundreds of lines of test setup code, you can describe your entire test scenario declaratively:

```yaml
# Create a project with initial content
- project:
    name: "math_course"
    title: "Mathematics 101"
    root:
      children:
        - page: "Introduction"
        - container: "Module 1"
          children:
            - page: "Lesson 1"

# Create a course section from the project
- section:
    name: "spring_2024"
    from: "math_course"
    title: "Math 101 - Spring 2024"

# Apply operations to modify the project
- manipulate:
    to: "math_course"
    ops:
      - add_page:
          title: "Lesson 2"
          to: "Module 1"

# Publish the changes
- publish:
    to: "math_course"
    description: "Adding Lesson 2"

# Apply the update to the section
- update:
    from: "math_course"
    to: "spring_2024"

# Verify the section now has the updated content
- verify:
    to: "spring_2024"
    structure:
      root:
        children:
          - page: "Introduction"
          - container: "Module 1"
            children:
              - page: "Lesson 1"
              - page: "Lesson 2"
```


test/support/scenarios/README.md
```diff
@@ -0,0 +1,354 @@
...
+  - add_container:
+      title: "Module 2"
+      to: "root"  # Optional, defaults to root
+  ```
+
+### Content Organization
+- **`move`**: Moves a resource to a different parent container
+  ```yaml
+  - move:
+      from: "Page 1"
```
nicocirio, line 122: WDYT about renaming this `target` instead of `from`?
I am not a native English speaker, but with the current approach, I read "move something from Page 1 to Module 2" instead of "move Page 1 (the target) to Module 2".
If `target` makes more sense, then we should make the same change in the following actions

test/support/scenarios/engine.ex
```diff
@@ -0,0 +1,236 @@
...
+
+  defp execute_directive(%PublishDirective{} = directive, state) do
+    PublishHandler.handle(directive, state)
+  end
+
+  defp execute_directive(%VerifyDirective{} = directive, state) do
+    case VerifyHandler.handle(directive, state) do
+      {:ok, state, verification} -> {:ok, state, verification}
+      error -> error
+    end
```
nicocirio: This `case` can be removed, since we are not doing anything with the result.
We could simply call `VerifyHandler.handle(directive, state)`

test/support/scenarios/engine.ex
```diff
@@ -0,0 +1,236 @@
...
+  def put_user(state, name, user) do
+    %{state | users: Map.put(state.users, name, user)}
+  end
+
+  @doc """
+  Updates an institution in the state.
+  """
+  def put_institution(state, name, institution) do
+    %{state | institutions: Map.put(state.institutions, name, institution)}
+  end
```
nicocirio: The doc and the function names might be confusing. Perhaps updating the docs to be something like `Adds or updates <something> in the state` would be clearer, wdyt?

test/support/scenarios/directive_parser.ex
nicocirio, line 1: To achieve a more defensive code in a future iteration, we could expand the DirectiveParser module to validate that required fields are provided for each directive. Or even go further and validate the ops for a ManipulateDirective (for example, that only the `before` or the `after` key is provided for a `reorder` operation)

test/support/scenarios/directives/enrollment_handler.ex
```diff
@@ -0,0 +1,43 @@
...
+      # Note: For simplicity, we'll just note that enrollment was requested
+      # Actual enrollment requires complex role setup
+      case role do
+        :instructor ->
+          # Would normally call Sections.enroll with instructor role
+          :ok
+
+        :student ->
+          # Would normally call Sections.enroll with student role
+          :ok
```
nicocirio, line 31: We are missing the implementation to actually enroll the user in the given section with the given role.

test/support/scenarios/directives/institution_handler.ex
```diff
@@ -0,0 +1,39 @@
...
+          country_code: country_code,
+          institution_email: inst_email,
+          institution_url: inst_url
+        })
+
+      # Store institution in state
+      new_state = Engine.put_institution(state, name, institution)
+
+      # Optionally set as current institution
+      new_state = %{new_state | current_institution: institution}
```
nicocirio, line 31: The created institution will always be set as the current one. If we want to set it optionally, we will have to extend the InstitutionDirective definition to support opts.

test/support/scenarios/directives/product_handler.ex
```diff
@@ -0,0 +1,60 @@
...
+             ) do
+          {:ok, blueprint} ->
+            # Store the product in state
+            updated_state = Engine.put_product(state, name, blueprint)
+            {:ok, updated_state}
+
+          {:error, reason} ->
+            {:error, "Failed to create product '#{name}': #{inspect(reason)}"}
+        end
+    end
```
nicocirio: Although this works, I think nesting `case` is not easy to read. I think we may want to adjust Claude's rules to prefer using `with` instead.

test/support/scenarios/directives/publish_handler.ex
```diff
@@ -0,0 +1,37 @@
...
+      {:ok, _publication} =
+        Publishing.publish_project(
+          built_project.project,
+          pub_description,
+          state.current_author.id
+        )
+
+      # No longer automatically apply updates to sections
+      # Use the update directive instead
+      # Also no longer storing publications in state - they're fetched when needed
```
nicocirio, line 29: These comments seem to be from a Claude's previous iteration of the feature.

test/support/scenarios/directives/remix_handler.ex
```diff
@@ -0,0 +1,143 @@
...
+      # Get the target section or product
+      section =
+        case Engine.get_product(state, section_name) do
+          nil ->
+            Engine.get_section(state, section_name) ||
+              raise "Section or product '#{section_name}' not found"
+
+          product ->
+            product
+        end
```
nicocirio, line 33: There is an edge case where a test may handle both a section and a product that may have the same name. In that scenario, this code will assume we are referring to the product.
So, we probably want to be careful and always name sections and products with unique names.

test/support/scenarios/directives/update_handler.ex
```diff
@@ -0,0 +1,70 @@
...
+          Oli.Delivery.Sections.SectionCache.clear(updated_section.slug)
+
+          # Small delay to ensure all async processing is complete
+          Process.sleep(100)
+
+          refreshed_section = Oli.Delivery.Sections.get_section!(updated_section.id)
+
+          # Update the section in state with the refreshed section
+          updated_state = Engine.put_section(state, section_name, refreshed_section)
+          {:ok, updated_state}
```
nicocirio: The `Oli.Delivery.Sections.Updates.apply_publication_update` already clears the SectionCache, so I think we do not need to clear it again here, nor wait (Process.sleep). We can directly assign the `updated_section` to the state.

test/support/scenarios/directives/remix_handler.ex
```diff
@@ -0,0 +1,143 @@
...
+
+      # Rebuild the section curriculum with the modified hierarchy
+      Sections.rebuild_section_curriculum(section, updated_hierarchy, updated_pinned_publications)
+
+      # Clear cache and reload section
+      SectionCache.clear(section.slug)
+      Process.sleep(100)
+
+      refreshed_section = Sections.get_section!(section.id)
```
nicocirio: `Sections.rebuild_section_curriculum` already clears the cache. So there is no need to repeat that here, nor wait before fetching the refreshed_section

test/support/scenarios/directives/verify_handler.ex
```diff
@@ -0,0 +1,409 @@
...
+
+          section ->
+            {:section, section}
+        end
+
+      product ->
+        # Products are sections behind the scenes
+        {:section, product}
+    end
+  end
```
nicocirio, line 87: I've seen many handlers doing the same check as here to check if the given name matches a product, section, or project. So maybe we can have a handler_helpers file to reuse this code. 
On the other hand, we must be sure to provide unique names for products, projects, and sections when we define the YAML. 
For future iterations of the feature, we may want to check on that uniqueness restriction as a first step when we parse the YAML.

test/oli/scenarios/section_customization.scenario.yaml
```diff
@@ -0,0 +1,132 @@
+# SKIP: This scenario is currently being developed
```
nicocirio: We should remove this comment since this scenario is being considered when running `mix test test/oli/scenarios/scenario_runner_test.exs --trace`

test/support/scenarios/builder.ex
nicocirio, line 1: Should we use `Repo.insert` in the Builder, or should we rely on `insert(:something)` to use `factory.ex`?

test/support/scenarios/ops.ex
```diff
@@ -0,0 +1,455 @@
...
+    else
+      dest
+    end
+  end
+
+  defp revise!(dest, target, set_params) do
+    %{project: proj} = dest
+    rev = dest.rev_by_title[target]
+
+    if rev do
```
nicocirio: Here we start an `if` that ends up with many levels of nested `if`s, combined with `case`, etc. We might want to update our Claude rules to write this kind of logic in a more "elixir-way", maybe using `with` combined with helper functions.

nicocirio (changes requested):
This new approach for writing integration tests is really great! Looking forward to apply it combined with the new Claude `scenario.md`.

I left some notes regarding things we might need to change, and some other ideas for future iterations of this new feature.

Regarding the Builder, I sometimes experimented tests using Repo.insert failing randomly when running async with other tests. I have had better results using factories instead, but it is just an opinion.

test/support/scenarios/directives/enrollment_handler.ex
darrensiegel: yeah this is still under development and not needed at the moment

test/support/scenarios/README.md
darrensiegel: I went back and forth during the impl.  At first I had a mix of "source / target" and "from / to".  Since plan is to have Claude Code write all of the future scenarios based on natural language description - I wanted to keep the interface across directives and ops consistent and simple, so i settled on "from / to" everywhere

test/support/scenarios/directives/institution_handler.ex
darrensiegel: Yeah good catch, at the moment institution is not used (future directives might need it)

test/support/scenarios/directives/product_handler.ex
darrensiegel: Yes, good call.  I'll add a note to the CLAUDE.md file

test/support/scenarios/builder.ex
darrensiegel: Actually, it should use neither of those approaches.  The entire point of the Scenarios infra is that it uses the actual Oli infra structure code for creating and manipulating things in the system.  This is one area where it does not - and relies on manually creating a Project.  I'm going to change this so that it uses whatever the "Add Project" feature uses to create.

test/support/scenarios/directives/verify_handler.ex
darrensiegel: Good call, i'm going to add a "uniqueness restrition" check in the next round of work.  I've made a note of this.

test/support/scenarios/builder.ex
nicocirio: makes more sense 😄

nicocirio (approved):
Thanks for addressing those suggestions!

PR #5869, opened 2025-09-08
[FEATURE] [MER-4751] Add LTI tooltip icons

simonchoxx:
[MER-4751](https://eliterate.atlassian.net/browse/MER-4751)

This PR implements the display of the LTI icon and an informational tooltip on hover in the following sections:

- Assessment (Quiz) scores
- Scored Activities
- Practice Activities

Whenever an activity contains an LTI 1.3 external tool, its title will be accompanied by the LTI icon and a tooltip with the label “LTI 1.3 External Tool”.

Captura de pantalla 2025-09-08 a la(s) 10 00 51 a  m

Captura de pantalla 2025-09-08 a la(s) 10 00 39 a  m

[MER-4751]: https://eliterate.atlassian.net/browse/MER-4751?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/oli_web/components/delivery/practice_activities/practice_activities.ex
```diff
@@ -32,7 +33,14 @@ defmodule OliWeb.Components.Delivery.PracticeActivities do
   def update(assigns, socket) do
     params = decode_params(assigns.params)
 
-    {total_count, rows} = apply_filters(assigns.assessments, params)
+    lti_page_ids = Resources.section_resource_ids_with_lti_activities(assigns.section.id)
```
nicocirio: This query will be executed every time the component gets updated. I think we could move this to the `mount` callback to execute it only once

lib/oli_web/components/delivery/scored_activities/scored_activities.ex
```diff
@@ -567,11 +568,18 @@ defmodule OliWeb.Components.Delivery.ScoredActivities do
         Map.put(acc, resource_id, {total_attempts, avg_score})
       end)
 
+    lti_page_ids = Resources.section_resource_ids_with_lti_activities(section.id)
+
```
nicocirio: same suggestion here

lib/oli_web/components/delivery/quiz_scores/quiz_scores.ex
```diff
@@ -160,12 +161,19 @@ defmodule OliWeb.Components.Delivery.QuizScores do
 
     graded_pages = Oli.Delivery.Sections.SectionResourceDepot.graded_pages(section.id)
 
+    lti_page_ids = Resources.section_resource_ids_with_lti_activities(section.id)
+
```
nicocirio: same suggestion here

lib/oli/resources.ex
```diff
@@ -464,6 +464,45 @@ defmodule Oli.Resources do
...
   end
 
+  @doc """
+  Returns a list of resource ids that are LTI activities in the given section.
+  Example
+      iex> section_resource_ids_with_lti_activities(1)
+      [1, 2, 3]
+  """
+  @spec section_resource_ids_with_lti_activities(integer()) :: [integer()]
+  def section_resource_ids_with_lti_activities(section_id) do
```
nicocirio: The name of this function and/or the docs are confusing for me. We want, given a section_id, all page resource_ids that contain at least one LTI activity, right? If so, wdyt about `get_page_resource_ids_with_lti_activities(section_id)` and `Returns a list of resource IDs for pages containing LTI activities` for the docs.

lib/oli_web/components/delivery/practice_activities/practice_activities.ex
simonchoxx: Here, the proposed change can be applied

lib/oli_web/components/delivery/scored_activities/scored_activities.ex
simonchoxx: Not here, because the activities are obtained within the same component with the `get_activities/3` function.

lib/oli_web/components/delivery/quiz_scores/quiz_scores.ex
simonchoxx: Not here either, because the assessments are not provided by the parent liveview but are calculated in this same component with the function `get_grades_all_students/5`.

nicocirio: I might be wrong, but if we calculate lti_page_ids on `mount`, assign it to the socket, then in the  `get_grades_all_students/5` function we could grab it from the socket assigns. 
Actually, we could even go a step further; on `mount` we could calculate `graded_pages` and `lti_page_ids` to then calculate `graded_pages_with_lti_activity` and only assign the latter to the socket (since I believe that is the only one we need for the table model). With this refactor we would pass from calculating on every update `1 Depot fetch (graded_pages) + 1 query (lti_page_ids) + 1 post-processing (Enum.map for graded_pages_with_lti_activity)` to just doing it once on `mount`

nicocirio: This refactor is out of the ticket's scope, so we could definitely do it in a future tech-debt sprint

simonchoxx: I agree with everything you mention, and I also think we should work on it in a future tech debt sprint.

nicocirio (approved):
Thanks for addressing those comments! 😄

PR #5872, opened 2025-09-08
[ENHANCEMENT] [MER-4904] Student metrics assertions workflows

darrensiegel:
Towards adding `Oli.Scenarios` support for validating student performance metrics, this PR:

- Moves the bulk of the `Oli.Scenarios` impl into `lib/oli` in anticipation of production features that will use scenario directives (e.g. MCP server tools)
- Adds "objectives" and "tags" support during project creation
- Adds ability to target objectives and tags during activity creation
- Renames the `verify` directive to `assert` and implements `structure`, `resource`, `progress` and `proficiency` assertions
- Adds the `clone` directive to allow projects to be cloned
- Adds the `use` directive to allow scenario YAML files to be reusable (aka, like "imports" or "includes")
- Adds the `call` directive to allow one-off custom functions to be run to do things outside of the Scenario framework
- Implements a basic set of tests around cloned projects
- Implements a basic set of tests around activity bank fulfillment
- Implements basic progress and proficiency tests
- Adds `mix scenarios` to run all scenarios and `mix scenarios test/scenarios/one.scenario.yaml` to run exactly one

### DIFF
--- a/lib/oli/scenarios/directives/answer_question_handler.ex
+++ b/lib/oli/scenarios/directives/answer_question_handler.ex
@@ -88,57 +88,15 @@ defmodule Oli.Scenarios.Directives.AnswerQuestionHandler do
 
   # Find the ActivityAttempt for the given activity
   defp find_activity_attempt(attempt_state, activity_revision) do
-    # For inline activities, the resource_id in the revision might not match
-    # what's in the attempt_hierarchy. We need to find by matching the activity
-    # content or by iterating through all attempts.
-
-    # First try direct lookup by resource_id
     resource_id = activity_revision.resource_id
 
     case Map.get(attempt_state.attempt_hierarchy, resource_id) do
       nil ->
-        # If not found, try to find any activity attempt (for single activity pages)
-        # This is a simplification - in real scenarios we'd need better matching
-        case Map.values(attempt_state.attempt_hierarchy) do
-          [{%{attempt_guid: guid} = activity_attempt, part_attempts}] ->
-            # Single activity on page, use it with part_attempts included
-            activity_attempt_with_parts = %{
-              activity_attempt
-              | part_attempts: Map.values(part_attempts)
-            }
-
-            {:ok, %{attempt_guid: guid, activity_attempt: activity_attempt_with_parts}}
-
-          [%{attemptGuid: guid} = thin_info] ->
-            # Single activity on page (adaptive format)
-            {:ok, %{attempt_guid: guid, activity_attempt: thin_info}}
-
-          [] ->
-            {:error, "No activity attempts found in hierarchy"}
-
-          _ ->
-            # Multiple activities - need better matching logic
-            # For now, take the first one as a workaround
-            case List.first(Map.values(attempt_state.attempt_hierarchy)) do
-              {%{attempt_guid: guid} = activity_attempt, part_attempts} ->
-                activity_attempt_with_parts = %{
-                  activity_attempt
-                  | part_attempts: Map.values(part_attempts)
-                }
-
-                {:ok, %{attempt_guid: guid, activity_attempt: activity_attempt_with_parts}}
-
-              %{attemptGuid: guid} = thin_info ->
-                {:ok, %{attempt_guid: guid, activity_attempt: thin_info}}
-
-              _ ->
-                {:error, "Could not find matching activity attempt"}
-            end
-        end
+        # If not found by resource_id, search for matching activity by content
+        find_activity_by_content(attempt_state, activity_revision)
 
       # Basic page format: {%ActivityAttempt{}, part_attempts}
       {%{attempt_guid: guid} = activity_attempt, part_attempts} ->
-        # Include the part_attempts in the activity_attempt
         activity_attempt_with_parts = %{
           activity_attempt
           | part_attempts: Map.values(part_attempts)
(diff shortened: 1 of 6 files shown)


lib/oli/scenarios/directives/answer_question_handler.ex
```diff
@@ -0,0 +1,222 @@
...
+          [%{attemptGuid: guid} = thin_info] ->
+            # Single activity on page (adaptive format)
+            {:ok, %{attempt_guid: guid, activity_attempt: thin_info}}
+
+          [] ->
+            {:error, "No activity attempts found in hierarchy"}
+
+          _ ->
+            # Multiple activities - need better matching logic
+            # For now, take the first one as a workaround
```
nicocirio: Just to point out, this is a workaround Claude implemented, and we may want to improve the way to handle this case.

lib/oli/scenarios/directives/answer_question_handler.ex
```diff
@@ -0,0 +1,222 @@
...
+    # what's in the attempt_hierarchy. We need to find by matching the activity
+    # content or by iterating through all attempts.
+
+    # First try direct lookup by resource_id
+    resource_id = activity_revision.resource_id
+
+    case Map.get(attempt_state.attempt_hierarchy, resource_id) do
+      nil ->
+        # If not found, try to find any activity attempt (for single activity pages)
+        # This is a simplification - in real scenarios we'd need better matching
```
nicocirio: Just to point out, Claude is using a simplification to match the activity. I am not sure how this simplification may not apply to more complex cases.

lib/oli/scenarios/directives/answer_question_handler.ex
```diff
@@ -0,0 +1,222 @@
...
+        # Multiple choice responses need to be wrapped in brackets
+        {:ok, "#{response}"}
+
+      "oli_short_answer" ->
+        # Short answer is just the plain text
+        {:ok, response}
+
+      _ ->
+        # Default to plain response for unknown types
+        {:ok, response}
```
nicocirio: Here we have an oportunity to improve our Claude rules when writing a `case`; `oli_short_answer` and `_` both return the same tuple, so maybe we can remove `oli_short_answer` match and just use the catch-all one (`_`).

lib/oli/scenarios/directives/assert/proficiency_assertion.ex
```diff
@@ -0,0 +1,247 @@
...
+              summary.resource_id == ^objective.resource_id and
+              summary.resource_type_id == ^objective_type_id,
+          select: {
+            summary.num_first_attempts_correct,
+            summary.num_first_attempts,
+            summary.num_correct,
+            summary.num_attempts
+          }
+        )
+        |> Repo.all()
```
nicocirio: Instead of fetching the raw_data here, we may want to use the Metrics module (`raw_proficiency_per_learning_objective` seems to be a good candidate)

lib/oli/scenarios/directives/assert/structure_assertion.ex
```diff
@@ -0,0 +1,241 @@
...
+
+          section ->
+            {:section, section}
+        end
+
+      product ->
+        # Products are sections behind the scenes
+        {:section, product}
+    end
+  end
```
nicocirio, line 61: This helper function, as well as others, is used more than once across the `directives/assert` folder. To increase readability and simplify future refactors, we might want to declare them in a common helper file within that same folder

lib/oli/scenarios/directives/edit_page_handler.ex
```diff
@@ -0,0 +1,175 @@
...
+  defp validate_project_name(name) when is_binary(name), do: {:ok, name}
+  defp validate_project_name(_), do: {:error, "Project name must be a string"}
+
+  # Get project from state
+  defp get_project(project_name, state) do
+    case Map.get(state.projects, project_name) do
+      nil -> {:error, "Project '#{project_name}' not found"}
+      built_project -> {:ok, built_project}
+    end
+  end
```
nicocirio, line 62: These functions are the same as the ones we use on `lib/oli/scenarios/directives/activity_handler.ex`, so maybe they can be moved to a common helper functions module.

lib/oli/scenarios/directives/view_practice_page_handler.ex
```diff
@@ -0,0 +1,153 @@
...
+    end
+  end
+
+  # Get section from state
+  defp get_section(state, section_name) do
+    case Map.get(state.sections, section_name) do
+      nil -> {:error, "Section '#{section_name}' not found"}
+      section -> {:ok, section}
+    end
+  end
```
nicocirio, line 56: Here, we could use the functions defined on `Oli.Scenarios.Engine`
Those may need a minor refactor to return the tuples `{:ok, ...}` or `{:error, ...}`

lib/oli/scenarios/directives/view_practice_page_handler.ex
```diff
@@ -0,0 +1,153 @@
...
+      |> Map.values()
+      |> Enum.find(fn built_project ->
+        built_project.project.id == section.base_project_id
+      end)
+
+    case project do
+      nil -> {:error, "Source project for section not found"}
+      p -> {:ok, p}
+    end
+  end
```
nicocirio, line 107: Does it make sense to define these 2 functions on the `Oli.Scenarios.Engine` module? (After line 261, where we define the helper functions)

lib/oli/scenarios/hooks.ex
```diff
@@ -0,0 +1,219 @@
...
+
+  Example usage in YAML:
+    - hook:
+        function: "Oli.Scenarios.Hooks.delay_execution/1"
+  """
+  def delay_execution(%ExecutionState{} = state) do
+    Logger.info("Delaying execution for 1 second...")
+    Process.sleep(1000)
+    state
+  end
```
nicocirio, line 108: In a future iteration, we may want/need to define the number of milliseconds for `Process.sleep/1`

lib/oli/torus_doc/activities/short_answer_converter.ex
```diff
@@ -0,0 +1,69 @@
...
+        else
+          json
+        end
+
+      json =
+        if activity.tags && activity.tags != [] do
+          Map.put(json, "tags", activity.tags)
+        else
+          json
+        end
```
nicocirio, line 64: This is a pattern I've seen before in previous converter.ex files, so maybe it makes sense to have a helper function for them. Something like `def add_optional_fields(json, activity)` that returns the json with the optional tags and objectives added (if any)

lib/oli/torus_doc/activity_converter.ex
```diff
@@ -0,0 +1,181 @@
...
+      error -> error
+    end
+  end
+
+  defp convert_short_answer(activity) do
+    case Oli.TorusDoc.Activities.ShortAnswerConverter.convert(activity) do
+      {:ok, json} -> {:ok, json}
+      error -> error
+    end
+  end
```
nicocirio: These 2 functions have an unnecessary `case`, since we are returning the exact match without any post-processing.

lib/oli/torus_doc/activity_converter.ex
```diff
@@ -0,0 +1,181 @@
...
+  """
+  def generate_id do
+    "gen_" <> Base.encode16(:crypto.strong_rand_bytes(8), case: :lower)
+  end
+
+  @doc """
+  Convenience function that parses YAML and converts to Torus JSON in one step.
+  """
+  def from_yaml(yaml_string) do
+    alias Oli.TorusDoc.ActivityParser
```
nicocirio: Although this works, we generally put `alias` at the beginning of the module (this could be a rule we add to Claude)

lib/oli/torus_doc/activity_parser.ex
```diff
@@ -0,0 +1,157 @@
...
+    end
+  end
+
+  def parse_activity(_), do: {:error, "Invalid activity structure: expected a map"}
+
+  defp validate_activity_type(%{"type" => type}) when is_binary(type) do
+    cond do
+      type == "oli_multi_choice" -> {:ok, :mcq}
+      # Support both variants
+      type == "oli_multiple_choice" -> {:ok, :mcq}
```
nicocirio: These 2 could be merged into `type when type in ["oli_multi_choice", "oli_multiple_choice"] -> {:ok, :mcq}`.
I do not have a strong opinion about this suggestion, I just mention it for you to decide if you prefer this approach and update Claude rules.

lib/oli/torus_doc/activity_parser.ex
```diff
@@ -0,0 +1,157 @@
...
+    end
+  end
+
+  defp parse_hints(hints) when is_list(hints) do
+    Enum.map(hints, fn
+      %{"body_md" => body} -> %{body_md: body}
+      hint when is_binary(hint) -> %{body_md: hint}
+      _ -> nil
+    end)
+    |> Enum.reject(&is_nil/1)
```
nicocirio: I know that `hints` probably won't be a long list to refactor into a single `Enum.reduce`. I'm just pointing it out to consider updating our Claude rules to avoid nesting `Enum` functions when it can be achieved in a single `Enum.reduce`

lib/oli/torus_doc/page_parser.ex
```diff
@@ -0,0 +1,415 @@
...
+         "Activity block must have either 'activity' (inline), 'stem_md' (inline) or 'activity_id' (reference)"}
+    end
+  end
+
+  defp parse_inline_activity(activity_data, block_id, virtual_id) do
+    # Parse the activity definition using ActivityParser
+    # The block should contain all activity fields
+    alias Oli.TorusDoc.ActivityParser
+
+    case ActivityParser.parse_activity(activity_data) do
```
nicocirio, line 197: Just pointing out another example of an alias defined just before it is used instead of defining it at the beginning of the module (or simply/directly use `Oli.TorusDoc.ActivityParser.parse_activity(activity_data)`

PR #5882, opened 2025-09-12
[BUG FIX] [MER-4936] Google SSO rework

nicocirio:
This draft PR replicates the fix that was implemented in OliWeb.UserLoginLive in OliWeb.UserRegistrationLive.

On the other hand, both PRs implemented the fix for SSO in login and registration for Users. Shouldn't we apply the same fix for the Author's flow?


eliknebel:
> Shouldn't we apply the same fix for the Author's flow?

Good catch, yes I believe we should

nicocirio:
https://github.com/Simon-Initiative/oli-torus/pull/5884

PR #5891, opened 2025-09-15
[ENHANCEMENT] [MER-4957] Enhance LogicLab authoring and register outOf points

Ringenberg:
In order to better handle LogicLab activity content, this adds a file input element so that content authors can upload new activities generated in other tools instead of submitting them to the developers to manually add to the database.

It adds a stub for getting the maximum points available based on the selected activity.

It adds a toggle so authors can override a derivation activities tutoring flag which removes the need to have two copies of select activities in the database.

This requires the requested update of logiclab.oli.cmu.edu.

### DIFF
--- a/assets/src/components/activities/logic_lab/LogicLabAuthoring.tsx
+++ b/assets/src/components/activities/logic_lab/LogicLabAuthoring.tsx
@@ -21,7 +21,6 @@ import {
   AllActivityTypes,
   LabActivity,
   LogicLabModelSchema,
-  getLabServer,
   isLabActivity,
   maxPoints,
   translateActivityType,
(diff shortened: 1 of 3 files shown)


Ringenberg:
logiclab.oli.cmu.edu has been updated so that the activity information from it now includes the calculated maximumScore.  Activities already set in torus will have to be updated for this to take effect.
This can be done by selecting a different activity in the authoring dialog and then reselecting the original activity.

assets/src/components/activities/logic_lab/LogicLabAuthoring.tsx
```diff
@@ -5,86 +5,119 @@ As a full authoring interface is not yet available, this simple interface
...
+  }
   // Workaround limited MathJaxMinimal implementation.
   const math = window.MathJax as any;
-  if (typeof math.typeset === 'function') {
+  // For MathJax 3+ use typesetPromise if available (MathJax 4 only supports this).
+  if (typeof math.typesetPromise === 'function') {
+    math.typesetPromise().catch((err: unknown) => console.error('MathJax typeset failed', err));
+  }
+  // fallback to typeset if available.
+  else if (typeof math.typeset === 'function') {
```
andersweinstein: To avoid possible concurrency issues we have had in the past, Torus code should use this idiom for async MathJax typesetting:
```
 let lastPromise = getGlobalLastPromise();
 lastPromise = lastPromise.then(() => window.MathJax.typesetPromise());
 setGlobalLastPromise(lastPromise);
```
which makes use of torus utility functions to manage the one and only global promise used for asynchronous MathJax typesetting. 

I have made this change in the code. Also taken out fallback since torus uses only typesetPromise elsewhere.

PR #5894, opened 2025-09-16
[BUG FIX] [MER-4914] focus is reset and user is kicked out of input when typing in AA component panel

dtiwarATS:

**Advanced mode: Fix focus loss & debounce updates**
- Changes applied only in Expert Mode (`isExpertMode=true`)
- Wrapped the form inside a parent `<div>` to intercept and handle Backspace key presses
- Prevented focus loss when typing in inputs
- Delayed save calls with 500ms debounce for smoother updates while typing
- Backspace-triggered deletes now save only on blur instead of mid-typing
- Preserved existing behavior for Basic Mode (no changes)

### DIFF
--- a/assets/src/apps/authoring/components/PropertyEditor/PropertyEditor.tsx
+++ b/assets/src/apps/authoring/components/PropertyEditor/PropertyEditor.tsx
@@ -112,7 +112,7 @@ const PropertyEditor: React.FC<PropertyEditorProps> = ({
 
           setFormData(updatedData);
 
-          // If backspace triggered, skip calling handler until blur
+          // If backspace triggered skip calling handler until blur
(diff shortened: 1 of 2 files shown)


assets/src/apps/authoring/components/PropertyEditor/PropertyEditor.tsx
```diff
@@ -71,71 +76,96 @@ const PropertyEditor: React.FC<PropertyEditorProps> = ({
     setFormData(value);
   }, [value]);
 
-  return (
-    <Form
```
dtiwarATS, line 75: These are formatting changes showing up because we wrapped them around a `<div>`. Changes are only in the `onChange` handler

bsparks (approved):

PR #5899, opened 2025-09-16
[BUG FIX] [MER-4958] Skip enrollment check in LiveSession plug if section does not require enrollment.

manelli:
If section doesn't require enrollment and the user or author is logged in:
1) [If user is logged in] enrolls it to the section (this is needed for future steps like checking if the user has visited that section)
2) Assigns `is_enrolled: true` and continues

See: https://eliterate.atlassian.net/browse/TRIAGE-1882

### DIFF
--- a/lib/oli_web/live_session_plugs/require_enrollment.ex
+++ b/lib/oli_web/live_session_plugs/require_enrollment.ex
@@ -14,12 +14,11 @@ defmodule OliWeb.LiveSessionPlugs.RequireEnrollment do
         %{
           assigns: %{
             current_user: user,
-            current_author: author,
             section: %Sections.Section{requires_enrollment: false} = section
           }
         } = socket
       )
-      when not is_nil(user) or not is_nil(author) do
+      when not is_nil(user) do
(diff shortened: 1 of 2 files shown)


lib/oli_web/live_session_plugs/require_enrollment.ex
```diff
@@ -5,6 +5,24 @@ defmodule OliWeb.LiveSessionPlugs.RequireEnrollment do
...
+        _session,
+        %{
+          assigns: %{
+            current_user: user,
+            current_author: author,
+            section: %Sections.Section{requires_enrollment: false} = section
+          }
+        } = socket
+      )
+      when not is_nil(user) or not is_nil(author) do
```
eliknebel: Should we have `or not is_nil(author)` here? It seems like we should just let the function clause below handle that case to check if this is an admin that can access the section, no? Not all authors should be automatically designated as `is_enrolled: true`, they should have a valid user account to access the section

manelli: 👍 Fixed in 5b21c171157cf9707cdd34ad1246600c615e6539

eliknebel (approved):

PR #5900, opened 2025-09-17
[FEATURE] [MER-4701] Related activities for objectives and subobjectives

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-4701) to the ticket

https://github.com/user-attachments/assets/c5dfde6d-e999-4b7e-913d-a1b56728841b

Other features added: 
- Search through activities
- keep state of "learning objectives" for the "Back to Learning Objectives" link


lib/oli/delivery/sections.ex
```diff
@@ -5007,11 +5013,78 @@ defmodule Oli.Delivery.Sections do
...
-        })
-      end)
+      if include_related_activities_count do
+        # This logic won't be excecuted for the student dashboard.
+        # Calculate related activities count for each objective using a single query with same SQL logic as detail view
+        activity_type_id = Oli.Resources.ResourceType.id_for_activity()
+        objective_ids = Enum.map(objectives, & &1.resource_id)
+
+        # Single query to get all activities and their objective attachments
+        activities_with_objectives =
```
darrensiegel: In a large course section like chemistry this query is going to take longer than what we expect (there will be thousands of activities), which will produce noticeable lag in the UI.  

I think instead what we should add is a new attribute on `SectionResource` called "related_activities" which is populated during course section creation and publication update time with an array of activity resource ids.   This will then allow the `SectionResourceDepot` to be used in this code to instantly retrieve the learning objectives and subobjectives, and then derive the count in memory from the `related_activities` array.

lib/oli/delivery/sections.ex
```diff
@@ -5609,4 +5682,158 @@ defmodule Oli.Delivery.Sections do
...
+
+  @doc """
+  Gets all evaluated activities that have a specific objective attached to them in the given section
+  """
+  def get_activities_for_objective(section, resource_id) do
+    activity_type_id = Oli.Resources.ResourceType.id_for_activity()
+
+    # Query activities that have the objective in their JSONB objectives field
+    query =
+      from([rev: rev, sr: sr, s: s] in DeliveryResolver.section_resource_revisions(section.slug))
```
darrensiegel: Once you make the change described above to pre-calculate and store activity ids in the `related_activities` field of the SectionResoruce - this function becomes much simplier.  You can just pass in the list of resource_ids and resolve their revisions and go from there.

lib/oli/delivery/sections.ex
```diff
@@ -5609,4 +5682,158 @@ defmodule Oli.Delivery.Sections do
...
+
+      _ ->
+        "No question stem available"
+    end
+  rescue
+    _ -> "No question stem available"
+  end
+
+  defp extract_question_stem(content) when is_map(content) do
+    case Map.get(content, "stem") do
```
darrensiegel, line 5881: I believe we have this "extract text from stem" written already in another part of the codebase.  Please track down where that is, ensure it is in a "reusable" module and then use it here.

darrensiegel (changes requested):

lib/oli/delivery/sections.ex
```diff
@@ -4931,25 +4932,61 @@ defmodule Oli.Delivery.Sections do
...
+        %{
+          title: sr.title,
+          resource_id: sr.resource_id,
+          revision_id: sr.revision_id,
+          children: [],
+          related_activities: sr.related_activities || []
+        }
+      end)
+
+    # Since section_resources of type objective do not have their children
```
darrensiegel: We should change this - there really is no reason why the "children" attr of LO SectionResources should not be populated.  That would eliminate the DB query here.

lib/oli/delivery/sections/post_processing.ex
```diff
@@ -75,4 +81,112 @@ defmodule Oli.Delivery.Sections.PostProcessing do
...
+        end)
+        |> Enum.join(" ")
+
+      update_query = """
+        UPDATE section_resources
+        SET related_activities = CASE #{case_statements} END
+        WHERE id = ANY($1)
+      """
+
+      Repo.query(update_query, [section_resource_ids])
```
darrensiegel: If this query fails for some reason - we will have no idea.  We should at least be logging the error.

PR #5923, opened 2025-09-23
[BUG FIX] [MER-4693] Direct delivery course has several students with multiple accounts

nicocirio:
[Link](https://eliterate.atlassian.net/browse/MER-4963) to the ticket

When an LMS user tries to access a non-lms course enrollment, it will be redirected to this page:

Screenshot 2025-09-23 at 11 45 47 AM

### DIFF
--- a/lib/oli_web/live/lms_user_instructions_live.ex
+++ b/lib/oli_web/live/lms_user_instructions_live.ex
@@ -2,6 +2,7 @@ defmodule OliWeb.LmsUserInstructionsLive do
   use OliWeb, :live_view
 
   alias Oli.Delivery.Sections
+  alias Oli.VendorProperties
 
   on_mount {OliWeb.UserAuth, :mount_current_user}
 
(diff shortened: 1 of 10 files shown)


lib/oli_web/live/lms_user_instructions_live.ex
```diff
@@ -0,0 +1,66 @@
...
+              you use to access these courses
+              <i>
+                <%= if @lms_course_titles != [],
+                  do: "- #{Enum.join(@lms_course_titles, ", ")} -" %>
+              </i>
+              only works with single sign-on through your school’s LMS.
+            </p>
+            <p>
+              <strong><%= "#{@section_title}" %></strong>
+              requires you to use a login account. Please log out of your current account with the button below, visit the enrollment URL again, and create a new Torus account using either email and password, or Sign In With Google.
```
darrensiegel: "Torus" should not be hardcoded here - instead use a vendor property to get the application name.

PR #5926, opened 2025-09-24
[FEATURE] [MER-4666] Better search for "Browse all" type tables in Admin workspace

manelli:
- Expanded admin search to cover IDs, titles, owners, collaborators, institutions, and base products/projects across Projects, Sections, and Products, with 3-char minimum 
- Added themed span-based highlighting for matched terms.

https://github.com/user-attachments/assets/1b07c382-099a-47a3-8745-2d1007029a16

See: https://eliterate.atlassian.net/browse/MER-4666

### DIFF
--- a/assets/css/app.css
+++ b/assets/css/app.css
@@ -43,6 +43,19 @@ a.download::after {
   content: '\f019';
 }
 
+.search-highlight {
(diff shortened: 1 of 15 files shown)


nicocirio (approved):
🚀

PR #5927, opened 2025-09-24
[FIX] Removes prefetching logiclab in delivery.

Ringenberg:
In the delivery component, the probing of the logiclab url is probably unnecessary and thus removed.  This also greatly simplifies the code and is an extension of the work for #5891.  Most of the probing checks were to verify that ACTIVITY_LOGICLAB_URL was properly set but adding a hard-coded default value mitigates the need for these checks.  Eventually, the hard coded default should be removed, but that requires changes in deployment to guarantee that this environment variable is set.  If it is not set, the logiclab activity should probably not be made available as an activity option or some other indication that it is not set in the activity configuration page.


Ringenberg:
As noted in the comments, the LogicLab interface is not reactive thus needing to set minimum dimensions that work with the LogicLab's layout.

andersweinstein (approved):

PR #5930, opened 2025-09-25
[BUG FIX] [MER-4992] fix right-to-left text in mcq/CATA/ordering choice delivery

andersweinstein:
Choices in multiple choice, CATA and ordering questions were appearing not to respect the right-to-left text direction setting in preview and delivery. This has a severe impact on Arabic courses that rely on right-to-left text. 

This happened because the `dir="rtl"` attribute was being set at the wrong level, at the choice content itself (leading to it being set on, say, a `<p>` within which the choice content is rendered), but NOT the containing div that groups the relevant icon (e.g. radio button for mcq) and choice content as a unit. The choice content box winds up laid out within a flex box container that lays out left to right and sizes its children to fit their content. Because choice content is rendered within such a left-aligned, sized-to-fit layout box, it makes no difference that it is set to rtl within this box.

This occurred in the ChoicesDelivery component; the ChoicesAuthoring component uses an editor so is different. Ordering choices go through a different component supporting draggable items, but the issue is similar.

Fix is to set the attribute at the containing div level for each choice. This has the effect that the icon (radio button or check box or drag handle) associated with the choice now appears on the right, rather than the left, when rtl order is used, which seems appropriate. Note this allows a mix of right-to-left and left-to-right choices within the same choice list, which may be odd, but is allowed in the model, which has the text direction as an attribute of the individual choice.

Also adjusted style so text margin from mcq icon (radio button) is specified in a text-direction-neutral way. 

Old (these have right-to-left setting on choices):
Screenshot 2025-09-25 at 4 59 12 PM

New:
Screenshot 2025-09-25 at 5 02 33 PM

### DIFF
--- a/assets/src/components/activities/common/choices/delivery/ChoicesDelivery.modules.scss
+++ b/assets/src/components/activities/common/choices/delivery/ChoicesDelivery.modules.scss
@@ -23,7 +23,7 @@
 }
 
 .choicesChoiceContent {
-  margin-left: 0.6rem;
+  margin-inline-start: 0.6rem;
   flex-grow: 1;
 
   /* When inside a MCQ/CATA, we want the images near the checkbox, not centered on the page */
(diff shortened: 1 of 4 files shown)


andersweinstein:
Note ordering choices show as if in a numbered list with their current position prepended. In left-to-right it looks like this:
Screenshot 2025-09-26 at 11 57 54 AM

With this change, right-to-left choices now 
Screenshot 2025-09-26 at 11 59 26 AM

PR #5942, opened 2025-10-07
[BUG FIX] [MER-5013] Fix spacebar in input

darrensiegel:
This is the fix that @codex suggested to allow HTML5 videos to ignore spacebars when clicked inside a text input or text area.  I was unable to test this locally as I don't have an understanding of how to use minio and I could not override it to use the real AWS S3 dev bucket. 

We need a dev to verify that this this indeed fixes the problem.  To do this, create a page with multiple HTML5 video elements on it.  Also include a couple of single response questions.  Verify that from the actual student interface that you can type into the text inputs the spacebar and it does NOT cause the videos on the page to play or pause (if they are playing).  

Disable this and verify that the above steps DO lead to videos being played and paused.

### DIFF
--- a/assets/src/components/video_player/VideoPlayer.tsx
+++ b/assets/src/components/video_player/VideoPlayer.tsx
@@ -8,6 +8,7 @@ import {
   Player,
   PlayerState,
   ProgressControl,
+  Shortcut,
   TimeDivider,
 } from 'video-react';
 import { PointMarkerContext, maybePointMarkerAttr } from 'data/content/utils';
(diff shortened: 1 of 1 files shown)


andersweinstein:
Spacebar is disabled, but seems there are OTHER shortcuts to deal with as well. K key also functions as play/pause (I think in either case). Have to find out all of them.

andersweinstein:
Have added alternate solution via codex to suppress keyboard shortcuts while allowing keyboard events to get through. This depends on some details of this component:

The player allows client to include a Shortcut child component with parameters which allows for overriding they player's built-in shortcuts. This Shortcut component evidently only handles events when the player state is "active". Some state handling is then also installed to prevent the player from ever going active, so that no keyboard events are processed at all. The player's sense of "active" state should not affect anything else about it, per codex.

With this magic it does not actually matter which shortcuts are defined in the `<Shortcut>` component parameters, since they never get invoked anyway. However, it is still necessary to include the `<Shortcut>` component to override the built-in keyboard handling.

andersweinstein (approved):

PR #5943, opened 2025-10-07
[BUG FIX] [MER-5004] guard auto enrollment with check

eliknebel:
https://eliterate.atlassian.net/browse/MER-5004

Fixes an issue regarding a recent change where independent student user accounts may be automatically enrolled in LTI sections simply by visiting the section URL.

This fix puts the ensure_direct_delivery_enrollment_allowed check in place before auto enrolling.

However, I think there is still a fundamental issue with the approach here. Throughout the rest of the system, we display an "Enroll" dialog before enrolling any student into a course to prevent this sort of inadvertent enrollment from occurring and as a confirmation/signal to users before joining a course. It seems to me that we should be doing that here as well.

For the hotfix I simply addressed the immediate issue but we may want to revisit how this is actually working in the future.

### DIFF
--- a/lib/oli_web/live_session_plugs/require_enrollment.ex
+++ b/lib/oli_web/live_session_plugs/require_enrollment.ex
@@ -4,7 +4,9 @@ defmodule OliWeb.LiveSessionPlugs.RequireEnrollment do
   import Phoenix.Component, only: [assign: 2]
   import Phoenix.LiveView, only: [redirect: 2, put_flash: 3]
 
+  alias Oli.Accounts.User
   alias Oli.Delivery.Sections
+  alias Oli.Delivery.Sections.Section
   alias Lti_1p3.Roles.ContextRoles
 
   def on_mount(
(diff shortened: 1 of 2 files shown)


lib/oli_web/live_session_plugs/require_enrollment.ex
```diff
@@ -13,14 +15,22 @@ defmodule OliWeb.LiveSessionPlugs.RequireEnrollment do
...
-            current_user: user,
-            section: %Sections.Section{requires_enrollment: false} = section
+            current_user: %User{} = user,
+            section: %Section{requires_enrollment: false} = section
           }
         } = socket
-      )
-      when not is_nil(user) do
-    if user, do: Sections.enroll(user.id, section.id, [ContextRoles.get_role(:context_learner)])
-    {:cont, assign(socket, is_enrolled: true)}
```
darrensiegel, line 23: Something isn't right here.  The is a missing `Sections.has_enrollment?` call that is present in the 31.8 branch.  I'm concerned that we do not see that here in the diff.

darrensiegel: We have to have that has_enrollment guard - otherwise we expose ourselves again to the "automatic downgrade of instructor to student" issue

darrensiegel (changes requested):

lib/oli_web/live_session_plugs/require_enrollment.ex
```diff
@@ -13,14 +15,26 @@ defmodule OliWeb.LiveSessionPlugs.RequireEnrollment do
...
+      ) do
+    if Sections.has_enrollment?(user.id, section.slug) do
+      {:cont, assign(socket, is_enrolled: true)}
+    else
+      with :ok <- Sections.ensure_direct_delivery_enrollment_allowed(user, section) do
```
darrensiegel: It seems like the current logic for `ensure_direct_delivery_enrollment_allowed` is going to allow DD users to enroll in an LTI section:

```  
def ensure_direct_delivery_enrollment_allowed(_user, %Section{open_and_free: false}), do: :ok

  def ensure_direct_delivery_enrollment_allowed(nil, %Section{}), do: :ok

  def ensure_direct_delivery_enrollment_allowed(%User{} = user, %Section{open_and_free: true}) do
    if user.independent_learner do
      :ok
    else
      {:error, :non_independent_user}
    end
  end
```

darrensiegel (changes requested):

PR #6260, opened 2026-03-02
[FEATURE] [MER-4943] Enable/Disable DOT on an Adaptive Page Level

manelli:
Add page-level DOT control for both adaptive pages

- Added `ai_enabled` to:
  - `revisions`
  - `section_resources`
- Added migration with graded-based backfill defaults:
  - scored (`graded=true`) -> `ai_enabled=false`
  - practice (`graded=false`) -> `ai_enabled=true`

- Section resource creation/update/migration paths now copy and keep `ai_enabled` in sync with pinned revisions (major/minor update paths included).

- Added page option checkbox: **Enable AI Assistant (DOT)** in Page Options modal.
- Added adaptive lesson panel toggle (under appearance settings) for DOT enablement.

---

## How To Create A DOT Adaptive Page

1. **Enable section assistant (as admin)**
   - Go to `/sections/<section_slug>/manage`
   - Turn on **Enable AI Assistant**

2. **Create adaptive page**
   - In authoring, create an **Adaptive Page**
  
3. **Enable page DOT + Navigation**
   - Open adaptive page editor
   - Right panel -> **Lesson**
   - Turn on:
     - **Enable AI Assistant (DOT)**
     - **Display Torus Navigation** (required for DOT UI visibility right now)

4. **Publish + update section**
   - Publish project changes
   - Apply/update section to latest publication

5. **Verify**
   - Open `/sections/<section_slug>/adaptive_lesson/<page_slug>`
   - DOT should be visible if both section and page toggles are enabled.

https://github.com/user-attachments/assets/5787f591-2cd3-48da-8fe0-bf2bbd89a634

See: https://eliterate.atlassian.net/browse/MER-4943


nicocirio:
> ### AI Review — performance
> ### Unnecessary page-context lookup when assistant is disabled at section level
> file: lib/oli_web/live/dialogue/window_live.ex line: 145 Description: `page_ai_context/3` is executed before checking `Sections.assistant_enabled?(section)`, so every mount does `SectionResourceDepot.get_section_resource/2` and parsing work even when section-level assistant is off. This adds avoidable per-request overhead on a hot LiveView path. Suggestion: Short-circuit on section-level flag first (for example, branch before calling `page_ai_context/3`), and only compute page context when `Sections.assistant_enabled?(section)` is true.
> 
> ### Duplicate assistant eligibility computation in the same render
> file: lib/oli_web/components/layouts/student_delivery_lesson.html.heex line: 136 Description: `Oli.Delivery.Sections.assistant_enabled_for_page?/2` is called in both the layout class expression and again for conditional rendering, repeating the same checks during a single render cycle. Suggestion: Compute the predicate once (for example, assign a local like `assistant_available?`) and reuse it in both places to avoid redundant work.

I think we could consider these 2 AI suggestions, especially the first one

nicocirio (approved):
🚀

PR #6428, opened 2026-04-08
[FEATURE] [MER-5344] lambda impl of eval engine

darrensiegel:
This PR migrates the `Simon-Initiative/authoring-eval` service into Torus as an AWS Lambda implementation, replacing the separate Docker/Express deployment path (the HTTP REST impl) with a Lambda-hosted evaluator that now lives directly in this repository. The migrated runtime is under `assets/src/eval_engine`, retains vm2 for the initial migration, and is wired into Torus through the `VariableSubstitution.LambdaImpl` path.

For packaging and deployment, this PR adds the `eval-bundle` build flow in `assets/package.json` and `assets/scripts/eval-bundle.js`. Running that script produces `priv/node/eval.zip`, which is the artifact intended for AWS Lambda deployment. The zip includes the built `eval.js` bundle, the Lambda entrypoint wrapper, and the required vm2 sidecar files.

This was tested locally against both practice and graded pages and is working well. If you want to exercise the live Lambda path during code review, contact me for the AWS credentials needed so your local Torus instance can invoke the function.

## Enabling the Lambda Eval Engine

The current implementation has been deployed in CMU AWS infrastructure as function `eval` in region `us-east-1`.  To use it, enable the Lambda-backed evaluator instead of the legacy HTTP REST implementation by setting these in `oli.env`:

```
VARIABLE_SUBSTITUTION_PROVIDER=LambdaImpl
VARIABLE_SUBSTITUTION_LAMBDA_FN_NAME=eval
VARIABLE_SUBSTITUTION_LAMBDA_REGION=us-east-1
```
  
To go back to the REST-based implementation, set:

```
VARIABLE_SUBSTITUTION_PROVIDER=RestImpl
```


eliknebel (commented):
Tested and works as expected. Changes look good, but my only comment here would be that our other lambda function implementations live under the cloud/ directory.

Since this is also a more standalone implementation from the typical torus assets, perhaps it would be better there as well to keep package.json dependencies and concerns isolated.

eliknebel (approved):

PR #6430, opened 2026-04-09
[FEATURE] [MER-5256] Add student support parameter modal UI

nicocirio:
## Summary

Adds section-scoped Student Support parameter customization, including persistence, save/reprojection flow, and the modal UI for editing inactivity and bucket thresholds.

## Changes

- add section-scoped persistence for Student Support parameter settings
- add save flow and Student Support reprojection after parameter updates
- add Student Support parameter modal UI and matrix interaction hook
- sync shared high-progress threshold between Struggling and Excelling
- add backend, LiveView, and hook test coverage

## Jira

- https://eliterate.atlassian.net/browse/MER-5256

### Change `last visit` threshold

https://github.com/user-attachments/assets/3c3cab40-1233-4a19-b2e5-bce6f612592e

### Change thresholds using numeric inputs

https://github.com/user-attachments/assets/cf655ea3-63ea-4a62-9212-2229d2bbc391

### Change thresholds by dragging limits
 

https://github.com/user-attachments/assets/8626c4ca-b4f3-4568-b770-e7e6e1324fb3

### Change thresholds using arrow keys (left/right for vertical limits, up/down for horizontal limits)

https://github.com/user-attachments/assets/0ce81b47-a697-4e57-be63-49a47b3316c8

### Light mode

https://github.com/user-attachments/assets/3e313fde-dd40-43f4-9693-3f166ec69a03


nicocirio:
On the projection settings lookup: these settings are section-scoped editable configuration rather than oracle payload. For example, an instructor can change the threshold values in the modal and then save, and the reprojection needs to use those newly persisted section settings immediately. We intentionally resolve the latest persisted settings in the projection path so save/reprojection reflects the current section truth. Moving that resolution fully upstream would be a separate architectural pass rather than a follow-up fix in this PR.

nicocirio:
On the authorization point: this is already enforced at the LiveView boundary. The instructor dashboard route is mounted behind authenticated instructor/admin access, and Mount.for/2 rejects unauthorized users before this event can run. The save path also persists against socket.assigns.section.id rather than a client-supplied section id.

nicocirio:
This is intentional in the current design: the visualization is meant to represent individual students, and the UI copy explicitly states that each dot represents one student. Capping, sampling, or aggregating into a heatmap/bin view would change the semantics of the visualization, so we are not taking that in this PR. If this proves to be a real performance issue at larger scales, we can evaluate alternative representations in a future iteration, but that would likely need to be worked through together with design.

eliknebel (approved):

PR #6433, opened 2026-04-09
[FEATURE] [MER-5260] Allow specific units to be excluded from item numbering

andersweinstein:
**Summary**
  This PR implements [MER-5260 ](https://eliterate.atlassian.net/browse/MER-5260)by allowing templates to select specific top-level Units to be displayed unnumbered while keeping curriculum numbering enabled elsewhere, so that units for things like introductory or orientation material can be used without affecting numbering in the substantive course content. 

  The feature is section/template-scoped via `unnumbered_unit_ids`. It does not change canonical curriculum ordering or structural numbering. Instead, it adds a display-only numbering overlay so special Units like introductions or prefatory material can appear without a Unit number, and numbers shown on later Units/Modules compress accordingly for display purposes .
  
**What Changed**
  Data model and settings:
  - Added `unnumbered_unit_ids `to sections/templates
  - Added validation to ensure only top-level Unit resource ids from the current template/section can be stored

  Template Overview UI:
  - Added Exclude the following units to the Template Overview content settings
  - Reused Scored Page multi-selection dropdown from Certificate Settings for consistency
  - Minor polish: selected-item chip x icons changed to render in white for better contrast
  - Built the selector from top-level Unit section_resource records

  Display-numbering infrastructure:
  - Added lib/oli/delivery/sections/display_numbering.ex to decorate hierarchies with display_numbering
  - Added a fast path so the normal case (`unnumbered_unit_ids == []`) returns the hierarchy unchanged
  - Added lib/oli/delivery/sections/display_labels.ex to centralize display-aware label/title formatting below the web
    layer

  Delivery/rendering updates:
  - Standard delivery helpers now use display-aware numbering/titles
  - Student delivery surfaces were updated to respect unnumbered-unit subtrees while preserving page numbering behavior where intended:
      - Home / Upcoming Agenda / My Assignments
      - Learn page Outline and Gallery
      - lesson popup outline
      - scored page headers (prologue/review)
      - Explorations
      - Practice Pages
      - standard container/outline views

  **Behavior**
  When global curriculum numbering is enabled:

  - Units listed in unnumbered_unit_ids display without a Unit prefix
  - numbering of any containers inside those unnumbered Unit subtrees is also suppressed
  - numbering of later visible Units and Modules compresses for display
  - page numbering remains unchanged

  Example:

  - Introduction: Preparatory information
  - Unit 1: First Substantive Unit
  - Unit 2: Second Substantive Unit

  instead of:

  - Unit 1: Introduction: Preparatory information
  - Unit 2: First Substantive Unit
  - Unit 3: Second Substantive Unit

  **Testing**
  Added and updated regression coverage for:
  - section persistence/validation
  - Template Overview save/load behavior
  - display-numbering overlay logic
  - standard delivery rendering
  - Learn outline/gallery
  - Home page tiles
  - lesson outline/header
  - scored-page prologue/review
  - Explorations / Practice Pages


nicocirio (changes requested):
I tested this locally from the UI.

On the authoring/template side, the feature appears to work as expected:

- I can check and uncheck `Display curriculum item numbers`
- When that setting is enabled, I can choose which units to exclude
- The selected units persist correctly on the template

I was not able to complete end-to-end validation in delivery, because creating a course section from the configured template failed and blocked further testing.

This is the error that occurred when trying to create a section from that template:

```text
Failed to create new section: #Ecto.Changeset<... errors: [unnumbered_unit_ids: {"can only be set after section resources are created", []}], ...>
```

Image

From the logs, this seems related to `unnumbered_unit_ids` being validated before the new section has its own section resources created.

Hypothesis only: when the section is created from the template, the template's `unnumbered_unit_ids` may be getting copied into the new section too early in the flow. If that happens before the new section's top-level unit resources exist, the validation would fail with:

```text
unnumbered_unit_ids: {"can only be set after section resources are created", []}
```

We could then add a test for “create section from configured template with unit exclusions” to cover this case.

andersweinstein:
Hmm, I think I tested with the new template preview function only (since that is quick to use), will investigate.

andersweinstein:
Fixed by removing validation step from section creation. Units are validated at authoring time, and delivery-time application only uses the unit  ids if they are recognized top-level units in the section, so does not seem necessary.

I realize there is a data hygiene issue, the stored unit ids are never cleaned up if the referenced units are removed from curriculum. So database can continue to hold stale unit ids, though their presence is harmless. I don't think this is a big worry since courses will be stable by the time folks want to use this. But could perhaps be cleaned up later.

nicocirio (approved):

PR #6436, opened 2026-04-10
[ENHANCEMENT] [MER-5551] Add module level logging to torus admin logging and feature flags

eliknebel:
https://eliterate.atlassian.net/browse/MER-5551

Adds admin-managed module-level runtime log overrides on `/admin/features`.

This change lets authorized admins apply and clear a Logger override for a specific loaded Elixir module without changing the global application log level.

This observability is necessary to help diagnose future LTI launch issues as well as other issues that may occur in production without flooding the entire system with logs.

# What Changed

- Added `Oli.RuntimeLogOverrides` for module-level override validation, apply, clear, and state listing.
- Added `Oli.RuntimeLogOverrides.Registry` as a supervised local registry for active module overrides.
- Wired the registry into application supervision.
- Extended `OliWeb.Features.FeaturesLive` with:
  - a module override form
  - active override rendering
  - clear actions
  - safer global log-level parsing using `String.to_existing_atom/1`
- Added backend tests for module override set/list/clear behavior and validation failures.
- Added LiveView tests for:
  - admin visibility
  - invalid module handling
  - clear behavior
  - non-admin access denial
  - regression coverage for the existing global logging control
- Updated the work-item spec pack to reflect the final module-level delivery.

### DIFF
--- a/docs/exec-plans/current/module-level-log-controls/fdd.md
+++ b/docs/exec-plans/current/module-level-log-controls/fdd.md
@@ -1,0 +1,248 @@
+# Module-Level Log Controls - Functional Design Document
+
+## 1. Executive Summary
+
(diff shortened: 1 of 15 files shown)


darrensiegel (approved):
This is a great addition.  But we just have to remember the limitation here - this appears to be single node only and not cluster wide.  So when / if we migrate to a clustered set up, this will have to be extended to work properly

PR #6440, opened 2026-04-13
[FEATURE] [MER-5506] Template update badge and banner

andersweinstein:
• ## Summary

 Implements [MER-5506](https://eliterate.atlassian.net/browse/MER-5506) adding template update indicators to template overview and workspace navigation.

  This change introduces a dismissible update banner on both template overview pages and adds a workspace sidebar badge when a project has templates with available updates. The sidebar badge reflects the number of templates with updates, while each individual template overview page shows the exact number of updates available for that template.

  ## What Changed

  - Added a dismissible template updates banner to both template overview implementations:
      - workspace route
      - admin/browse-all-templates route
  - Added session-scoped banner dismissal using a LiveView hook backed by sessionStorage
  - Added workspace sidebar badges for template updates:
      - badge appears on Publish when the submenu is closed
      - badge appears on Templates when the submenu is expanded
      - badge is a project-level signal independent of the currently displayed page.
  - Kept page-level and sidebar-level count semantics distinct:
      - overview page banner: exact number of available updates for that template
      - workspace sidebar badge: number of templates in the project that have at least one available update   
  - Added a per-row update indicator on the template list pages so authors can quickly see which templates have
    pending updates without opening each template overview first.
      - shown on both workspace route: Publish -> Templates & Admin -> Browse All Templates
      - the indicator appears next to the title of templates with available updates and is computed in a single
        batch query for the currently visible page of results

  ## Testing
  mix test test/oli_web/components/delivery/layouts_test.exs \
    test/oli_web/live/workspaces/course_author/products/details_live_test.exs \
    test/oli_web/live/products/details_view_test.exs

[MER-5506]: https://eliterate.atlassian.net/browse/MER-5506?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli/delivery/sections.ex
+++ b/lib/oli/delivery/sections.ex
@@ -3409,17 +3409,26 @@ defmodule Oli.Delivery.Sections do
   """
   def count_available_blueprint_updates(%Project{id: project_id}) do
     from(s in Section,
+      as: :section,
       where: s.base_project_id == ^project_id and s.type == :blueprint and s.status == :active,
-      select: s
+      join: spp in SectionsProjectsPublications,
+      as: :spp,
+      on: spp.section_id == s.id,
+      join: current_pub in Publication,
+      on: current_pub.id == spp.publication_id,
+      inner_lateral_join:
+        latest_pub in subquery(
+          from(p in Publication,
+            where: p.project_id == parent_as(:spp).project_id and not is_nil(p.published),
+            order_by: [desc: p.published, desc: p.id],
+            limit: 1
+          )
+        ),
+      on: true,
+      where: current_pub.id != latest_pub.id,
+      select: count(s.id, :distinct)
     )
-    |> Repo.all()
-    |> Enum.reduce(0, fn blueprint, total ->
-      if map_size(check_for_available_publication_updates(blueprint)) > 0 do
-        total + 1
-      else
-        total
-      end
-    end)
+    |> Repo.one()
   end
 
   @doc """
--- a/lib/oli/delivery/sections.ex
+++ b/lib/oli/delivery/sections.ex
@@ -3432,6 +3432,40 @@ defmodule Oli.Delivery.Sections do
   end
 
   @doc """
+  Returns a set of blueprint section ids from the given list that have at least
+  one available publication update.
+  """
+  def blueprint_ids_with_available_updates([]), do: MapSet.new()
+
+  def blueprint_ids_with_available_updates(blueprints) when is_list(blueprints) do
+    section_ids = Enum.map(blueprints, & &1.id)
+    project_ids = Enum.map(blueprints, & &1.base_project_id) |> Enum.uniq()
+
+    latest_publications =
+      from(p in Publication,
+        where: p.project_id in ^project_ids and not is_nil(p.published),
+        distinct: p.project_id,
+        order_by: [asc: p.project_id, desc: p.published, desc: p.id],
+        select: %{project_id: p.project_id, id: p.id}
+      )
+
+    from(s in Section,
+      where: s.id in ^section_ids,
+      join: spp in SectionsProjectsPublications,
+      on: spp.section_id == s.id,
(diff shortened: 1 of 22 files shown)


andersweinstein:
Codex replies:
```
  Performance: badge-count query in mount / lateral join / per-page recomputation
  I’m leaving this as-is for this PR. The current implementation removed the original N+1 issue and keeps the extra
  work bounded: the workspace badge uses one count query for the current project, and the template-list indicator
  uses one batch query for the currently visible page of templates rather than per-row or system-wide work. I’d
  prefer to revisit deeper optimization here with profiling/explain data in a follow-up if this path shows up as
  hot.

  UI: banner hidden by default and shown by JS
  Not taking this in this PR. Rendering the banner hidden by default was an intentional fix for the visible flash
  on reload after dismissal, which we confirmed in the UI. Reverting that would bring the flash back, so I’m
  keeping the current no-flash behavior.

  UI: banner copy is hard-coded / not localized
  Deferring for now. The localization point is valid, but it widens the scope of this feature beyond the update-
  indicator work, so I’d prefer to handle i18n alignment for this surface in a dedicated follow-up.

  Elixir: unsafe Remix.init/2 pattern match
  Not taking this in this PR. Remix.init/2is already used elsewhere with the same{:ok, state}match, and the current
  implementation flows through an initialization path that returns{:ok, %State{}}. I don’t see a concrete failing
  case introduced by this change.
```

lib/oli/delivery/sections.ex
```diff
@@ -3403,6 +3403,68 @@ defmodule Oli.Delivery.Sections do
...
   end
 
+  @doc """
+  Counts active blueprint sections for the given project that have at least one
+  available publication update.
+  """
+  def count_available_blueprint_updates(%Project{id: project_id}) do
+    from(s in Section,
+      as: :section,
+      where: s.base_project_id == ^project_id and s.type == :blueprint and s.status == :active,
```
nicocirio, line 3413: We are limiting the count to `active` blueprints, which makes sense.
But then, when we list the templates, `deleted` blueprints have the icon indicator next to them.

In the screenshot, we count 14 templates in the menu (deleted is excluded), but in the list we show 15 with the icon (see the deleted `Elixir Product 2`).

I think the icon should be excluded from `deleted` templates, to make the count match. WDYT?

Image

nicocirio (commented):
Nice work.
I tested it locally and found two things that we might want to consider:

1. Should the badge disappear when we navigate to another page, rather than the 'Templates' one?

https://github.com/user-attachments/assets/5d2890ef-a095-45e1-9021-e399b12f3562

2. The badge count does not match the count of icons in the list (the badge excludes `deleted` and the list does not).

lib/oli/delivery/sections.ex
andersweinstein: Good catch! I will check with Jess about this but I tend to agree the counts should match and we don't have to show the update indicator on deleted templates.

andersweinstein:
> 1. Should the badge disappear when we navigate to another page, rather than the 'Templates' one?

Thanks! That's a bug, now fixed.

nicocirio:
> > 1. Should the badge disappear when we navigate to another page, rather than the 'Templates' one?
> 
> I think the badge should stay to remind you that the project has Templates with updates even if you are on another page in the project. If you had an email folder with new messages indicated by a badge, it would show even if you were not on that folder. The purpose is to force you to notice that templates need updates, and this is a project-wide property. Remember also the badge is supposed to show on "Publish" when Publish is not expanded.
> 
> (could not play uploaded video, something seems wrong).

@andersweinstein ok, so then we should fix that, since the badge disappears when you navigate to other Publish subsection rather than the templates one

nicocirio (approved):

PR #6443, opened 2026-04-13
[BUG FIX] [MER-5548] REWORK - Fix author account linking redirect 500

eliknebel:
https://eliterate.atlassian.net/browse/MER-5548

Fix an issue identified in QA where redirecting a user after linking an authoring account resulted in a 500 error. This fix is 2-fold:
1. If no return to path is specified, the redirect will default to the account settings page to prevent 500.
2. The link authoring account feature now tracks the local route where a user came from and redirected them back to that page instead of always landing them back on the account settings page, which really no longer makes sense.

### DIFF
--- a/lib/oli_web/author_auth.ex
+++ b/lib/oli_web/author_auth.ex
@@ -30,7 +30,7 @@ defmodule OliWeb.AuthorAuth do
     # Get the return path AFTER account linking to respect any changes made
     author_return_to =
       build_redirect_path(params["request_path"]) ||
-        get_session(conn, :author_return_to) ||
+        build_redirect_path(get_session(conn, :author_return_to)) ||
         signed_in_path(conn)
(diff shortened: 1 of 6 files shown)


eliknebel:
1. "Cancel" is fine here and is better than the previously misleading label "Back to Account Settings"
2. Plug.Conn normalizes session keys to strings for both put_session/3 and get_session/2, so these are equivalent:
```
  put_session(conn, "author_return_to", path)
  get_session(conn, :author_return_to)
```
3. The impact is overstated. Phoenix redirect(to: ...) explicitly rejects protocol-relative paths like "//". So this is not an open redirect to an external origin. The actual failure mode is another server-side crash / ArgumentError. Crash case has been addressed.

darrensiegel (approved):

PR #6448, opened 2026-04-14
[FEATURE] [MER-5249] Summary Tile & AI Recommendation

simonchoxx:
[MER-5249](https://eliterate.atlassian.net/browse/MER-5249)

## Summary

This PR adds the Summary Tile to the Instructor Intelligent Dashboard. The tile gives instructors a scoped summary of the current dashboard context and an AI recommendation, while keeping the UI resilient to partial or delayed data.

### What changed

- Added a new Summary Tile at the top of the dashboard, directly below the global scope filter.
- Rendered scoped summary metrics for:
    - average student progress
    - average class proficiency
    - average assessment score
- Kept each metric card optional, so the tile can render partial data without showing empty placeholders.
- Added a normalized recommendation view model so the LiveView does not depend on provider-specific
  payloads.
- Supported recommendation states for:
    - loading / thinking
    - beginning-of-course fallback
    - ready / populated
    - unavailable / failure-safe
- Added thumbs up, thumbs down, and regenerate controls for the recommendation.
- Preserved the current recommendation while a regenerate request is in flight.
- Added server-side protection so sentiment and regenerate actions are rejected while the tile is already
  busy.
- Added projection and LiveView tests for the new tile behavior and the interaction state transitions.

### Notes

- The implementation keeps data shaping in the dashboard/projection layer instead of the HEEx templates.
- The PR includes the UI surface, recommendation plumbing, and the state handling needed to keep the tile
  stable during scope changes and async updates.
- Follow-up feedback around broader task supervision and additional copy/i18n cleanup was kept out of
  this slice to avoid widening the scope unnecessarily.

### Verification

- `mix test test/oli_web/live/delivery/instructor_dashboard/intelligent_dashboard_tab_test.exs`
- `mix format` on the touched Elixir files

Captura de pantalla 2026-04-15 a la(s) 12 49 37 p  m
&nbsp;
Captura de pantalla 2026-04-15 a la(s) 12 49 23 p  m

[MER-5249]: https://eliterate.atlassian.net/browse/MER-5249?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


simonchoxx:
Reply to the elixir review comment: I implemented the server-side in-flight check for sentiment and also hardened regenerate so it returns `:not_allowed` instead of raising when the tile is already busy. The generic `handle_event/3` fallback is already present in `instructor_dashboard_live.ex`. The broader Task supervision / bounded concurrency change is still deferred to a separate follow-up, because that affects the async orchestration model rather than just the guard.

simonchoxx:
Reply to the ui review comment: I am deferring the i18n cleanup for the loading/unavailable strings in the validation payload. Those strings live in the dev-only dashboard validation UI, not in the user-facing summary tile component, so I kept the scope to the server-side interaction fix in this pass.

simonchoxx:
Reply to the performance review comment: I am deferring the `inspect/2` payload cleanup and the async task orchestration change. The `summary_text`/validation payload here is still part of the prototype/dev-only surface, and the Task supervision change is a broader refactor than this review fix. I kept the scope to the server-side guard that was low-risk and directly tied to the reported race.

simonchoxx:
Regarding the requirements comment: the summary metric tooltips already satisfy AC-004. The triggers use aria-describedby, and the tooltip is revealed on hover and keyboard focus via group-hover and group-focus-within. The current test coverage verifies the accessible wiring; if you want a browser-level hover/focus smoke test as well, I would treat that as a follow-up.

simonchoxx:
Update: the recommendation interaction flow is now using supervised/monitored tasks instead of fire-and-forget. Both regenerate and sentiment now clean up correctly if the task exits early, so the UI no longer gets stuck in an in-flight state. I also changed the summary projector average calculation to a single-pass reducer.

nicocirio (commented):
Awesome work! Before merging the PR, I consider it is worth addressing the following AI comments:

1. Repeated Map.keys/1 list allocation in the same update pipeline (and maybe refactor maybe_put_dashboard_field/4 that is piped 13 times in the same pipeline, one for each dashboard field)
2. Progress aggregation can crash on missing progress_pct
3. Silent Failure On User Actions

nicocirio (approved):
🚀

PR #6455, opened 2026-04-15
[ENHANCEMENT] [MER-5469] [MER-5471] [MER-5475] [MER-5477] [MER-5479] [MER-5481] [MER-5482] Assessment settings and student exception Scenario tests

darrensiegel:
This PR adds infrastructure and `Oli.Scenarios` support to then be able to execute YAML scenario integration tests for the full collection of Assessment Settings and Student Exception features.   

New YAML tests are included here exercise:
- Available date
- Due date
- Late submission policies
- Timed assessment (both auto submit and submit late)
- Retake mode
- Review mode (view feedback, view answers)
- Scoring strategies (best, average, most recent)
- Max attempts
- Password protection
- Student exceptions for due date, max attempts, timer

Also introduced here is support for annotating scenario tests that **only** run during the nightly run against `master`.  There are two tests that use this new directive, two that exercise timed assessment tests.  These tests literally wait 2 minutes for the auto submit timer to fire, and so we don't want these running on every PR.


lib/oli_web/live/delivery/student/utils.ex
```diff
@@ -192,6 +192,15 @@ defmodule OliWeb.Delivery.Student.Utils do
...
   def page_terms(assigns) do
+    terms =
+      Oli.Delivery.Page.PrologueTerms.build(
+        assigns.effective_settings,
+        assigns.ctx,
+        assigns.has_scheduled_resources?
+      )
+
+    assigns = assign(assigns, :terms, terms)
+
```
nicocirio: `page_terms/1` recalculates `PrologueTerms.build/3` on every LiveView re-render. Since these terms are already derived in `Oli.Delivery.Page.PrologueState.build/4`, it would be better to extend `OliWeb.LiveSessionPlugs.InitPage.prologue_assigns/2` to assign `terms` once and pass them into the component, so the component stays presentational and avoids unnecessary recomputation.

nicocirio (commented):
I manually exercised the instructor flows for editing assessment settings to validate that the refactors did not break the expected baseline behavior, verified that the changes are reflected in the student prologue `page_terms`, and confirmed that the attempt-start behavior respects those conditions in an assessment attempt.

On the other hand, I think these AI comments are worth addressing in this PR:

- `lib/oli_web/components/delivery/quiz_scores/quiz_scores.ex:183` - avoid materializing the full gradebook when only the current page is needed.
- `lib/oli_web/live/sections/assessment_settings/student_exceptions_table.ex:611` - parse `student_id` defensively to avoid crashing the LiveView on malformed input.
- `lib/oli/delivery/settings/student_exceptions.ex:124` - replace `String.to_atom/1` on incoming keys to avoid atom table growth.
- `lib/oli/scenarios/directive_parser.ex:1506` - same atom-creation issue when parsing scenario input.
- `lib/oli/delivery/page/prologue_terms.ex:156` - fix the `submit_term/1` guard so `nil` does not produce the wrong term.

nicocirio (approved):

PR #6466, opened 2026-04-17
[BUG FIX] [MER-5575] Ensure preview banner shows when launched as current user

andersweinstein:
This fixes inconsistent template preview behavior where the Preview Mode banner and Exit Preview button could be missing depending on how preview was launched.

  Root cause:

  - Template preview had two launch paths in the product details LiveViews.
  - When preview used :hidden_instructor, it opened /authoring/products/:slug/preview_launch, which sets the template-preview session markers used by delivery layouts to render the banner.
  - When preview used :current_user, it opened /sections/:slug directly, bypassing that controller path and therefore skipping the session markers.
  - As a result, preview content loaded normally, but the banner could disappear for users who already had a delivery-side current_user session.

  Fix:

  - Unified both launch identities onto the controller endpoint /authoring/products/:slug/preview_launch
  - This makes the controller the single canonical entry point for template preview, regardless of launch identity
  - The controller continues to handle:
      - preview authorization
      - launch preparation
      - template preview session markers
      - hidden-instructor login when needed
      - redirect into delivery

  Tests:

  - Updated author product-details LiveView preview tests to expect the unified controller launch URL
  - Updated admin product-details LiveView preview tests to expect the unified controller launch URL
  - Re-ran controller preview tests to verify the preview session markers are still set correctly

  Verification:

  - mix test test/oli_web/live/workspaces/course_author/products/details_live_test.exs
  - mix test test/oli_web/live/products/details_view_test.exs
  - mix test test/oli_web/controllers/products_controller_test.exs

### DIFF
--- a/lib/oli_web/live/products/details_view.ex
+++ b/lib/oli_web/live/products/details_view.ex
@@ -622,9 +622,8 @@ defmodule OliWeb.Products.DetailsView do
     "Template preview could not be prepared"
   end
 
-  defp preview_launch_url(_socket, section_slug, :current_user), do: ~p"/sections/#{section_slug}"
-
-  defp preview_launch_url(socket, _section_slug, :hidden_instructor) do
+  defp preview_launch_url(socket, _section_slug, launch_identity)
+       when launch_identity in [:current_user, :hidden_instructor] do
     ~p"/authoring/products/#{socket.assigns.product.slug}/preview_launch"
   end
 end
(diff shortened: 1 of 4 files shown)


andersweinstein:
Codex response to comments:

Security thread short reply:
` This does not merge the two identities’ behavior; it only unifies the entry URL. The controller still branches on launch_identity and preserves :current_user vs :hidden_instructor handling. The reason for the change is that the old :current_user fast path bypassed put_template_preview_session/3, so template_preview_mode was never present in delivery session state and the banner disappeared.`

 UI thread short reply:
 ```
The direct /sections/:slug path was the bug. It skipped the controller that sets the preview-session markers used 
by delivery layouts to show the Preview Mode banner. Routing both identities through/authoring/products/:slug/preview_launch restores consistent preview UI while keeping the controller’s 
distinct handling for :current_user and :hidden_instructor.
```

darrensiegel (approved):

PR #6471, opened 2026-04-20
[ENHANCEMENT] [MER-5573] edit template remix containers

andersweinstein:
• Summary

  Adds a template-remix-only Options flow for blueprint-created units. Authors can now set draft unit metadata from the remix screen, including title, introduction content, poster image, and intro video, without leaving the page.

  This also adjusts the UX so creating a new unit immediately opens the options modal, and the modal submit button reads Apply to reflect that changes are applied to the current remix draft rather than persisted until the page-level Save.

  What Changed

  - Added an Options button for template-created blueprint units in template remix.
  - Reused the existing container options modal UI in the remix context.
  - Kept modal edits draft-only in remix state until the page-wide Save.
  - On remix save, blueprint unit edits create a new revision and update the published-resource mappings across the base project’s publication set.
  - Creating a new unit now auto-opens the options modal.
  - The modal submit button is Apply in template remix.

  Behavior Notes

  - Options is only shown for blueprint-scoped containers created in template remix. Existing project-origin containers do not get the button.
  - Modal Apply updates the current draft on screen and marks remix as dirty.
  - Page-level Save is still required to persist changes.

  Why

  Template authors almost always need to rename a newly created container immediately, so auto-opening the modal reduces friction. Using Apply instead of Save makes the draft-vs-persist distinction clearer in remix.

  Testing

  - Added LiveView coverage for:
      - Options button visibility rules
      - auto-opening the modal on unit creation
      - Apply label in template remix
      - in-session draft updates through the modal
  - Added domain coverage for:
      - blueprint unit edits creating a new revision
      - publication mappings being updated across the base project publication set

  Verification

  - mix test test/oli_web/live/remix_section_test.exs
  - mix test test/oli/delivery/remix/remix_container_test.exs test/oli_web/live/remix_section_test.exs

### DIFF
--- a/lib/oli/delivery/remix.ex
+++ b/lib/oli/delivery/remix.ex
@@ -341,18 +341,19 @@ defmodule Oli.Delivery.Remix do
          base_project,
          %Author{id: author_id}
        ) do
-    edited_nodes = edited_blueprint_nodes(hierarchy, previous_hierarchy)
+    previous_by_resource_id =
+      previous_hierarchy
+      |> Hierarchy.flatten_hierarchy()
+      |> Map.new(&{&1.resource_id, &1})
+
+    edited_nodes = edited_blueprint_nodes(hierarchy, previous_by_resource_id)
 
     if edited_nodes == [] do
       {:ok, hierarchy}
     else
       Repo.transaction(fn ->
         Enum.reduce_while(edited_nodes, %{}, fn %HierarchyNode{} = node, updated_nodes ->
-          previous_node =
-            Hierarchy.find_in_hierarchy(previous_hierarchy, fn n ->
-              n.resource_id == node.resource_id
-            end)
-
+          previous_node = Map.fetch!(previous_by_resource_id, node.resource_id)
           previous_revision = Repo.get!(Revision, previous_node.revision.id)
 
           attrs = editable_attrs(node.revision)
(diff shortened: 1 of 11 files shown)


lib/oli/delivery/remix.ex
```diff
@@ -273,7 +309,15 @@ defmodule Oli.Delivery.Remix do
   def save(%State{} = state, author \\ nil) do
     %Section{base_project: base_project} = section = Repo.preload(state.section, :base_project)
 
-    with {:ok, hierarchy} <- ContainerCreation.materialize(state.hierarchy, base_project, author) do
+    with {:ok, hierarchy} <- ContainerCreation.materialize(state.hierarchy, base_project, author),
+         {:ok, hierarchy} <-
+           persist_blueprint_container_edits(
```
darrensiegel: Seems like the transaction inside of persist_blueprint_container_edits should be more broad and include the remainder of this function, specifically `Hierarchy.finalize` and `Sections.rebuild_section_curriculum`

darrensiegel (changes requested):

lib/oli/delivery/remix.ex
andersweinstein: OK, `Remix.save/2 `now owns the transaction boundary so container materialization, blueprint revision updates, and the section curriculum rebuild all succeed or fail together. Also fixed the save path to propagate rebuild failure instead of always returning success.

darrensiegel (approved):

PR #6491, opened 2026-04-23
[FEATURE] [MER-4082] Adaptive page duplication

darrensiegel:
This PR implements page duplication for adaptive pages.  Given that this is a high risk feature, the entire feature is hidden behind a canary feature flag that defaults to 100% available - so if after deployment we realize something isn't fully correct we can easily toggle it back to 0% available to remove access.


darrensiegel:
Addressed review comments

rgachuhi (approved):

PR #6494, opened 2026-04-23
[FEATURE] [MER-5250] AI Recommendations Feedback & Regeneration

simonchoxx:
[MER-5250](https://eliterate.atlassian.net/browse/MER-5250)

## Overview

This PR completes the AI recommendation feedback flow in the instructor dashboard, aligns the experience with the approved Figma design, and addresses privacy concerns raised during review by moving feedback details to internal storage and limiting what is sent to external Slack channels.

### What changes

- Added the full feedback flow for a recommendation:
  - Thumbs up/down sentiment
  - Optional additional text feedback
  - Visual confirmation after submission
- Kept recommendation regeneration behavior with a loading state and fallback handling, without losing the currently visible recommendation.
- Consolidated persistence for recommendation and feedback traceability.
- Updated the UI so it reflects the correct state after feedback is submitted, including after navigation or reload.
- Improved visual and interaction consistency across controls and modals to match the design.

### Privacy and data handling updates

- Reworked feedback notification behavior to avoid exposing sensitive details in Slack.
- Updated Slack notifications so external channels only receive a minimal signal that new custom feedback was captured, without personal or contextual details.

### Admin visibility

- Added a new Admin LiveView to review all captured custom feedback.
- Implemented a pageable, sortable table for operational review and follow-up.

### Functional impact

The feedback and regeneration experience is now consistent both functionally and visually, with stable behavior across recommendation states.

https://github.com/user-attachments/assets/c01016dd-c31d-47a6-b73b-596e346bf9ae

[MER-5250]: https://eliterate.atlassian.net/browse/MER-5250?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/exec-plans/current/epics/intelligent_dashboard/feedback_ui/fdd.md
+++ b/docs/exec-plans/current/epics/intelligent_dashboard/feedback_ui/fdd.md
@@ -1,0 +1,204 @@
+# AI Recommendation Feedback UI - Functional Design Document
+
+## 1. Executive Summary
+This feature completes the instructor-facing recommendation interaction loop inside the Intelligent Dashboard summary tile. The design extends the existing recommendation lifecycle and summary-tile interaction flow to support three additional behaviors: post-sentiment UI state transitions, qualitative feedback submission through an accessible modal, and persistence of the exact `original_prompt` payload used for each recommendation generation.
+
+The implementation stays within existing repository boundaries: LiveView and component state remain in the instructor dashboard surface, recommendation persistence and authorization stay in `Oli.InstructorDashboard.Recommendations`, and Slack delivery is mediated through `Oli.Slack`. The design intentionally preserves the existing `recommendation_instance_id` relationship for feedback while adding prompt-persistence fidelity at the recommendation-instance layer rather than denormalizing feedback storage.
+
+## 2. Requirements & Assumptions
+- Functional requirements:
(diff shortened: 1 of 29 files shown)


darrensiegel (changes requested):
The Slack notification feature has some serious privacy implications.  

I think we need to change this feature so that the details of the feedback (user, test, section, etc) get stored in a new table in the DB.  Then, the only thing that gets raised to the external Slack channel is the indication that "a new custom feedback has been captured".

We likely also need a new and simple Admin LiveView that shows ALL custom feedback, in a pageable, sortable table.

simonchoxx:
Thanks for the review, @darrensiegel. This flow would be adjusted to address the privacy concern:

- Slack would send only a generic notification: `“A new custom feedback has been captured.”`
- No user/section/recommendation/feedback-text details would be sent externally.
- The full details would remain in the database and would be available internally through a new LiveView for system admins only, with a pageable/sortable table of all custom feedback.
- It would be accessible from `Admin → System Management → View AI Custom Feedback `(/admin/ai_recommendation_feedback).

On storage: we would reuse the existing tables (`instructor_dashboard_recommendation_feedback + instructor_dashboard_recommendation_instances`) since they already capture the required level of detail, instead of adding a new table.

Would this solution work for you?

darrensiegel (approved):

PR #6495, opened 2026-04-23
[BUG FIX] [MER-5586] explorations begin button not working

andersweinstein:
## Summary

  Fixes [MER-5586](https://eliterate.atlassian.net/browse/MER-5586) by restoring adaptive part attempts for stateful non-scorable parts.

  Adaptive attempt creation was narrowed in #6041  to only create part attempts for scorable parts and non-scorable parts referenced by scoring rules. That excluded parts like janus-navigation-button, which still use the normal adaptive persistence pipeline. As a result, explorations could fail to start or navigate because the client could not find a partAttemptGuid for those parts.

  This change adds a distinct set of stateful non-scorable adaptive parts that should receive part attempts for persistence, while remaining non-scorable.  It also separates adaptive persisted-part semantics from analytics-tracked semantics, so restoring runtime persistence does not broaden adaptive insights/response-summary behavior.

  This PR also incidentally includes a fix for stale load-testing-only finalize route references in delivery templates that could break local exploration launch when load testing mode is enabled*.

  ## Changes

  - add @stateful_non_scorable_part_types in Oli.Activities.AdaptiveParts
  - include these parts in adaptive tracked part attempt creation:
      - janus-navigation-button
      - janus-popup
      - janus-audio
      - janus-video
      - janus-image-carousel
      - janus-capi-iframe
 - split adaptive part classification into:
      - scorable parts
      - rule-scored parts
      - persisted stateful non-scorable parts
  - keep analytics / response-summary tracking scoped to scorable ∪ rule-scored
  - keep manual grading scoped to scorable parts only
  - update adaptive attempt creation to use persisted-part helpers rather than analytics-tracked helpers
  - expand adaptive parts and hierarchy tests to verify:
      - stateful non-scorable parts get part attempts
      - analytics-tracked adaptive parts do not broaden to include persisted-only stateful parts
      - display-only parts still do not get part attempts
  - replace stale load-testing template references to removed :finalize_attempt page delivery routes with the current page lifecycle transition route

  ## Root Cause

  Before #6041 , adaptive activities got part attempts from revision_parts, so all adaptive parts with authored part ids received attempts. After that change, adaptive activities only created part attempts for:

  - scorable adaptive parts
  - non-scorable parts referenced by scoring-relevant rules

  That excluded stateful UI parts that call onInit, onSave, or onSubmit but are not scored. For Real Chem explorations, the Begin button (janus-navigation-button) fell into that gap, causing missing attemptGuid values, failed /active saves, and broken navigation.
  
  A follow-on issue in the initial fix was that reusing the existing adaptive tracked_part? concept for persistence also broadened analytics semantics, because adaptive response summaries already use tracked_part? as their filter.  Updated version corrects that by introducing a separate persisted-part classification for runtime state compatibility.

  ## Verification
  Verified with focused tests:
  - mix test test/oli/delivery/attempts/hiearchy_test.exs

  ## Notes
  *The load-testing finalization route fix is separate from MER-5586 behavior. It only affects environments where load testing mode is enabled. It was surfacing during local adaptive exploration launch because the load testing config option had been set some time ago by a developer as a convenience because it has the effect of disabling the Captcha. Given changes in the system, there is no longer any URL that can do the job the stale route did during real load testing. The fix is just to prevent a crash when testing adaptive page delivery with this setting.

[MER-5586]: https://eliterate.atlassian.net/browse/MER-5586?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli/activities/adaptive_parts.ex
+++ b/lib/oli/activities/adaptive_parts.ex
@@ -190,6 +190,7 @@ defmodule Oli.Activities.AdaptiveParts do
     content
     |> stateful_non_scorable_part_definitions()
     |> Enum.map(&Map.get(&1, "id"))
+    |> Enum.reject(&is_nil/1)
     |> MapSet.new()
   end
 
--- a/lib/oli/activities/adaptive_parts.ex
+++ b/lib/oli/activities/adaptive_parts.ex
@@ -12,16 +12,35 @@ defmodule Oli.Activities.AdaptiveParts do
                          "janus-text-slider",
                          "janus-fill-blanks"
                        ])
+  @stateful_non_scorable_part_types MapSet.new([
+                                      "janus-navigation-button",
+                                      "janus-popup",
+                                      "janus-audio",
+                                      "janus-video",
+                                      "janus-image-carousel",
+                                      "janus-capi-iframe"
+                                    ])
 
   def scorable_part_types, do: @scorable_part_types
+  def stateful_non_scorable_part_types, do: @stateful_non_scorable_part_types
 
   def scorable_part_type?(type) when is_binary(type),
     do: MapSet.member?(@scorable_part_types, type)
 
   def scorable_part_type?(_), do: false
 
+  def stateful_non_scorable_part_type?(type) when is_binary(type),
+    do: MapSet.member?(@stateful_non_scorable_part_types, type)
+
+  def stateful_non_scorable_part_type?(_), do: false
+
   def scorable_part?(part) when is_map(part), do: scorable_part_type?(Map.get(part, "type"))
   def scorable_part?(_), do: false
+
+  def stateful_non_scorable_part?(part) when is_map(part),
+    do: stateful_non_scorable_part_type?(Map.get(part, "type"))
+
+  def stateful_non_scorable_part?(_), do: false
 
   def tracked_part?(content, part_id) when is_map(content) and is_binary(part_id) do
     MapSet.member?(tracked_part_ids(content), part_id)
@@ -124,7 +143,31 @@ defmodule Oli.Activities.AdaptiveParts do
   def tracked_part_definitions(_), do: []
 
   def tracked_part_ids(content) do
-    MapSet.union(scorable_part_ids(content), rule_scored_part_ids(content))
+    [
+      scorable_part_ids(content),
+      stateful_non_scorable_part_ids(content),
+      rule_scored_part_ids(content)
+    ]
+    |> Enum.reduce(MapSet.new(), &MapSet.union/2)
+  end
+
+  def stateful_non_scorable_part_definitions(content) when is_map(content) do
+    merged_parts = merged_parts_by_id(content)
+
+    content
+    |> ordered_part_ids()
+    |> Enum.map(&Map.get(merged_parts, &1))
+    |> Enum.reject(&is_nil/1)
+    |> Enum.filter(&stateful_non_scorable_part?/1)
+  end
+
+  def stateful_non_scorable_part_definitions(_), do: []
+
+  def stateful_non_scorable_part_ids(content) do
+    content
+    |> stateful_non_scorable_part_definitions()
+    |> Enum.map(&Map.get(&1, "id"))
+    |> MapSet.new()
   end
 
   def rule_scored_part_ids(content) when is_map(content) do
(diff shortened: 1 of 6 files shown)


andersweinstein:
Codex responses to AI review comments: 
 
 1. Combined reply for the two load-testing marker comments (ui and security)

  The __FINALIZATION_URL__ token is only emitted when Oli.Utils.LoadTesting.enabled?() is true, and it is not part of the normal learner finalization flow. The actual delivery client still finalizes through finalizePageAttempt(...), which POSTs section_slug, revision_slug, and attempt_guid to PageLifecycleController.transition/2, where finalization is handled server-side.

  That said, the concern about the marker contract is fair. This change does alter the legacy load-testing marker value from the old attempt-scoped route to the current lifecycle endpoint. I could not find any in-repo consumer for __FINALIZATION_URL__, and the previous :finalize_attempt route reference appears to have been stale for years. I fixed it here because that stale helper was crashing the load-testing-enabled render path.

  If compatibility with an older external harness that expects a directly navigable finalize URL is still required, that likely needs either a dedicated testing endpoint or a revised marker contract in follow-up work.

  2. AI Review — performance on adaptive_parts.ex

  Fair point. The current implementation does an extra pass when collecting stateful_non_scorable_part_ids/1. I kept the change small and explicit for the regression fix, but agree this helper could be tightened to compute the ID set directly in one pass. I’d prefer to leave that as follow-up unless we want to broaden this PR beyond the bug fix.

  3. AI Review — elixir “Non-scorable parts are now being classified as scored”

  I don’t think this applies to the current diff. This change does not widen a “scored parts” API. The new union is in tracked_part_ids/1, which is used for adaptive attempt creation. The parts added here still retain non-scorable behavior; this change is only restoring persisted attempts for stateful adaptive parts so persistence and rules continue to work.

  4. AI Review — elixir “Hard-coded Janus type list will silently miss future stateful widgets”

  Agreed that a richer source of truth would be better if Janus part metadata exposed “stateful but non-scorable” directly. I didn’t find an existing manifest-level flag for that, so this PR keeps an explicit allowlist in AdaptiveParts to restore the broken persistence behavior without introducing a larger metadata refactor.
  Centralizing that metadata would make sense as follow-up work.

rgachuhi:
Analysis

 The underlying regression diagnosis is correct. The generic adaptive runtime does require a real partAttemptGuid for components that go through onInit / onSave / onSetData / onGetData; otherwise it logs partvattempt guid for ... not found! and the component cannot persist state. You can see that in assets/src/components/activities/adaptive/AdaptiveDelivery.tsx:298.

On that narrow point, janus-navigation-button absolutely needs a backing part_attempt. The same is also true, technically, for popup/audio/video/carousel/iframe: they all use the same generic persistence path, so giving them part_attempts is internally consistent.

The problem is where the branch draws the boundary. In master, tracked_part? effectively means “adaptive part that matters for scoring/insights tracking.” In this branch, it now also means “adaptive part that merely needs persisted runtime state.” Those are not the same concept. Because the codebase already reuses tracked_part? in insights/summary code, this fix leaks into downstream behavior.

  So my assessment is:

  - it fixes the Begin-button bug mechanism correctly
  - it does expand scope beyond that bug
  - it does introduce a semantic inconsistency by overloading tracked_part?
  - the iframe handling is not arbitrary, but it is still too broad for the current abstraction because it changes analytics classification, not just persistence

 What I would recommend

  Keep the behavioral idea, but split the concept:

  - add a separate predicate for “needs persisted adaptive part attempt” or similar
  - use that predicate only in adaptive hierarchy/attempt creation
  - leave tracked_part? for scorable/rule-scored analytics semantics, or introduce a distinct analytics predicate if needed

That would restore nav-button behavior and keep iframe/popup/audio/video/carousel persistence working, without silently changing what adaptive insights think an input is.

Non-finding but important scope note

Manual grading itself looks safe from this specific leak. lib/oli_web/live/manual_grading/manual_grading_view.ex:620 still filters adaptive parts by AdaptiveParts.scorable_part_ids(...), so these extra stateful non-scorable attempts should not suddenly appear in the grading list. The main risk is analytics/insights, not grading.

Bottom line

No, I would not say this branch fixes the issue appropriately as-is. It fixes the regression, but it does so by broadening tracked_part? in a way that changes adaptive analytics semantics and pulls janus-capi-iframe into that broader tracked-input bucket. The right fix is the same narrow runtime compatibility idea, but with a separate “stateful/persisted” classification instead of reusing “tracked.”

rgachuhi (changes requested):
Take a look at my comment

andersweinstein:
> Take a look at my comment

Addressed in latest commits.

rgachuhi (approved):

PR #6509, opened 2026-04-28
[BUG FIX] [MER-5588] Intelligent Dashboard Confusion Points

simonchoxx:
[MER-5588](https://eliterate.atlassian.net/browse/MER-5588)

## Context

In the Intelligent Dashboard, the Progress tile for the Entire course scope is built by combining scope_resources with progress bins to define the x-axis and bar values. scope_resources is also consumed by other dashboard projections.

## Problem

scope_resources may include descendants that are deeper than the level expected on the progress chart axis, while progress bins are computed for the effective axis of that view. This mismatch caused the tile to include items without matching bins, which were rendered as zero and made the Entire course view confusing (content shown with no data in the chart).

## Fix

The Progress projection was updated to align axis items with the resource_ids present in by_resource_bins, while preserving the original scope_resources order.
The change was implemented in the shared projection layer (progress/projector) rather than in tile presentation, so progress logic remains centralized in one source of truth.

[MER-5588]: https://eliterate.atlassian.net/browse/MER-5588?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli/instructor_dashboard/data_snapshot/projections/progress/projector.ex
+++ b/lib/oli/instructor_dashboard/data_snapshot/projections/progress/projector.ex
@@ -13,6 +13,8 @@ defmodule Oli.InstructorDashboard.DataSnapshot.Projections.Progress.Projector do
   @spec build(Scope.t(), map(), map(), keyword()) :: map()
   def build(%Scope{} = scope, progress_bins_payload, scope_resources_payload, opts \\ []) do
     items = Map.get(scope_resources_payload, :items, [])
+    by_resource_bins = by_resource_bins(progress_bins_payload)
+    items = align_items_with_bins(items, by_resource_bins)
     total_students = Map.get(progress_bins_payload, :total_students, 0)
 
     %{
@@ -20,7 +22,7 @@ defmodule Oli.InstructorDashboard.DataSnapshot.Projections.Progress.Projector do
       class_size: total_students,
       completion_threshold: @default_completion_threshold,
       y_axis_mode: :count,
-      series_all: build_series(items, progress_bins_payload, total_students),
+      series_all: build_series(items, by_resource_bins, total_students),
       schedule_context: schedule_context(Keyword.get(opts, :schedule)),
       page_window: %{page: 1, per_page: @default_per_page, total_items: 0, total_pages: 0},
       schedule_marker: @empty_schedule,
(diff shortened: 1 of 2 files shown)


lib/oli/instructor_dashboard/data_snapshot/projections/progress/projector.ex
```diff
@@ -80,6 +75,22 @@ defmodule Oli.InstructorDashboard.DataSnapshot.Projections.Progress.Projector do
     end)
   end
 
+  defp by_resource_bins(progress_bins_payload) do
+    Map.get(
+      progress_bins_payload,
+      :by_resource_bins,
+      Map.get(progress_bins_payload, :by_container_bins, %{})
+    )
+  end
```
Francisco-Castro, line 84: NIT

```elixir
  defp by_resource_bins(%{by_resource_bins: bins}), do: bins
  defp by_resource_bins(%{by_container_bins: bins}), do: bins
  defp by_resource_bins(_), do: %{}
```

Francisco-Castro (approved):
## Review notes

Fix is correct and minimal. The mismatch between `progress_bins` (depth-1 axis keys) and `scope_resources.items` (flat recursive descendants) is real, and aligning at the projector layer is the right call given that `progress/projector.ex` is the only consumer that joins these two oracles together.

A few things to consider before merging:

  ### 1. Document the cross-module invariant

The semantic difference between the two oracles isn't stated anywhere. A future tile that consumes both will rediscover this bug. Suggest expanding the `@moduledoc` on each oracle:

**`lib/oli/instructor_dashboard/oracles/scope_resources.ex`**
```elixir
  @moduledoc """
  Returns course title and scoped resources for the requested scope.

  `items` is a flat, pre-order list of ALL descendants under the scoped 
  node (pages and containers at every depth). Consumers that need only 
  direct children — for example, when joining against axis-keyed data 
  from `Oli.InstructorDashboard.Oracles.ProgressBins` — must filter 
  accordingly.
"""
```

  lib/oli/instructor_dashboard/oracles/progress_bins.ex (currently no @moduledoc)
```elixir
  @moduledoc """
  Computes progress bin counts keyed by the direct children of the
  scoped node (depth-1 only). `by_resource_bins` is NOT a superset of
  `ScopeResources.items`; callers joining the two payloads must align 
  keys before iterating.
  """
```

###  2. Test gaps in align_items_with_bins/2

Two paths in the new helper aren't covered:

  - Empty bins skip (the cb6ef5a961 commit): `by_resource_bins == %{}` with non-empty items should return items unchanged. No regression test asserts this.
  - Zero-overlap fallback (`if aligned_items == [], do: items`): unreachable today given the upstream contract (bins keys ⊆ scope_resources items), but if you keep it, please cover it. If you'd rather drop it and let empty_state(:no_scope_children) trigger when the invariant breaks, that's also reasonable — the fallback silently re-introduces the original zero-bar render if upstream ever drifts.

PR #6515, opened 2026-04-28
[BUG FIX] [MER-5266] adjust position/style of download CSV button

andersweinstein:
Adjusts the Intelligent Dashboard  "Download Dashboard Data (CSV)" button to match the MER-5266 Figma design more closely.

  The download action is now presented as a text link with a leading download icon, uses the same tokenized blue treatment as nearby dashboard navigation controls, and sits above the scope navigator with tighter vertical alignment to the Insights tab row.

Screenshot 2026-04-29 at 6 04 00 AM

### DIFF
--- a/lib/oli_web/components/delivery/instructor_dashboard/intelligent_dashboard/shell.ex
+++ b/lib/oli_web/components/delivery/instructor_dashboard/intelligent_dashboard/shell.ex
@@ -27,7 +27,32 @@ defmodule OliWeb.Components.Delivery.InstructorDashboard.IntelligentDashboard.Sh
 
     ~H"""
     <div id="learning-dashboard" class="container mx-auto mb-10" phx-hook="Scroller">
-      <div class="mb-4">
+      <div class="-mt-9 mb-4 space-y-1.5">
+        <div class="flex justify-end">
+          <form
+            id="intelligent-dashboard-download-form"
+            phx-hook="BrowserTimezoneForm"
+            action={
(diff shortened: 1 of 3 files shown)


simonchoxx:
I've been reviewing your changes. Everything seems fine. I just have two things to mention.

In the Figma designs, the download button appears to be on the same row as the tabs (Dashboard, Content, etc.). In your branch, the download button is slightly lower. I'm not sure if this will pass QA, but I understand that we can't have the button on the exact same row as the tab buttons.

The "Download dashboard data (CSV)" text in Figma has a font weight of 700, and in your branch, it appears to have a font weight of 500. These are minor issues, but they might cause it to fail QA again.

andersweinstein:
> I've been reviewing your changes. Everything seems fine. I just have two things to mention.
> 
> In the Figma designs, the download button appears to be on the same row as the tabs (Dashboard, Content, etc.). In your branch, the download button is slightly lower. I'm not sure if this will pass QA, but I understand that we can't have the button on the exact same row as the tab buttons.

Yes, I noticed this too. In fact it could float higher, I think, the problem is if you narrow the window so that it would overlap the tab buttons and be ugly. The figma didn't address this, so I just left it with a little more space. I could run it by designer Jess I suppose.

> The "Download dashboard data (CSV)" text in Figma has a font weight of 700, and in your branch, it appears to have a font weight of 500. These are minor issues, but they might cause it to fail QA again.

I will adjust the weight. I believe the most important thing for this Redo was the placement/style of the button which was completely wrong.

darrensiegel (approved):

PR #6517, opened 2026-04-28
[BUG FIX] [MER-5559] Scope student exception date edits to current section

darrensiegel:
Changes:
- Updated `lib/oli_web/live/sections/assessment_settings/student_exceptions_table.ex` so date editing reloads the selected exception by section_id, resource_id, and user_id.
- Added a regression test in `test/oli_web/live/sections/assessment_settings/settings_live_test.exs` covering same student/resource exceptions across two sections and verifying only the current section updates.

### DIFF
--- a/lib/oli_web/live/sections/assessment_settings/student_exceptions_table.ex
+++ b/lib/oli_web/live/sections/assessment_settings/student_exceptions_table.ex
@@ -472,10 +472,22 @@ defmodule OliWeb.Sections.AssessmentSettings.StudentExceptionsTable do
   end
 
   def handle_event("edit_date", %{"user_id" => user_id}, socket) do
+    user_id = String.to_integer(user_id)
+
     selected_setting =
       Enum.find(socket.assigns.table_model.rows, fn s ->
-        s.user_id == String.to_integer(user_id)
+        s.user_id == user_id
       end)
+      |> case do
+        nil ->
+          nil
+
+        %{resource_id: resource_id} ->
+          case Settings.get_student_exception(resource_id, socket.assigns.section.id, user_id) do
+            nil -> nil
+            selected_setting -> Repo.preload(selected_setting, :user)
+          end
+      end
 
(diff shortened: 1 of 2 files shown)


andersweinstein:
Hmm, I had codex investigate MER-5559 and it said issue no longer existed in master (though it offered to add missing regression test for it). It said bug was corrected with the fix for [MER-5567](https://eliterate.atlassian.net/browse/MER-5567) via [PR #6479](https://github.com/Simon-Initiative/oli-torus/pull/6479), which corrected the student-exception date edit path to use the section-scoped exception update flow and eliminated the cross-section Ecto.MultipleResultsError seen in AppSignal.

I *think* the code change in this PR is at best defensive, but unnecessary to prevent a bug, though regression test is a worthwhile addition. Codex:
```
The production change in this PR:

  - takes the row the user clicked in handle_event("edit_date", ...)
  - extracts resource_id and user_id
  - then re-fetches the exception with Settings.get_student_exception(resource_id, section.id,
    user_id)

  That does make the modal-open step explicitly section-scoped. But in current master, the clicked
  row is already coming from AssessmentSettings.get_student_exceptions(section.id), so the table data
  is already section-scoped before this PR.

  So I do not see it correcting a second real cross-section bug beyond the historical AppSignal
  issue.

  - perform_edits/5 was the real scoping bug behind the AppSignal crash.
  - This PR hardens handle_event("edit_date", ...), but I don’t see evidence that current master has
    a separate scoping defect there.

```

Maybe drop the redundant re-fetch (which got a performance comment), but keep the regression test as a useful addition to test coverage?

[MER-5567]: https://eliterate.atlassian.net/browse/MER-5567?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #6537, opened 2026-04-29
[FEATURE] [MER-5355] Consolidate instructor recommendation prompt flow

simonchoxx:
[MER-5355](https://eliterate.atlassian.net/browse/MER-5355)

## Summary

This PR unifies how the instructor recommendation prompt is built so there is a single editable template that is processed at runtime with real datasets.

## What changed

- Consolidated the main recommendation prompt into a single template.
- Added post-processing to replace `#{data}` with the actual dataset block before sending the request to the model.
- Updated message construction so the main content lives in the rendered prompt.
- Kept a robust fallback when a custom template does not include `#{data}` (datasets are appended at the end).
- Updated tests to validate:
  - `#{data}` interpolation,
  - custom template behavior,
  - fallback behavior for blank templates or templates without the placeholder,
  - original prompt persistence under the new format.

## Note

The `no_signal` path keeps its dedicated branch to preserve current behavior when there is not enough data signal.

[MER-5355]: https://eliterate.atlassian.net/browse/MER-5355?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli/instructor_dashboard/recommendations/prompt.ex
+++ b/lib/oli/instructor_dashboard/recommendations/prompt.ex
@@ -4,6 +4,7 @@ defmodule Oli.InstructorDashboard.Recommendations.Prompt do
   """
 
   @version "recommendation_prompt_v1"
+  @data_placeholder "\#{data}"
   @default_no_signal_prompt """
   You are an expert learning engineer and instructor dashboard analyst. You will be given 1-5 datasets exported from an LMS/assessment system. Each dataset is presented as:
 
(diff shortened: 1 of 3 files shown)


lib/oli/instructor_dashboard/recommendations/prompt.ex
```diff
@@ -24,34 +25,64 @@ defmodule Oli.InstructorDashboard.Recommendations.Prompt do
...
+
+  When forming the insight, consider (implicitly -- do not enumerate):
 
   - Patterns across datasets (performance, attempts, hints, timing, engagement).
   - Evidence from specific dataset descriptors and column names.
   - Whether the issue concerns students, groups, concepts, or engagement behaviors.
   - What concrete instructor action would most directly address the issue.
+  - Where is the class in the schedule? Recommendations are most helpful when they relate to what we are now covering in the course.
+
+  ###Required structure
```
darrensiegel: A space here is needed to make this an actual markdown header

lib/oli/instructor_dashboard/recommendations/prompt.ex
```diff
@@ -24,34 +25,64 @@ defmodule Oli.InstructorDashboard.Recommendations.Prompt do
...
+  - What's happening (signal)
+  - Why it matters (interpretation)
+  - What to consider doing (suggested action)
+
+  ###Tone guidelines
```
darrensiegel: Space

lib/oli/instructor_dashboard/recommendations/prompt.ex
```diff
@@ -24,34 +25,64 @@ defmodule Oli.InstructorDashboard.Recommendations.Prompt do
...
+  - If evidence is suggestive but not definitive, label it clearly as an inference within the sentence.
   - If no meaningful insight can be supported by the data, say so explicitly in one sentence.
+  - Round numbers to the hundredth decimal place
 
-  Rules:
+  #### Example format (illustrative only)
+
+  "Several students are progressing into Unit 4, but a sizable group is showing low proficiency on core plant biology objectives, particularly cell structure and photosynthesis. You may want to revisit these concepts or assign additional practice before moving further into the unit."
+
+  ###Example format (illustrative only)
```
darrensiegel: Space

lib/oli/instructor_dashboard/recommendations/prompt.ex
```diff
@@ -62,9 +93,11 @@ defmodule Oli.InstructorDashboard.Recommendations.Prompt do
 
   @spec build_messages(map(), keyword()) :: [map()]
   def build_messages(input_contract, opts \\ []) when is_map(input_contract) do
+    rendered_prompt = input_contract |> system_prompt(opts) |> render_prompt(input_contract)
+
     [
-      %{role: :system, content: system_prompt(input_contract, opts)},
-      %{role: :user, content: user_prompt(input_contract)}
+      %{role: :system, content: rendered_prompt},
+      %{role: :user, content: "Begin now."}
```
darrensiegel: Seems like we have "Begin now." included twice.  Once here, and once above as part of the prompt. This "Begin now" here will not be overridable - please remove it and see if you can make the recommendations work without a user prompt

darrensiegel (changes requested):

lib/oli/instructor_dashboard/recommendations/prompt.ex
simonchoxx: Done. I removed the extra user message %{role: :user, content: "Begin now."} so we now send only the rendered system prompt in `build_messages/2`.
I also validated behavior by generating recommendations with this change, and recommendations are still generated correctly.
Additionally, I updated the related tests to reflect the single-message prompt contract.

darrensiegel (approved):

PR #6556, opened 2026-05-06
[FEATURE] [MER-5257] AI Email Capabilities & updates

Francisco-Castro:
## Summary

Delivers the non-UI backend that powers the instructor-dashboard AI email feature: a normalized context contract, AI draft generation, whitelist placeholder substitution, per-recipient template realization, send-time validation, and one Oban job per recipient via the existing Torus mailer. The modal LiveComponent (PR 2) and entry-point launchers (PR 3) follow.

Jira: [MER-5257](https://eliterate.atlassian.net/browse/MER-5257)

## Requisites covered

This PR delivers backend support for the following MER-5257 requisites. UI-bound requisites land in PR 2 (modal) and PR 3 (entry-point integrations).

| # | Requisite | Status |
|---|---|---|
| FR-001 | Normalized initiation context from supported dashboard entry points | Covered |
| FR-002 | Stable situation contract for prompt semantics | Covered |
| FR-003 | Initial AI draft generation with default neutral tone | Covered |
| FR-008 | Block send when recipients or required fields are invalid | Covered |
| FR-009 | Per-recipient realization via whitelist placeholder substitution | Covered |
| FR-010 | Dispatch via existing Torus email with one job per recipient | Covered |
| FR-013 | Dedicated GenAI Feature Config "Instructor Email" | Covered |
| FR-015 | Send-time placeholder validation; block on invalid | Covered |
| FR-004 | Manual subject/body editing | Phase 4 (PR 2) |
| FR-005 | Tone selection without auto-regenerate | Phase 4 |
| FR-006 | Regenerate with selected tone | Phase 4 |
| FR-007 | Recipient chips with truncation + manual add | Phase 4 |
| FR-011 | Recoverable failures with retry-friendly modal | Phase 4 (backend returns structured errors) |
| FR-012 | Modal accessibility (keyboard / screen-reader) | Phase 4 |
| FR-014 | Body hyperlinks via Slate `RichTextEditor` | Phase 4 |

## Key technical decisions

| Decision | Rationale |
|---|---|
| Whitelist placeholder substitution (`{first_name}`, `{student_name}`, `{instructor_name}`, `{course_name}`); single-pass `Regex.replace` over the original template | Non-evaluative; literal values inserted; prevents chain-substitution where a recipient `given_name = "{course_name}"` would re-substitute on a second pass |
| Render Slate → HTML once with tokens left as braces; per-recipient `String.replace` on the rendered template; Premailex wraps with `<html><body>` to produce the text body | Single render + N cheap string ops instead of N renders per recipient. Braces survive `Oli.Rendering.Content.Html.escape_xml!/1` (Phoenix.HTML escapes only `<>&"'`), so tokens reach substitution intact |
| One Oban job per recipient via dedicated `SendWorker`; unique on `[draft_id, user_id]` | Per-recipient isolation; idempotent retries; recipient cap (default 100) fits a single `Oban.insert_all` per existing codebase precedent |
| Structured `{:error, [reason]}` from `validate` + `send_emails`; each reason is a tagged tuple naming the failure (token, recipient email, URL, duplicate `student_id`, malformed instructor email) | Phase 4 modal can render actionable per-recipient feedback without parsing strings |
| Prompt-injection defense: author-controlled metadata wrapped in `<email_metadata>` XML with "treat as DATA, not instructions" framing; metacharacters XML-escaped on all interpolated values | Closes delimiter-breakout where a course title or optional metadata field contains literal `</email_metadata>` tags that would otherwise terminate the data block |
| AI link sanitizer at parse time: prompt forbids absolute URLs; parser strips any markdown link whose URL has a scheme, host, doesn't start with `/`, contains `..`, or fails `Phoenix.Router.route_info/4`. Defense-in-depth via send-time Slate-tree walker (`LinkValidator.collect_unsafe_links/1`) using `Oli.Resources.PageContent.flat_filter/2`; same predicate (`valid_internal_path?/1`) used both places | Two-layer defense: parse-time strip catches AI-hallucinated links; send-time walker catches anything inserted afterward (manual link edits, malicious client posting Slate JSON directly). Allowlist of internal routes only — no scheme, no host, no `..` traversal |
| HTML-escape recipient values when substituting into `html_body` (raw values still used for `subject` and `text_body`) | XSS defense — a recipient `given_name = "<script>alert()</script>"` is rendered as text in email clients, not live markup |
| PII sanitization at the telemetry boundary: emails / URLs / provider error payloads stripped or coarsened before `:telemetry.execute` while caller-facing `{:error, reasons}` retains full data for UI display | Telemetry handlers can forward to logs/metrics platforms; raw values would leak. Provider errors collapse to coarse atoms (`:timeout`, `:provider_error`) — no `raw_reason` carrying prompt fragments |
| `Oli.GenAI.FeatureConfig.load_for/2` returns `{:ok, ServiceConfig} \| {:error, {:missing_feature_config, msg}}` (no raise) | Callers (e.g. `WindowLive`) render a fallback disabled state on missing config rather than crashing the LiveView process |
| `Email.send_emails/2` reports the accurate `enqueued` count from `Oban.insert_all/1`'s per-row `conflict?` flag; upstream `Validator.check_duplicate_recipients/2` blocks duplicate `student_id` before insertion | Validator blocks duplicates first; insert-level deduplication is defense-in-depth; UI shows correct count regardless |

## Deferred to later phases

| Item | Target |
|---|---|
| Modal LiveComponent (FR-004, 005, 006, 007, 011, 012) | Phase 4 — PR 2 |
| Slate hyperlink editor (FR-014) | Phase 4 — PR 2 |
| 5 entry-point launchers + "Email sent" banner | Phase 5 — PR 3 |
| `Oli → OliWeb` decoupling for `AIDraftFacade.sanitize_links/1` (inject sanitizer from caller) | Phase 4 — natural boundary |
| Per-entry-point URL allowlist + student-appropriate route allowlist | Phase 5+ — URL-menu backlog |

[MER-5257]: https://eliterate.atlassian.net/browse/MER-5257?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


priv/repo/seeds.exs
```diff
@@ -239,6 +239,23 @@ case Oli.Repo.all(RegisteredModel) do
       section_id: nil
     })
 
+    # Dedicated ServiceConfig for the instructor email feature (MER-5257).
+    # Points at the same primary_model as standard-no-backup today; decoupled
+    # so the model can be swapped via the admin UI (OliWeb.GenAI.FeatureConfigsView)
+    # without affecting other GenAI features.
+    email_service_config =
```
darrensiegel, line 247: These insertions seem to be in a spot where they will **not** execute when there is already an existing RegisteredModel (which is the case for servers running previous isntances of the code, like Tokamaka, Proton). 

For this feature, the reliable place is a DB migration. The migration should insert the
  new completions_service_configs row and the global gen_ai_feature_configs
  row idempotently, likely by deriving primary_model_id from the existing
  "standard-no-backup" config or its primary model. It should also use
  WHERE NOT EXISTS style logic because the current unique index on
  [:section_id, :feature] does not reliably prevent duplicate global rows
  when section_id is NULL in Postgres.

darrensiegel (changes requested):

priv/repo/seeds.exs
Francisco-Castro: @darrensiegel Addressed in https://github.com/Simon-Initiative/oli-torus/pull/6556/changes/574d5255eea2fad0bc09711d1e7d5e77ae66ca3d new migration priv/repo/migrations/20260513120000_add_instructor_email_feature_config.exs

> - Idempotent via WHERE NOT EXISTS on both inserts
> - Explicit section_id IS NULL check (Postgres NULL-distinctness guard)
> - primary_model_id derived from standard-no-backup config at migrate time
> - Seeds insertions retained for fresh-install path (migration sees empty DB at first mix ecto.migrate, seeds run after)

darrensiegel (approved):

PR #6559, opened 2026-05-06
[BUG FIX] [MER-5612] Implement student progress CSV download

simonchoxx:
[MER-5612](https://eliterate.atlassian.net/browse/MER-5612)

## Summary

This PR implements CSV download support for the `Progress` tab in the `Student Dashboard`. Previously, the `Download CSV` action was not implemented, so clicking it did nothing.

## What changed

- Added a student-specific progress download endpoint.
- Wired the Download CSV button in the Progress tab to that endpoint.
- Implemented CSV generation with one row per course page, including:
  - resource
  - type
  - score
  - attempts
  - accesses
  - first visited / last visited
- Consolidated progress row-building logic into a shared function to avoid duplication between LiveView and controller.
- Added access validation to ensure the requested student belongs to the section.

## Result

CSV download for `Student Dashboard > Progress` now works correctly and is no longer a placeholder/no-op action.

[MER-5612]: https://eliterate.atlassian.net/browse/MER-5612?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli/delivery/sections.ex
+++ b/lib/oli/delivery/sections.ex
@@ -1837,9 +1837,9 @@ defmodule Oli.Delivery.Sections do
   Returns per-page progress rows for a student in a section, including
   score/access/attempt metadata and timestamps.
   """
-  def student_progress_rows(section_slug, student_id) do
+  def student_progress_rows(%Section{} = section, student_id) do
     resource_accesses =
-      Oli.Delivery.Attempts.Core.get_resource_accesses(section_slug, student_id)
+      Oli.Delivery.Attempts.Core.get_resource_accesses(section.slug, student_id)
       |> Enum.reduce(%{}, fn r, m ->
         resource_access =
           case r.score do
@@ -1851,8 +1851,8 @@ defmodule Oli.Delivery.Sections do
       end)
 
     page_nodes =
-      section_slug
-      |> DeliveryResolver.full_hierarchy()
+      section
+      |> SectionResourceDepot.get_delivery_resolver_full_hierarchy()
       |> Hierarchy.flatten()
       |> Enum.filter(&(&1.revision.resource_type_id == ResourceType.id_for_page()))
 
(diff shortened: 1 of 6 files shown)


lib/oli/delivery/sections.ex
```diff
@@ -1833,6 +1833,81 @@ defmodule Oli.Delivery.Sections do
...
+            nil -> r
+            _ -> Map.put(r, :score, round_score(r.score))
+          end
+
+        Map.put(m, r.resource_id, resource_access)
+      end)
+
+    page_nodes =
+      section_slug
+      |> DeliveryResolver.full_hierarchy()
```
darrensiegel: We should almost NEVER be using the `DeliveryResolver.full_hierarchy/1` directly.  For large courses this can be an expensive query to execute.   

Instead, use `SectionResourceDepot.get_delivery_resolver_full_hierarchy/1`

lib/oli/delivery/sections.ex
```diff
@@ -1833,6 +1833,81 @@ defmodule Oli.Delivery.Sections do
...
+          number_attempts: if(is_nil(ra), do: 0, else: ra.resource_attempts_count),
+          number_accesses: if(is_nil(ra), do: 0, else: ra.access_count),
+          updated_at: if(is_nil(ra), do: nil, else: ra.updated_at),
+          inserted_at: if(is_nil(ra), do: nil, else: ra.inserted_at)
+        }
+      end)
+
+    unordered_rows =
+      section_slug
+      |> fetch_all_pages(nil, :resource_id)
```
darrensiegel: The call to `fetch_all_pages` runs a query that returns the full Revision, but below only `title`, `graded`, `resource_id` are used. 

Those three attributes exist directly on the `SectionResource` record and can be drastically faster (and with much less memory impact) to retrieve through `SectionResourceDepot`, filtered by page resource type.

darrensiegel (changes requested):

darrensiegel (approved):

PR #6581, opened 2026-05-12
[BUG FIX] [MER-5615] Adjust same lineage remix error messages

andersweinstein:
This PR adjusts the error messages shown for same lineage remix errors in response to review by QA & designer. Main message is now:  `This content can't be added here because content from the same lineage already exists. Choose a different source project, then try again.`

For the (perhaps unusual, but possible) case of attempting to remix a _set_ of sources from different projects that share lineage in one operation, the message is now:  `This content can't be added here because the selected sources combine content from the same lineage. Select content from one source project at a time, then try again.`


darrensiegel (changes requested):
I don't think an end user will understand "lineage".  It is not a term that we use anywhere else in the UI.   I think we can leave the error message as-is.

andersweinstein:
> I don't think an end user will understand "lineage". It is not a term that we use anywhere else in the UI. I think we can leave the error message as-is.

Well QA sent ticket for rework requesting this message (after consulting designer, I *think*). But I will leave as is.

PR #6602, opened 2026-05-26
[ENHANCEMENT] [MER-5654] Removed non-functional 'font-size' field from Advancedd Author

kjsegovi:
This pull request makes a small change to the `janus-mcq` component schema by removing the `fontSize` property from both the `McqModel` TypeScript interface and the JSON schema definition. 

The font size should be adjusted at the option level for each option in the MCQ. So this quick fix removes the fontSize option from the component panel. 

Link to ticket: https://eliterate.atlassian.net/browse/MER-5654


darrensiegel (changes requested):
We should remove this new `package-lock.json` file, I think.

darrensiegel (approved):

PR #6606, opened 2026-05-27
[ENHANCEMENT] [MER-5642] Context-aware email draft modal UI (Phases 4-6)

Francisco-Castro:
## Update — instructor link-editing defect fixed

Manual testing with an **instructor** account (vs. the admin/author account used earlier, which masked it) exposed a real defect: link editing in the Draft Email body was non-functional for instructors. The Settings modal hit an author-only endpoint they can't reach ("Failed to initialize"), and inserted links got an external `http://` href that the email's internal-only validator rejects at send. Root cause: the modal reused the authoring `LinkModal`.

**Now fixed and verified end-to-end** (commits `7786bdda53` shared-modal hardening, `c524b7baf6` link picker). Instructors get an internal **course-page link picker** sourced server-side, saved as `/course/link/:revision_slug` and rewritten to the section delivery URL on send. See **Instructor link picker** below.

## Summary

Delivers the context-aware **Draft Email** modal for the instructor dashboard (MER-5642, frontend / Phases 4–6). A reusable `DraftEmailModal` LiveComponent lets instructors generate an AI draft from dashboard context, edit it, and send via the existing Torus email pipeline —without leaving their current view. Builds on the merged backend PR #6556 (Phases 1–2).

Jira: [MER-5257](https://eliterate.atlassian.net/browse/MER-5257) ([MER-5642](https://eliterate.atlassian.net/browse/MER-5642) frontend)

## What's included

- **`DraftEmailModal` LiveComponent** — recipient chips (with overflow + remove), tone selection, "Generate New Draft" (AI), subject input, Slate rich-text body (inline + link only), Send/Cancel, loading/error/empty states, focus trap, and live-region announcements.
- Replaces and retires the legacy `Students.EmailModal`.
- **AI draft generation** through the existing GenAI stack via a thin facade; tone + situation drive the prompt. Tone change does **not** auto-regenerate (explicit Generate only).
- **Send pipeline reuse** — per-recipient Oban `SendWorker`, placeholder realization, and send-time whitelist validation (from PR #6556).

## Instructor link picker (link-editing fix)

The body editor's link feature now works for instructors, who are not authors:

- **Picker-first, internal-only.** Inserting a link (toolbar or ⌘L) opens a course-page picker sourced server-side via `SectionResourceDepot.get_lessons/1` (non-hidden lessons), passed to the editor as a prop — no author-only endpoint, no external URLs. Selecting a page wraps the selection as `/course/link/:revision_slug`, which the email renderer rewrites to a section delivery URL (`/sections/:slug/lesson/:revision_slug`) and the send-time validator accepts.
- **Mode-gated.** A new optional `CommandContext.linkContext` switches the shared `LinkModal`/`LinkCmd` into email mode; authoring behavior is unchanged when it's absent.
- **Shared modal hardening** (commit `7786bdda53`): nested-Escape containment (Esc closes only the inner picker, not the draft), preferred initial focus on the page select, and blur-before-`aria-hidden` to clear a focus a11y warning.

## Async draft generation (review follow-up)

AI draft generation runs via LiveView `start_async`/`handle_async` (replacing a `Task.Supervisor.async_nolink`): the task is tied to the LiveView, so it's auto-cancelled on navigate-away, and a crash is delivered as an error instead of taking down the dashboard. Closing the modal cancels the in-flight task. Each generation is tagged with a unique **request id** so a late or superseded result is dropped deterministically — a result is applied only if it matches the modal's current generation, and close/reopen/regenerate invalidate prior ids. Commits `7aa00109ed`, `9868797947`.

## Entry points (all 5)

Student Support tile · Assessments tile ("Email Students Not Completed", auto-populated) · Student Overview → student list · Content → student list · Learning Objectives → proficiency student list. Each opens the same modal scoped to the originating context/recipients.

## Accessibility

Dialog semantics + trapped focus + Esc-to-close + focus return; keyboard-reachable chip removal with announced excluded (no-email) recipients; `aria-required` subject; `aria-disabled` Send with click-to-validate; group labels via `aria-labelledby` (not dangling `<label>`s).

## Supporting / infra changes in this PR

- GenAI `provider_opts` + JSON-mode plumbing (`AIDraftFacade` structured responses).
- `Button` primitive: additive `aria_disabled` + atom-key `@rest` globals fix + `:close` aria-label fix.
- Flash container `z-50` (renders above sticky table headers).
- `RichTextEditor` optional debounce (`onEditDebounceMs`), flushed on blur/Send.
- `MarkdownToSlate` module; single-query `get_section_with_base_project/1` slug resolution.
- **Server-side email-body Slate sanitization** (`SlateSanitizer`): the `update_body_slate` client payload and the send-time render are allowlisted to paragraphs, text (+ marks), and validator-passing internal links; the email renders with `render_errors: false`. Prevents HTML injection via tampered client Slate (e.g. an unsupported node whose raw type would otherwise be interpolated into the email HTML).
- **AppSignal telemetry on AI draft generation** (`Email.Telemetry`): generated/failed/link-stripped counts + duration, tagged by bounded `situation_key`/`tone`/`reason` (no PII).

## Follow-ups (deferred)

Non-blocking server-side hardening + perf follow-ups — none are active vulnerabilities (verified in review) — are tracked in [TRIAGE-2370](https://eliterate.atlassian.net/browse/TRIAGE-2370): section-scoped internal-link allowlist, `send_email` authz revalidation, scoped/projected section helper, and draft-generation rate-limit + actor attribution.

UI polish: the Slate placeholder/caret alignment (a slate-react default missing a `left` offset) is fixed scoped to the email editor here; generalizing the same fix to the other shared Slate editors is tracked in [TRIAGE-2373](https://eliterate.atlassian.net/browse/TRIAGE-2373).

## Testing

Manual (browser) testing performed:

- All 5 entry points open the modal with the correct recipients / context
- Tone selection (no auto-regenerate) + AI generate / regenerate
- Subject + body editing with Send-gating (Send enables only with recipient + subject + body)
- Recipient removal + the excluded "no email on file" note
- Keyboard navigation, focus trap, and Esc-to-close
- Cancel & X close
- End-to-end send — Oban job enqueued, "Email sent" flash, page stays interactive (clean teardown)

Instructor link picker (verified as an instructor account):

- Picker opens (toolbar) with focus on the page select; inserts `/course/link/:revision_slug`
- Settings reopens the picker with the current page preselected
- End-to-end send — link rendered as `<a class="internal-link" href="/sections/.../lesson/...">` and delivered (passes send-time validation)
- Esc closes only the picker (draft preserved); focus trap stays within the picker; no console a11y warning

Automated: jest `893` passing (full suite, link-picker + shared-modal specs included); targeted Elixir suites green (`draft_email_modal`, `instructor_dashboard_live`, `draft_email_async` — incl. request-id scoping + close/reopen/regenerate lifecycle); `mix format` / prettier / eslint clean.

## How to test

Instructor (or admin) on a section dashboard → any entry point above → select students → **Email** → Generate a draft → edit → **Send**. Confirm the "Email sent" flash and that the page stays interactive. To exercise the link picker: select text in the body → link toolbar button (or ⌘L) → choose a course page → Save → Send.

## Screenshots
**Draft Email modal — recipient chips + tone selection (Neutral / Encouraging / Firm)**
01-modal-open-recipients-tones

**AI-generated draft — subject + body populated, ready to edit**
02-ai-generated-draft

**Send gating — Send is `aria-disabled` and validates on click ("Add a subject before sending.")**
03-send-disabled-validation

**Excluded recipient — "no email on file" note + empty-recipient state**
04-excluded-no-email-note

**Body rich-text editor — insert link (inline + link only)**
05-insert-link

**End-to-end send — "Email sent to 1 student(s)" confirmation flash**
06-send-success-flash

**Entry point — Learning Objectives → proficiency band → Draft Email**
07-entry-learning-objectives

**Modal overlays the page cleanly (objectives entry point)**
08-objectives-modal-overlay

  ## Docs / housekeeping

Feature planning + spec docs live under `docs/exec-plans/current/epics/intelligent_dashboard/email_sending/` (PRD, gaps, progress, figma-audit). On merge, this folder will be moved `current/ → archive/` per the exec-plans convention — a docs-only move, no code impact.

[MER-5257]: https://eliterate.atlassian.net/browse/MER-5257?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MER-5642]: https://eliterate.atlassian.net/browse/MER-5642?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


assets/src/components/misc/DotDistributionChart.tsx
```diff
@@ -888,13 +888,17 @@ function renderDots(
             // Render as vertical tower of individual dots
             return (
               <g key={`${level}-tower-${groupedDot.proficiency_value}-${index}`}>
-                {groupedDot.tower_dots.map((towerDot) => {
-                  const dotId = `tower-${level}-${towerDot.student_id}`;
+                {groupedDot.tower_dots.map((towerDot, towerIndex) => {
```
eliknebel: This change seems unrelated to the feature in the ticket

assets/src/components/editing/elements/link/LinkElement.tsx
```diff
@@ -23,6 +23,7 @@ export const LinkEditor = (props: Props) => {
       {...props.attributes}
       id={props.model.id}
       href="#"
+      onClick={(e) => e.preventDefault()}
```
eliknebel, line 26: This change seems unrelated to the feature in the ticket. Are we sure this doesn't adversly affect how link elements work?

assets/src/components/editing/elements/link/LinkModal.tsx
```diff
@@ -69,6 +69,15 @@ export const LinkModal = ({ onDone, onCancel, model, commandContext, projectSlug
 
         setPages(result);
       },
+      () => {
```
eliknebel, line 94: While showing an error message seems better here, this change seems unrelated to the feature in the ticket

lib/oli_web/components/layouts/instructor_dashboard.html.heex
```diff
@@ -14,7 +14,7 @@
   />
 
   <div class="relative flex-1 flex flex-col pt-4 pb-[60px]">
-    <div id="flash_container" class="container mx-auto sticky top-16 z-40">
+    <div id="flash_container" class="container mx-auto sticky top-16 z-50">
```
eliknebel, line 17: Can we make sure this change doesn't clash with the user menu z-index? It looks like it is also set to z-50

lib/oli_web/live/delivery/instructor_dashboard/instructor_dashboard_live.ex
```diff
@@ -1247,6 +1250,53 @@ defmodule OliWeb.Delivery.InstructorDashboard.InstructorDashboardLive do
     {:noreply, socket}
   end
 
+  def handle_info({:generate_draft, component_id, email_context}, socket) do
+    if email_context do
+      task =
+        Task.Supervisor.async_nolink(Oli.TaskSupervisor, fn ->
```
eliknebel: Is there a particular reason to be using `async_nolink` here? I would think that if the LiveView process dies (e.g. user navigates away) then we would want this process to go away as well, no? Otherwise this process could compelte and have no process to reply to

eliknebel (changes requested):

assets/src/components/misc/DotDistributionChart.tsx
Francisco-Castro: Agreed — pre-existing, unrelated to this feature ([MER-4703](https://eliterate.atlassian.net/browse/MER-4703) changed the payload to id; [MER-5157](https://eliterate.atlassian.net/browse/MER-5157)'s tower code still read student_id → the undefined-key warning). Removed it from this PR; proper fix tracked in [TRIAGE-2352](https://eliterate.atlassian.net/browse/TRIAGE-2352) (caused-by [MER-4703](https://eliterate.atlassian.net/browse/MER-4703)/[MER-5157](https://eliterate.atlassian.net/browse/MER-5157), with a regression test attached).

assets/src/components/editing/elements/link/LinkElement.tsx
Francisco-Castro: Confirmed safe. This preventDefault is on the editor's placeholder anchor only — LinkEditor renders `<a href="#">` for links inside the Slate editor (and read-only previews), and is imported solely by `modelEditorDispatch.tsx`. Without it, clicking a link while editing makes the browser jump to # (top of page). It can't affect how real links behave: delivered/published links are rendered by separate code — `data/content/writers/html.tsx` (internal-link/external-link) and server-side `rendering/content/html.ex` — neither of which uses LinkEditor. End-to-end verified: an instructor-inserted course-page link sends and renders correctly as `<a class="internal-link" href="/sections/…/lesson/…">`. This change is part of the link-editing flow this PR builds on (now a full internal-page picker — see the updated description).

assets/src/components/editing/elements/link/LinkModal.tsx
Francisco-Castro: It's part of the link-editing flow (same change set). This is the error branch of the link modal's page-fetch — without it, a failed fetch leaves the modal spinning on "Loading…" indefinitely; the callback surfaces a clear error instead (as you note, better UX). It's actually central to this work: this exact failure path is what revealed that the modal's page-fetch hit an author-only endpoint instructors can't reach — which this PR now fixes at the root (email mode sources pages server-side and skips that fetch). The callback remains the correct fallback for authoring-mode fetch failures. No effect on link rendering.

lib/oli_web/components/layouts/instructor_dashboard.html.heex
Francisco-Castro: Verified there's no clash — including a local test. The user menu's z-50 is on its dropdown, which lives inside #instructor-dashboard-header, and that header is sticky top-0 z-[90], so it forms its own stacking context. z-index only competes within the same stacking context, so the menu's z-50 is scoped inside the z-90 header and always renders above the flash; the flash's z-50 is a separate root-level context sitting under z-90. They never compete — the menu stays on top regardless.

Tested locally: injected a tall flash into #flash_container and opened the account menu over it — the dropdown renders fully on top of the flash (screenshot below). The 40→50 bump is just so the flash clears the sticky table header (z-[40]). Same result on both mobile (fixed) and desktop (sm:absolute), since both are inside the z-90 header.

lib/oli_web/live/delivery/instructor_dashboard/instructor_dashboard_live.ex
Francisco-Castro: Good catch — reworked the draft task lifecycle. Switched from Task.Supervisor.async_nolink to LiveView's start_async/handle_async, so the task is tied to the LiveView: it's automatically cancelled on navigate-away (no more orphaned task with nothing to reply to), and a crash arrives as {:exit, reason} in handle_async (surfaced as an error, not a dashboard crash). Closing the modal also explicitly cancels the in-flight task.

To fully prevent a late/superseded draft from landing in the wrong place, each generation is tagged with a unique request id: a delivered result is applied only if it matches the modal's current generation, and close/reopen/regenerate all invalidate the previous id. So a result from a closed, reopened, or superseded generation is dropped deterministically — no timing window.

Covered by tests for the full handle_info → start_async → handle_async → deliver chain, crash handling, cancel-on-close, request-id scoping (matching applied; stale/duplicate dropped), and the close→reopen→regenerate lifecycle (the reopen-reset is mutation-verified). Commits 7aa00109ed + 9868797947.

eliknebel (changes requested):
While testing this, I noticed a couple UI issues that I think we want to address

Image

1. The actual email body input only fits to a single line/row in the input, which could make it confusing/difficult for users to actually select the input. The actual input should span the entire available height of the white background
2. The email and Subject input backgrounds are the same as the modal background, which makes these appear to be 'disabled'. We should use the same white (and dark when in dark mode) as the email body for these input fields. If we adjust the email input background, we might need to also adjust the chips color so that they stand out against the white/dark input background. We should ensure these changes also work as expected when in dark mode.
3. It seems to me that it would make more sense for the AI generation buttons to sit directly on the modal background, and not within a white-background box which appears like an input box, when it is not

Francisco-Castro:
@eliknebel here is how it looks now. Notice that the placeholder text and the cursor are aligned, and the placeholder text makes more sense.

image


image

eliknebel (approved):

PR #6610, opened 2026-05-28
[BUG FIX] [MER-5565] Fix adaptive scoring attempt count across page retakes

manelli:
Fixes adaptive attempt-based scoring so first tries on later page attempts are scored as attempt 1.

Adaptive activity attempts use an absolute attempt number that advances along with the page/resource attempt. For example, when a learner starts page attempt 3, the first adaptive activity attempt on that page can also have `attempt_number: 3`.

That absolute activity attempt number is useful for persistence, ordering, and analytics, but it is not the same thing as the learner’s try number within the current page attempt. Attempt-based adaptive scoring needs the screen-local try count:

- Page attempt 3, first try on the screen should score as try 1.
- Page attempt 3, second try on the screen should score as try 2.

Previously, adaptive scoring passed the absolute activity attempt number directly as `currentAttemptNumber`, so a first correct answer on page attempt 3 could be treated as try 3 and receive reduced credit.

This changes adaptive scoring to calculate the screen-local attempt number as:

`activity_attempt_number - resource_attempt_number + 1`

The same relative calculation is applied in the client-side adaptive check path.

See: https://eliterate.atlassian.net/browse/MER-5665

### DIFF
--- a/assets/src/apps/delivery/store/features/adaptivity/actions/triggerCheck.ts
+++ b/assets/src/apps/delivery/store/features/adaptivity/actions/triggerCheck.ts
@@ -34,6 +34,7 @@ import {
 import {
   selectPreviewMode,
   selectResourceAttemptGuid,
+  selectResourceAttemptNumber,
   selectReviewMode,
   selectSectionSlug,
 } from '../../page/slice';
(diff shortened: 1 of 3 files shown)


manelli:
Closing. duplicate of https://github.com/Simon-Initiative/oli-torus/pull/6611

PR #6620, opened 2026-06-02
[ENHANCEMENT] [MER-5619] New Entry Points to Instructor View

gastonabella:
[MER-5619](https://eliterate.atlassian.net/browse/MER-5619)

### Summary
Added an Edit action to Customize Content remix page entries, now pointing to the new LiveView-based Instructor Preview route introduced in [Simon-Initiative/oli-torus#6635](https://github.com/Simon-Initiative/oli-torus/pull/6635).

Added unsaved changes protection when navigating from Customize Content to Instructor Preview, including a save-and-continue modal with dynamic messaging based on the navigation target.

Updated Assessment Settings so assessment titles link to the new Instructor Preview route, passing the current page as `return_to` so users can navigate back to their original context.

Aligned the unsaved changes modal and remix action icons with the Figma designs.

### Notes
This PR now depends on [MER-5617](https://eliterate.atlassian.net/browse/MER-5617) / [Simon-Initiative/oli-torus#6635](https://github.com/Simon-Initiative/oli-torus/pull/6635), which adds the LiveView-based Instructor Preview route used by these links.

[MER-5619]: https://eliterate.atlassian.net/browse/MER-5619?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MER-5617]: https://eliterate.atlassian.net/browse/MER-5617?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

https://github.com/user-attachments/assets/27152dd0-ae39-4978-87b3-894276832958


gastonabella:
Addressed the UI review feedback by increasing the unsaved changes modal close button hit area while keeping the icon visually the same size.

I did not make changes for the other AI review comments:
- The redirect target is already validated with `valid_internal_path?/1`, which only allows internal paths.
- `push_navigate` is now used for the Instructor Preview path after the MER-5617 LiveView route integration; the remaining `redirect` fallback is intentional for non-Instructor Preview navigation.
- I did not restore a “leave without saving” action because the current modal design/spec only includes `Cancel` and `Save and continue`.

lib/oli_web/live/delivery/remix/actions.ex
```diff
@@ -12,10 +12,26 @@ defmodule OliWeb.Delivery.Remix.Actions do
...
+        :if={@resource_type == ResourceType.id_for_page() and not is_nil(@edit_url)}
+        variant={:secondary}
+        size={:sm}
+        class="!px-4 no-underline hover:no-underline"
+        href={@edit_url}
+        aria-label={@edit_label}
+        data-unsaved-changes-reason="instructor_view"
+      >
+        <:icon_left>
+          <OliWeb.Icons.edit_instructor_view class="h-5 w-5" />
```
nicocirio, line 31: We already have an edit icon in `OliWeb.Icons`. What do you think about reusing that instead of creating a new one?
I think that maybe a small tweak may be needed to size it as the Figma design, but I like that approach better than creating a specific new icon for this use case, which will probably won't be used again elsewhere.

lib/oli_web/live/delivery/remix_section.ex
```diff
@@ -1281,6 +1299,27 @@ defmodule OliWeb.Delivery.RemixSection do
 
   defp valid_internal_path?(_), do: false
 
+  defp navigate_to_pending_target(socket, target, :instructor_view) do
+    push_navigate(socket, to: target)
+  end
+
+  defp navigate_to_pending_target(socket, target, _reason) do
+    redirect(socket, to: target)
+  end
```
nicocirio, line 1308: I get the need for this helper function due to the comments you made on the PR, and now understand why sometimes we push_navigate, and sometimes we redirect. 
I believe these kinds of comments are the ones we should encourage our new `documentation SKILL` to handle.

lib/oli_web/live/sections/assessment_settings/settings_live.ex
```diff
@@ -107,12 +108,23 @@ defmodule OliWeb.Sections.AssessmentSettings.SettingsLive do
...
 
+  defp current_path(url) do
+    %{path: path, query: query} = URI.parse(url)
+
+    case query do
+      nil -> path
+      "" -> path
+      query -> "#{path}?#{query}"
+    end
+  end
```
nicocirio, line 126: We have the same helper function defined on `remix_section.ex`.
This is an example of the discussion we've been having lately, about how AI tends to duplicate code with small helper functions instead of abstracting those into a common/reusable module.
Maybe we should think about developing a new "clean up/refactor" SKILL to be executed as a final step of our harness workflow. It's just an idea that we can follow up on offline.

nicocirio (approved):
LGTM!
I left a small NIT about an icon, but feel free to merge it as is.

PR #6625, opened 2026-06-03
[BUG FIX] [MER-5695] [MER-5694] Fix manual grading rendering of legacy adaptive reports

andersweinstein:
Fixes issues in manual grading review rendering for legacy adaptive/EverApp CAPI reports, as seen in the old BioBeyond Blue Planet report which, unlike newer version, stores data in EverApp.

## Root Causes ##
- Blank report: the legacy CAPI iframe content was loaded, but Torus constrained the nested iframe/host sizing so the rendered content collapsed or was clipped.
- Editable report in instructor review: EverApp initialized in VIEWER mode even during manual grading review, and the legacy report hydrated its persisted readonly:false value unless Torus explicitly forced read-only review behavior.
- Potential instructor data mutation: EverApp save/submit handlers were not guarded for review mode, so instructor interaction could flow through the normal student persistence path.
  
## Direct Fixes ##
 - Fix nested CAPI iframe sizing so legacy report content is visible instead of collapsing/clipping in manual grading.
  - Make EverApp initialize activities with REVIEW mode during instructor review/manual grading.
  - Suppress EverApp save/submit handlers in review mode so instructor interaction cannot persist as student work.
  - Force CAPI readonly: true in review mode so legacy reports render read-only even when the stored student attempt has readonly: false.
  
 ## Defensive Hardening ##
  - Resolve legacy report data from review snapshots, including aggregate app.<simId>.data values.
  - Add stable IDs/namespacing for manual grading adaptive preview iframes to avoid LiveView hook and duplicate ID issues.

  ## Testing
  - Manually verified the legacy Blue Planet report renders in manual grading review.
  - Manually verified the rendered report is read-only in instructor review.

Screenshot 2026-06-02 at 9 12 24 PM

### DIFF
--- a/assets/src/apps/delivery/components/ActivityRenderer.tsx
+++ b/assets/src/apps/delivery/components/ActivityRenderer.tsx
@@ -76,6 +76,106 @@ const AllAttemptStateList: {
   attemptGuid: string;
   attempt: unknown;
 }[] = [];
+
+const normalizeReviewDataKey = (key: string) => key.replace(/[^a-zA-Z0-9]/g, '').toLowerCase();
+
+const valueForKey = (snapshot: Record<string, unknown>, key: string) =>
+  Object.prototype.hasOwnProperty.call(snapshot, key) ? snapshot[key] : undefined;
+
+const valueFromAggregateData = (
+  snapshot: Record<string, unknown>,
+  simId: string,
+  key: string,
+): { found: boolean; value?: unknown } => {
+  const data = valueForKey(snapshot, `app.${simId}.data`);
+
+  if (data === undefined) {
+    return { found: false };
+  }
+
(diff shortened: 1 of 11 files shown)


andersweinstein:
Regarding the AI UI note about iframe width/overflow: this behavior is intentional for this hotfix. The legacy Blue Planet report was loading but rendering blank because the nested CAPI iframe/host sizing was being constrained or clipped. Preserving the authored iframe dimensions and allowing the manual-grading wrapper to provide the scroll boundary is the part of the fix that restored the report rendering in local verification. I agree broader responsive containment for arbitrary CAPI activities is worth considering separately, but reintroducing max-width/max-height clipping here risks recreating MER-5695.

darrensiegel (approved):

PR #6628, opened 2026-06-03
[BUG FIX] [MER-5698] fix remix move button crash

andersweinstein:
## Summary

  Fixes a crash when moving remix items with the Move button/modal in section and template customize-content pages.

  ## Details

  Selecting a destination container in the Move modal reused the shared HierarchyPicker.update_active handler. That handler attempted to clear :error_message with map update syntax, but Move modal assigns do not include :error_message, causing a KeyError.

  The fix uses Map.put/3 so the handler works for both Add Materials and Move modal state.

  ## Regression Source

  This was introduced by 47754de102 ([BUG FIX] [MER-5615] Prevent remixing projects with shared resources (#6577)), which added error_message: nil handling for Add Materials modal errors. That commit is included in both v0.33.0 and v0.33.1.

  ## Testing

  Added regression coverage for:

  - Section remix Move modal moving an item into a selected container
  - Template/product remix Move modal moving an item into a nested selected container and saving

  Ran:

  mix test test/oli_web/live/remix_section_test.exs:37 test/oli_web/live/remix_section_create_container_live_test.exs:103

 

[MER-5615]: https://eliterate.atlassian.net/browse/MER-5615?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli_web/live/delivery/remix_section.ex
+++ b/lib/oli_web/live/delivery/remix_section.ex
@@ -753,11 +753,10 @@ defmodule OliWeb.Delivery.RemixSection do
 
     active = Hierarchy.find_in_hierarchy(hierarchy, uuid)
 
-    modal_assigns = %{
+    modal_assigns =
       modal_assigns
-      | active: active,
-        error_message: nil
-    }
+      |> Map.put(:active, active)
+      |> Map.put(:error_message, nil)
 
     {:noreply, assign(socket, modal_assigns: modal_assigns)}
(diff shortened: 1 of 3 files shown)


andersweinstein:
Redoing with correct target to avoid spurious comments.

PR #6635, opened 2026-06-04
[ENHANCEMENT] [MER-5617] Add instructor preview shell and relink delivery navigation

nicocirio:
Jira ticket: https://eliterate.atlassian.net/browse/MER-5617

## Summary
- Add the instructor preview shell for delivery pages and lesson preview.
- Preserve `return_to` and `sidebar_expanded` across preview navigation.
- Relink delivery entry points from the deprecated controller-backed UI to the NG23 LiveView surfaces.
- Remove dead delivery code that is no longer rendered.

## Implementation notes
- Added a dedicated instructor preview lesson LiveView plus shared preview-return resolution.
- Extended `LearnLive` to handle nested containers so instructor/student entry points land on the NG23 course view.
- Updated delivery navigation helpers, layouts, and preview routes to keep preview context stable.
- Documented the architecture and the phase-by-phase rollout in the instructor view shell work item docs.

### Preview header: "Return To" link is preserved throughout delivery preview navigation

https://github.com/user-attachments/assets/d7a9f12f-44cd-48c5-a24e-5a18d0481ece

### Overview > Course Content: When accessing a container, the link was updated to point to the NG23 UI instead of the deprecated controller view.

https://github.com/user-attachments/assets/a64ed8ea-45d5-431f-9e46-2a450500dace

## Other
- Removed the dead `CourseLatestVisitedPage` component.
- Relinked `Course Content` actions to the new NG23 UI.
- Fixed the navigation bar alignment against the side curves.
#### Before

https://github.com/user-attachments/assets/7c90455f-9539-46db-8dfa-f238f329d37e

#### After

https://github.com/user-attachments/assets/3a1251f8-59f9-497a-a2aa-dc5f0f29d7dc

- Fixed the `/dev/icons` crash caused by an icon attr missing a default value.
#### Before
image

#### After

https://github.com/user-attachments/assets/a9d6e0db-0f41-4563-9906-2c79b945fd0c

## Validation
- `mix format`
- `mix compile`
- `mix test test/oli_web/components/delivery/assignments/assignment_card_test.exs test/oli_web/components/delivery/deliberate_practice_card_test.exs test/oli_web/components/delivery/exploration_card_test.exs test/oli_web/components/delivery/layouts_test.exs test/oli_web/controllers/page_delivery_controller_test.exs test/oli_web/live/delivery/instructor/preview_lesson_live_test.exs test/oli_web/live/delivery/student/assignments_live_test.exs test/oli_web/live/delivery/student/explorations_live_test.exs test/oli_web/live/delivery/student/index_live_test.exs test/oli_web/live/delivery/student/learn_live_test.exs test/oli_web/live/delivery/student/schedule_live_test.exs test/oli_web/live/delivery/student/utils_test.exs test/oli_web/live/delivery/student_dashboard/course_content_live_test.exs test/scenarios/run_all_test.exs test/scenarios/scenario_runner_test.exs`
- Result: `446 tests, 0 failures (42 excluded)`


nicocirio:
Follow-up to https://github.com/Simon-Initiative/oli-torus/pull/6635#issuecomment-4625967453

Valid observation, but I am deferring this for now.

This PR is already large, and moving preview page-context construction to async assigns or lazily loading hierarchy / annotations would expand the scope beyond the hardening follow-up we wanted here. Those are good candidates for a focused performance follow-up once this preview shell work lands.

nicocirio:
Follow-up to https://github.com/Simon-Initiative/oli-torus/pull/6635#issuecomment-4625967453

Valid observation, but I am deferring this for now.

This PR is already large, and moving preview page-context construction to async assigns or lazily loading hierarchy / annotations would expand the scope beyond the hardening follow-up we wanted here. Those are good candidates for a focused performance follow-up once this preview shell work lands.

nicocirio:
Follow-up to https://github.com/Simon-Initiative/oli-torus/pull/6635#issuecomment-4625974892

Covered by the same hardening change taken for the security review comment.

PreviewLessonLive now sanitizes request_path independently instead of preserving it wholesale when return_to is valid.

nicocirio:
Follow-up to https://github.com/Simon-Initiative/oli-torus/pull/6635#issuecomment-4625967227

Covered by the same scroller hardening change taken for the security review comment.

The dynamic selector lookup now uses escaped, quoted attribute selectors instead of interpolating raw values into querySelector.

nicocirio:
Follow-up to https://github.com/Simon-Initiative/oli-torus/pull/6635#issuecomment-4625967242

Addressed.

- Added a <main id="main"> landmark to PreviewLessonLive.
- Replaced the invalid role="section_*_target" DOM hook with data-scroll-target for nested section targets.
- Updated the scroller to support that new hook without changing the existing id-based cases used by modules and page cards.

lib/oli_web/components/delivery/layouts.ex
gastonabella, line 1320: Nit, but this function got bigger, and now it's a little bit more complicated to follow the logic with the nested case/if statements. I wonder if it'd make it more readable to pattern match the function on the `preview_mode` param (`true` or `false`), or even find a way to refactor it, since both scenarios have a very similar code block.

Another thought I have about AI-generated code is that it doesn't tend to add comments in parts of the code where context is sometimes needed to understand what's happening. We should find a way to refine the skills so they take this into account, without going overboard and adding comments everywhere.

gastonabella (approved):
It looks great! I just left a small comment that doesn't prevent this PR from being merged.

PR #6648, opened 2026-06-09
[ENHANCEMENT] [MER-5627] Update assignment terms page

andersweinstein:
## Summary

  Implements the [MER-5627](https://eliterate.atlassian.net/browse/MER-5627) Assignment Terms redesign for the student prologue page.

  This PR replaces the legacy prologue terms/attempts presentation with the Figma-style grouped
  Assignment Terms layout, backed by a dedicated view model for schedule, time limit, scoring, and
  attempts state.

  ## Changes

  - Added `Oli.Delivery.Page.AssignmentTerms` to centralize Assignment Terms card visibility, labels,
  copy, and historical attempt formatting.
  - Reworked the student prologue UI to render:
    - Schedule card
    - Time Limit card
    - Scoring card
    - Attempts summary/history card
    - Late submission warning/default states
  - Preserved begin-attempt behavior, password modal behavior, disabled attempt state, review links, and
  adaptive feedback.
  - Hid duplicate assignment/schedule metadata from the prologue page header for this redesigned view.
  - Added accessibility labels/landmarks for the new grouped cards and attempt-history controls.
  - Added focused coverage for normal, disabled, unscheduled, read-by, unlimited-attempt, SAYG, and
  ungraded-attempt edge cases.

  ## Notes

  - Existing grace-period and retake-mode settings were not displayed in the pre-change prologue UI, so
  this PR does not introduce visible copy for those settings.
  - Figma is treated as the UI source of truth where it differs from the old ticket/body behavior.

  ## Verification

  - `mix format lib/oli_web/live/delivery/student/prologue_live.ex test/oli_web/live/delivery/student/
  prologue_live_test.exs`
  - `mix test test/oli/delivery/page/assignment_terms_test.exs`
  - `mix test test/oli_web/live/delivery/student/prologue_live_test.exs`
  
  
Screenshot 2026-06-09 at 11 05 58 AM

[MER-5627]: https://eliterate.atlassian.net/browse/MER-5627?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


andersweinstein:
[Codex:] The blocking-gate behavior is already preserved in the current head: `assignment_terms` is rendered with `:if={!@show_blocking_gates?}`.

For the past-attempts performance note, the latest follow-up removed full `ResourceAttempt` structs from the assignment terms model and stores only compact display fields plus feedback text. Capping/paginating past attempts would change the visible behavior of this terms page, so I would leave that for a separate product decision if needed.

lib/oli/delivery/attempts/feedback_text.ex
```diff
@@ -0,0 +1,182 @@
...
+
+  defp manual_part_attempt_feedback_query(resource_attempt_ids) do
+    from(pa in PartAttempt,
+      join: aa in ActivityAttempt,
+      on: pa.activity_attempt_id == aa.id,
+      where:
+        aa.resource_attempt_id in ^resource_attempt_ids and pa.grading_approach == :manual and
+          not is_nil(pa.feedback),
+      order_by: [asc: aa.id, asc: pa.id],
+      select: {aa.resource_attempt_id, pa.feedback}
```
darrensiegel, line 179: It seems like this is assuming that for each resource attempt, that there is at most ONE manually scored part?

lib/oli/delivery/attempts/feedback_text.ex
```diff
@@ -0,0 +1,182 @@
...
+    resource_attempt_ids = Enum.map(resource_attempts, & &1.id)
+
+    feedback_texts_by_resource_attempt_id =
+      resource_attempt_ids
+      |> manual_part_attempt_feedback_query()
+      |> Repo.all()
+      |> Enum.reduce(%{}, fn {resource_attempt_id, feedback}, acc ->
+        feedback_texts = extract_from_part_attempt(%{feedback: feedback})
+
+        Map.update(acc, resource_attempt_id, [feedback_texts], fn existing ->
```
darrensiegel, line 47: If we do have more than one part (per resource attempt id) with manually scored feedback, here is where we will silently drop them.

darrensiegel (dismissed):

lib/oli/delivery/attempts/feedback_text.ex
andersweinstein: [Codex:] It should not be assuming that. The query can return multiple manual part-attempt rows for the same resource attempt id, and the reducer stores a list of extracted feedback lists per resource attempt (`[feedback_texts | existing]`). The final `Map.new/2` path then `Enum.reverse()`s, `List.flatten()`s, and `Enum.uniq()`s that accumulated list, so multiple manually scored parts for the same resource attempt should all be represented in the returned feedback text list.

lib/oli/delivery/attempts/feedback_text.ex
andersweinstein: [Codex:] The current accumulator should preserve these rather than drop them: the initial value is `[feedback_texts]`, and each additional row for the same resource attempt prepends another feedback-text list with `[feedback_texts | existing]`. Later we reverse, flatten, and de-dupe before returning `{attempt_guid, feedback_texts}`. So multiple manually scored parts under one resource attempt id should be included; adding a focused regression test for that case would be a reasonable follow-up if we want the behavior documented in coverage.

andersweinstein: Review comment in error. Added regression test verifying multiple manually-scored part handling.

eliknebel (approved):

PR #6657, opened 2026-06-10
[ENHANCEMENT] [MER-5633] One at a time mode updates

andersweinstein:
## Summary
Updates one-at-a-time assessment layout to match the revised Figma direction. [MER-5633](url)

  - Removes the inner scrollable question box so question content can use the page layout normally.
  - Removes the right-side per-part point rail while preserving the question-wide points display.
  - Aligns question-list rail in margin left of question content area.
  
  This also fixes two other pre-existing issues discovered in the course of testing One-at-a-Time mode:
 
  ## One-at-a-Time Feedback Rendering fix

  While validating the layout change, we found that in the case of one-at-a-time presentation with score-at-the-end (SATE),  multi-input questions could render duplicate feedback after evaluation/reload. This was a pre-existing bug. This does not arise with one-at-a-time/Score-as-You-Go (SAYG).

  This happened because although activities are generally responsible for rendering their own feedback, in the case of one-at-a-time/SATE, question submission is initiated by LiveView event handler which evaluates server-side and renders feedback from LiveView state. Activities using common components like `EvaluationConnected` avoided rendering feedback in this case. But the feedback-controlling logic in the custom multi-input activity feedback components resulted in that activity _also_ rendering its own feedback after remounting with evaluated state in this state.
  
   To avoid duplicate feedback, a common predicate was defined to ensure activity-owned feedback and graded points are consistently suppressed for the case of graded one-at-a-time/score-at-the-end delivery. SAYG and review behavior are unchanged.
  
 ## One-at-a-Time Finish Attempt Routing fix

  While validating the one-at-a-time/SATE flow, we found that the one-at-a-time component had its own `finalize_attempt` handler that had drifted from the parent `LessonLive` finalization behavior. In particular, it always redirected back to lesson delivery and did not honor the usual review-submission redirect.

  This appears to be a pre-existing one-at-a-time bug rather than a regression due to the layout changes, but fix is included here because this PR updates the same specialized presentation mode. The component now routes Finish Attempt through the parent `LessonLive` finalizer, so it shares the same finalization and review routing behavior as the normal scored page submit flow.

  ## Testing

  - `mix test test/oli_web/live/delivery/student/lesson/components/one_at_a_time_question_test.exs`
  - `mix test test/oli_web/live/delivery/student/lesson_live_test.exs`
  - `yarn test test/delivery/activity_feedback_visibility_test.ts --runInBand`
  - `yarn test test/multi_input/multi_input_delivery_test.tsx --runInBand`
  
  Browser MCP verified that a non-SAYG one-at-a-time multi-input question renders LiveView feedback/points without duplicate activity feedback/points after reload.
  
  ## One-at-a-Time + Score-As-You-Go ##
  
Screenshot 2026-06-11 at 10 23 20 AM

  ## One-at-a-Time + Score-at-the-End ##
  
Screenshot 2026-06-11 at 10 30 38 AM

[MER-5633]: https://eliterate.atlassian.net/browse/MER-5633?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/assets/src/components/activities/common/delivery/evaluation/EvaluationConnected.tsx
+++ b/assets/src/components/activities/common/delivery/evaluation/EvaluationConnected.tsx
@@ -1,23 +1,37 @@
 import React from 'react';
 import { useSelector } from 'react-redux';
+import { ActivityContext } from 'components/activities/DeliveryElement';
 import { useDeliveryElementContext } from 'components/activities/DeliveryElementProvider';
 import { Evaluation } from 'components/activities/common/delivery/evaluation/Evaluation';
 import { Submission } from 'components/activities/common/delivery/evaluation/Submission';
+import { DeliveryMode } from 'components/activities/types';
 import { ActivityDeliveryState, isEvaluated } from 'data/activities/DeliveryState';
 
+export const isOneAtATimeScoreAtTheEndDelivery = (context: ActivityContext, mode: DeliveryMode) =>
+  context.graded && mode !== 'review' && context.oneAtATime && context.batchScoring;
+
+export const shouldShowActivityFeedback = (
+  context: ActivityContext,
+  mode: DeliveryMode,
+  evaluated: boolean,
+) =>
+  context.showFeedback == true &&
+  evaluated &&
+  context.surveyId === null &&
+  !isOneAtATimeScoreAtTheEndDelivery(context, mode);
+
 export const EvaluationConnected: React.FC = () => {
-  const { context, writerContext } = useDeliveryElementContext();
-  const { surveyId } = context;
+  const { context, mode, writerContext } = useDeliveryElementContext();
   const uiState = useSelector((state: ActivityDeliveryState) => state);
 
   return (
     <>
       <Evaluation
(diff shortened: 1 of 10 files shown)


andersweinstein:
Re AI review feedback change concerns:

This feedback predicate change is intentional. The intended behavior change is limited to the one-at-a-time + score-at-the-end case, where LiveView owns the per-question submit/evaluation flow and already renders the feedback. In that mode, activity-rendered feedback must be suppressed to avoid duplicate feedback after reload/remount.

  For normal scored delivery, this should _not_ cause early feedback exposure: feedback is still gated on the presence of an evaluated attempt state, and the client does not receive evaluated state before the page/attempt is submitted. So the helper does _not_ make feedback appear during an in-progress normal scored assessment; it only prevents activity-owned duplicate feedback in the one-at-a-time score-at-end path. 

  SAYG remains unaffected because it is not batch-scored, and review remains unaffected because feedback in review is explicitly allowed.

eliknebel (approved):

PR #6660, opened 2026-06-12
[FEATURE] [MER-5630] [MER-5707] Lock score mode changes if student attempts exist

andersweinstein:
Implements [MER-5630](https://eliterate.atlassian.net/browse/MER-5630) scoring-mode setting lock and related instructor guidance.

  - Locks scoring mode once learner attempts exist for an assessment, while remaining editable over instructor-only attempts to prevent scoring-mode updates and bulk-apply scoring-mode changes for learner-started assessments.
  - Include warning on bulk-apply settings that scoring mode will not be applied to assessments where student attempts exist.
  - Renders locked setting with lock icon and explanatory tooltip
  - Adds info icon with tooltip to column header
  - Adds post-section-creation Manage-page setup recommendation with a Review Settings path, replacing the “section created successfully” flash since it includes that message.
  - Adds assessment-settings warning banner with per-user "Don't show again" flag persisted in user preferences.
  
 This also aligns assessment-settings/student exceptions UI with updated Figma color tokens [MER-5707](https://eliterate.atlassian.net/browse/MER-5707):
  - Use Background/bg-secondary for the table container and sticky first-column backgrounds.
  - Use Table/table-top-row for table header and sticky header backgrounds.
  - Use Text/text-high for primary table text, including sorted column headers.
  - Use Text/text-button for enabled select field text, underline, and dropdown chevron.
  - Text/text-low-alpha on disabled/locked select-style fields 
  
  ## Testing

  - mix test test/oli_web/live/sections/assessment_settings/settings_live_test.exs
  - mix test test/oli_web/live/delivery/instructor_dashboard/manage/manage_tab_test.exs test/oli_web/controllers/
    open_and_free_controller_test.exs
  - mix test test/oli_web/live/new_course/course_details_test.exs
  - yarn run check-types

 ## New Section Creation Warning ##
Screenshot 2026-06-12 at 2 14 07 PM

## Assessment Settings Warning ##
Screenshot 2026-06-15 at 11 22 00 AM

## Locked & Disabled Settings ##
Screenshot 2026-06-15 at 11 29 21 AM

## Dark Mode ##
Screenshot 2026-06-15 at 11 32 59 AM

## Bulk Apply Warning ##
Screenshot 2026-06-12 at 2 18 49 PM

[MER-5630]: https://eliterate.atlassian.net/browse/MER-5630?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[MER-5707]: https://eliterate.atlassian.net/browse/MER-5707?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


assets/css/table.css
```diff
@@ -18,8 +18,9 @@
   @apply h-[50px] text-left font-normal text-xs sm:text-sm relative;
 }
 
-.instructor_dashboard_table th[data-sort-column='true'] {
```
eliknebel, line 21: Ideally we would use tailwind for all styles moving forward, however since these styles already exist here this is probably the right place for these changes barring a tech debt refactor

andersweinstein: Noted. Yes, seems outside scope of this ticket to migrate away from existing css files.

lib/oli/delivery/settings/assessment_settings.ex
```diff
@@ -186,8 +186,8 @@ defmodule Oli.Delivery.Settings.AssessmentSettings do
         name_with_container_label: name_with_container_label,
         revision_slug: rev.slug,
         is_adaptive: Map.get(rev.content || %{}, "advancedDelivery") == true,
-        student_attempts_count: student_attempts_count,
-        has_student_attempts: student_attempts_count > 0,
+        student_attempts_count: if(has_student_attempts, do: 1, else: 0),
+        has_student_attempts: has_student_attempts,
```
nicocirio: `student_attempts_count` no longer appears to be needed here. In this PR it is no longer used as a real count, only as a boolean-like `0/1` flag, since the bulk apply warning no longer says `"X assignments"` and now uses a generic message instead. At that point, `student_attempts_count` is just duplicating `has_student_attempts`, so it seems we could remove it and rely only on `has_student_attempts` for the scoring-mode lock / refresh logic.

lib/oli_web/live/sections/overview_view.ex
```diff
@@ -63,6 +65,11 @@ defmodule OliWeb.Sections.OverviewView do
 
         %{base_project: base_project} = section |> Repo.preload(:base_project)
 
+        show_section_created_setup =
+          Phoenix.Flash.get(socket.assigns.flash, :section_created_setup) == true
+
+        socket = clear_flash(socket, :section_created_setup)
+
```
nicocirio: `section_created_setup` looks fragile here. The card is derived from flash during `mount/3`, but `mount/3` runs twice (disconnected + connected). On the initial HTTP render the controller flash is present, so the banner appears; on the connected mount the flash is no longer there, so `show_section_created_setup` becomes false and the banner disappears. That matches the flicker seen after section creation. This likely needs a persistence mechanism other than flash-for-mount, or at least a way to survive the connected remount.

https://github.com/user-attachments/assets/283b7307-40ec-49a8-be51-f147aeb78648

nicocirio (changes requested):
Nice work! We should ensure the message is shown during section creation, and the PR should be ready for merge.

lib/oli_web/live/sections/overview_view.ex
andersweinstein: Thanks for catching! This worked, then there was a change in response to AI review, but ... did not retest. 😳

lib/oli/delivery/settings/assessment_settings.ex
andersweinstein: Done

lib/oli_web/live/sections/overview_view.ex
andersweinstein: - Restored section-created setup card signaling to ?section_created=true.
 - Added SectionCreatedUrlCleanup hook to remove that query param after mount without hiding the card.
 - Updated creation/manage tests for the query-param flow.

andersweinstein:
[Codex:] Remaining AI review replies:

- Elixir: "Missing URL cleanup hook implementation" appears stale/incorrect. The hook exists in `assets/src/hooks/section_created_url_cleanup.ts` and is registered in `assets/src/hooks/index.ts`; it removes only `section_created` via `history.replaceState` after mount.

- UI: "Tooltip id collision across table instances" is valid as a general shared-component concern, but the concrete tooltip introduced by this PR supplies an explicit `tooltip_id` (`assessment-settings-batch-scoring-column-tooltip`). I would treat requiring table-level ids for every future tooltip-icon use as follow-up hardening rather than necessary for this PR.

nicocirio (approved):

PR #6661, opened 2026-06-15
[FEATURE] [MER-5632] Navigating away from SAYG assignment

simonchoxx:
[MER-5632](https://eliterate.atlassian.net/browse/MER-5632)

## Summary

This PR adds a saved-work notice for students when they leave a score-as-you-go assignment. The notice is shown on the next page they navigate to so students know their work was saved and that they can resume the assignment before the due date.

### Changes

- Adds a client-side hook to track when a student is leaving a score-as-you-go lesson and show the saved-work notice after navigation.
- Renders the notice mount point across student delivery layouts and supporting pages where students can land after leaving a lesson.
- Adds the notice source marker to score-as-you-go lesson pages so navigation can be detected reliably.
- Keeps the notice dismissible and avoids showing it repeatedly once consumed.
- Adjusts layout placement for delivery pages, settings, research consent, and student workspace pages so the notice does not block key content or navigation.

### Tests

- Adds hook coverage for the saved-work notice behavior.
- Adds LiveView/controller coverage for the notice mount points and SAYG lesson marker behavior.
- Adds Playwright coverage for leaving a SAYG lesson through lesson back navigation, bottom lesson navigation, and user menu destinations.

### Visual Changes

https://github.com/user-attachments/assets/17a90447-c3c5-4fe8-9ee5-2959c9083015

---

https://github.com/user-attachments/assets/02662b7d-c814-489a-8e57-7213db653b6a

[MER-5632]: https://eliterate.atlassian.net/browse/MER-5632?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli_web/components/layouts/student_delivery.html.heex
+++ b/lib/oli_web/components/layouts/student_delivery.html.heex
@@ -56,12 +56,12 @@
       <.flash_group flash={@flash} />
     </div>
     <Components.Delivery.Layouts.sayg_saved_work_notice
-      :if={assigns[:active_tab] == :learn}
-      container_class="relative z-[45] mx-4 mt-4 mb-6 md:mx-6"
-    />
-    <Components.Delivery.Layouts.sayg_saved_work_notice
-      :if={assigns[:active_tab] != :learn}
-      class="top-2"
+      container_class={
+        if assigns[:active_tab] == :learn,
+          do: "relative z-[45] mx-4 mt-4 mb-6 md:mx-6",
+          else: "absolute left-4 right-4 z-[45] md:left-6 md:right-6"
+      }
+      class={if(assigns[:active_tab] != :learn, do: "top-2", else: nil)}
     />
     <div
       :if={assigns[:paywall_summary] && OliWeb.LayoutView.show_pay_early(@paywall_summary)}
--- a/lib/oli_web/components/layouts/student_delivery.html.heex
+++ b/lib/oli_web/components/layouts/student_delivery.html.heex
@@ -61,7 +61,7 @@
     />
     <Components.Delivery.Layouts.sayg_saved_work_notice
       :if={assigns[:active_tab] != :learn}
-      class="top-4"
+      class="top-2"
     />
     <div
       :if={assigns[:paywall_summary] && OliWeb.LayoutView.show_pay_early(@paywall_summary)}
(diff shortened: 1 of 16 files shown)


simonchoxx:
> ### AI Review — performance
> ### Bulk mark scenario visits instead of nested writes
> file: lib/oli/scenarios/hooks.ex line: 157 Description: This hook performs a DB write once per enrolled student inside a per-section loop. For scenarios with many sections or students, setup time scales poorly and can produce many small transactions/queries. Suggestion: Add a bulk helper in `Oli.Delivery.Sections` that accepts the scenario section slugs/ids, fetches enrolled students in one set operation, and marks visits with a single `insert_all`/upsert or batched operation.

This hook is only used as scenario setup for Playwright coverage and currently seeds a single section/student pair, so the nested writes stay small and isolated to test data setup.

A bulk helper in `Oli.Delivery.Sections` would require new shared API around enrollment state updates, which is larger than this change. I’d prefer to keep this hook simple here and revisit a bulk path separately if we end up using it for larger generated datasets.

lib/oli_web/components/layouts/student_delivery.html.heex
```diff
@@ -55,6 +55,14 @@
...
       <.flash_group flash={@flash} />
     </div>
+    <Components.Delivery.Layouts.sayg_saved_work_notice
+      :if={assigns[:active_tab] == :learn}
+      container_class="relative z-[45] mx-4 mt-4 mb-6 md:mx-6"
+    />
+    <Components.Delivery.Layouts.sayg_saved_work_notice
+      :if={assigns[:active_tab] != :learn}
+      class="top-2"
+    />
```
nicocirio, line 65: NIT: Since the two `:if` branches are mutually exclusive and only differ in container_class / class, this could be simplified to a single component invocation with conditional attrs:

```  

<Components.Delivery.Layouts.sayg_saved_wokk_notice
    container_class={
      if assigns[:active_tab] == :learn,
        do: "relative z-[45] mx-4 mt-4 mb-6 md:mx-6",
        else: "absolute left-4 right-4 z-[45] md:left-6 md:right-6"
    }
    class={if(assigns[:active_tab] != :learn, do: "top-2", else: nil)}
  />
```

That would reduce duplication and make future changes less likely to land in one branch but not the other.

nicocirio (approved):
Awesome work 👏 👏 ! I tested it locally and it works as expected
🚀 🚀

PR #6663, opened 2026-06-17
[ENHANCEMENT] [MER-5701] Comprehensive automated CAPI component tests

Francisco-Castro:
---

## Ticket

- **MER-5701** — https://eliterate.atlassian.net/browse/MER-5701  (epic **MER-5367** — Automated Testing)
- Requirement (verbatim): "If one does not exist already, this task is to create a fully comprehensive set of automated tests for the CAPI component (which allows simulations to communicate with a host adaptive page)."
- **Scope guidance (Eli, covering for Darren):** automated tests = Playwright driving the headless browser and asserting; **do not change or refactor production code** unless the testing uncovers a bug or a change is strictly required to enable a test (e.g. DOM test handles). This PR honors that — no CAPI/delivery runtime code changed; the only Elixir app change is the prod-gated `edit_adaptive_page` scenario-seeding directive.
- **Acceptance test cases (Kevin Segovia, in-ticket):** TC1 (external iframe loads + scrolls) and TC2 (CAPI variable reaches the rules engine via a trapstate condition). Captured verbatim at `docs/exec-plans/current/epics/automated_testing/capi/kevin-test-cases.md`. TC2 is reflected by Playwright test 3; TC1's iframe-load portion is covered by the stub iframe load exercised across the Playwright suite (scroll-specific behavior is not separately asserted here — it is unit-tested in `assets/test/delivery/janus_capi_iframe_test.ts`).

## What

Comprehensive automated test coverage for the CAPI component / interface — the `postMessage` protocol between an embedded simulation and the host adaptive page (`assets/src/components/parts/janus-capi-iframe/`). No prior coverage existed for the protocol itself; this adds it across three layers.

### Playwright (10) — `assets/automation/tests/torus/capi/capi.spec.ts`
1. Complete the CAPI handshake and assert the host echoes the requestToken + config (context/section/user/question). See `completes handshake with the sim`.
2. Drive ON_READY and assert the host pushes initial configData to the sim (page→sim VALUE_CHANGE) then INITIAL_SETUP_COMPLETE. See `pushes initial state then INITIAL_SETUP_COMPLETE after ON_READY`.
3. Set a variable, send CHECK_REQUEST, and assert the full check round-trip (CHECK_START_RESPONSE + CHECK_COMPLETE_RESPONSE) plus the observable trapstate effect (`acknowledged`), proving the variable reached the rules engine. See `runs the check lifecycle on CHECK_REQUEST`.
4. Set sim state, revisit the lesson, and assert the saved value is pushed back on re-init. See `restores saved sim state on revisit`.
5. Exercise SET_DATA_REQUEST and assert a success SET_DATA_RESPONSE echoing the value. See `acknowledges SET_DATA_REQUEST with a success response`.
6. Exercise GET_DATA_REQUEST for a stored key (hit) and an unknown key (miss → `exists:false`). See `returns stored data on GET_DATA_REQUEST and a miss for unknown keys`.
7. Exercise RESIZE_PARENT_CONTAINER_REQUEST (absolute + relative) and assert the response echoes messageId. See `responds to RESIZE_PARENT_CONTAINER_REQUEST (absolute and relative)`.
8. Send malformed / non-JSON / pre-handshake traffic and assert the listener survives and still handshakes. See `survives malformed and pre-handshake messages, then handshakes normally`.
9. Post a well-formed CAPI message from the top window and assert it is ignored (source check). See `ignores a CAPI message posted from a foreign source`.
10. Send a post-handshake message with a mismatched requestToken and assert it is still processed — characterizing the current boundary (source enforced, token not). See `documents current gap: same-iframe messages with a mismatched requestToken are still processed`.

### Jest (17) — `assets/test/adaptivity/capi_variable_test.ts`
1. Verify `capi.ts` type coercion and inference: `getCapiType`, `coerceCapiValue` (incl. ENUM-invalid throw), `parseCapiValue`, `isSpecialArrayString`, and the `CapiVariable` constructor.

### ExUnit (4) — `test/oli/scenarios/edit_adaptive_page_test.exs`
1. Verify the new `edit_adaptive_page` directive builds adaptive page content referencing the activity (happy path, `graded`, unknown activity, unknown page).

## How it's built

- Deterministic in-repo **stub sim** (`support/stub-sim.html` + `capiStub.ts`) loaded via `page.route()` — no live simulation, fully offline/CI-safe.
- Seeded through the scenario framework via a new **`edit_adaptive_page`** directive (`lib/oli/scenarios/`) that builds an adaptive page referencing a created activity. Test infrastructure only; the scenario endpoint is gated off in prod.

## Notable

- **No CAPI/delivery runtime code changed** — `git diff` on `assets/src`, `lib/oli_web`, and `lib/oli/delivery` is empty (per Eli's scope guidance above). The only Elixir app changes are scenario-seeding infrastructure (`lib/oli/scenarios/` + schema) used through the prod-disabled `/test/scenario-yaml` endpoint.
- Docs: `assets/automation/tests/torus/capi/capi.md` (run + coverage) and `docs/exec-plans/current/epics/automated_testing/capi/` (research, plan, findings).

## Findings surfaced (separate follow-up tickets, not addressed here)

- **F2 — no-rules check crash** (`triggerCheck` → `processResults` on undefined; medium): **TRIAGE-2359**.
- **`requestToken` not validated after handshake** (`ExternalActivity.tsx:1076` TODO; characterized by the requestToken test): **TRIAGE-2360**.
- **Duplicate `id="app"`** in `chromeless.html.heex`: **TRIAGE-2361**.
- **Sibling `edit_page_handler` swallows the same upsert error** fixed in the new `edit_adaptive_page` handler: **TRIAGE-2362**.
- **Scenario page-edit handlers mutate revisions in place** (theoretical post-publish concern; not triggered by the pre-publish seed): **TRIAGE-2365**.

## Test plan

```
PLAYWRIGHT_SCENARIO_TOKEN=my-token mix phx.server
cd assets/automation && npx playwright test tests/torus/capi/capi.spec.ts   # 10 pass, ~5 min
cd assets && npx jest test/adaptivity/capi_variable_test.ts                 # 17 pass
mix test test/oli/scenarios/edit_adaptive_page_test.exs                      # 4 pass
```

### DIFF
--- a/assets/automation/tests/torus/capi/capi.md
+++ b/assets/automation/tests/torus/capi/capi.md
@@ -18,30 +18,31 @@ The host (`ExternalActivity`) is the ground truth for the wire format — there 
 
 ## Test layers
 
-| Layer | File | Run |
-|---|---|---|
-| Protocol E2E (Playwright) | `capi.spec.ts` | `npx playwright test tests/torus/capi/capi.spec.ts` |
-| Value coercion (Jest) | `assets/test/adaptivity/capi_variable_test.ts` | `cd assets && npx jest test/adaptivity/capi_variable_test.ts` |
-| Seeding directive (ExUnit) | `test/oli/scenarios/edit_adaptive_page_test.exs` | `mix test test/oli/scenarios/edit_adaptive_page_test.exs` |
+| Layer                      | File                                             | Run                                                           |
+| -------------------------- | ------------------------------------------------ | ------------------------------------------------------------- |
+| Protocol E2E (Playwright)  | `capi.spec.ts`                                   | `npx playwright test tests/torus/capi/capi.spec.ts`           |
+| Value coercion (Jest)      | `assets/test/adaptivity/capi_variable_test.ts`   | `cd assets && npx jest test/adaptivity/capi_variable_test.ts` |
+| Seeding directive (ExUnit) | `test/oli/scenarios/edit_adaptive_page_test.exs` | `mix test test/oli/scenarios/edit_adaptive_page_test.exs`     |
 
 ## Playwright coverage
 
-| # | Test | Asserts |
-|---|---|---|
-| 1 | handshake | HANDSHAKE_REQUEST → HANDSHAKE_RESPONSE, requestToken echoed, config has context/sectionSlug/userId/questionId |
-| 2 | init sequence | ON_READY → page→sim VALUE_CHANGE (carries seeded var) → INITIAL_SETUP_COMPLETE |
-| 5 | state restore | sim sets a var, revisit lesson, value pushed back on re-init |
-| 6 | SET_DATA | SET_DATA_REQUEST → SET_DATA_RESPONSE success, value echoed |
-| 7 | GET_DATA | stored value returns `exists:true`; unknown key returns `exists:false`, value `'[]'` |
-| 8 | RESIZE | RESIZE_PARENT_CONTAINER_REQUEST (absolute + relative) → RESPONSE echoes messageId |
-| 9 | robustness | malformed / non-JSON / pre-handshake traffic does not break the listener; handshake still works after |
-| 10 | source check | a CAPI message posted from the top window (not the iframe) is ignored |
-| 11 | token gap (characterization) | a same-iframe post-handshake message with a *mismatched* `requestToken` is still processed — documents the current boundary (source enforced, token not), see finding #2 |
-| 3 | check lifecycle | CHECK_REQUEST → deck check → CHECK_START_RESPONSE + CHECK_COMPLETE_RESPONSE (seed carries a non-navigating trapstate rule) |
+| #   | Test                         | Asserts                                                                                                                                                                  |
+| --- | ---------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| 1   | handshake                    | HANDSHAKE_REQUEST → HANDSHAKE_RESPONSE, requestToken echoed, config has context/sectionSlug/userId/questionId                                                            |
+| 2   | init sequence                | ON_READY → page→sim VALUE_CHANGE (carries seeded var) → INITIAL_SETUP_COMPLETE                                                                                           |
+| 5   | state restore                | sim sets a var, revisit lesson, value pushed back on re-init                                                                                                             |
+| 6   | SET_DATA                     | SET_DATA_REQUEST → SET_DATA_RESPONSE success, value echoed                                                                                                               |
+| 7   | GET_DATA                     | stored value returns `exists:true`; unknown key returns `exists:false`, value `'[]'`                                                                                     |
+| 8   | RESIZE                       | RESIZE_PARENT_CONTAINER_REQUEST (absolute + relative) → RESPONSE echoes messageId; iframe bounding box follows (800×600 → 500×400 → 550×400)                             |
+| 9   | robustness                   | malformed / non-JSON / pre-handshake traffic does not break the listener; handshake still works after                                                                    |
+| 10  | source check                 | a CAPI message posted from the top window (not the iframe) is ignored                                                                                                    |
+| 11  | token gap (characterization) | a same-iframe post-handshake message with a _mismatched_ `requestToken` is still processed — documents the current boundary (source enforced, token not), see finding #2 |
+| 3   | check lifecycle              | CHECK_REQUEST → deck check → CHECK_START_RESPONSE + CHECK_COMPLETE_RESPONSE (seed carries a non-navigating trapstate rule)                                               |
 
 Out of this Playwright suite's scope (deliberately): outbound `CONFIG_CHANGE` and review-mode
 behavior (VALUE_CHANGE suppressed / `readonly` pushed) — these need a context-change/review flow
-beyond the protocol round-trip this suite covers, and aren't asserted here.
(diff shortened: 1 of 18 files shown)


assets/automation/tests/torus/capi/capi.md
```diff
@@ -0,0 +1,98 @@
...
+
+No external services (no Canvas, no real sim host). Runtime ≈ 5 min for the 10 active tests
+(single worker, per `playwright.config.ts`).
+
+If you hit `Playwright Test did not expect test.describe()` / "two versions of @playwright/test"
+on a full-file run, it's a stale transform cache — `rm -rf node_modules/.cache` and re-run.
+
+## CI placement
+
+Only a nightly Playwright workflow exists today (`--grep @nightly`). These tests, like the other
```
gastonabella, line 75: Could we wire these CAPI Playwright tests into an automated run, or at least link the tracking follow-up here? The main MER-5701 coverage lives in `capi.spec.ts`, but the current nightly workflow only runs `--grep @nightly`, and this spec is untagged. Without CI/nightly wiring, the highest-value regression coverage may remain local/manual rather than automated.

assets/automation/tests/torus/capi/capi.md
```diff
@@ -0,0 +1,98 @@
...
+| 5 | state restore | sim sets a var, revisit lesson, value pushed back on re-init |
+| 6 | SET_DATA | SET_DATA_REQUEST → SET_DATA_RESPONSE success, value echoed |
+| 7 | GET_DATA | stored value returns `exists:true`; unknown key returns `exists:false`, value `'[]'` |
+| 8 | RESIZE | RESIZE_PARENT_CONTAINER_REQUEST (absolute + relative) → RESPONSE echoes messageId |
+| 9 | robustness | malformed / non-JSON / pre-handshake traffic does not break the listener; handshake still works after |
+| 10 | source check | a CAPI message posted from the top window (not the iframe) is ignored |
+| 11 | token gap (characterization) | a same-iframe post-handshake message with a *mismatched* `requestToken` is still processed — documents the current boundary (source enforced, token not), see finding #2 |
+| 3 | check lifecycle | CHECK_REQUEST → deck check → CHECK_START_RESPONSE + CHECK_COMPLETE_RESPONSE (seed carries a non-navigating trapstate rule) |
+
+Out of this Playwright suite's scope (deliberately): outbound `CONFIG_CHANGE` and review-mode
```
gastonabella, line 42: Could we add coverage for outbound `CONFIG_CHANGE`, or explicitly link a follow-up ticket for it? The research notes define the comprehensive CAPI scope as including `CONFIG_CHANGE`, but the committed suite documents it as out of scope. Given the ticket asks for comprehensive interface coverage, I think we should either cover this message type here or make the deferral visible and tracked.

assets/automation/tests/torus/capi/capi.spec.ts
```diff
@@ -0,0 +1,382 @@
...
+  }) => {
+    const { frame } = await setup(page, seedScenario);
+    await handshakeAndReady(frame);
+
+    await sendFromStub(frame, CapiType.RESIZE_PARENT_CONTAINER_REQUEST, {
+      messageId: 'resize-abs',
+      width: { type: 'absolute', value: '500' },
+      height: { type: 'absolute', value: '400' },
+    });
+    await expect
```
gastonabella, line 299: Could we assert the actual iframe/container dimensions after `RESIZE_PARENT_CONTAINER_REQUEST`? Right now this test only checks that the host returns a success response with the messageId. Since the resize request’s user-visible behavior is changing the parent container size, a DOM/bounding-box assertion would make this regression test much stronger.

gastonabella (commented):
Looks good overall, thanks for putting this coverage together. I left a few comments around CI/nightly wiring and a couple of coverage gaps that may be worth clarifying or tracking.

assets/automation/tests/torus/capi/capi.md
Francisco-Castro: Agreed — nothing runs this Playwright spec in CI today. Tagging it @nightly alone wouldn’t help: the Playwright nightly job only runs the LTI smoke test against Canvas/Tokamak and does not start Torus + Postgres or enable the /test/scenario-yaml seeding endpoint that these tests depend on. The separate nightly scenario job runs ExUnit scenarios, not Playwright specs.

The follow-up is [MER-5777](https://eliterate.atlassian.net/browse/MER-5777), which adds a CI job with Torus + Postgres + the seeding endpoint. I’ve commented there so the CAPI suite is included, and I’ll link the ticket from capi.md.

assets/automation/tests/torus/capi/capi.md
Francisco-Castro: Good catch — the deferral was documented but not tracked. Filed [TRIAGE-2412](https://eliterate.atlassian.net/browse/TRIAGE-2412) (linked to [MER-5701](https://eliterate.atlassian.net/browse/MER-5701)) and added the reference to `capi.md`.

Why not in this PR: `CONFIG_CHANGE` only fires when the deck navigates screens while the sim stays mounted on a layer. That needs a multi-screen fixture the current `edit_adaptive_page` scenario seeding can't build yet — the ticket carries the full analysis (fixture shape, seeding extension, ~2–3 days). The ticket covers only `CONFIG_CHANGE`; review-mode testing was marked optional in the original plan, so it has no ticket.

gastonabella (approved):
LGTM! Thanks for the changes

PR #6665, opened 2026-06-18
[BUG FIX] [MER-5716] Add default instructor dashboard recommendation config

simonchoxx:
[MER-5716](https://eliterate.atlassian.net/browse/MER-5716)

## Summary
Add an idempotent seed for the global `instructor_dashboard_recommendation` GenAI feature config.

### Changes
- Added a separate idempotent check in `priv/repo/seeds.exs` for the missing global `gen_ai_feature_configs` row
- The seed reuses the existing `standard-no-backup` service config
- Kept the setup aligned with the existing seed flow so new installs and deployments both get the default

### Verification
- `mix test test/oli/gen_ai/feature_config_test.exs test/oli/gen_ai_test.exs`

[MER-5716]: https://eliterate.atlassian.net/browse/MER-5716?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/priv/repo/seeds.exs
+++ b/priv/repo/seeds.exs
@@ -233,12 +233,6 @@ case Oli.Repo.all(RegisteredModel) do
       section_id: nil
     })
 
-    Oli.Repo.insert!(%FeatureConfig{
-      feature: :instructor_dashboard_recommendation,
-      service_config_id: service_config.id,
-      section_id: nil
-    })
-
     # Dedicated ServiceConfig for the instructor email feature (MER-5257).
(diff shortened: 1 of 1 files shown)


eliknebel (changes requested):
This looks good, but I think we want this to actually live in the `seeds.exs` script where other records are seeded, not as a migration.

The seeds script is assumed to be idempotent and is run on every deployment so as long as that property holds with this change, then it probably is better suited there.

eliknebel (approved):

PR #6670, opened 2026-06-19
[FEATURE] [MER-5622] Add instructor bank selection manager

nicocirio:
## Jira

- Ticket: https://eliterate.atlassian.net/browse/MER-5622

## Summary

Adds the full instructor-facing bank selection manager workflow for preview mode. This delivers the dedicated selection route, the Figma-aligned management surface, preview-shell integration, candidate paging, right-panel question preview, remove/restore flows, invalid-removal warning handling, whole-bank removal, and final QA hardening across the full five-phase plan.

## What Was Added

- Added a standalone preview-session LiveView route for managing one bank selection on one preview page.
- Reused the Instructor View shell and delivery header while adding the manager-local back behavior.
- Implemented incremental candidate loading, removed-state rendering, and active-question count updates.
- Added right-panel candidate preview rendering through the existing preview activity pipeline.
- Wired preview-driven `Remove` / `Restore` actions for `bank_candidate` targets into LiveView mutations.
- Added the invalid-removal warning modal, including the whole-bank removal path and redirect back to the preview page with flash feedback.
- Fixed modal lifecycle and pending-action reply handling so warning dismissal clears React-side updating state and the modal can be reopened reliably.
- Hardened preview lesson layout behavior to avoid the transient default live layout flash on redirect.
- Added and updated targeted LiveView/component coverage plus phase execution records for phases 4 and 5.

## Validation

- `mix format`
- `mix compile`
- `mix test test/oli_web/live/delivery/instructor/bank_selection_manager_live_test.exs`
- `mix test test/oli/delivery/instructor_customizations/write_api_test.exs`
- `mix test test/oli_web/live/delivery/instructor test/oli_web/components/delivery`
- Manual browser QA against the Figma manager and modal designs, including modal reopen, dismiss paths, focus/keyboard behavior, remove/restore flows, and whole-bank removal redirect feedback

## Video

https://github.com/user-attachments/assets/f9c7dedd-0c0a-40f0-82a5-f5c895417a63

### DIFF
--- a/lib/oli_web/live/delivery/instructor/bank_selection_manager_live.ex
+++ b/lib/oli_web/live/delivery/instructor/bank_selection_manager_live.ex
@@ -148,26 +148,16 @@ defmodule OliWeb.Delivery.Instructor.BankSelectionManagerLive do
   end
 
   def handle_event("toggle_all_candidate_checkboxes", _params, socket) do
+    visible_candidate_ids = visible_candidate_ids(socket.assigns.candidates)
+
     checked_candidate_ids =
       if all_visible_candidates_checked?(
            socket.assigns.candidates,
            socket.assigns.checked_candidate_ids
          ) do
-        Enum.reduce(
-          visible_candidate_ids(socket.assigns.candidates),
-          socket.assigns.checked_candidate_ids,
-          fn id, acc ->
-            MapSet.delete(acc, id)
-          end
-        )
+        MapSet.new()
       else
-        Enum.reduce(
-          visible_candidate_ids(socket.assigns.candidates),
-          socket.assigns.checked_candidate_ids,
-          fn id, acc ->
-            MapSet.put(acc, id)
-          end
-        )
+        MapSet.new(visible_candidate_ids)
       end
 
     {:noreply, assign(socket, :checked_candidate_ids, checked_candidate_ids)}
--- a/lib/oli_web/live/delivery/instructor/bank_selection_manager_live.ex
+++ b/lib/oli_web/live/delivery/instructor/bank_selection_manager_live.ex
@@ -1,0 +1,140 @@
+defmodule OliWeb.Delivery.Instructor.BankSelectionManagerLive do
+  use OliWeb, :live_view
+
+  alias Oli.Delivery.InstructorCustomizations
+  alias OliWeb.Components.Delivery.Layouts
+  alias OliWeb.Delivery.Instructor.{PreviewReturn, PreviewRoutes}
+
+  def mount(
+        %{"revision_slug" => revision_slug, "selection_id" => selection_id} = params,
+        _session,
+        socket
+      ) do
+    section = socket.assigns.section
+    navigation_params = navigation_params(params, section.slug)
+
+    case InstructorCustomizations.resolve_bank_selection_preview_target(
+           section,
+           revision_slug,
+           selection_id
+         ) do
+      {:error, {:not_found, :page}} ->
+        {:ok, redirect(socket, to: PreviewRoutes.learn_path(section.slug, navigation_params))}
+
+      {:error, {:invalid_page_type, :adaptive}} ->
+        {:ok,
+         redirect(
+           socket,
+           to:
+             PreviewRoutes.page_path(
+               section.slug,
+               revision_slug,
+               adaptive_redirect_params(params)
+             )
+         )}
+
+      {:error, {:not_found, :selection}} ->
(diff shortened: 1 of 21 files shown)


nicocirio:
Follow-up on https://github.com/Simon-Initiative/oli-torus/pull/6670#issuecomment-4754108128\n\nThe point about the checkbox selection controls is valid, but the bulk-action side effect is intentionally sequenced into MER-5623 rather than MER-5622. We kept the current preparatory selection state in place and added an inline code comment to make that follow-up ownership explicit, instead of re-scoping this PR into the next ticket in the epic.

nicocirio:
Follow-up on https://github.com/Simon-Initiative/oli-torus/pull/6670#issuecomment-4754109709

We are not taking the suggested visualState/statusPill change for the right-hand preview in this PR. In the current MER-5622 design, the removed state is communicated in the left candidate list, and we do not want to introduce an additional Removed pill or treatment in the preview panel that diverges from the approved Figma for this screen.

lib/oli_web/live/delivery/instructor/bank_selection_manager_live.ex
```diff
@@ -0,0 +1,1199 @@
...
+    do: collect_selection_resource_ids(conditions)
+
+  defp collect_selection_resource_ids(_), do: []
+
+  defp activity_type_titles do
+    Activities.list_activity_registrations()
+    |> Map.new(fn registration -> {registration.id, registration.title} end)
+  end
+
+  defp criteria_label("tags"), do: "Tags:"
```
gastonabella, line 1135: Small consistency note: in MER-5620 we had to make a similar call for how to display authored selection criteria in the Activity Bank Selection preview. I briefly checked the direction with Jess, and the approach we landed on there was a simpler instructor-facing presentation.

For affirmative operators like `contains` / `equals`, we avoid showing the operator and just display the criterion label and values, e.g. `Tags`, `Learning Objectives`, `Activity Types`. For negative operators like `does_not_contain` / `does_not_equal`, we show that in the label instead, e.g. `Excluded Tags`, `Excluded Learning Objectives`, `Excluded Activity Types`.

That said, the design did not deeply specify these more complex criteria cases, so it may be worth revisiting the final wording. My approach may be a little too simplified and could be hiding useful authoring logic from instructors.

Since the collapsed Activity Bank Selection preview and this expanded Manage Questions view are closely related, I think we should try to align the criteria presentation between both. Even if the UI rendering is different — React in MER-5620 and LiveView here — the criteria-label/value normalization could probably live in shared Elixir logic so the two screens do not drift.

lib/oli_web/live/delivery/instructor/bank_selection_manager_live.ex
```diff
@@ -0,0 +1,1199 @@
...
+         &toggle_checked_candidate_id(&1, candidate_id)
+       )}
+    else
+      {:noreply, socket}
+    end
+  end
+
+  def handle_event("toggle_all_candidate_checkboxes", _params, socket) do
+    checked_candidate_ids =
+      if all_visible_candidates_checked?(
```
gastonabella, line 154: Since checked_candidate_ids is normalized to visible candidates, could this handler return `MapSet.new()` when all visible candidates are checked, and `MapSet.new(visible_candidate_ids)` otherwise? The current reduce preserves non-visible ids, but those ids appear to be intentionally discarded by `normalize_checked_candidate_ids/2`, so the simpler version may better reflect the current selection model.

gastonabella (commented):
I tested it locally, and it looks great! Left a couple of comments, and especially one that we need to agree on.

eliknebel (approved):

lib/oli_web/live/delivery/instructor/bank_selection_manager_live.ex
nicocirio: Thanks for your feedback! I completely agree with standardizing how we present the selection criteria, applying what you and Jess already agreed upon.

I was thinking that perhaps the most efficient way to do this is by merging this pull request as it is now, and then doing pair programming this afternoon to unify those criteria in the pull request you have in draft. Let me know what you think, thanks!

lib/oli_web/live/delivery/instructor/bank_selection_manager_live.ex
nicocirio: 🚀

gastonabella (approved):

PR #6672, opened 2026-06-22
[FEATURE] [MER-5620] Activity Bank Selection Remove and Restore

gastonabella:
## Summary

This PR implements [MER-5620](https://eliterate.atlassian.net/browse/MER-5620), adding instructor customization support for Activity Bank Selections in Instructor Preview.

It introduces a new Activity Bank Selection preview UI in Instructor Preview, replacing the previous inline selection display in that surface. The new preview shows key selection metadata, authored selection criteria, and a sampled question rendered through the existing activity preview flow when available, with fallback support for activity types that do not yet have a dedicated preview component.

The PR also adds remove/restore support for entire Activity Bank Selections through the existing LiveView/React customization bridge, while keeping embedded activity remove/restore aligned with the same contract. When students already have scored attempts or practice page visits, the preview now shows the corresponding warning banner and confirmation modal to make clear that changes only affect future attempts.

On top of the core customization flow, this work links the preview card’s **Manage questions** action to the expanded Activity Bank management view, improves sampled-question selection, and ensures sampled activities preserve the expected preview header and learning objective details.

Finally, the selection-criteria display was consolidated and refined so both Instructor Preview and the expanded management view use the same server-side presentation logic. That includes clearer handling of more complex authored criteria, helper text for the `all` / `any` distinction, and several UI polish updates to better match the intended design.

As part of that cleanup, the shared selection-criteria header/content is now rendered through a single Phoenix function component for the LiveView-based surface, and the same server-rendered HTML is passed into the React preview card for consistent display there as well.

This also works in the existing Product/Template Customize Content preview flow, and template-level activity exclusions continue to propagate only to future course sections created from the customized template.

https://github.com/user-attachments/assets/ba731dfd-ca1d-46cc-a305-bc08cf70bffe

[MER-5620]: https://eliterate.atlassian.net/browse/MER-5620?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -40,6 +40,13 @@ jobs:
         run: |
           cp oli.example.env oli.env
 
+      - name: 🔧 Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: "20"
+          cache: yarn
+          cache-dependency-path: assets/yarn.lock
+
       - name: 💾 Restore the deps cache
         id: mix-deps-cache
         uses: actions/cache@v3
@@ -127,9 +134,11 @@ jobs:
           token: ${{ secrets.SIMON_BOT_PERSONAL_ACCESS_TOKEN }}
 
       - name: 🔧 Configure
-        uses: actions/setup-node@v1
+        uses: actions/setup-node@v4
         with:
-          node-version: "16.14.2"
+          node-version: "20"
+          cache: yarn
+          cache-dependency-path: assets/yarn.lock
 
       - name: 🧪 Setup BEAM toolchain
         uses: erlef/setup-beam@v1
@@ -175,9 +184,11 @@ jobs:
           ref: ${{ github.event.pull_request.head.ref }}
 
       - name: 🔧 Configure
-        uses: actions/setup-node@v1
+        uses: actions/setup-node@v4
         with:
-          node-version: "16.14.2"
+          node-version: "20"
+          cache: yarn
+          cache-dependency-path: assets/yarn.lock
 
       - name: 🧪 Setup BEAM toolchain
         uses: erlef/setup-beam@v1
--- a/lib/oli/delivery/attempts/core.ex
+++ b/lib/oli/delivery/attempts/core.ex
@@ -66,6 +66,30 @@ defmodule Oli.Delivery.Attempts.Core do
     Repo.one(query) > 0
   end
 
+  @doc """
+  For a given section and page resource id, determine whether any learner resource attempts are
+  present without exposing learner-specific details.
+  """
+  def has_any_resource_attempts?(%Section{id: section_id}, resource_id) do
+    from(access in ResourceAccess,
+      join: attempt in ResourceAttempt,
+      on: access.id == attempt.resource_access_id,
+      where: access.section_id == ^section_id and access.resource_id == ^resource_id
+    )
+    |> Repo.exists?()
+  end
+
+  @doc """
+  For a given section and page resource id, determine whether any learner page access exists
+  without exposing learner-specific details.
+  """
+  def has_any_resource_accesses?(%Section{id: section_id}, resource_id) do
+    from(access in ResourceAccess,
+      where: access.section_id == ^section_id and access.resource_id == ^resource_id
+    )
+    |> Repo.exists?()
+  end
+
   def browse_lms_grade_updates(
         %Paging{limit: limit, offset: offset},
         %Sorting{field: field, direction: direction},
--- a/lib/oli/rendering/content/activity_bank_selection_criteria.ex
+++ b/lib/oli/rendering/content/activity_bank_selection_criteria.ex
@@ -2,21 +2,21 @@ defmodule Oli.Rendering.Content.ActivityBankSelectionCriteria do
   @moduledoc """
   Builds the instructor-facing presentation for Activity Bank selection criteria.
 
-  This module intentionally translates authored selection logic into friendlier labels for
-  instructor preview surfaces. The goal is to keep the criteria understandable without exposing
-  raw query-style wording.
+  This module translates authored selection logic into instructor-facing labels for preview
+  surfaces.
 
   Current UX rules:
 
   - show top-level clause context as "Activities must match all of the following." /
     "Activities may match any of the following."
   - for tags and learning objectives:
-    - `contains` -> "Must include ..."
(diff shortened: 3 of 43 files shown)


gastonabella:
> ### AI Review — performance
> ### Bank selection preview builds run serial per-selection queries
> file: lib/oli_web/delivery/instructor/activity_bank_selection_preview.ex line: 29 Description: `build_preview_map/3` calls `build_preview/4` once per selection, and each preview path resolves customization state, counts candidates, samples a candidate, resolves the sampled activity, and may resolve objective/criteria labels. On pages with several bank selections this creates a serial N+1 load path during instructor preview rendering, directly increasing TTFB. Suggestion: Add a bulk preview builder that resolves summaries/samples/revisions/labels for all selections in one pass, or prefetch the needed selection summaries and candidate revisions before the `Enum.map/2`.
>

Good callout. I agree the current builder still scales per selection and a fully bulked preview builder would be better for pages with many bank selections.

I’d prefer to defer that as a follow-up because doing it well means adding batch APIs for selection summaries/samples, sampled revisions, and criteria labels rather than only reshuffling this function. This PR already reduced one duplicated path with `get_bank_selection_summary/3`, but I agree there is room for a broader bulk optimization.

lib/oli/delivery/attempts/core.ex
```diff
@@ -66,6 +66,30 @@ defmodule Oli.Delivery.Attempts.Core do
...
+  @doc """
+  For a given section and page resource id, determine whether any learner page access exists
+  without exposing learner-specific details.
+  """
+  def has_any_resource_accesses?(%Section{id: section_id}, resource_id) do
+    from(access in ResourceAccess,
+      where: access.section_id == ^section_id and access.resource_id == ^resource_id
+    )
+    |> Repo.exists?()
+  end
```
nicocirio, line 91: can we add tests for these 2 new functions?

lib/oli/rendering/content/activity_bank_selection_criteria.ex
```diff
@@ -0,0 +1,278 @@
...
+      |> Enum.flat_map(fn {_selection, parsed_selection} ->
+        criteria_resource_ids(parsed_selection)
+      end)
+      |> Enum.uniq()
+
+    if resource_ids == [] do
+      %{}
+    else
+      section_slug
+      |> DeliveryResolver.from_resource_id(resource_ids)
```
nicocirio, line 78: One thought here: this title resolution might be able to use a hybrid approach instead of going entirely through DeliveryResolver. It looks like objective titles could probably come from SectionResourceDepot, since objectives are already cached there with the title we need, while tags still seem to require `DeliveryResolver` because they are not part of the depot today. I’m not convinced this is worth the added split logic, but I'm leaving the idea in case it's a good fit for you.

lib/oli_web/components/delivery/activity_bank_selection_criteria.ex
```diff
@@ -0,0 +1,100 @@
...
+    """
+  end
+
+  @doc """
+  Renders the selection criteria block to an HTML string.
+
+  This is used by instructor preview pages that embed the criteria markup into a client-owned
+  preview payload while keeping the criteria presentation logic server-side.
+  """
+  def selection_criteria_html(rows, opts \\ []) do
```
nicocirio, line 77: nice! ❤️

lib/oli_web/live/delivery/instructor/bank_selection_manager_live.ex
```diff
@@ -560,16 +569,11 @@ defmodule OliWeb.Delivery.Instructor.BankSelectionManagerLive do
...
-                    </div>
-                    <div class="bg-Specially-Tokens-Fill-fill-input-focused rounded-md px-2.5 py-2 text-Text-text-high text-base font-semibold leading-6">
-                      {row.value}
-                    </div>
-                  </div>
+                <div class="mt-5">
+                  <ActivityBankSelectionCriteriaComponent.selection_criteria
+                    rows={@selection_criteria_rows}
+                    helper_text={@selection_criteria_helper_text}
+                  />
```
nicocirio, line 576: Thanks for creating this component!

lib/oli_web/live/delivery/instructor/preview_lesson_live.ex
```diff
@@ -292,6 +309,113 @@ defmodule OliWeb.Delivery.Instructor.PreviewLessonLive do
...
+        </button>
+      </div>
+    </div>
+    """
+  end
+
+  attr :pending, :map, default: nil
+  attr :warning, :map, default: nil
+
+  defp preview_attempt_warning_modal(assigns) do
```
nicocirio, line 346: Instead of creating a modal from scratch (with esc to close, click away, focus wrap, etc), we could have reused the existing one at `OliWeb.Components.Modal`

.github/workflows/build.yml
nicocirio, line 1: Do the .github workflow changes need to be part of this PR? They look unrelated to the MER-5620 behavior change

nicocirio (commented):
Awesome work! Great abstractions and very well organized code.
I left some comments, please let me know what you think

.github/workflows/build.yml
gastonabella: These workflow changes were needed to fix CI failures caused by outdated Node versions in our GitHub Actions jobs. Some jobs were still pinned to Node 16.14.2, but the current frontend dependency tree now requires Node >= 18 during `yarn --cwd assets --frozen-lockfile`. Updating those workflows to Node 20 (and modern setup-node) was necessary to get asset install/typecheck/build steps passing again.

lib/oli/rendering/content/activity_bank_selection_criteria.ex
gastonabella: Good idea. I considered keeping objective title resolution closer to the depot boundary as well, but I’d prefer to leave this path unified for now unless we see a clearer benefit from splitting objectives and tags into separate resolution flows. Since this presenter was already going through a few cleanup iterations, I kept the simpler single-path version here.

gastonabella:
Thanks @nicocirio ! I pushed the updates for those comments just now.

nicocirio (approved):
🚀

PR #6673, opened 2026-06-22
[BUG FIX] [MER-5696] pass paywall check as if instructor in template preview

andersweinstein:
## Summary

  Fixes [MER-5696](https://eliterate.atlassian.net/browse/MER-5696), where paid template preview could redirect authors to the payment page when the same browser session also had a delivery `current_user` login.

  Template preview usually avoids the paywall when it launches through the hidden instructor user, which is the same mechanism commonly used to give admins instructor-style delivery access. However, when a non-hidden delivery `current_user` already exists in the session, template preview uses that user instead and enrolls them into the blueprint section for preview. Because that enrollment is a learner enrollment, normal paywall enforcement can treat the preview like unpaid student access and redirect to /payment.

  ## Changes

  - Added `OliWeb.TemplatePreviewMode` to centralize active template preview detection.
  - Updated `Oli.Plugs.EnforcePaywal`l to bypass paywall enforcement only for active template preview sessions where:
      - `template_preview_mode` is true
      - the preview section slug matches the current section
      - the current section is a blueprint

  - Updated `OliWeb.LiveSessionPlugs.SetPaywallSummary` to assign an instructor-style paywall summary during active template preview.
      - This keeps LiveView delivery state consistent with the plug.
      - It also suppresses payment/grace-period chrome during preview.

  - Left the existing preview enrollment/login behavior unchanged.
  
Note: also added a small drive-by CI stabilization for the unrelated dataset initiator values test by making its assertion order-independent, removing its :flaky tag since it is now deterministic.

  ## Verification

  - mix test test/oli_web/plugs/plug_order_integration_test.exs test/oli_web/live_session_plugs/set_paywall_summary_test.exs
  - mix test test/oli_web/controllers/products_controller_test.exs

[MER-5696]: https://eliterate.atlassian.net/browse/MER-5696?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli/analytics/datasets.ex
+++ b/lib/oli/analytics/datasets.ex
@@ -345,7 +345,6 @@ defmodule Oli.Analytics.Datasets do
       DatasetJob
       |> join(:left, [j], proj in Oli.Authoring.Course.Project, on: j.project_id == proj.id)
       |> distinct(true)
-      |> order_by([_j, proj], asc: proj.title, asc: proj.id)
       |> select([_j, proj], %{id: proj.id, title: proj.title})
 
     Repo.all(query)
@@ -366,7 +365,6 @@ defmodule Oli.Analytics.Datasets do
       |> join(:left, [j], u in Oli.Accounts.Author, on: j.initiated_by_id == u.id)
       |> distinct(true)
       |> where(^filter_by_project_id)
-      |> order_by([_j, u], asc: u.email, asc: u.id)
       |> select([_j, u], %{id: u.id, email: u.email})
 
     Repo.all(query)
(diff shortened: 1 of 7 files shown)


andersweinstein:
[Codex:] Re: AI security review comment “Session flag bypasses paywall without authorization recheck”

This PR intentionally keeps template preview authorization at the preview-launch boundary, which is how the existing hidden-instructor preview path already works. The bypass only applies when server-side template preview session state is active, the stored preview slug matches the current section, and the section is a blueprint, so it does not bypass payment for normal enrollable course sections or provide a path for student users to self-authorize access.

Rechecking author authorization in the delivery paywall plug would be a broader defense-in-depth change to the preview session model, not necessary for MER-5696.

nicocirio (approved):

PR #6675, opened 2026-06-23
[BUG FIX] [MER-4630] Preserve product brand when creating sections from templates

simonchoxx:
[MER-4630](https://eliterate.atlassian.net/browse/MER-4630)

## Summary

This PR fixes the section creation flow from templates so the template’s assigned brand is preserved on the new section.

### Context

When a template has a brand selected in section details, creating a new section from that template was dropping the brand on the duplicated section.

The root cause was in the internal duplication path used by “create section from template”: `Blueprint.duplicate/2` was explicitly clearing `brand_id`.

### What changed

- Updated `Blueprint.duplicate/2` to carry forward the source template’s `brand_id` unless a caller explicitly overrides it.
- Added a regression test that verifies a section created from a branded template keeps that brand.

### Verification

- `mix format test/oli/delivery/sections/blueprint_test.exs`
- `mix test test/oli/delivery/sections/blueprint_test.exs:260`

### Notes

The fix is intentionally scoped to template-to-section duplication and does not change unrelated section creation flows.

[MER-4630]: https://eliterate.atlassian.net/browse/MER-4630?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nicocirio (approved):
🚀

PR #6684, opened 2026-06-25
[FEATURE] [MER-5623] Add bulk actions to the bank selection manager

nicocirio:
## Jira

- Ticket: https://eliterate.atlassian.net/browse/MER-5623

## Summary

- add same-state multi-select behavior to the bank selection manager
- add contextual bulk remove and restore actions for the currently shown query rows
- wire the manager to an atomic backend bulk mutation flow with plural invalid-removal handling
- keep the preview action state, refreshed candidate rows, and checked-row normalization aligned after mutations

## Improvement

- preserves a clean state model where the visible candidate query stays separate from ephemeral checked-row UI state
- batches candidate mutation work in InstructorCustomizations instead of looping through single-row writes
- precomputes bulk-selection render state and caches visible preview dependencies and payloads to reduce repeated work during selection changes

## Validation

- mix test test/oli_web/live/delivery/instructor/bank_selection_manager_live_test.exs test/oli/delivery/instructor_customizations/write_api_test.exs test/oli/delivery/instructor_customizations/target_resolver_test.exs
- mix format
- mix compile
- yarn --cwd assets run format
- yarn --cwd assets run check-types
- yarn --cwd assets run deploy

## Demo

https://github.com/user-attachments/assets/f8bf2c70-bf3a-4e2f-8842-67bd31fb1692


nicocirio:
Follow-up on the UI review comment (preview remount / bulk CTA hover): I’m not taking that change in this pass. The current preview-shell id churn is intentional in this branch because it is part of how the ignored preview subtree stays aligned with bulk-selection state transitions. The hover-style difference on the bulk action button is cosmetic, so I’m leaving it out of this follow-up rather than changing the preview sync path opportunistically.

nicocirio:
Follow-up on the TypeScript review comment about stale disabled fallback buttons: I’m not taking that change because I don’t see a matching authoritative server state for it in this PR. Successful preview-customization replies currently carry the next action, while unsuccessful replies are intentionally non-authoritative and should not clear the existing control. If we later introduce a success reply that intentionally omits actions, that would be the right time to add fuller reconciliation in the hook.

nicocirio:
Follow-up on the performance review comment about eager preview payloads: agreed this is a real tradeoff, but I’m deferring it from this review-follow-up. In the current implementation the cache is bounded to the visible candidate batch, and that eager encoding was chosen to avoid repeating the same work during rapid preview switches. If profiling shows the assign-size/TTFB cost is material, I’d rather address it in a focused pass that reworks the preview-data lifecycle deliberately.

gastonabella (approved):
Looks great! There are just a few conflicts to resolve before merging.

PR #6656, opened 2026-06-10
[FEATURE] [MER-5685] Add Flashcards component to Advanced Authoring

kjsegovi:
This PR adds the Authoring Flashcard Component to the core components list in Advanced Authoring only. 
All the required entries from the [JIRA ticket](https://eliterate.atlassian.net/browse/MER-5685) have been met. 

I took heavy inspiration from two components. 
- janus-Carousel
- janus-fill-blanks

Authors can add a deck of flip cards to an adaptive screen, configure front/back content in a fullscreen editor, and tune layout and behavior from the property panel. Learners can flip cards in delivery with keyboard and screen reader support, and adaptivity rules can react to flip states for all cards flipped or the number of cards that have been flipped. I've also added an adaptivitySchema for the index of the card flipped to be followed.

**New Flashcards Component Implementation:**

- Added a fully functional delivery component (`Flashcard.tsx`) that manages flashcard flipping state, handles notifications, and synchronizes adaptivity variables for use in runtime environments.
- Implemented a highly interactive authoring component (`FlashcardAuthor.tsx`) that allows content creators to add, edit, and delete flashcards, including a Quill-based rich text editor for card faces. The authoring component uses React portals for a fullscreen editing experience and supports configuration events.
- Created a shared presentational component (`FlashcardsView.tsx`) used by both authoring and delivery, featuring responsive layout, keyboard accessibility, ARIA support, and support for randomization and custom CSS.

**Styling and Accessibility:**

- Added a dedicated CSS file (`Flashcard.css`) with responsive design, container queries, keyboard focus indicators, and support for both text and image content, ensuring the component is visually appealing and accessible.

**Integration and Registration:**

- Registered the authoring component as a custom element with the authoring platform, including schema and capability definitions, and mapped custom events for configuration.

**Authoring**

- Fullscreen configure UI to add/remove cards and edit front/back content
- Rich content via Quill/TextFlow (text + images)
- Property panel options: enabled, custom CSS/class, randomize, min/max cards per row, flip duration 

**Delivery**

- Responsive card grid (ResizeObserver + min/max cards-per-row bounds)
- Flip animation with configurable duration
- Click, Enter, and Space to flip
- Screen reader labels (aria-pressed, aria-describedby, visually hidden instructions)
- Optional card randomization on delivery

**Adaptivity (CAPI)**

- Exposes: flippedCards, hasCardBeenFlipped, allCardsFlipped
- Accepts: flipAllCards to flip the entire deck from adaptivity rules

PR #6687, opened 2026-06-25
[BUG FIX] [MER-5685] Updates Made to Janus-Flashcard Component

kjsegovi:
I made the fixes to the flashcard edit modal and updated how resizing works. It was a bit complicated and might need comprehensive testing of other widgets, as I changed LayoutEditor.tsx. I tested other components, and they behave as expected, but it's always good to have a second pair of eyes on some components. I also added a new icon to the Flashcards, as it was reusing the Fill in the Blank Icons."

**All Changes Made**
- Split layout height from card face height so that the card face height doesn't override the layout height when editing heights by dragging or in the component tab. Introduces cardHeight and schema helpers (resolveCardHeight, computeFlashcardsLayoutHeight, resolveCardHeightForLayout, withFlashcardsLayoutDimensions) so the ResizeContainer box (height) and per-card face height are no longer the same field. Updates defaults to match visual floors (480px width, 180px card height, 120px empty deck).
- Authoring auto-resize: FlashcardAuthor syncs layout dimensions on card count/width/row changes via onResize, preserves manual height resizes, and derives card face height when the box is dragged taller/shorter.
- Flashcard calls onResize after ready, so the delivery host gets explicit width/height when responsive layout is off (this is copying the same pattern as TextFlow).
- LayoutEditor onResize wiring: Generic handlePartAutoResize merges part onResize settings into part.custom, enabling live selection-box updates for dynamic-height parts
- Configure modal: Made the modal scrollable instead of having a size that responds to the height of the items in the modal. 
- Adds janus-flashcards to fit-content rules when responsive layout was enabled
- Image scaling in squished cards. Added CSS + stripFlashcardImageDimensions() so text+image cards scale proportionally instead of clipping.
- Made a Flashcard SVG instead of reusing the Fill in the Blank SVG.


kjsegovi:
The two GitHub review items were small. handlePartAutoResize was declared after renderPart, which shouldn’t have been a practical runtime issue since renderPart runs after handlers are initialized on each render, not when it’s defined, but I understand why the reviewer wanted it reordered for correctness.
onResize was added in LayoutEditor for flashcards auto-resize, but it applies to any part in the layout. Technically, a malicious part could try to override another part’s custom props, though that felt like a low concern in authoring. I updated it anyway by checking that the resize payload matches a part in the current parts list and only accepting width, height, and cardHeight.
The auto-resize behavior on skinnier/mobile-width platforms didn’t seem concerning during testing, so I left that alone for now, but I’m open to revisiting if we want better responsive behavior.

Francisco-Castro (changes requested):
Hi @kjsegovi ! Great work on this — the height/face split design is right, and I could verify a lot of it working. I did a deep pass with manual testing on a local adaptive page (plus some scripted browser testing), and I found one blocker that unfortunately keeps Mia's height issue alive, plus a few smaller items.

**Blocker**: the authoring auto-resize never runs (props.editMode is always undefined)

Both auto-resize effects in FlashcardAuthor.tsx guard on props.editMode:

```javascript
useEffect(() => {
  if (!props.editMode) {   // <-- always undefined inside the web component
    return;
  }
```

Across the custom-element boundary the prop key is lowercase `editmode`: React writes the `editMode` prop as the HTML attribute `editmode`, and `toCamelCase` in `customElementWrapper.ts` only converts dash-separated names `(/-(\w)/g)`, so it stays `editmode`. `props.editMode` is therefore always undefined and both effects return early — the layout-dimension sync is effectively dead code. (Note the same file already reads lowercase configuremode, which is why configure works.)

Repro (verified on this branch):
1. Add a Flashcards component (Height correctly starts at 120 — the new defaults work 👍).
2. Open the configure modal, add cards until you have 4, Save.
3. Canvas renders a 2×2 grid (~400px tall), but the COMPONENT tab Height stays frozen at the old value, and the selection box only wraps the first row — exactly Mia's original report. The stale height also persists to the revision (DB-verified), and reloading doesn't self-heal.

Delivery/preview still renders all cards correctly (fit-content rule + runtime resolveCardHeightForLayout re-derivation), so this is author-facing — but it's the core of the M2 fix.

Suggested fix: read the lowercase key like configuremode, e.g. const editMode = parseBoolean((props as any).editmode ?? props.editMode); in both effects (or teach the wrapper to map editmode → editMode).

**Should fix:** cardHeight is silently dropped on the layout-persistence path

LayoutEditor now includes cardHeight in the layout-only keys, routing it through onPartLayoutChange → the (deprecated) dragPart event → EditingCanvas.handlePositionChanged, whose whitelist only copies x/y/width/height/z — so cardHeight never reaches the store on that path. Today this is masked because resolveCardHeightForLayout deterministically re-derives it and the modal Save persists it, but it's a dead-end write that makes correctness depend on the derivation math staying identical in three files. One-liner: add cardHeight to the newPosition whitelist in EditingCanvas.tsx (~line 85).

**Should fix:** Configure modal: scrollable ✔, but still initializes small

Verified the scroll behavior works well (body scrolls, modal grows per card, caps at viewport). However on first open with 1 card the body is ~260px with ~320px of content, so the editor sits clipped behind the Save/Cancel footer — it reads as "cut off," which was Mia's original complaint ("initialize a little larger"). Suggest a min-height on the modal content (e.g. min(60vh, fit-content)) so it opens comfortably on large screens. Also note modal-dialog-scrollable now applies to every part's config modal — I spot-checked TextFlow's and it works fine.

**Should fix:** Test coverage

There are no tests for janus-flashcards. The new schema helpers (resolveCardHeight, computeFlashcardsLayoutHeight, resolveCardHeightForLayout, withFlashcardsLayoutDimensions, stripFlashcardImageDimensions) are pure functions and cheap to unit test — and a mounted-component test would have caught the editMode bug above. Would be great to add at least the helper tests in this PR since they encode the height fix.

**Verified working:** What I verified working

- New defaults (480×120) and empty-deck floor ✔
- Manual drag-resize + persistence across reloads ✔
- Modal save persisting cards + derived cardHeight ✔
- Delivery/preview: full layout, mouse flip, keyboard flip (Tab order, visible focus ring, Enter/Space) ✔
- Seeds registration fix confirmed the hard way — my dev DB didn't have janus_flashcards registered and the component was invisible until re-seeding, so good catch adding it 👍
- New dedicated icon ✔
- LayoutEditor regression sweep (your "second pair of eyes" ask): TextFlow kept exact dimensions across reloads, no spontaneous saves on editor open, other parts' config modals open/edit/persist fine. The handlePartAutoResize whitelist + clamping look solid.

**Separate-ticket candidates (pre-existing, found while testing)**

- The left Sequence panel has a transparent hitbox extending over the canvas: for parts near the left edge (including the default spawn at 0,0) the floating Edit/Copy toolbar is unclickable — clicks are intercepted by section.aa-panel.left-panel. Since flashcards is configure-heavy, this blocks the primary flow until the author drags the part right. Also, at y=0 the toolbar hides behind the icon bar (toolBarTopOffset = -38).
- Quill toolbar flashes unstyled (giant picker arrows) for a frame when the editor remounts on modal open / "Add card" — FOUC from the remount-per-card pattern, present since #6656.

Happy to re-verify quickly once the editMode fix is in — the repro above takes about a minute.

kjsegovi:
Thanks for the thorough review, @Francisco-Castro . 

I addressed all the actionable items:
- Updated authoring to handle the lowercase editmode attribute.
- Persisted cardHeight through the EditingCanvas layout-update path.
- Increased the initial Flashcards configuration modal height.
- Added coverage for layout calculations, responsive widths, custom-element behavior, and resize persistence.
Follow-up manual testing also exposed a resize feedback loop. Auto-layout now pauses during an active canvas resize and synchronizes once on release, preserving the selected width, height, and derived cardHeight.

The full frontend suite passes and I created some test coverage for the layouts. 

For this part of the comment: 
```
Separate-ticket candidates (pre-existing, found while testing)

The left Sequence panel has a transparent hitbox extending over the canvas: for parts near the left edge (including the default spawn at 0,0) the floating Edit/Copy toolbar is unclickable — clicks are intercepted by section.aa-panel.left-panel. Since flashcards is configure-heavy, this blocks the primary flow until the author drags the part right. Also, at y=0 the toolbar hides behind the icon bar (toolBarTopOffset = -38).
Quill toolbar flashes unstyled (giant picker arrows) for a frame when the editor remounts on modal open / "Add card" — FOUC from the remount-per-card pattern, present since https://github.com/Simon-Initiative/oli-torus/pull/6656.

```

I didn't work on it. Instead, I think the best option is to try and create a new ticket for the issues and then focus on the issues individually. 

Let me know if more changes need to be done or if you have any questions on the update.

Francisco-Castro:
Hi @kjsegovi  — great progress. All four items from round 1 are fixed and verified: the lowercase editmode fix works (auto-resize is alive — adding cards now updates height end-to-end), cardHeight survives the EditingCanvas save path, the config modal opens at a proper height, and drag-resize keeps the author's height while re-deriving card height. New tests pass 10/10 locally. Nice work.

Found one remaining issue while verifying in the browser, and it can corrupt data, so requesting one more round:

Selecting a flashcards part that has a stale stored height rewrites its layout.

Repro (no DB access needed):
1. Take a flashcards part whose stored height doesn't match its cards (any part authored before this fix has this — e.g. restore an old revision of a flashcards screen via Revision History).
2. Open the page. In read-only it looks perfect — your new auto-resize heals it on screen.
3. Toggle "Read only" off and just click the part to select it.
4. The healed height is discarded and the stale stored height comes back. Depending on the part's geometry one of two things happens:
  - Height field + selection box show the stale height while the cards overflow it (looks like the original MER-5685 symptom), or
  - the cards visibly shrink to fit the stale height and that shrunken layout is silently saved as a new revision — no resize, no Save clicked, and reload doesn't bring the old size back.

Cause: selection re-renders the canvas from the store, which still has the stale height (the load-time heal ran in read-only mode, where it can't persist). The anti-snap-back logic in the FlashcardAuthor resize effect then treats that stale value as a manual resize by the author and adopts it — and if the re-derived cardHeight differs, the change is persisted.

Suggested direction: let the effect tell a store re-push apart from a real user resize (e.g. if height regresses to the exact pre-heal value with the layout key unchanged, re-heal instead of adopting it), or re-run and persist the heal once when the page transitions from read-only to edit.

Everything else looks solid — expect this is a small, final round

kjsegovi:
Hey @Francisco-Castro. I think this would be an issue if the flashcard components had gone into production, as it seems like this issue only affects existing components, right? This is a whole new component that hasn't had any users, so technically we shouldn't see previously set heights disappear. 
Please let me know if I have misunderstood your comment, and I can look over the component again to see what I can do to fix the issue.

Francisco-Castro:
You are right!

Francisco-Castro (approved):
Nice work @kjsegovi 🏅

PR #6692, opened 2026-06-26
[BUG FIX] [MER-5721] Stop Preview as Student from showing instructor customization UI (rework)

andersweinstein:
## Summary

  This reworks the fix for [MER-5721](https://eliterate.atlassian.net/browse/MER-5721) by simplifying section preview behavior.

  The section Manage page action **Preview Course as Student** no longer enters a special section preview mode. It now opens the normal student delivery route for the section, equivalent in routing behavior to “Open as student.” That means student preview uses the same delivery pages, lesson routes,  prologue handling, activity rendering, and navigation paths that normal student delivery uses.

  The only remaining section `preview_mode` is instructor preview, which is the customization-oriented preview shell used by instructors.

  ## Root Cause

  The regression came from treating section preview as one generic mode even though two different experiences were sharing parts of the `/sections/:slug/preview...` route family.

  Student section preview was intended to show what a student would see. Instructor preview was intended to show the instructor customization preview shell and assessment customization UI. As preview navigation was moved toward shared preview routes, student preview links could land in instructor preview rendering paths. That exposed instructor-only UI, including the instructor preview header and assessment customization controls, in a flow that was supposed to present the student experience.

  A related issue was that some instructor preview links still targeted legacy `/preview/page/...` paths. Those could bypass the newer instructor preview LiveViews at `/preview/lesson/...`, so internal course links and activity-bank preview links did not consistently land in the instructor customization interface.

  ## Fix

  This change removes the special “student section preview” mode which had been entered via Manage/Preview Course as Student instead of trying to preserve two preview subtypes through all navigation.

  The new model is:

  - **Preview Course as Student** opens normal delivery: `/sections/:slug`.
  - **Open as Student** continues to use normal delivery routes.
  - **Instructor preview** is the only section preview mode and uses `/sections/:slug/preview...`.
  - Page links in instructor preview use `/sections/:slug/preview/lesson/:revision_slug`. [But see note]
  - Activity-bank selection preview links in instructor preview use `/sections/:slug/preview/lesson/:revision_slug/selection/:selection_id`.
  - The legacy `/preview/page/...` route is no longer generated for normal instructor page preview links. It remains only where still required by the existing adaptive preview controller path.
  
  *Note: Instructor preview now consistently uses `/sections/:slug/preview/lesson/:revision_slug` for ordinary pages so page rendering enters the instructor customization shell. But adaptive pages are an exception: they still require the legacy `/sections/:slug/preview/page/:revision_slug` controller path because adaptive delivery is rendered through the advanced page preview/controller flow rather than the LiveView instructor preview lesson shell. Since Explorations are adaptive pages, exploration links in instructor preview must continue to use `/preview/page/...`
  
  This means `/preview/page` is not fully obsolete yet. It remains necessary for adaptive page preview compatibility, while non-adaptive instructor preview pages should use `/preview/lesson`.

  As part of this simplification:

  - Removed the `section_preview_kind` discriminator from earlier fix.
  - Removed the `PreviewMode` predicate module from earlier fix that existed to distinguish student and instructor preview subtypes.
  - Updated preview route helpers so regular instructor preview page navigation targets `/preview/lesson`.
  - Removed the legacy `/preview/page/:revision_slug/selection/:selection_id` route.
  - Updated internal content links and activity-bank selection links to use the instructor preview LiveView routes.
  - Updated attempt-state redirects so preview requests stay in instructor preview routes.
  - Kept normal delivery behavior responsible for student-facing preview, including graded page prologue/lesson flow and delivery activity rendering.
  
### Instructor preview prologue cleanup

  This PR also removes the old `/sections/:slug/preview/prologue/:revision_slug` route. Instructor preview is intended to show the instructor customization interface for assessments, not the student prologue/attempt-start screen. Student-facing assessment prologues remain available through the normal delivery route used by “Preview Course as Student” and “Open as Student”.

  Preview-mode prologue URL generation and attempt-state redirect fallback logic now canonicalize to `/sections/:slug/preview/lesson/:revision_slug` instead of `/preview/prologue`. The attempt-state redirect change is defensive/consistency cleanup: current instructor preview routes do not run through `RedirectByAttemptState`, but if a preview-shaped request ever reaches that plug, it will now stay in the instructor preview lesson route rather than targeting the removed prologue route.
  
  ## Verification

  Ran focused tests covering:

  - normal delivery routing for Preview Course as Student behavior
  - instructor preview lesson routing
  - activity-bank selection preview routing
  - adaptive preview fallback routing
  - rendered internal course links
  - preview layout/header behavior
  - attempt-state redirects
  - controller fallback behavior

[MER-5721]: https://eliterate.atlassian.net/browse/MER-5721?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/oli_web/router.ex
```diff
@@ -1399,6 +1399,26 @@ defmodule OliWeb.Router do
...
+          OliWeb.LiveSessionPlugs.RequireInstructorPreview,
+          {OliWeb.LiveSessionPlugs.InitPage, :set_prologue_context},
+          OliWeb.LiveSessionPlugs.SetBrand,
+          OliWeb.LiveSessionPlugs.SetInstructorPreviewReturn,
+          OliWeb.LiveSessionPlugs.RequireEnrollment,
+          OliWeb.LiveSessionPlugs.SetRequestPath,
+          OliWeb.LiveSessionPlugs.SetPaywallSummary
+        ] do
+        live("/prologue/:revision_slug", Delivery.Student.PrologueLive, :preview)
+      end
```
nicocirio: I believe an instructor in preview mode will never (or should never) get to the prologue page. My understanding is that the prologue page is aimed at students, since it summarizes the page terms and all the attempts made by that student.
Instead, when instructors try to access a graded page in preview mode, they are directly routed to `OliWeb.Delivery.Instructor.PreviewLessonLive`, where they can customize (remove/restore) the activities of that page.
But I am not 100% sure about this definition; probably it is something we should ask Darren.
If he confirms the behaviour, then we can also remove the `OliWeb.LiveSessionPlugs.RequireInstructorPreview` live-session-plug, since it was created to handle/route this specific case.

lib/oli_web/delivery/instructor/preview_routes.ex
```diff
@@ -21,12 +21,12 @@ defmodule OliWeb.Delivery.Instructor.PreviewRoutes do
...
 
-  def page_path(section_slug, revision_slug, params \\ [])
+  def adaptive_page_path(section_slug, revision_slug, params \\ [])
 
-  def page_path(section_slug, revision_slug, params) when params in [[], %{}],
+  def adaptive_page_path(section_slug, revision_slug, params) when params in [[], %{}],
     do: ~p"/sections/#{section_slug}/preview/page/#{revision_slug}"
 
-  def page_path(section_slug, revision_slug, params),
+  def adaptive_page_path(section_slug, revision_slug, params),
```
nicocirio, line 29: These routes handle all page types, not only adaptive. So I think this new function name might be confusing, since it may suggest that it handles only adaptive pages. Is there a reason why the previous name was changed?

lib/oli/rendering/content/url_helpers.ex
```diff
@@ -0,0 +1,50 @@
+defmodule Oli.Rendering.Content.UrlHelpers do
+  @moduledoc """
+  Shared URL helpers for rendered content links.
+  """
```
nicocirio, line 4: This module kind of overlaps with `OliWeb.Delivery.Instructor.PreviewRoutes`. To keep preview-path-helpers centralized, do you think we can extend that module and remove this one?

nicocirio (commented):
Great work. I left a couple of suggestions, let me know what you think 😄

lib/oli_web/delivery/instructor/preview_routes.ex
andersweinstein: The rationale is that preview now uses /preview/lesson for basic pages so the /preview/page route is obsolete for basic pages (there is only instructor preview and regular delivery). /preview/page only remains for use on adaptive pages.

nicocirio: You're right, thanks for the explanation!

lib/oli/rendering/content/url_helpers.ex
andersweinstein: We previously had this closer to `OliWeb.Delivery.Instructor.PreviewRoutes`, but that introduced an 
`Oli -> OliWeb` dependency from the rendering layer which was flagged in AI review. This helper keeps the fallback URL construction inside `Oli.Rendering.Content` so rendered content does not depend on Phoenix route helpers. 

The web layer still owns the primary route generation via injected `internal_link_url` / `selection_preview_url`; `UrlHelpers` is just the fallback for encoding path segments and appending query params.

lib/oli_web/router.ex
andersweinstein: I agree it is never intended for instructor preview to hit the prologue page (if they want to see it they should preview as student). I think codex left this in defensively in case there was some path to this route from instructor preview or maybe just old deep links to it. (Also there may be tests dealing with this route)? If there truly is no path to it for an instructor, we don't need the whole route at all, /preview/prologue/ should not have a use.

andersweinstein: OK, I removed the /preview/prologue route completely (see note added to PR description).

andersweinstein:
[Codex:] Replying to the AI UI comment on “Internal lesson links drop preview state”: https://github.com/Simon-Initiative/oli-torus/pull/6692#issuecomment-4813140753

I checked the callers for `Student.Utils.build_html/3`. That helper is currently used by normal student delivery/review (`:delivery` and `:review` modes), not by instructor preview rendering.

Instructor preview page content is rendered through `PreviewPageContext.build/…`, which sets `mode: :instructor_preview` and injects `internal_link_url` via `PreviewRoutes.lesson_path(section_slug, revision_slug, navigation_params)`, preserving the preview route and return context. So this appears to be a false positive from inspecting the shared helper in isolation.

lib/oli/rendering/content/url_helpers.ex
nicocirio: Ok, that makes sense, we should definitely avoid `Oli -> OliWeb` dependency.
Thanks for the explanation!

nicocirio (approved):

PR #6695, opened 2026-06-29
[BUG FIX] [MER-5723] Mark adaptive practice pages complete on finalize

nicocirio:
# MER-5723 Adaptive pages not always showing 100% completion

Jira ticket: https://eliterate.atlassian.net/browse/MER-5723

## Summary

This fixes a regression where adaptive practice pages could finish successfully while still showing partial completion, such as `77%`, instead of `100%`.

The issue affected adaptive pages with branching paths, where learners can validly complete the lesson without visiting every screen.

## Root Cause

Adaptive ungraded page finalization was updating the `resource_attempt` lifecycle and rollup state, but it was no longer forcing the associated `resource_access.progress` to `1.0`.

Historically, that completion update was handled from the client-side adaptive finish flow. That path was removed in the adaptive payment/review flow refactor, and the backend finalize path did not take over responsibility for marking progress complete.

Regression introduced by:

- `MER-5049 / MER-5054` in PR `#6401`, which removed the client-side `mark_completed` adaptive finish behavior

## Changes

- Add a regression test proving that ungraded adaptive page finalization sets page progress to `100%`
- Cover both finalized adaptive outcomes:
  - evaluated attempts
  - submitted attempts with manual grading still pending
- Update the backend ungraded adaptive finalize path to mark the related `resource_access` as complete

## Why This Fix

The fix is implemented in the backend finalization path instead of restoring the client-side behavior.

That keeps the business rule in the authoritative server-side workflow and ensures adaptive practice completion does not depend on a specific UI path or frontend caller.

The completion update is also kept atomic with the existing `PageLifecycle.finalize/3` transaction, so adaptive ungraded finalization and progress completion now succeed or fail together.

## Verification

https://github.com/user-attachments/assets/c1a29b96-04c7-4106-b942-803327a78821

Executed:

```bash
mix test test/oli/delivery/attempts/page_lifecycle_test.exs
mix format lib/oli/delivery/attempts/page_lifecycle/ungraded.ex test/oli/delivery/attempts/page_lifecycle_test.exs
```

Result:

- regression reproduced before the fix with `resource_access.progress == 0.0`
- targeted test module passes after the fix

## Risk

Low.

The change is narrowly scoped to ungraded adaptive page finalization and does not alter the in-progress activity-based progress calculation during the attempt.

### DIFF
--- a/lib/oli/delivery/attempts/page_lifecycle/ungraded.ex
+++ b/lib/oli/delivery/attempts/page_lifecycle/ungraded.ex
@@ -245,9 +245,7 @@ defmodule Oli.Delivery.Attempts.PageLifecycle.Ungraded do
          revision: %{content: %{"advancedDelivery" => true}},
          resource_access_id: resource_access_id
        }) do
-    resource_access_id
-    |> Oli.Delivery.Attempts.Core.get_resource_access()
-    |> Oli.Delivery.Metrics.mark_progress_completed()
+    Oli.Delivery.Metrics.mark_progress_completed(resource_access_id)
   end
 
   defp maybe_mark_adaptive_progress_completed(_), do: {:ok, nil}
--- a/lib/oli/delivery/attempts/page_lifecycle/ungraded.ex
+++ b/lib/oli/delivery/attempts/page_lifecycle/ungraded.ex
@@ -58,6 +58,8 @@ defmodule Oli.Delivery.Attempts.PageLifecycle.Ungraded do
   end
 
   @impl Lifecycle
+  @spec finalize(FinalizationContext.t()) ::
+          {:ok, FinalizationSummary.t()} | {:error, term()}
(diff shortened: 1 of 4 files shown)


nicocirio:
Follow-up on the AI review comments:

- On the transaction/Ecto.Multi concern: the adaptive ungraded finalize path already executes inside the outer PageLifecycle.finalize/3 transaction. If Ungraded.finalize/1 returns `{:error, reason}`, the outer finalize path rolls the whole operation back, so this does not leave a partially committed resource attempt.

- On the extra ResourceAccess fetch: taken in follow-up. I added a direct Metrics.mark_progress_completed/1 path for resource_access_id and updated the adaptive finalize path to use it, avoiding the extra lookup. I also added direct coverage for the new resource_access_id variant in Metrics.ProgressTest.

gastonabella (approved):
LGTM!

PR #6699, opened 2026-06-30
[ENHANCEMENT] [MER-5625] Correct randomized-bank LO minimum coverage and query performance

manelli:
- Batch instructor preview bank candidate lookup across all selections on a page to avoid issuing one candidate query per randomized bank.
- Reuse the bounded candidate result handling for single-selection and batched preview summary paths.
- Correct randomized-bank LO minimum coverage so ranges reflect guaranteed coverage when the selection count forces at least one candidate for an objective.
- Remove a duplicate `Modal` alias in the instructor preview LiveView.

image

See: https://eliterate.atlassian.net/browse/MER-5625

### DIFF
--- a/lib/oli/activities/realizer/query/batch.ex
+++ b/lib/oli/activities/realizer/query/batch.ex
@@ -1,0 +1,105 @@
+defmodule Oli.Activities.Realizer.Query.Batch do
+  @moduledoc """
+  Executes multiple realizer queries in a single database round trip.
+  """
+
+  alias Oli.Activities.Realizer.Logic
+  alias Oli.Activities.Realizer.Query.Builder
+  alias Oli.Activities.Realizer.Query.Paging
+  alias Oli.Activities.Realizer.Query.Result
+  alias Oli.Activities.Realizer.Query.Source
+  alias Oli.Repo
+  alias Oli.Resources.Revision
+
+  @type query_id :: String.t()
+  @type query_spec :: {query_id(), %Logic{}, Source.t()}
+
+  @doc """
+  Executes query specs and returns `%Result{}` values keyed by query id.
+  """
+  @spec execute([query_spec()], Paging.t(), :paged | :random) ::
+          {:ok, %{query_id() => %Result{}}} | {:error, term()}
+  def execute([], %Paging{}, view_type) when view_type in [:paged, :random], do: {:ok, %{}}
+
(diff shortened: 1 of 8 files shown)


gastonabella (commented):
Great refactor overall; batching the candidate lookups is a nice performance improvement and should reduce unnecessary database round trips.

Suggestion: consider moving the batching/query-composition logic out of `TargetResolver` and into a reusable Realizer query module (i.e `Oli.Activities.Realizer.Query.Batch`).

`TargetResolver` currently knows quite a bit about how Realizer queries are built and decoded: shifting placeholders, wrapping each branch with `selection_id`, joining branches with `UNION ALL`, executing the combined query, and rebuilding `%Revision{}` structs from raw rows. That feels closer to `Oli.Activities.Realizer.Query` infrastructure than instructor-customization target resolution.

A small extraction could keep `TargetResolver` focused on domain concerns while making the batch behavior reusable:

```elixir
queries =
  Enum.reduce_while(selections, {:ok, []}, fn selection, {:ok, acc} ->
    selection_id = selection["id"]

    with {:ok, logic} <- parse_logic(selection) do
      excluded_ids =
        excluded_ids_by_selection_id
        |> Map.get(selection_id, MapSet.new())
        |> MapSet.to_list()

      source = %Source{
        publication_id: publication_id,
        section_slug: section.slug,
        blacklisted_activity_ids: excluded_ids
      }

      {:cont, {:ok, [{selection_id, logic, source} | acc]}}
    else
      error -> {:halt, error}
    end
  end)

with {:ok, queries} <- queries do
  Query.Batch.execute(Enum.reverse(queries), paging, :paged)
# => {:ok, %{"selection-a" => %Result{}, "selection-b" => %Result{}}}
end
```

That way this module would only decide which selection queries need to run, while the Realizer query layer would own how to batch, execute, and decode them.

manelli:
@gastonabella Implemented, thanks for the idea. `TargetResolver` now only prepares selection-specific query specs, and the SQL batching lives in `Oli.Activities.Realizer.Query.Batch`

See: 9c3ebfa623c3f3f9c1fdbc1b0f09964615aca362

gastonabella (approved):
Looks good to me, thanks for extracting this!

One small follow-up thought: since `Batch.execute/3` currently accepts `:random`, it may be worth either adding test coverage for that mode or narrowing the public contract to `:paged` until random batching is actually needed.

PR #6701, opened 2026-06-30
[BUG FIX] [MER-5756] fix automated reveal of paginated questions

andersweinstein:
## Summary

  Fixes MER-5756: basic-page branching questions using “reveal questions by automation” stopped revealing subsequent paginated questions after the first response.

  The regression was introduced by #6152 / `a6a03c4c141` (`[MER-4470] condense pagination links on mobile devices`). That PR added defensive clamping in `PaginationControls.onSelectPage` based on `pages.count()`. The clamp itself is reasonable, but the automation event listener for the `oli-show-content-page` event was registered only once during the initial render, picking up an instance of onSelectPage that closed over `pages` when `pages` was still empty. As a result, automated reveal events saw `pages.count() === 0` and clamped every requested page back to page `0`.

  This meant learner delivery received the branching `show_page` action correctly, but the requested next page was never shown.

  ## Changes

  - Keep the latest pagination page count in a ref so the one-time automation event listener can clamp
  against the initialized page count.
  - Add a regression test that dispatches `oli-show-content-page` after pagination initialization and
  verifies the next page is revealed.

  ## Drive-by cleanup

  - Skip pagination-state persistence when there is no resource attempt guid as happens in authoring preview. This prevents authoring-preview 404s like `/resource_attempt/?keys[]=paginationState` which, though in fact harmless, add noise to the console window. 
  - Add a small test for the empty-attempt-guid persistence guard.

  ## Verification

  - `cd assets && yarn test test/components/misc/PaginationControls_test.tsx test/data/persistence/
  pagination_test.ts --runInBand`
  - `cd assets && yarn run check-types`

[MER-4470]: https://eliterate.atlassian.net/browse/MER-4470?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/assets/src/components/misc/PaginationControls.tsx
+++ b/assets/src/components/misc/PaginationControls.tsx
@@ -18,6 +18,11 @@ export interface PaginationControlsProps {
 type Page = List<Element>;
 type DisplayItem = { type: 'page'; index: number; key: string } | { type: 'ellipsis'; key: string };
 
+const clampPageIndex = (pageIndex: number, pageCount: number) => {
+  const maxIndex = Math.max(0, pageCount - 1);
+  return Math.min(Math.max(0, pageIndex), maxIndex);
+};
+
 export const PaginationControls = (props: PaginationControlsProps) => {
   const controls = useRef<HTMLDivElement>(null);
   const pageCount = useRef(0);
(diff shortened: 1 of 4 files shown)


andersweinstein:
[codex: replying to https://github.com/Simon-Initiative/oli-torus/pull/6701#issuecomment-4847033792]

`onSelectPage` does not currently perform side effects beyond clamping and setting `active`; the shared behavior has been extracted into `clampPageIndex(...)` and is used by both the document event handler and `onSelectPage`.

The event handler intentionally does not call `onSelectPage` because a one-time document listener would capture the initial render’s callback, which closes over the initial empty `pages` list. That stale closure is the root cause of this regression. The event handler uses `pageCount.current` so it can remain registered once while clamping against the initialized page count.

eliknebel (approved):

PR #6706, opened 2026-07-01
[BUG FIX] [MER-5725] Fix various issues with deleted pages and learning objectives

darrensiegel:
This PR fixes three distinct problems:

**1. The Instructor Dashboard was continuing to show learning objectives that were no longer reachable from the course content.** 

This was easy to reproduce:

1. As an author, create an objective, place an activity on a page in the course project and attach that objective to that activity. 
2. Publish the project, apply the update to a course section and then as a student, answer the question.
3. Verify that in the Instructor Dashboard that you see results for that objective.
4. Then as an author, remove that page from the course.  Publish and apply the update to the section.
5. As instructor, look again at Instructor Dashboard "Learning Objectives" tab and see that the table still shows that objective. 

The fix here was to post-process the list of learning objectives and filter out any LOs that are no longer "reachable" from the course content.  With this fix in place, the LO no longer displays in step 5 above. 

**2. The "By Objective" option in the Authoring Insights view was showing results for all objectives, included those marked as deleted.** 

To reproduce:

1. Publish a course project with a learning objective attached to an activity. 
2. As a student, answer the question.
3. As author, in "Insights" view, verify you see data for that LO when you use "By Objective" option.
4. As author, delete the host page, and then Remove the objective from the Objectives editor.
5. As author, from "Insights" view, use "By Objective" option and you will still see that LO listed.

Fix was to only show non deleted objectives. 

**3. The "Remove" learning objective feature would crash if attempting to detach the learning objective from an "orphaned" activity.** 

 To reproduce this issue:

1. Create a page with an activity, attach an LO to that activity.
2. From Curriculum view, delete that page from the course. 
3. From the "Objectives" view, attempt to "Remove" that LO. 
4. You should see a modal dialog appear indicating that the LO is still attached to that Activity.
5. Click "Ok" to proceed (to detach the LO and delete).  

Without this fix the removal fails due to a BadMapError.  With this fix, the detachment succeeds and the LO is removed.

### DIFF
--- a/lib/oli_web/live/delivery/instructor_dashboard/instructor_dashboard_live.ex
+++ b/lib/oli_web/live/delivery/instructor_dashboard/instructor_dashboard_live.ex
@@ -119,7 +119,8 @@ defmodule OliWeb.Delivery.InstructorDashboard.InstructorDashboardLive do
             Sections.get_objectives_and_subobjectives(socket.assigns.section,
               exclude_sub_objectives: false,
               include_related_activities_count: true
-            ),
+            )
+            |> filter_root_contained_objectives(),
           navigator_items:
             Oli.Delivery.Sections.SectionResourceDepot.containers(socket.assigns.section.id,
               numbering_level: {:in, [1, 2]}
@@ -1686,4 +1687,31 @@ defmodule OliWeb.Delivery.InstructorDashboard.InstructorDashboardLive do
   end
 
   defp parse_dashboard_section_split(_split), do: :error
+
+  defp filter_root_contained_objectives(objectives) do
+    root_contained_ids =
+      objectives
+      |> Enum.filter(&root_contained_objective?/1)
+      |> Enum.map(& &1.resource_id)
+      |> MapSet.new()
+
+    parent_ids_for_root_contained_subobjectives =
+      objectives
+      |> Enum.filter(fn objective ->
+        not is_nil(objective.subobjective) and
+          MapSet.member?(root_contained_ids, objective.resource_id)
+      end)
+      |> Enum.map(& &1.objective_resource_id)
+      |> MapSet.new()
+
+    Enum.filter(objectives, fn objective ->
+      MapSet.member?(root_contained_ids, objective.resource_id) or
+        (is_nil(objective.subobjective) and
+           MapSet.member?(parent_ids_for_root_contained_subobjectives, objective.resource_id))
+    end)
+  end
+
+  defp root_contained_objective?(objective) do
+    nil in List.wrap(Map.get(objective, :container_ids))
+  end
 end
--- a/lib/oli/analytics/summary/browse_insights.ex
+++ b/lib/oli/analytics/summary/browse_insights.ex
@@ -133,14 +133,25 @@ defmodule Oli.Analytics.Summary.BrowseInsights do
          %Sorting{} = sorting,
          options
        ) do
-    total_count = get_total_count(query, options)
+    case objective_resource_type?(options) do
+      true ->
+        query
+        |> add_non_activity_select(0, options)
+        |> add_non_activity_order_by(sorting, options)
+        |> Repo.all()
+        |> filter_deleted_objectives()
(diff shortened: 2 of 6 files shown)


lib/oli_web/live/delivery/instructor_dashboard/instructor_dashboard_live.ex
```diff
@@ -1686,4 +1687,31 @@ defmodule OliWeb.Delivery.InstructorDashboard.InstructorDashboardLive do
...
+          MapSet.member?(root_contained_ids, objective.resource_id)
+      end)
+      |> Enum.map(& &1.objective_resource_id)
+      |> MapSet.new()
+
+    Enum.filter(objectives, fn objective ->
+      MapSet.member?(root_contained_ids, objective.resource_id) or
+        (is_nil(objective.subobjective) and
+           MapSet.member?(parent_ids_for_root_contained_subobjectives, objective.resource_id))
+    end)
```
nicocirio, line 1711: If in the future we find that this list can grow large enough for this helper to become a measurable hotspot, one possible follow-up would be to collapse the current 3 passes over `objectives` into 2 passes instead.

```elixir
defp filter_root_contained_objectives(objectives) do
  {root_contained_ids, parent_ids_for_root_contained_subobjectives} =
    Enum.reduce(objectives, {MapSet.new(), MapSet.new()}, fn objective, {root_ids, parent_ids} ->
      if root_contained_objective?(objective) do
        root_ids = MapSet.put(root_ids, objective.resource_id)

        parent_ids =
          if not is_nil(objective.subobjective) do
            MapSet.put(parent_ids, objective.objective_resource_id)
          else
            parent_ids
          end

        {root_ids, parent_ids}
      else
        {root_ids, parent_ids}
      end
    end)

  Enum.filter(objectives, fn objective ->
    MapSet.member?(root_contained_ids, objective.resource_id) or
      (is_nil(objective.subobjective) and
         MapSet.member?(parent_ids_for_root_contained_subobjectives, objective.resource_id))
  end)
end
```

That would make this helper traverse the list twice instead of three times.

nicocirio (approved):
LGTM

All of the following validation steps were completed, and in every case the fix behaved as expected:

1. As an author, create these learning objectives:
   - `LO Active`
   - `LO Parent`
   - `LO Child` as a subobjective of `LO Parent`
   - `LO Removed`
2. As an author, create two scored pages in the course:
   - `Page A`
   - `Page B`
3. On `Page A`, add two graded activities.
4. Attach `LO Active` to the first activity on `Page A`.
5. Attach only `LO Child` to the second activity on `Page A`.
6. Do not attach `LO Parent` directly to any page or activity.
7. On `Page B`, add one graded embedded activity.
8. Attach `LO Removed` to the activity on `Page B`.
9. Publish the project and apply the update to a section.
10. As a student, answer all three activities so analytics data exists.
11. As an instructor, open `Instructor Dashboard > Learning Objectives`.
12. Verify that all of the following are visible:
   - `LO Active`
   - `LO Child`
   - `LO Parent`
   - `LO Removed`
13. Specifically confirm that `LO Parent` appears even though only `LO Child` is directly attached to an activity.
14. As an author, open `Insights > By Objective`.
15. Verify that all of the following are visible:
   - `LO Active`
   - `LO Child`
   - `LO Removed`
16. As an author, remove `Page B` from the course.
17. Then go to `Objectives` and remove `LO Removed`.
18. If the delete/detach modal appears, confirm the action.
19. Verify that the removal completes successfully and that no crash occurs.
20. Publish the project again and apply the update to the same section.
21. As an instructor, return to `Instructor Dashboard > Learning Objectives`.
22. Verify that `LO Removed` is no longer shown.
23. Verify that all of the following are still shown:
   - `LO Active`
   - `LO Child`
   - `LO Parent`
24. Confirm again that `LO Parent` remains visible as context for the subobjective.
25. As an author, return to `Insights > By Objective`.
26. Verify that the following are still shown:
   - `LO Active`
   - `LO Child`

PR #6710, opened 2026-07-03
[ENHANCEMENT] [MER-5422] Add Playwright coverage for Simple Author adaptive branching

manelli:
- Adds automation support for creating Simple Author adaptive practice pages.
- Builds and publishes a three-screen Simple Author MCQ branching lesson:
  - Routing Question
  - Correct Terminal
  - Incorrect Terminal
- Verifies both correct and incorrect learner routes in delivery.
- Hardens adaptive authoring page objects for Simple Author flowchart/screen-panel mode, read-only toggling, MCQ insertion, screen renaming, and flowchart rule creation.
- Stabilizes student course entry by bounding welcome-page navigation to `domcontentloaded`.

See: https://eliterate.atlassian.net/browse/MER-5422

### DIFF
--- a/assets/automation/tests/torus/course_authoring/adaptive-authoring.spec.ts
+++ b/assets/automation/tests/torus/course_authoring/adaptive-authoring.spec.ts
@@ -1,6 +1,6 @@
 import { setRuntimeConfig } from '@core/runtimeConfig';
 import { test } from '@fixture/my-fixture';
-import { Browser, BrowserContext, Page, expect } from '@playwright/test';
+import { BrowserContext, Page, expect } from '@playwright/test';
 import { StudentCoursePO } from '@pom/course/StudentCoursePO';
 import { TYPE_USER } from '@pom/types/type-user';
 import { CurriculumTask } from '@tasks/CurriculumTask';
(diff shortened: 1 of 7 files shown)


assets/automation/tests/torus/course_authoring/adaptive-authoring.spec.ts
```diff
@@ -115,7 +123,165 @@ test('Author creates an adaptive page with MCQ and student resolves it', async (
...
+  test('author creates and publishes the lesson', async ({ browser }) => {
+    test.setTimeout(180_000);
+
+    await buildAndPublishBranchingLesson(browser);
+  });
+
+  test('student follows the correct route', async ({ browser }) => {
+    test.setTimeout(120_000);
+
+    const context = await browser.newContext();
```
nicocirio: I think these tests are opening an extra Playwright page that never gets used.

`forEachTest` in `assets/automation/src/core/fixture/my-fixture.ts` always navigates the implicit fixture `page` to `/`, but these cases also create their own `browser.newContext()` + `newPage()` in the spec. That seems to leave one idle window/page per test.

This does not look functionally wrong, but it does add noise and unnecessary overhead. It may be worth either disabling the auto fixture for this file or consolidating the flow so each test uses a single page/context.

https://github.com/user-attachments/assets/151d878d-6219-4754-a709-796c72aaae06

nicocirio (approved):
Nice work on this. I noticed one small improvement opportunity: these tests seem to open an extra idle browser window/page in each case. If it is a simple adjustment, it would be great to clean that up and avoid the extra overhead. It is not blocking from my side, so I’m approving the PR as is.

assets/automation/tests/torus/course_authoring/adaptive-authoring.spec.ts
manelli: Thanks Nico for the review! Fixed in 927ea37a94

PR #6712, opened 2026-07-06
[ENHANCEMENT] [MER-5446] Add delivery automation coverage for activities

nicocirio:
Jira: [MER-5446](https://eliterate.atlassian.net/browse/MER-5446)

Summary:
- Add Playwright delivery coverage for image hotspot, image coding, vlab, logic lab, and embedded activities.
- Reuse shared student-delivery helpers and scenario seeds to keep specs focused.

Implementation notes:
- Grouped related assertions with test.step blocks to reduce repeated setup.
- Added a deterministic embedded runtime stub for automation-only launch coverage.
- Validated embedded attempt state through Torus state API instead of the database.

Validation:
- mix format
- mix compile
- yarn --cwd assets run format
- yarn --cwd assets run check-types
- yarn --cwd assets run deploy

Evidence:
Screenshot 2026-07-07 at 9 53 28 AM

[MER-5446]: https://eliterate.atlassian.net/browse/MER-5446?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nicocirio:
Following up on the bot comments:

1. Embedded runtime stub route
This route is already gated behind `:enable_playwright_scenarios`, which is disabled in `prod` and only enabled for local automation-oriented environments. We added it to support deterministic embedded delivery coverage when the legacy runtime assets are unavailable locally. I'm leaving any additional hardening of the dev-only exposure as follow-up work so this PR stays focused on the automation coverage itself.

2. Runtime URL fallback scheme
Addressed in `75997f1a70`. The fallback no longer downgrades generated runtime URLs to `http`; it now falls back to `https` instead.

andersweinstein:
Note the rather specialized image_coding activity uses three different types of javascript programs: plain text output programs, image processing programs which load and process an image resource using a little library, and "table" (spreadsheet) processing programs which do the same for a spreadsheet resource (csv file). The test here only tests the first type. The plain text type is certainly the simplest and suffices to test a lot of the functionality of the activity. But I wonder if it should also test at least an image processing example, as that involves some different aspects (loading the image resource, and a more involved mode of evaluation which compares the image resulting from student code with the result of running the solution code). in the special course that gave rise to this activity, the main examples are the image processing ones (that is why it gets this name). I can point you to examples if needed.

nicocirio:
> Note the rather specialized image_coding activity uses three different types of javascript programs: plain text output programs, image processing programs which load and process an image resource using a little library, and "table" (spreadsheet) processing programs which do the same for a spreadsheet resource (csv file). The test here only tests the first type. The plain text type is certainly the simplest and suffices to test a lot of the functionality of the activity. But I wonder if it should also test at least an image processing example, as that involves some different aspects (loading the image resource, and a more involved mode of evaluation which compares the image resulting from student code with the result of running the solution code). in the special course that gave rise to this activity, the main examples are the image processing ones (that is why it gets this name). I can point you to examples if needed.

@andersweinstein Thanks for the insight! And yes, sure, please share those examples with me to understand all the "flavours" an image coding activity may have, so I can expand the test cases.

Francisco-Castro:
Hi @nicocirio! 👋 While building [MER-5443](https://eliterate.atlassian.net/browse/MER-5443), I reused the support scaffolding and `StudentCoursePO` changes from this PR — thanks, they slotted in cleanly.

During Codex cross-review, two small `StudentCoursePO` improvements came up. Since that helper is shared across the Lane 4 delivery specs, I wanted to flag them here in case they're easy to fold into this PR. If you'd rather keep scope tight, totally fine — my branch is expected to merge after yours, so I can pick them up in [MER-5443](https://eliterate.atlassian.net/browse/MER-5443) after this lands.

1. `openPage` outline matcher may be ambiguous

```typescript
Current matcher:
const outlineTitle = this.page.getByRole('button', {
  name: new RegExp(`\\b${escapeRegExp(pageName)}\\b`),
});
```
This can match multiple outline buttons when page names overlap, like Quiz and Quiz 2, which could lead to strict-mode failures or the wrong page being selected. Suggested change, assuming the button accessible name is exactly the page title:

```typescript
const outlineTitle = this.page.getByRole('button', { name: pageName, exact: true });
```
2. `findVisibleCourseHome` uses a manual polling loop

The current loop adds repeated timeout entries to Playwright traces and duplicates behavior Playwright already handles well with locator auto-waiting. A combined locator wait may be clearer:
```typescript
await this.page
  .locator('#home-continue-learning, #home-assignments, #view_selector')
  .first()
  .waitFor({ state: 'visible', timeout });
```

Both seem low-code but worth verifying against the delivery suite because `StudentCoursePO` is shared navigation infrastructure. No pressure to take them here — happy to carry them in [MER-5443](https://eliterate.atlassian.net/browse/MER-5443) if that's cleaner.

[MER-5443]: https://eliterate.atlassian.net/browse/MER-5443?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MER-5443]: https://eliterate.atlassian.net/browse/MER-5443?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MER-5443]: https://eliterate.atlassian.net/browse/MER-5443?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

nicocirio:
@Francisco-Castro Thanks, Francisco. I took a version of the `openPage` matcher suggestion here.

Instead of matching the button by exact accessible name, I tightened it inside `StudentCoursePO` by looking for a button that contains the exact page title text. That keeps the helper API stable for the existing specs while avoiding the ambiguous partial-match behavior you pointed out.

For `findVisibleCourseHome`, I’d prefer not to change it in this PR. That polling helper predates this branch, is currently working for the delivery flows covered here, and we didn’t introduce or modify that behavior as part of MER-5446. To keep scope tight, I’d rather leave that refactor for a follow-up if we want to revisit the shared navigation infrastructure more broadly.

darrensiegel (approved):

PR #6715, opened 2026-07-08
[ENHANCEMENT] [MER-5495] Add scenario coverage for course discussions

andersweinstein:
## Summary

  Adds scenario-level coverage for Course Discussion participation and moderation workflows for [MER-5495](https://eliterate.atlassian.net/browse/MER-5495).

  This extends `Oli.Scenarios` with reusable discussion primitives:

  - `discussion_config`
  - named `discussion_post`
  - `discussion_moderation`
  - `discussion_delete`
  - `assert.discussion`

  The new scenario covers:

  - Course Discussion enabled/disabled inheritance from product to section
  - student post creation
  - visibility to another student
  - student deletion
  - pending posts when auto-accept is disabled
  - instructor approval
  - instructor rejection/deleted state

  Existing certificate scenario behavior is preserved: `discussion_post` still synchronously processes certificate eligibility side effects.

  ## Why Scenario Coverage

  The manual regression tests describe end-to-end user workflows, but the behavior under test is primarily application state and LiveView/server event behavior, not browser-specific rendering or external integration.

  Per the automated testing plan, the preferred hierarchy is:

  1. Scenario tests for integrated application workflows
  2. LiveView tests where UI event wiring needs direct coverage
  3. Playwright only when browser behavior or cross-system behavior is essential

  For this work, scenario tests are sufficient because they exercise the real Torus application paths for section discussion
  configuration, post creation, moderation, deletion, visibility, and persistence without relying on brittle browser automation. No Canvas/LTI flow, browser-only interaction, or client-heavy behavior is required for the covered acceptance cases.

  ## Validation

  - `mix compile`
  - `MIX_ENV=test mix test test/scenarios/discussions/discussions_test.exs test/scenarios/certificates/certificates_test.exs test/oli/
  scenarios/certificate_directives_test.exs test/oli/scenarios/certificate_parser_test.exs test/scenarios/validation/
  invalid_attributes_test.exs test/scenarios/validation/schema_validation_test.exs`

[MER-5495]: https://eliterate.atlassian.net/browse/MER-5495?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli/scenarios/directives/assert/discussion_assertion.ex
+++ b/lib/oli/scenarios/directives/assert/discussion_assertion.ex
@@ -56,6 +56,7 @@ defmodule Oli.Scenarios.Directives.Assert.DiscussionAssertion do
           discussion_data.anonymous_posting,
           "anonymous_posting"
         ),
+        check_post_scope(section, post),
         compare_field(post && post.status, discussion_data.status, "status"),
         check_visibility(section, student, post, discussion_data.visible)
       ]
@@ -84,6 +85,18 @@ defmodule Oli.Scenarios.Directives.Assert.DiscussionAssertion do
     if actual == expected,
       do: :ok,
       else: "expected #{label}=#{inspect(expected)}, got #{inspect(actual)}"
+  end
+
+  defp check_post_scope(_section, nil), do: :ok
+
+  defp check_post_scope(section, post) do
+    root_resource_id = section.root_section_resource && section.root_section_resource.resource_id
+
+    if post.section_id == section.id and post.resource_id == root_resource_id do
+      :ok
+    else
+      "post does not belong to section '#{section.slug}'"
+    end
   end
 
   defp check_visibility(_section, _student, _post, nil), do: :ok
--- a/lib/oli/scenarios/directives/assert/discussion_assertion.ex
+++ b/lib/oli/scenarios/directives/assert/discussion_assertion.ex
@@ -1,0 +1,121 @@
+defmodule Oli.Scenarios.Directives.Assert.DiscussionAssertion do
+  @moduledoc """
+  Handles assertions for course discussion configuration and named posts.
+  """
+
+  alias Oli.Repo
+  alias Oli.Resources.Collaboration
+  alias Oli.Scenarios.DirectiveTypes.{AssertDirective, VerificationResult}
+  alias Oli.Scenarios.Directives.Assert.Helpers
+  alias Oli.Scenarios.Engine
+
+  def assert(%AssertDirective{discussion: discussion_data}, state)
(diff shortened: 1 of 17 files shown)


andersweinstein:
[Codex:] Responding to the remaining Elixir review note about `discussion_assertion.ex` treating `:deleted` posts as visible: I left this behavior unchanged intentionally. In the current delivery UI, deleted public discussion posts remain in the rendered discussion list as deleted/tombstone posts rather than disappearing entirely, and the manual moderation expectation for rejected posts is that the student sees the post as deleted. In this scenario assertion, `visible` is therefore modeling whether the named post is visible in the discussion surface, not whether its non-deleted content is visible. The separate `status: "deleted"` assertion covers the tombstone state.

lib/oli/scenarios/directives/discussion_post_handler.ex
```diff
@@ -5,34 +5,104 @@ defmodule Oli.Scenarios.Directives.DiscussionPostHandler do
...
+          name: post_name,
+          student: student_name,
+          section: section_name,
+          body: body,
+          reply_to: reply_to,
+          anonymous: anonymous
+        },
         %ExecutionState{} = state
       ) do
     with {:ok, student} <- fetch_user(state, student_name),
```
gastonabella, line 27: Could we make `discussion_post` verify that the student is enrolled in the target section before creating the post? Since this directive is meant to model student discussion participation, allowing any existing user to post could let a scenario pass even though the real delivery workflow would not be reachable for that actor.

lib/oli/scenarios/directives/assert/discussion_assertion.ex
```diff
@@ -0,0 +1,121 @@
...
+          discussion_data.contains_discussions,
+          "contains_discussions"
+        ),
+        compare_field(config && config.auto_accept, discussion_data.auto_accept, "auto_accept"),
+        compare_field(
+          config && config.anonymous_posting,
+          discussion_data.anonymous_posting,
+          "anonymous_posting"
+        ),
+        compare_field(post && post.status, discussion_data.status, "status"),
```
gastonabella, line 60: Could we have discussion assertions verify that a named post belongs to the asserted section/resource whenever `post` is provided? Right now `status` compares the named post directly, so a scenario could accidentally reference a post from another section and still pass unless it also asserts `visible`.

andersweinstein: Good eye! Both done.

gastonabella (approved):
LGTM 🏅

PR #6719, opened 2026-07-08
[BUG FIX] [MER-5767] Fix practice page progress after reset

eliknebel:
## Summary

Fixes MER-5767 by allowing practice/ungraded page progress to be recomputed from reset activity attempts. Previously, page progress updates only ran for the first scoreable activity attempt, so a learner who reset a partially completed practice page could complete the visible reset attempts while the page stayed stuck below 100%.

## Query Change Details

`Oli.Delivery.Metrics.update_page_progress/1` now treats any scoreable activity attempt as eligible to trigger progress recomputation, instead of only scoreable attempts where `attempt_number = 1`. This is necessary because reset activity attempts normally have a higher attempt number.

The SQL update in `lib/oli/delivery/metrics.ex` was reworked into CTEs:

- `target` resolves the current activity attempt guid to its parent page attempt, revision, and `resource_access` row. This keeps the update scoped to exactly one learner/page access record.
- `latest_attempts` selects the latest scoreable attempt for each activity resource on that same page attempt using `SELECT DISTINCT ON (aa2.resource_id)` ordered by `aa2.id DESC`. This replaces the old `aa2.attempt_number = 1` filter, which excluded reset attempts.
- `counts` computes completion using only those latest scoreable attempts: submitted or evaluated attempts count as complete, and all latest scoreable attempts count toward the denominator.

The final `UPDATE resource_accesses` still preserves the existing monotonic behavior with `GREATEST(calculated_progress, resource_accesses.progress)` and caps progress with `LEAST(1.0, ...)`, so recomputation can repair stuck progress but will not reduce already-complete progress.

## Tests

- Added regression coverage for a partially completed page where reset attempts complete the remaining scoreable activities and page progress reaches `1.0`.

## Verification

- `mix format lib/oli/delivery/metrics.ex test/oli/delivery/metrics/progress_test.exs`
- `mix test test/oli/delivery/metrics/progress_test.exs`
- `git diff --check`

## Risk

Low to moderate. The update remains scoped to a single `resource_access` row, but progress recomputation now runs for scoreable reset attempts and evaluates latest attempts per activity resource rather than first attempts only.

### DIFF
--- a/lib/oli/delivery/metrics.ex
+++ b/lib/oli/delivery/metrics.ex
@@ -1212,17 +1212,17 @@ defmodule Oli.Delivery.Metrics do
   @doc """
   For an activity attempt specified by an attempt guid, calculate and set in the corresponding resource access
   record, the percentage complete for the related page. This calculation only needs to be performed after the
-  evaluation of the first attempt for given activity.  This method should update exactly one record, the resource
+  evaluation of a scoreable attempt for given activity.  This method should update exactly one record, the resource
   access for the page that this activity attempt ultimately pertains to (through its parent resource attempt).
 
   Can return one of:
   {:ok, :updated} -> Progress calculated and set
-  {:ok, :noop} -> Nothing needed to be done, since the attempt number was greater than 1
+  {:ok, :noop} -> Nothing needed to be done, since the attempt was not scoreable
   {:error, :unexpected_update_count} -> 0 or more than 1 record would have been updated, rolled back
   {:error, e} -> An other error occurred, rolled back
   """
   def update_page_progress(activity_attempt_guid) when is_binary(activity_attempt_guid) do
-    if Core.is_scoreable_first_attempt?(activity_attempt_guid) do
+    if scoreable_activity_attempt?(activity_attempt_guid) do
       do_update(activity_attempt_guid)
     else
       {:ok, :noop}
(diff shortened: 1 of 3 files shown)


lib/oli/delivery/metrics.ex
```diff
@@ -1241,9 +1240,42 @@ defmodule Oli.Delivery.Metrics do
...
+          LIMIT 1
+        ),
+        latest_attempts AS (
+          SELECT DISTINCT ON (aa2.resource_id)
+            aa2.id,
+            aa2.lifecycle_state
+          FROM activity_attempts AS aa2
+          JOIN target ON target.resource_attempt_id = aa2.resource_attempt_id
+          WHERE aa2.scoreable = true
+          ORDER BY aa2.resource_id, aa2.id DESC
```
nicocirio: Non-blocking: It may be safer to order by `aa2.attempt_number DESC, aa2.id DESC` instead, since the business concept here is latest attempt number, and `id DESC` is only an implicit proxy. This is probably fine in normal operation, but in a very unlikely case such as a migration, backfill, or restore inserting rows out of the original sequence, a higher `id` could end up pointing to an older attempt number.

nicocirio (approved):

lib/oli/delivery/metrics.ex
eliknebel: Agreed, I will make this change

PR #6721, opened 2026-07-08
[BUG FIX] [MER-5766] Restore manual grading for adaptive CAPI iframes

kjsegovi:
## Summary

Fixes MER-5766: manually scored `janus-capi-iframe` attempts were not appearing in the
Instructor manual grading queue and were grading out of 1 instead of the screen max.

- Re-expose **Requires Manual Grading** for `janus-capi-iframe` in Advanced Author
  (without treating it as an auto-scored adaptive input)
- Honor an explicit manual-grading flag even when the iframe is referenced by a
  scoring/trap-state rule (`persisted_part_grading_approach`)
- Submit manual adaptive attempts to the grading queue only when the rules engine
  returns a correct result, so students can retry on incorrect
- On incorrect, leave all part attempts active (do not persist automatic parts early)
  to avoid stale scores on retry
- Backfill instructor **Out Of** from the screen `maxScore` for manual non-scorable
  adaptive parts (iframe) at grading time

Screen `maxScore` remains the single source of truth for iframe scoring; authors do not set a separate part-level max.

## Test plan
I have done testing for each component in Advanced Author and confirmed that there is no regression. The fix also works as expected. 

That being said, QA testing on hotfixes needs to be diligent as well. Each component needs to be tested to ensure functionality, and the iframe component needs to be tested extensively with various URLs linked in the iframe. 

## Automated tests

- `assets/test/advanced_authoring/adaptive_scorable_part_types_test.ts`
- `test/oli/activities/adaptive_parts_test.exs`
- `test/oli/delivery/attempts/activity_lifecycle/evaluate_test.exs`

### DIFF
--- a/assets/src/apps/authoring/components/PropertyEditor/schemas/part.ts
+++ b/assets/src/apps/authoring/components/PropertyEditor/schemas/part.ts
@@ -15,8 +15,23 @@ export const adaptiveScorablePartTypes = new Set([
   'janus-fill-blanks',
 ]);
 
+// Stateful, non-auto-scored parts that can still be flagged for manual grading.
+// These are NOT auto-scored adaptive inputs (so they do not receive adaptive feedback
+// defaults and do not contribute an automatic screen score), but authors need to be able
+// to mark them as requiring manual grading (e.g. an embedded report inside an iframe).
+export const manualGradablePartTypes = new Set(['janus-capi-iframe']);
+
 export const isAdaptiveScorablePartType = (type?: string | null) =>
   !!type && adaptiveScorablePartTypes.has(type);
+
+export const isManualGradablePartType = (type?: string | null) =>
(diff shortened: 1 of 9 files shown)


Francisco-Castro (approved):
Approving. Fix addresses both regression roots (#6401 UI gating, #6495 grading approach override). Tests pass locally. CI bot comments about `maxScore` in `required` are invalid — no Scoring schema has a `required` array.

Notes:
1. **QA**: attempts now enter the grading queue only on a "correct" rules result (new behavior). A manual-iframe screen with no reachable correct rule would never reach the instructor — please test that case and confirm the design.
2. **Follow-ups (non-blocking)**: Simple Author shows the iframe toggle on responsive pages but not non-responsive (pre-existing gap); `apply_adaptive_manual_out_of` backfill has no test coverage.

PR #6725, opened 2026-07-13
[FEATURE] [MER-5795] Native AB test experiments support

eliknebel:
https://eliterate.atlassian.net/browse/MER-5795

This PR replaces Torus's external UpGrade dependency with native A/B/N experimentation. It introduces a Torus-owned experiment domain, authoring and lifecycle workflows, section participation controls, sticky learner assignment, weighted-random and Thompson Sampling policies, delivery-time exposure/reward handling, and experiment-aware xAPI/ClickHouse ingestion.

The implementation preserves published content immutability: experiments select among alternatives at delivery time without modifying published revisions. Native experiments start fresh; existing UpGrade experiments and learner assignments are not migrated.

### 1. Authoring an experiment and its decision points

Creating and managing an A/B decision point, defining conditions, configuring an experiment and policy, selecting participating sections, and exercising lifecycle controls.

https://github.com/user-attachments/assets/5b60f95d-0c0e-4b03-976f-d01e745fd826

### 2. Section delivery

Participating vs. non-participating behavior, learner assignment, sticky condition reuse, rendered alternatives, and reward flow.

https://github.com/user-attachments/assets/b24aff78-ec50-4915-87ba-1f483a257538

### 3. Experiment OLAP results in ClickHouse

Ingested experiment attribution rows and representative ClickHouse queries/results.

https://github.com/user-attachments/assets/0806e7ec-89d0-428e-bb14-9d93f3eb2631

### Native experiment domain

- Adds Torus-owned persistence and context APIs for experiments, decision points, conditions, participating sections, assignments, policy state, and runtime evidence.
- Enforces project, section, enrollment, and authorization scope at the experiment boundary.
- Keeps experiment schemas private to the owning context and exposes stable request/response contracts.
- Adds lifecycle validation for draft, active, paused, completed, and archived experiments.

### Authoring and section participation

- Adds native decision-point and condition management to the project Experiments workspace.
- Supports A/B/N experiment creation, editing, policy configuration, and lifecycle actions.
- Adds a dedicated experiment detail surface for policy details and eligible-section participation.
- Supports immediate participation updates and paginated section selection.
- Separates experiment decision points from learner-preference alternatives in authoring.
- Derives whether a project contains native experiments instead of relying on manual experiment toggles.

### Assignment policies

- Supports deterministic weighted-random assignment for baseline experiments.
- Supports non-contextual Thompson Sampling with a Beta-Bernoulli reward model.
- Persists explicit per-condition posterior state and policy-version metadata.
- Preserves sticky assignments before considering a new adaptive sample.
- Applies MVP guardrails and idempotent reward updates so retries do not update a posterior twice.

### Delivery runtime

- Routes delivery assignment and exposure recording through `Oli.Experiments`.
- Restricts experiment behavior to explicitly participating sections.
- Uses the first alternative as the safe fallback when no applicable active experiment exists or the section is not participating.
- Associates evaluated activity attempts with experiment outcomes/rewards and updates adaptive policy state after successful reward handling.
- Adds end-to-end scenario coverage for authoring, publishing, participation, delivery, stickiness, fallback, rewards, and posterior updates.

### UpGrade removal

- Removes UpGrade assignment, mark, metric-log, transport, builder, worker, and active configuration paths.
- Removes the UpGrade JSON export workflow; retained legacy export routes return `410 Gone` during the cutover window.
- Keeps legacy UpGrade-authored groups visible as read-only content with explicit non-migration messaging.
- Does **not** import legacy experiments or preserve legacy learner assignments.

### xAPI and ClickHouse foundation

- Attaches zero or more experiment attributions to existing page-view, evaluated-attempt, rollup, and media xAPI statements.
- Sends statements through the existing asynchronous xAPI/S3 pipeline; learner-facing transactions do not perform direct S3 or ClickHouse work.
- Extracts attribution-level rows into ClickHouse while preserving a hash link to the original raw event.
- Supports production ETL, local direct upload, and deterministic backfill/replay paths.
- Adds scoped OLAP query contracts and observability for experiment-attributed events.
- Excludes learner names, LMS identifiers, raw learner responses, and full adaptive-policy state from experiment attribution payloads.

## xAPI experiment attribution shape

Experiment evidence is represented by the `http://oli.cmu.edu/extensions/experiment_attributions` array on an existing xAPI host statement. A single page, attempt, rollup, or media event can carry multiple attributions when it is associated with more than one decision point or experiment.

### Complete evaluated-attempt example

This abbreviated host statement carries a reward attribution:

```json
{
  "timestamp": "2026-07-14T12:00:00Z",
  "actor": {
    "objectType": "Agent",
    "account": { "name": "123", "homePage": "https://proton.oli.cmu.edu" }
  },
  "verb": {
    "id": "http://adlnet.gov/expapi/verbs/completed",
    "display": { "en-US": "completed" }
  },
  "object": {
    "id": "https://proton.oli.cmu.edu/parts/part-1",
    "definition": {
      "type": "http://adlnet.gov/expapi/activities/question",
      "name": { "en-US": "Part 1" }
    }
  },
  "result": {
    "score": { "min": 0, "max": 1, "raw": 1.0 },
    "extensions": {
      "http://oli.cmu.edu/extensions/reward_source": "activity_attempt:full_credit"
    }
  },
  "context": {
    "extensions": {
      "http://oli.cmu.edu/extensions/project_id": 1001,
      "http://oli.cmu.edu/extensions/section_id": 2001,
      "http://oli.cmu.edu/extensions/publication_id": 3001,
      "http://oli.cmu.edu/extensions/activity_id": 606,
      "http://oli.cmu.edu/extensions/part_id": "part-1",
      "http://oli.cmu.edu/extensions/part_attempt_guid": "part-guid",
      "http://oli.cmu.edu/extensions/experiment_attributions": [
        {
          "role": "reward",
          "experiment_id": 101,
          "experiment_uuid": "11111111-2222-3333-4444-555555555555",
          "decision_point_id": 202,
          "condition_id": 303,
          "condition_code": "condition-a",
          "assignment_id": 404,
          "assignment_key": "101:202:505",
          "enrollment_id": 505,
          "algorithm": "thompson_sampling",
          "policy_version": "thompson_sampling:v2",
          "key": "reward-key",
          "reward_source": "activity_attempt:full_credit"
        }
      ]
    }
  }
}
```

### Page-view exposure

A page-view statement identifies the assigned condition that was actually rendered:

```json
{
  "http://oli.cmu.edu/extensions/experiment_attributions": [
    {
      "role": "exposure",
      "experiment_id": 101,
      "experiment_uuid": "11111111-2222-3333-4444-555555555555",
      "decision_point_id": 202,
      "decision_point_key": "worked-example-format",
      "condition_id": 303,
      "condition_code": "condition-a",
      "assignment_id": 404,
      "assignment_key": "101:202:505",
      "section_id": 2001,
      "enrollment_id": 505,
      "content_revision_id": 7001,
      "algorithm": "weighted_random",
      "policy_version": "weighted_random:v1",
      "key": "exposure:404:7001",
      "recorded_at": "2026-07-14T11:55:00Z"
    }
  ]
}
```

### Outcome and reward on one evaluated attempt

An evaluated attempt may carry both the measured outcome and the reward derived from it:

```json
{
  "http://oli.cmu.edu/extensions/experiment_attributions": [
    {
      "role": "outcome",
      "experiment_id": 101,
      "decision_point_id": 202,
      "condition_id": 303,
      "condition_code": "condition-a",
      "assignment_id": 404,
      "assignment_key": "101:202:505",
      "activity_attempt_id": 8001,
      "resource_attempt_id": 8002,
      "activity_resource_id": 606,
      "score": 1.0,
      "out_of": 1.0,
      "key": "outcome:8001:404"
    },
    {
      "role": "reward",
      "experiment_id": 101,
      "decision_point_id": 202,
      "condition_id": 303,
      "condition_code": "condition-a",
      "assignment_id": 404,
      "assignment_key": "101:202:505",
      "outcome_key": "outcome:8001:404",
      "reward_value": 1.0,
      "reward_source": "activity_attempt:full_credit",
      "algorithm": "thompson_sampling",
      "policy_version": "thompson_sampling:v2",
      "key": "reward:8001:404"
    }
  ]
}
```

### Multiple experiments on one rollup event

When a page or activity contains content attributed to multiple decision points, one rollup statement contains one entry per assignment. This example associates the same host event with two experiments:

```json
{
  "http://oli.cmu.edu/extensions/experiment_attributions": [
    {
      "role": "rollup",
      "experiment_id": 101,
      "experiment_uuid": "11111111-2222-3333-4444-555555555555",
      "decision_point_id": 202,
      "condition_id": 303,
      "condition_code": "condition-a",
      "assignment_id": 404,
      "assignment_key": "101:202:505",
      "enrollment_id": 505,
      "algorithm": "thompson_sampling",
      "policy_version": "thompson_sampling:v2",
      "key": "reward:8001:404"
    },
    {
      "role": "rollup",
      "experiment_id": 102,
      "experiment_uuid": "66666666-7777-8888-9999-000000000000",
      "decision_point_id": 212,
      "condition_id": 313,
      "condition_code": "condition-b",
      "assignment_id": 414,
      "assignment_key": "102:212:505",
      "enrollment_id": 505,
      "algorithm": "weighted_random",
      "policy_version": "weighted_random:v1",
      "key": "outcome:8001:414"
    }
  ]
}
```

### Media interaction with multiple attributions

A media event inside nested or overlapping experimental content retains its media host data and changes the attribution role to `media_interaction`:

```json
{
  "http://oli.cmu.edu/extensions/video_url": "https://media.example.edu/lesson-1.mp4",
  "http://oli.cmu.edu/extensions/video_time": 42.5,
  "http://oli.cmu.edu/extensions/experiment_attributions": [
    {
      "role": "media_interaction",
      "experiment_id": 101,
      "decision_point_id": 202,
      "condition_id": 303,
      "condition_code": "condition-a",
      "assignment_id": 404,
      "assignment_key": "101:202:505",
      "key": "exposure:404:7001"
    },
    {
      "role": "media_interaction",
      "experiment_id": 102,
      "decision_point_id": 212,
      "condition_id": 313,
      "condition_code": "condition-b",
      "assignment_id": 414,
      "assignment_key": "102:212:505",
      "key": "exposure:414:7001"
    }
  ]
}
```

The `role` describes how the host event participates in analysis (`assignment`, `exposure`, `outcome`, `reward`, `rollup`, or `media_interaction`). The attribution `key` provides event-level idempotency, while `assignment_key` links evidence to the learner's sticky assignment. ClickHouse expands every array entry into its own attribution row and links it to the raw host statement by event hash.

## MVP scope not included in this PR

The following work remains intentionally incomplete and will be delivered separately:

- **Built-in experiment analytics UI:** the dashboards and product-facing analytics defined in `docs/exec-plans/current/epics/ab_testing/analytics/` are not implemented. This PR provides the underlying xAPI, ClickHouse projection, and query foundation only.
- **Dataset creation support:** dataset generation has not yet been updated to expose the new experiment-related fields carried by xAPI statements. Raw statement emission, ETL extraction, ClickHouse attribution storage, and query support are present, but experiment-aware dataset creation remains follow-up work.
- **Legacy migration:** existing UpGrade experiment definitions, history, and learner assignments are intentionally not migrated.

## Verification

Implementation slices include focused ExUnit, LiveView, TypeScript, ETL, ClickHouse uploader/backfill, schema validation, and `Oli.Scenarios` coverage. Recorded verification includes:

- formatting and warnings-as-errors compilation;
- experiment context, policy, persistence, runtime, analytics-contract, privacy, and coupling tests;
- authoring and experiment-detail LiveView tests;
- delivery reward, fallback, section-participation, stickiness, and Thompson Sampling scenario tests;
- xAPI schema, local uploader, production ETL fixture, ClickHouse projection, and backfill-query tests;
- requirements traceability and work-item validation for the implemented epic slices.

## Rollout notes

- Native experiments are a hard cutover for newly authored experiments.
- Existing UpGrade-backed content is not migrated and remains read-only where surfaced.
- Obsolete UpGrade secrets/configuration can be removed after the cutover code is deployed.
- The ClickHouse migration and xAPI ETL changes must be deployed with the application changes before experiment-attributed analytics queries are used.
- This work includes new configurable env variables that allow email verification requirement to be disabled for author and user accounts upon creation. This is to help facilitate easier account creation on test instances only and defaults to enabled when not set.

There is a test instance deployed at https://ab-testing.plasma.oli.cmu.edu that is configured to use the XAPI clickhouse direct uploader and linked up to our staging clickhouse instance. You may want to use this instance for easier testing of this feature. An Andrew google account is required for access. For access to the staging clickhouse DB, please contact me.


lib/oli/experiments/schemas/experiment_definition.ex
```diff
@@ -0,0 +1,78 @@
...
+    field :description, :string
+    field :state, Ecto.Enum, values: @states, default: :draft
+    field :assignment_unit, Ecto.Enum, values: @assignment_units, default: :enrollment
+    field :algorithm, Ecto.Enum, values: @algorithms
+    field :policy_config, :map, default: %{}
+    field :started_at, :utc_datetime
+    field :ended_at, :utc_datetime
+
+    belongs_to :project, Project
+    belongs_to :section, Section
```
darrensiegel: This cardinality surprised me.  So an experiment can only apply to exactly one course section?

lib/oli/experiments/schemas/decision_point.ex
```diff
@@ -0,0 +1,49 @@
...
+  alias Oli.Resources.{Resource, Revision}
+
+  schema "experiment_decision_points" do
+    field :decision_point_key, :string
+    field :title, :string
+    field :position, :integer, default: 0
+
+    belongs_to :experiment, ExperimentDefinition
+    belongs_to :alternatives_resource, Resource
+    belongs_to :alternatives_revision, Revision
```
darrensiegel: I would be interested in understanding why we pin a specific revision here?

lib/oli/experiments.ex
```diff
@@ -0,0 +1,2758 @@
+defmodule Oli.Experiments do
+  @moduledoc """
```
darrensiegel, line 2: In looking at this, I immediately got the sense that something wasn't right with how revisions were being accessed.  This impl seems to side-step the publication model (i.e. it makes no use of Resolvers).  I posed this question to Codex and it confirmed: 

 Yes—there is a real architectural flaw, but it is more precise than “all direct Revision access is wrong.”

  The core problem is that Oli.Experiments mixes three different meanings of a revision:

  1. The currently editable authoring revision.
  2. The revision deployed to a particular section.
  3. An immutable historical revision recorded as evidence.

  Only the third should normally be accessed by exact revision ID. The first two should be resolved through the publication model.

  ## Main findings

  ### 1. Authoring candidate discovery bypasses the working publication

  lib/oli/experiments.ex:203 joins Revision directly to ProjectResource. ProjectResource establishes that a resource belongs to a project; it does not identify the revision currently
  mapped by the project’s working publication.

  By contrast, lib/oli/publishing/authoring_resolver.ex:48 resolves through:

  project → working publication → published_resources → revision

  The experiment query effectively does:

  project → project_resources → every historical revision

  The subsequent ordering and Enum.uniq_by/2 do not repair this. In fact, ordering first by title means it does not reliably select the highest revision ID when titles have changed.

  Consequences include:

  - Displaying obsolete alternatives groups.
  - Displaying option definitions that are no longer current.
  - Selecting a revision that is not present in the working publication.
  - Increasing query and memory cost as revision history grows.

  This is a definite publication-model violation.

  ### 2. Authoring validation proves resource membership, not revision membership

  lib/oli/experiments.ex:2181 verifies:

  - The revision exists.
  - It is an alternatives revision.
  - Its resource_id matches the request.
  - The resource belongs to the project.

  It does not prove that the revision is the one mapped in that project’s working publication. Therefore a stale revision—or potentially a revision associated with a shared resource
  through another project/publication history—can be accepted.

  The authoritative authoring question should be:

  > What revision does this resource resolve to in this project’s working publication?

  The current code instead asks:

  > Does this exact revision exist, and does its resource belong to this project?

  Those are materially different guarantees.

  ### 3. Activation validates stale content while runtime deliberately floats forward

  Activation reads the stored revision directly in lib/oli/experiments.ex:1810. If the alternatives resource was edited after experiment creation, activation still validates the
  experiment against the old revision.

  Runtime does the opposite. lib/oli/experiments.ex:856 ignores request.alternatives_revision_id and matches only:

  - project
  - optional section
  - alternatives resource
  - decision-point key

  The test test/oli/experiments/runtime_test.exs:53 confirms that floating across revisions is intentional.

  That creates an inconsistent lifecycle:

  Creation/activation validation → stored historical revision
  Delivery execution           → currently deployed revision

  The experiment is therefore certified against one piece of content but may execute against another.

  Compatible edits happen to work. Incompatible edits can produce less obvious behavior:

  - If no condition codes overlap, delivery falls back.
  - If only some codes overlap, the experiment may continue with a reduced condition set.
  - A changed option meaning that retains the same ID may silently receive the old condition’s treatment.

  This is the most important architectural inconsistency.

  ### 4. Delivery obtains the right revision—but Oli.Experiments trusts the caller

  The rendering path is publication-aware:

  - lib/oli/publishing/delivery_resolver.ex:32 resolves through the section’s SectionsProjectsPublications mapping.
  - lib/oli/resources.ex:402 converts those resolved revisions into decision-point data.
  - lib/oli/resources/alternatives/decision_point_strategy.ex:119 passes the resolved resource ID, revision ID, and options into Oli.Experiments.

  So the normal UI/runtime caller supplies correct information. But Oli.Experiments is documented as the public context boundary and does not independently verify it.

  In particular:

  - validate_publication/1 only verifies that the publication belongs to the project; it does not verify that the section is currently deployed to it.
  - active_experiment_match/2 does not constrain by publication or resolved revision.
  - record_exposure/1 accepts an arbitrary existing content_revision_id without proving it belongs to the decision-point resource or supplied publication.

  That is primarily an integrity and tenancy-boundary weakness rather than an immediate exploit in the current rendering path.

  ### 5. Section-scoped authoring is also ambiguous

  list_available_decision_points/1 accepts a scope containing a section, but still scans project revision history. For a section-scoped experiment, there are two plausible authorities:

  - The project’s working revision via AuthoringResolver.
  - The section’s deployed revision via DeliveryResolver.

  The current implementation uses neither. This matters when authoring has moved ahead of the section’s deployed publication.

  ### 6. Historical revision IDs are appropriate for exposures

  lib/oli/experiments.ex:1223 should remain an exact revision reference conceptually. An exposure is historical evidence: it must record what the learner actually saw and must not
  later resolve to a newer revision.

  The issue is not storing that ID. The issue is failing to verify at ingestion that it is the revision resolved for the relevant section/publication/resource.

  The same distinction applies to DecisionPoint.alternatives_revision_id: it can legitimately record “the revision this experiment was configured against,” but it should not
  simultaneously act as the current authoring or delivery authority.

  ## Architectural conclusion

  The implementation currently combines two incompatible models:

  - A pinned experiment, suggested by storing and validating alternatives_revision_id.
  - A resource-level floating experiment, demonstrated by runtime matching across compatible revisions.

  The existing runtime test indicates that the intended model is probably resource-level floating. Under that model:

  - alternatives_resource_id is the durable experiment target.
  - alternatives_revision_id is configuration provenance, not the current authority.
  - Authoring operations should use AuthoringResolver.
  - Delivery operations should use DeliveryResolver or equivalent publication-scoped queries.
  - Compatibility with experiment conditions must be checked against the resolved revision that will actually execute.
  - Exposure records should pin the exact delivered revision after validating its publication/resource relationship.

  So your concern is correct: lib/oli/experiments.ex bypasses the publication model in its authoring discovery and validation, and it applies inconsistent revision semantics between
  activation and delivery. The direct revision access used for historical exposure provenance is appropriate; the direct access used to determine current authoring or delivery content
  is not.

darrensiegel (changes requested):

lib/oli/experiments/schemas/experiment_definition.ex
eliknebel: Experiment section relationship is now many-to-many and modeled by experiment_sections table

lib/oli/experiments/schemas/decision_point.ex
eliknebel: specific revision pin removed, resource revision is now loaded according to the publication model

lib/oli/experiments.ex
eliknebel: revisions are now resolved properly using the publication model

darrensiegel:
I've done a high level review and exploration of the implementation (aided a lot by Codex in a Q&A interaction, which really finally helped me fully understand the "adaptive" nature of the experimentation platform)

My main concerns are around the "reward" implementation.  

1. There is a **synchronous** call to `Oli.Delivery.Experiments.RewardHandoff.record_evaluated_activity` from within the evaluated activity RollUp.   I'm concerned about the impact here on course sections that both DO and DO NOT have experiments, and more so about the latter.   Following the logic and control flow here, I see at least 5 serial DB reads just from within `Oli.Experiments.validate_delivery_participation_scope`:

```
  defp validate_delivery_participation_scope(%Scope{} = scope) do
    with {:ok, scope} <- validate_institution(scope),
         {:ok, scope} <- validate_project(scope),
         {:ok, scope} <- validate_participation_section(scope),
         {:ok, scope} <- validate_user(scope),
         {:ok, scope} <- validate_enrollment(scope) do
      {:ok, scope}
    end
  end
```

Each of those `validate_*` functions can make a serial DB read.  Some of them seemingly add no real value (like the Institution one).   This all happens for every single project every single attempt evaluation, regardless of whether there is or is not an experiment present in the project / section. 

## My suggestion here:
- Create a cheap and fast way to determine if a section HAS or does NOT HAVE an experiment present.  This could just be an EXISTS style query across `ExperimentSection` looking for any record for that course section id.  Maybe call it "has_experiment?" and hang it off of `Oli.Sections`?    Make sure an index exists on section_id in that table (I think it does). 
- Use that `Oli.Sections.has_experiment?` to guard the execution of the entire reward impl in the roll up. 
- Make the execution of the reward impl from the RollUp async by using the `Oli.Delivery.Experiments.RewardHandoffWorker` 
- Streamline, if not eliminate, the `validate_delivery_participation_scope` impl to reduce as much as possible these serial reads 


2. The "bulk reward" implementation is asynchronous and launches from the Oban worker defined in `Oli.Delivery.Experiments.RewardHandoffWorker`.   That implementation ultimately is just an `Enum.reduce` over each activity attempt id from the collection.  For each of those, we repeatedly fetch a lot of the same information (driven from the multi-join query here:  

```
defp load_evaluated_attempt_contexts(activity_attempt_ids) do
    from(activity_attempt in ActivityAttempt,
      join: resource_attempt in ResourceAttempt,
      on: resource_attempt.id == activity_attempt.resource_attempt_id,
      join: resource_access in ResourceAccess,
      on: resource_access.id == resource_attempt.resource_access_id,
      join: section in Section,
      on: section.id == resource_access.section_id,
      left_join: spp in SectionsProjectsPublications,
      on: spp.section_id == section.id and spp.project_id == section.base_project_id,
      left_join: published_resource in PublishedResource,
      on:
        published_resource.publication_id == spp.publication_id and
          published_resource.resource_id == resource_access.resource_id,
      where: activity_attempt.id in ^activity_attempt_ids,
      select: %{
        activity_attempt: activity_attempt,
        page_content: resource_attempt.content,
        scope: %Scope{
          institution_id: section.institution_id,
          project_id: section.base_project_id,
          publication_id: spp.publication_id,
          section_id: section.id,
          user_id: resource_access.user_id,
          enrollment_id:
            fragment(
              "(SELECT e.id FROM enrollments e WHERE e.section_id = ? AND e.user_id = ? LIMIT 1)",
              section.id,
              resource_access.user_id
            )
        },
        published_revision_id: published_resource.revision_id
      }
    )
    |> Repo.all()
  end
```

## My suggestion here: 

Rework the `Oli.Delivery.Experiments.RewardHandoffWorker` so that it issues EXACTLY TWO queries regardless of the number of activity attempt ids present.  The first query would be a BULK fetch of all needed activity attempts.  The second is then a query for the CONTEXT (page content, scope). **

The net result here will be a reworked `Oli.Delivery.Experiments.RewardHandoffWorker` impl that supports TWO paths:  a single attempt id path and an optimized path for a collection of attempt ids.

darrensiegel (approved):

PR #6726, opened 2026-07-13
[ENHANCEMENT] [MER-5500] Add paywall and payment-provider smoke coverage

nicocirio:
## Jira

- Ticket: https://eliterate.atlassian.net/browse/MER-5500
- Related: https://eliterate.atlassian.net/browse/MER-5498

## Summary

- add scenario-driven Playwright coverage for paid-section paywall behavior and payment unlock flows
- extend scenario support so paid-course setup can be expressed from product defaults, section overrides, and institution-qualified pricing
- add test-only bootstrap support for guest enrollment and pending provider payments used by the smoke suite

## Improvement

- covers both `MER-5498` and `MER-5500` in one lane because they share the same project -> product -> section bootstrap and learner-facing payment path
- adds isolated scenario files plus Playwright specs for no-grace, active-grace, expired-grace, guest, discount, payment-code, Stripe, and Cashnet cases
- documents the delivered DSL/hook split and stabilizes unlock specs by seeding an isolated world per test

## Validation

- `python3 /Users/nicocirio/.local/share/harness/skills/validate/scripts/validate_work_item.py docs/exec-plans/current/epics/automated_testing/paywall-and-provider-smoke --check all`
- `mix test test/scenarios/validation test/oli/scenarios`
- `npx playwright test tests/torus/student_payment --reporter=line`

## Demo
Screenshot 2026-07-13 at 6 19 59 PM


nicocirio:
Follow-up on the AI review comments:

- I addressed the actionable hardening items in the latest push:
  - `PlaywrightSessionController` now accepts the scenario token only via header, compares it with `Plug.Crypto.secure_compare/2`, and rejects ambiguous redirect targets more strictly.
  - `DirectiveParser` now validates atom inputs for `payment_options`, `grace_period_strategy`, and institution discount `type` against explicit allowlists instead of accepting arbitrary atoms.

- I did **not** move `GET /test/log_in_user` under a strict `MIX_ENV=test` guard in this PR. Our Playwright automation runs against a dev-like runtime rather than a `MIX_ENV=test` server, so making that route test-env-only would break the current automation model.

- That said, the route is still gated today behind `Application.compile_env(:oli, :enable_playwright_scenarios, false)` and also requires the shared Playwright scenario token to be present in the request header before it will create a session. We can revisit a stricter environment-level restriction separately if we decide to change how PW environments are provisioned.

darrensiegel (approved):

PR #6727, opened 2026-07-13
[BUG FIX] [MER-5625] Preserve preview customization UI state across updates (rework)

andersweinstein:
## Issue

  MER-5625 made the learning-objective and available-points summaries update after instructors remove or restore questions. Those LiveView updates can transiently reconnect the custom elements rendering the assessment content. This was causing presentation errors due to use of stale state after sequences of actions like the following:

  1. Remove several activity-bank selections.
  2. Dismiss the success banner.
  3. Restore one selection.

  The other removed selections incorrectly appear included: their buttons change to Remove, and their Removed pill and red rail disappear. Reloading shows that persistence was correct; only the client presentation had been reset.

  ## Root cause

  Each bank-selection React root independently stored its remove/restore state. When LiveView updated the new counters, the custom elements could reconnect and render again using their original server payload.

  That payload represented the state when the page initially loaded. It did not contain removals performed since then. Re-rendering therefore reset every component to stale initial state, after which the restore response updated only the selected item.

  React-backed embedded activities used the same general local-state pattern and were vulnerable to the same failure.

  ## Why state management was reworked

  A narrow fix could prevent one custom element from unmounting or teach one component to ignore repeated props. That would fix the reported bank-selection sequence, but it would leave state ownership distributed across independent React roots and preserve the same vulnerability for embedded activities and other preview surfaces.

  The implementation instead introduces a small page-scoped semantic store shared by:

  - Activity-bank selections
  - Embedded activities
  - Bank candidates
  - Server-rendered activity fallbacks

  The store owns only semantic interaction state: included or removed, pending action, whether toggling is currently allowed, and the bank’s available count. Buttons, rails, pills, backgrounds, and accessibility attributes are derived from that state.

  Initial server payloads seed targets only once. LiveView replies update the store, while reconnecting components subscribe to the existing state instead of overwriting it with stale payloads. Reloading still reconstructs the store from authoritative persisted server state.

  This is broader than the minimum symptom fix, but it removes the shared failure mechanism and gives all remove/restore presentations one consistent source of truth.
  
  ## Regression coverage

  - Added a bank-selection test reproducing the failed sequence: remove multiple selections, rerender from stale initial payloads, restore one, and verify the others retain their Restore buttons, Removed state, and red rails.

  - Added a custom-element reconnect test confirming bank-selection state survives transient disconnect/reconnect cycles.
  - Added an equivalent embedded-activity test confirming multiple removals survive stale preview-context rerenders and restoring one activity does not reset another.

  - Added coverage for bank-candidate and server-rendered fallback variants, including their distinct visual and disabled-state behavior.

  The relevant tests are in:

  - `assets/test/activities/preview/activity_bank_selection_preview_test.tsx`
  - `assets/test/activities/preview/preview_foundation_test.tsx`

  The embedded-activity regression test is named:

  `preserves each embedded activity removal through stale preview-context rerenders`

  ## Verification

  - 25 instructor-preview Jest tests pass.
  - TypeScript, ESLint, and Prettier checks pass.
  - The reported browser sequence passes, and a subsequent reload confirms the same persisted state.
  
  ## Demo Video

https://github.com/user-attachments/assets/50a2e210-1193-4fc8-aa2a-2353bf4d9ced

### DIFF
--- a/assets/src/components/activities/common/preview/ActivityPreviewCard.tsx
+++ b/assets/src/components/activities/common/preview/ActivityPreviewCard.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import { PreviewContext } from 'components/activities/types';
+import { usePreviewCustomizationState } from 'components/instructor_preview/preview_customization_store';
 import { LearningObjectiveList } from './LearningObjectiveList';
 import { PreviewDetailsToggle } from './PreviewDetailsToggle';
 import { PreviewHeader } from './PreviewHeader';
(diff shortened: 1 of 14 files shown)


andersweinstein:
[codex:]

**Reply to AI UI comment: “Async updates have no completion or failure announcement”**

The polite live region is intentionally scoped to the pending state. Completion is reflected through the updated action label and semantic card state, along with the existing server-authored flash. Actual mutation failures render their specific message through `preview_flash`, which uses `role="alert"`. Also, `ok: false` includes non-error outcomes such as confirmation-required and modal dismissal, so it cannot reliably drive a generic failure announcement.

**Reply to AI TypeScript comment: “Page-backed stores survive hook teardown”**

Preserving the host-backed store across hook teardown is intentional. LiveView can transiently destroy and remount this hook while retaining the same preview page element; clearing the host store there would recreate the stale-state regression this PR fixes. The store is attached to `#instructor-preview-lesson`, so genuine page removal makes it unreachable with the DOM node, while a different page resource ID replaces it on lookup. Only fallback stores live in a global map and require explicit teardown cleanup, which `clearFallbackPreviewCustomizationStore` already performs.

darrensiegel (approved):

PR #6730, opened 2026-07-16
[BUG FIX] [MER-4525] [MER-5724] Ensure activity authoring UI non-editable in read-only contexts

andersweinstein:
## Summary

Fixes activity authoring UI that remained editable when rendered with `editMode=false`.

This showed up in two places:

- **Activity Bank locked mode**, where locked activities should render as read-only. [MER-4525]
- **Instructor Preview,** where activities without custom preview components (e.g. SingleResponse) fall back to using authoring components, similarly exposing edit affordances where should render read-only. [MER-5724], which notes Hints and Explanation tabs showing in a SingleResponse, though there are other possible instances.

The fix covers two distinct kinds of editability:

- React activity authoring controls that were not receiving or honoring `editMode=false`.
- Slate-embedded authoring affordances, where settings/edit modals could be reached through hover toolbars or direct clicks on embedded elements such as images, popups, links, formulas.

## Details

- Propagates and honors `editMode` across common activity authoring surfaces, including stems, hints, explanations, answer keys, response/feedback editors, targeted feedback, and activity-specific controls.
- Keeps existing activity-specific conditions intact, including Image Coding's distinction between image-evaluated and regex-evaluated problems, while also applying read-only state.
- Adds centralized Slate read-only protection:
  - `HoverContainer` does not open authoring popovers when Slate is read-only.
  - `Toolbar` renders nothing when Slate is read-only.
  - The Slate `Editor` suppresses read-only clicks/mousedowns on authoring controls such as buttons, form controls, formula spans, and resize handles.
- Allows native audio/video playback controls to remain interactive in read-only Slate, since playback is not an authoring edit.
- Preserves locked Activity Bank visual parity for choice authoring where practical: drag handles and delete/add controls remain visible but disabled, rather than disappearing.
- Prevents disabled answer-key radio/checkbox icons from visually toggling when clicked.
- Disables formula click affordances/cursor styling in read-only Slate.
- Adds keyboard reorder protection for disabled draggable choice/order controls.
- Adds targeted regression coverage for read-only authoring controls and disabled answer-key choice controls.

## Verification

- `yarn test test/activities/readonly_authoring_test.tsx --runInBand`
- `yarn test test/components/editing/readonly_slate_elements_test.tsx test/activities/readonly_authoring_test.tsx --runInBand`
- `yarn test test/activities/choices_delivery_test.tsx test/activities/readonly_authoring_test.tsx --runInBand`
  - `yarn check-types`

## Sample Demo

https://github.com/user-attachments/assets/2569f967-5461-4e8e-be94-152c607fab68

[MER-4525]: https://eliterate.atlassian.net/browse/MER-4525?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MER-5724]: https://eliterate.atlassian.net/browse/MER-5724?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/assets/src/components/activities/check_all_that_apply/CheckAllThatApplyAuthoring.tsx
+++ b/assets/src/components/activities/check_all_that_apply/CheckAllThatApplyAuthoring.tsx
@@ -134,6 +134,7 @@ const CheckAllThatApply = () => {
     projectSlug: projectSlug,
   });
   const isInstructorPreview = mode === 'instructor_preview';
+  const readOnly = !editMode || isInstructorPreview;
 
   return (
     <ControlledTabs isInstructorPreview={isInstructorPreview}>
@@ -174,7 +175,7 @@ const CheckAllThatApply = () => {
           onSelect={(id) => dispatch(CATAActions.toggleChoiceCorrectness(id))}
           isEvaluated={false}
           context={writerContext}
-          disabled={isInstructorPreview}
+          disabled={readOnly}
           multiSelect
         />
         <SimpleFeedback partId={model.authoring.parts[0].id} />
@@ -194,7 +195,7 @@ const CheckAllThatApply = () => {
           addTargetedResponse={() => dispatch(CATAActions.addTargetedFeedback())}
           unselectedIcon={<Checkbox.Unchecked />}
           selectedIcon={<Checkbox.Checked />}
-          disabled={isInstructorPreview}
+          disabled={readOnly}
           multiSelect
         />
       </TabbedNavigation.Tab>
--- a/assets/src/components/activities/common/authoring/RemoveButton.tsx
+++ b/assets/src/components/activities/common/authoring/RemoveButton.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import { useAuthoringElementContext } from 'components/activities/AuthoringElementProvider';
 import { AuthoringButton } from 'components/activities/common/authoring/AuthoringButton';
+import { classNames } from 'utils/classNames';
 
 type Props = {
   onClick: () => void;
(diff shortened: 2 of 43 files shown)


andersweinstein:
[Codex:] Responding to the duplicated AI review comments on ordering drag behavior and Slate read-only keyboard semantics.

**Ordering disabled drag behavior:** this appears to be based on the `Draggable.Column` call site but misses how this helper is structured. `Draggable.Column` does not currently expose an `isDragDisabled` prop; disabling is applied on each `Draggable.Item`. `ResponseChoices` passes `isDragDisabled={disabled ?? false}` to both `Draggable.Item` and `Draggable.DragIndicator`, and `DraggableColumn` also suppresses the keyboard reorder path when `isDragDisabled` is set. The guarded `setItems` callback is retained as a defensive backstop, not as the only protection.

**Slate read-only keyboard semantics:** agreed that the central Slate mouse/click suppression is not a complete semantic disabling mechanism for arbitrary embedded form controls. In this PR, the intended primary fix is to propagate `editMode=false` into the known activity authoring components so their own controls render disabled/read-only. The Slate-level capture is a fallback for our embedded authoring affordances that open settings/edit UI through clicks or hover toolbars, while intentionally preserving non-editing interactions such as audio/video playback. If we identify a concrete Torus embedded element that remains keyboard-operable and mutates content in read-only mode, I would handle that at the element renderer/component boundary rather than broadening Slate into a general form-control disabler.

darrensiegel (approved):

PR #6740, opened 2026-07-21
[FEATURE] [MER-5796] Remix templates support

darrensiegel:
Implemented support for remixing from product/template sources in course remix, with focused handling for admin-authored enrollable sections and hidden-instructor sessions.

### High-level approach:

- Extended remix source resolution to include product/template sources alongside project sources.
- Added product-source visibility rules so regular instructors respect community membership, while hidden instructors used for admin-authored enrollable sections can access the relevant community-visible templates.
- Adjusted Remix LiveView identity selection so enrollable section remix prioritizes the hidden instructor path, while preserving author-admin behavior for template/product remix.
- Added product-source selection handling that resolves selected materials through the product’s pinned publications.
- Used SectionResourceDepot for product/template hierarchy visibility so hidden section resources and descendants are filtered consistently.
- Added scenario DSL support for source_product so integration scenarios can cover product/template remix flows.

### Fixed review issues:
- telemetry events now execute synchronously instead of spawning unbounded tasks;
- product-source preselection compares existing materials by resource_id when pinned publication ids differ;
- All Pages sorting now supports publication_date for both project and product sources.

Validation includes focused unit, LiveView, and scenario coverage for source visibility, admin/hidden-instructor remix access, product-source hierarchy filtering, material selection, pinned-publication edge cases, and All Pages sorting.


priv/schemas/v0-1-0/scenario.schema.json
```diff
@@ -1568,6 +1568,14 @@
                   "from": {
```
gastonabella, line 1568: Could we clean up this schema hunk? It looks like `from_product` and `user` were added to the `move` operation, and then duplicated in the same object. I think these belong only on the top-level `remix` directive schema below, so removing them here would keep the scenario schema aligned with the parser and avoid confusing generated docs/validation.

gastonabella (approved):

PR #6741, opened 2026-07-22
[ENHANCEMENT] [MER-5416] Expand MIXED workflow coverage Part 2/3

nicocirio:
## Jira

- Ticket: https://eliterate.atlassian.net/browse/MER-5416

## Summary

Expands the Playwright plus scenario workflow coverage for the MIXED regression matrix.

This was initially planned as the second implementation part of the work. We decided to split it into Part 2/3 and Part 3/3 because including all remaining matrix rows in one PR would make the review surface and CI diagnosis unnecessarily large.

Part 2/3 adds coverage for:

- CORE text editing.
- INLINE formatting, internal and external links, foreign text, popup content, and inline callouts.
- LIST indentation and bullet styles.
- TABLE structure, cell formatting, alternating rows, and hidden borders.
- IMAGE insertion and replacement, caption, alt text, and width settings.
- FIGURE title and nested content.

The remaining Part 3/3 coverage will address:

- YouTube, video, webpage, and formula workflows.
- CodeBlock follow-up behaviors.
- Definition, dialog, conjugation, description-list, and theorem workflows.
- The two matrix rows currently marked needs-triage.

The [MIXED coverage matrix](https://github.com/Simon-Initiative/oli-torus/blob/MER-5416-automated-basic-page-authored-PART-2/docs/exec-plans/current/epics/automated_testing/mer-5416/mixed_coverage_matrix.md) tracks each spreadsheet row, its Editing/Persisted/Preview/Delivery status, and the associated workflow test.

## Technical Notes

- Keeps the scenario to Playwright authoring to scenario assertions pattern established in Part 1.
- Adds focused author-preview and published-delivery assertions for each covered workflow group.
- Cleans up obsolete grouped helpers and makes image attribute assertions explicit.

## Validation

- `npm run pw -- tests/torus/course_authoring/mixed-workflow.spec.ts --reporter=line`
- Targeted IMAGE-B/C/D/E/F validation passed after the cleanup.
- `mix format`
- `mix compile`

## Evidence
Screenshot 2026-07-22 at 12 33 20 PM

### DIFF
--- a/assets/automation/tests/torus/course_authoring/mixed_workflow/actions.ts
+++ b/assets/automation/tests/torus/course_authoring/mixed_workflow/actions.ts
@@ -138,8 +138,16 @@ export const mixedWorkflowActions: WorkflowActionRegistry = {
     await page.goto(editorPath(projectSlug, pageRevisionSlug), { waitUntil: 'load' });
 
     await insertBlockImage(page);
-    await uploadImage(page, path.resolve(process.cwd(), 'tests/resources/media_files', jpgName), jpgName);
-    await uploadImage(page, path.resolve(process.cwd(), 'tests/torus/student_delivery/support', pngName), pngName);
+    await uploadImage(
+      page,
+      path.resolve(process.cwd(), 'tests/resources/media_files', jpgName),
+      jpgName,
+    );
+    await uploadImage(
+      page,
+      path.resolve(process.cwd(), 'tests/torus/student_delivery/support', pngName),
+      pngName,
+    );
     await page.locator('.name').getByText(jpgName, { exact: true }).click();
     await page.getByRole('button', { name: 'Select', exact: true }).click();
     await expect(page.locator('[data-slate-editor="true"] img')).toHaveAttribute(
@@ -210,33 +218,6 @@ export const mixedWorkflowActions: WorkflowActionRegistry = {
     };
   },
 
-  async author_inline_embeds({ curriculumTask, homeTask, page }, params) {
-    const projectSlug = asString(params.project_slug, 'project_slug');
-    const pageRevisionSlug = asString(params.page_revision_slug, 'page_revision_slug');
-    const foreignText = 'INLINE-N foreign text';
-    const popupTrigger = 'INLINE-O popup trigger';
-    const popupContent = 'INLINE-O popup content';
-    const calloutText = 'INLINE-R inline callout text';
-
-    await test.step('author foreign text, popup content, and inline callout', async () => {
-      await homeTask.login('author');
-      await page.goto(editorPath(projectSlug, pageRevisionSlug), { waitUntil: 'load' });
-
-      await curriculumTask.addForeignToolbar(foreignText, 'arabic', false);
-      await curriculumTask.addPopUpToolbar(popupTrigger, popupContent, false);
-      await curriculumTask.addCalloutToolbar(calloutText, false);
-      await previewFlush(() => curriculumTask.openPreview());
-    });
-
-    return {
-      callout_text: calloutText,
-      foreign_text: foreignText,
-      page_revision_slug: pageRevisionSlug,
-      popup_content: popupContent,
-      popup_trigger: popupTrigger,
-    };
-  },
-
   async author_inline_external_link({ curriculumTask, homeTask, page }, params) {
     const projectSlug = asString(params.project_slug, 'project_slug');
     const pageRevisionSlug = asString(params.page_revision_slug, 'page_revision_slug');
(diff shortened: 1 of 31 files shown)


assets/automation/tests/torus/course_authoring/mixed_workflow/actions.ts
```diff
@@ -16,6 +18,343 @@ const previewFlush = async (openPreview: () => Promise<{ close(): Promise<void>
...
+    const pageRevisionSlug = asString(params.page_revision_slug, 'page_revision_slug');
+    const typedText = 'Mixed workflow typing stays responsive across a couple of sentences.';
+    const editor = page.locator('[data-slate-editor="true"]');
+
+    await test.step('sign in and type the CORE-A text without editor lag', async () => {
+      await homeTask.login('author');
+      await page.goto(editorPath(projectSlug, pageRevisionSlug), { waitUntil: 'load' });
+
+      await editor.click();
+      await page.keyboard.press('Meta+A');
```
gastonabella: Could we use `ControlOrMeta+A` here instead of `Meta+A`? Playwright runs in CI on non-macOS environments, and `Meta+A` may not select all text there. Using the cross-platform shortcut would make CORE-A less dependent on the runner OS.

lib/oli/scenarios/features/mixed_content_hooks.ex
```diff
@@ -86,6 +98,361 @@ defmodule Oli.Scenarios.Features.MixedContentHooks do
...
+  @doc """
+  Asserts that an inline callout renders in author preview.
+  """
+  def assert_author_preview_inline_callout(%ExecutionState{} = state),
+    do: assert_preview_text(state, "EXPECTED_TEXT")
+
+  @doc """
+  Asserts that inline foreign text persists in delivery content.
+  """
+  def assert_student_delivery_inline_foreign(%ExecutionState{} = state),
```
gastonabella: Could we assert the selected foreign language as well as the text? The Playwright action selects Arabic for INLINE-N, but the scenario hook only checks that a `foreign` element contains the expected text. If the `lang` attribute stopped persisting or rendering, this row would still pass even though the spreadsheet behavior specifically includes selecting a language.

lib/oli/scenarios/features/mixed_content_hooks.ex
```diff
@@ -86,6 +98,361 @@ defmodule Oli.Scenarios.Features.MixedContentHooks do
...
+           |> length() >= 3,
+           "Expected author-preview table to contain the added column"
+
+    state
+  end
+
+  @doc """
+  Asserts that table structure and cell formatting persist in delivery content.
+  """
+  def assert_student_delivery_table_structure(%ExecutionState{} = state) do
```
gastonabella, line 327: Could we add a delivery-content assertion for the added column in TABLE-B? The author-preview hook checks that the first row has at least three cells, but `assert_student_delivery_table_structure/1` only checks row count, header, colspan, and alignment. A delivery serialization regression that dropped the inserted column could still pass this coverage row.

gastonabella (commented):
I ran the targeted MIXED workflow suite locally and got 12 passing tests, with the same two workflows failing: `TABLE-G/H` and `IMAGE-B/C/D/E/F`.

For `TABLE-G/H`, the failure looks a bit timing/state-sensitive. In one run it timed out waiting for the `Insert Table` button during the repeated table insertion flow, while in another run the workflow reached the scenario assertions but did not find the hidden-border table in preview/delivery. Could we take another look at the focus/menu handling around the repeated table insertion and hidden-border selection? It may just need a little more synchronization to be reliable as regression coverage.

For `IMAGE-B/C/D/E/F`, the upload step did not show `img-mock-05-16-2025.jpg` in the media manager after `setFiles`; in one run the UI showed `Error: Could not upload file.` Could we clarify whether there is a local setup prerequisite for media uploads, or make the workflow surface that prerequisite more clearly? Right now the targeted command from the PR body does not pass locally for me.

nicocirio:
Addressed the review comments — CORE-A now uses `ControlOrMeta+A`, INLINE-N's delivery check also asserts the persisted `lang`, and TABLE-B's delivery check now covers the added column. For the flaky failures reported here, here's a summary:

| Test | Failure | Fix | Validated |
|---|---|---|---|
| `TABLE-B/C/D/E/F` | `Insert...` button timeout | Retry with an actual selection change instead of a single click | 8/8 repeated runs + full spec |
| `TABLE-G/H` | Same timeout, plus hidden-border table missing from assertions | Same fix, plus retry the full click sequence until the style is confirmed | 10/10 repeated runs + full spec |
| `IMAGE-B/C/D/E/F` | Ambiguous timeout waiting for uploaded filename | Longer wait + fails fast with the app's real error if upload banner appears | 8/8 with minio running locally |

gastonabella (approved):

PR #6743, opened 2026-07-23
[BUG FIX] [MER-5836] Fix admin/external tools view timeout

andersweinstein:
## Summary

  Improves the performance and resilience of the admin LTI External Tools page, which could load very slowly or return a 504 Gateway Timeout.

  ## Root cause

  The page calculated section-usage counts by querying every matching Section record, selecting all section columns, transferring those records into Elixir, and then grouping and counting them in application code.

  The same expensive query was also executed twice during the initial static LiveView request: once from mount/3 and again from handle_params/3.

  A [production AppSignal sample](https://appsignal.com/open-learning-initiative/sites/618539ab2cf81d7e3cd051ca/performance/incidents/960/samples/618539ab2cf81d7e3cd051ca-149321194841248426331784752380) confirmed this behavior:

  - Total request time: 42.13 seconds
  - First execution: 28.15 seconds
  - Second execution: 13.94 seconds
  - Combined database time: 42.09 seconds

  Query performance varied substantially with database load, cached data, planner statistics, and the volume of delivery resources. In some cases, a single execution ran longer than the gateway allowed the HTTP request to remain open, resulting in a 504 response.

  ## Changes

  - Replaced loading complete Section records with a PostgreSQL COUNT(DISTINCT section_id) grouped by external tool.
  - Removed now-unnecessary joins to sections and activity_registrations, which were previously needed only to materialize complete activity-registration and section records.
  - Removed the duplicate query execution between mount/3 and handle_params/3.
  - Moved usage-count loading into an asynchronous LiveView task.
  - Rendered the page immediately with `Calculating...` while counts load.
  - Displayed `Unknown` if the optional count query fails, allowing the rest of the page to remain usable.
  - Cancelled stale count tasks when pagination, filtering, or sorting changes.
  - Instrumented the asynchronous operation as an AppSignal background action so its Ecto query timing remains observable. The AppSignal action is named:

  `OliWeb.Admin.ExternalTools.ExternalToolsView#load_usage_counts`

  ## Verification

  - Added context-level coverage for PostgreSQL usage-count aggregation.
  - Added LiveView coverage for asynchronous loading and failure handling.
  - Focused context and LiveView tests pass.
  - Confirmed formatting and git diff --check.

### DIFF
--- a/lib/oli/lti/platform_external_tools.ex
+++ b/lib/oli/lti/platform_external_tools.ex
@@ -236,30 +236,7 @@ defmodule Oli.Lti.PlatformExternalTools do
           query
       end
 
-    platform_instances = Repo.all(query)
-
-    platform_instances_usage_count =
-      platform_instances
-      |> Enum.map(& &1.id)
-      |> get_sections_grouped_by_platform_instance_ids()
-      |> Enum.map(fn {id, sections} -> {id, length(sections)} end)
-      |> Enum.into(%{})
-
-    platform_instances =
-      platform_instances
-      |> Enum.map(fn p ->
-        Map.put(
-          p,
-          :usage_count,
-          Map.get(platform_instances_usage_count, p.id, 0)
-        )
-      end)
-
-    if field == :usage_count do
-      Enum.sort_by(platform_instances, & &1.usage_count, direction)
-    else
-      platform_instances
-    end
+    Repo.all(query)
   end
 
   @doc """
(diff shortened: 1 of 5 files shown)


eliknebel (approved):
Changes look good. Tested page and still loads as expected

nicocirio:
I agree, also tested it locally, and it worked as expected. If the query is still expensive in Stellarator, I would consider adding a composite index on `section_resources(activity_type_id, section_id)` and validating the plan with `EXPLAIN ANALYZE`, since the count filters by `activity_type_id` and uses `COUNT(DISTINCT section_id)`.

PR #6750, opened 2026-07-30
[ENHANCEMENT] [MER-5671] Bio Beyond >  Designer Planet Playwright Test

gastonabella:
[MER-5671](https://eliterate.atlassian.net/browse/MER-5671)

## Summary

This PR adds an end-to-end Playwright test for the BioBeyond Unit 7 Designer Planet adaptive lesson.

The test imports the private BioBeyond course archive through the automation setup API, creates a disposable course section and student, fetches the private answer key from the Playwright assets bucket, completes the Designer Planet happy path, and runs best-effort teardown afterward.

It also hardens automation setup/teardown handling for this flow, including transient setup retries, teardown warnings, and cleanup of section/project/publication mappings.

## How to Run Locally

Start MinIO and make sure the private assets are uploaded:

- `mer-5671/biobeyond-designer-planet-course.zip`
- `mer-5671/answers.json`

Start Torus with the Playwright asset bucket configured:

```bash
AWS_S3_ACCESS_KEY_ID=devuser \
AWS_S3_SECRET_ACCESS_KEY=devpassword123 \
PLAYWRIGHT_SCENARIO_TOKEN=my-token \
PLAYWRIGHT_ASSETS_BUCKET=torus-playwright-assets-dev \
mix phx.server
```

In another terminal, run the Playwright test:
```bash
cd assets/automation

PLAYWRIGHT_AUTOMATION_API_KEY=<automation-api-key> \
PLAYWRIGHT_SCENARIO_TOKEN=my-token \
PLAYWRIGHT_BASE_URL=http://127.0.0.1 \
npx playwright test tests/torus/student_delivery/designer-planet-adaptive.spec.ts
```

`PLAYWRIGHT_AUTOMATION_API_KEY` must be an API key with automation_setup_enabled enabled.

[MER-5671]: https://eliterate.atlassian.net/browse/MER-5671?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


simonchoxx (commented):
I ran the Designer Planet spec locally a few times and it passed consistently. While looking through the setup path, I noticed that `BioBeyondDesignerPlanetHooks` and `designer-planet-adaptive.scenario.yaml` do not seem to be used by the Playwright spec added in this PR. I also ran the spec twice with both files removed from their repo paths, and the test still passed because setup is done via `/api/v1/automation_setup` using `importArchiveAndCreateSection`.

Do you think we should remove these two files from the PR unless they are intentionally meant to support a separate scenario runner? If we remove them, the `seedStudentDeliveryScenario(..., params = {})` change in `support/common.ts` may not be needed either.

gastonabella:
> I ran the Designer Planet spec locally a few times and it passed consistently. While looking through the setup path, I noticed that `BioBeyondDesignerPlanetHooks` and `designer-planet-adaptive.scenario.yaml` do not seem to be used by the Playwright spec added in this PR. I also ran the spec twice with both files removed from their repo paths, and the test still passed because setup is done via `/api/v1/automation_setup` using `importArchiveAndCreateSection`.
> 
> Do you think we should remove these two files from the PR unless they are intentionally meant to support a separate scenario runner? If we remove them, the `seedStudentDeliveryScenario(..., params = {})` change in `support/common.ts` may not be needed either.

@simonchoxx Good catch, thanks! Those files were from an earlier scenario-seed setup path, but the final spec now uses `/api/v1/automation_setup` with the private asset bucket. I’m a bit surprised our change-cleanup pass didn’t catch them as stale, but I agree they’re no longer needed here. I’ll remove the scenario YAML/hooks and revert the now-unneeded common helper change.

simonchoxx (approved):
Nice work 🚀

PR #6752, opened 2026-07-30
[ENHANCEMENT] [MER-5676] Adaptive Lesson: HW Stellar Life Cycles - Training (Playwright)

simonchoxx:
[MER-5676](https://eliterate.atlassian.net/browse/MER-5676)

## Summary

Adds the Playwright test for the adaptive lesson **"Stellar Life Cycles - Training"** (Unit 2: R\*: Stars → Module 13, Habitable Worlds), which drives it through its happy path against a real imported course.

It reuses the existing private-asset endpoint and automation helpers (`AutomationAssetsTask`, `buildAutomationLoginData`), so it adds **no new infrastructure**. It also hardens `AdaptiveLessonTask`, which every adaptive spec depends on to reach a lesson, and fixes the automation teardown so a project carrying analytics rollups and imported media can actually be deleted.

## Context

The project is a full course; the test navigates to the correct lesson by title. The lesson totals 35 screens, of which **32 are on the path this test takes** — the other three exist only on the simulator branch.

That branch is optional by design: the lesson offers an authored **"Sim not working? Click here"** alternate path on screen 17 (Tutorial), and the ticket asks for the simulation to be attempted first while explicitly allowing either route. It is attempted with a real 30-second budget and never comes up under automation — the bundle downloads fine and WebGL is available in headless Chrome, but the CAPI app renders no canvas and logs nothing, so it never paints its controls. The alternate path is taken and the lesson completes through it.

Interactive screens are identified by their **authored janus part id** rather than by body text, which survives copy changes and never collides the way prose matching does. The deck stacks a screen on its `layerRef` ancestors and keeps their parts mounted, so a probe identifies "this screen or one of its descendants" — the rule order in the spec is load-bearing and documented as such. That is also why this lesson does not reuse the existing `completeAdaptiveHappyPath` runner, which identifies screens through `scanScreen()` widget heuristics instead.

The lesson title and every correct answer live in a **private answers JSON** that must never be committed, stored in the assets bucket under `mer-5676/`.

## What changed

**New spec:**

- `stellar-life-cycles-training.spec.ts` — each screen is claimed by the first rule whose probe is on screen, answered from the answers JSON, then the deck advances. It asserts that **every interactive screen on the happy path was answered**, so a silently skipped screen cannot reach the end of the deck and look like a pass.

**`AdaptiveLessonTask` hardening — shared by every adaptive spec:**

- The course outline retry is keyed on **the lesson card rendering**, not on landing on `/learn`: a server error renders at that same URL, so the previous check reported success and then failed 30s later on a locator timeout with a misleading message. Retrying is what gets past a transient one.
- The onboarding "Go to course" button is matched **by role, scoped to the wizard**. The stepper also renders a hidden `#automation-go-to-course` helper whose label contains that text, which a CSS text match picks up first.
- The prologue's Begin button now gets a **real wait**: `locator.isVisible()` ignores its `timeout` option (deprecated upstream), so the previous check never actually waited and burned a retry with a full page reload.
- Failure budgets tightened: a broken prologue used to take about five minutes to report.

**`AutomationSetupTask`:**

- The teardown request is wrapped so a **transport failure** (connection refused, socket hang up, request timeout) is reported as a warning instead of throwing. This runs in `afterAll`, where throwing fails a test that already passed. The existing per-entity payload reporting is untouched.

**Product change (`lib/oli/automation_setup.ex`, +30/−9):**

- `teardown_project/1` now deletes the analytics rollups (`resource_summary`, `response_summary`, `resource_part_responses`) and `media_items`. All of them reference the project or its resources with `ON DELETE NO ACTION`, so without this the entire teardown transaction fails on a foreign key and the project is stranded. **The order is load-bearing** and commented: response summaries also reference resource part responses, so they precede those.
- The function's doc comment listed an incomplete (and partly misnamed) inventory of what gets deleted; replaced with an accurate, shorter description.

## Verification

- **`stellar-life-cycles-training`: 7 green end-to-end runs** against a real local server (MinIO + Postgres), 9.7–12.0 min each, completing all 32 screens and asserting the end-of-lesson dialog.
- **`automation-teardown-warnings.spec.ts` (6 cases): all pass** with the transport-failure wrapper added here.
- `tsc`, `eslint` and `prettier` clean on every touched file; `mix format` and `mix compile --warnings-as-errors` clean.
- No adaptive spec is tagged `@nightly`, so **CI does not run them** — they require the private bucket and the automation API key.

## Running the test locally

**1. Required environment variables** (no new ones for this test):

```
PLAYWRIGHT_SCENARIO_TOKEN=<a_value_you_choose>
PLAYWRIGHT_ASSETS_BUCKET=<your_bucket_name>
PLAYWRIGHT_AUTOMATION_API_KEY=<key_generated_at_/admin/api_keys>
```

**2. Seed the bucket once.** Inside the bucket, create a `mer-5676/` folder with these two files:

```
<your-bucket>/
└── mer-5676/
    ├── HW_stellar_life_cycle.zip
    └── answers.json
```

(ask the team for both — they are private content, not in the repo). Only the **bucket name** is configurable; the `mer-5676/` folder and both file names are hardcoded in this spec, and the answers JSON schema is specific to this lesson.

**3. Start the server** with those env vars loaded, then run:

```bash
cd assets/automation
npx playwright test stellar-life-cycles-training
```

The test takes 10-12 minutes (32 screens with server-side rule evaluation on each, plus a 30-second simulator boot attempt). If any env var is missing, the test is skipped with an explicit message.

[MER-5676]: https://eliterate.atlassian.net/browse/MER-5676?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/assets/automation/src/systems/torus/tasks/AdaptiveLessonTask.ts
+++ b/assets/automation/src/systems/torus/tasks/AdaptiveLessonTask.ts
@@ -1,4 +1,4 @@
-import { expect, Page } from '@playwright/test';
+import { expect, Locator, Page } from '@playwright/test';
 import { AdaptiveDeckPO } from '@pom/delivery/AdaptiveDeckPO';
 
 /**
@@ -20,26 +20,41 @@ export class AdaptiveLessonTask {
       `/sections/${sectionSlug}/learn?selected_view=outline` +
       `&search_term=${encodeURIComponent(searchTerm)}`;
 
-    await this.page.goto(outlineUrl, { waitUntil: 'domcontentloaded' });
-    await this.acceptResearchConsentIfPresent();
-    await this.passWelcomeGateIfPresent();
-
-    if (!this.page.url().includes('/learn')) {
-      await this.page.goto(outlineUrl, { waitUntil: 'domcontentloaded' });
-    }
-
     const lessonCard = this.page
       .locator('button[phx-click="navigate_to_resource"]')
       .filter({ hasText: lessonTitle })
       .first();
 
-    await this.waitForLiveView();
-    await lessonCard.waitFor({ state: 'visible', timeout: 30_000 });
+    // Consent and the welcome gate each bounce the first load away from the
+    // outline. Landing on /learn is not enough — a server error renders at that
+    // URL too — so re-request until the card itself shows up.
+    let outlineReady = false;
+
+    for (let attempt = 0; attempt < 3 && !outlineReady; attempt += 1) {
+      await this.page.goto(outlineUrl, { waitUntil: 'domcontentloaded' });
+      await this.page.waitForLoadState('networkidle', { timeout: 10_000 }).catch(() => undefined);
+      await this.acceptResearchConsentIfPresent();
+      await this.passWelcomeGateIfPresent();
+      await this.waitForLiveView();
+
+      outlineReady = await lessonCard
+        .waitFor({ state: 'visible', timeout: 15_000 })
+        .then(() => true)
+        .catch(() => false);
+    }
+
+    if (!outlineReady) {
+      throw new Error(
+        `Lesson "${lessonTitle}" never appeared in the outline at ${this.page.url()}`,
+      );
+    }
+
     await Promise.all([
       this.page.waitForURL(/\/(prologue|adaptive_lesson)\//, { timeout: 20_000 }),
       lessonCard.click(),
     ]);
 
+    await this.page.waitForLoadState('networkidle', { timeout: 10_000 }).catch(() => undefined);
     await this.acceptResearchConsentIfPresent();
     await this.startAttemptFromPrologue();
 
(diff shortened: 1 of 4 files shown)


nicocirio:
When I ran the test it failed on screen 10:

```
Error: Stuck on screen 10 "Stellar Lifecycles" (content screen (no interaction)). Feedback: Incorrect, please try again.

Unknown Part Component Type: janus-formula
Unknown Part Component Type: janus-formula
```

Also, we should rename the bucket folder per what we agreed on:

```
Playwright tests bucket structure:

[course_name]_[adaptive_page_name]
  course.zip
  answers.json
```

nicocirio:
LGTM! I tested it locally and it worked as expected
image

nicocirio (approved):
🚀

PR #6758, opened 2026-08-04
[ENHANCEMENT] [MER-5835] Relax payment code constraints

darrensiegel:
This PR relaxes the payment code redemption constraints - removing the constraint that a payment code for a section must have been generated by a template that created that section. 

The only constraints remaining are:
1. Payment code has to be unredeemed
2. Payment code must be of a value equal to or greater than the cost of the section

### DIFF
--- a/lib/oli/delivery/paywall.ex
+++ b/lib/oli/delivery/paywall.ex
@@ -322,8 +322,8 @@ defmodule Oli.Delivery.Paywall do
   {:error, {:not_enrolled}} if the student is not enrolled in the section
   {:error, {:unknown_section}} when the section slug does not pertain to a valid section
   {:error, {:unknown_code}} when no deferred payment record is found for `code`
-  {:error, {:invalid_code}} when the code is invalid, whether it has already been redeemed or
-    if it doesn't pertain to this section or blueprint product
+  {:error, {:invalid_code}} when the code is invalid or has already been redeemed
+  {:error, {:insufficient_value}} when the code amount does not cover the section cost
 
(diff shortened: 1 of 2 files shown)


lib/oli/delivery/paywall.ex
```diff
@@ -727,26 +750,12 @@ defmodule Oli.Delivery.Paywall do
...
-        dynamic(
-          [p],
-          p.enrollment_id == ^enrollment_id and
-            (p.section_id == ^section.id or p.section_id == ^section.blueprint_id) and
-            p.type != :invalidated
-        )
-      end
-
-    from(p in Payment, where: ^where_condition)
+  def get_active_payment_for(enrollment_id, _section) do
```
gastonabella, line 753: Could we either keep using the `section` argument as a defensive check or remove it from this function’s signature? 

Right now `get_active_payment_for/2` accepts `section` but ignores it, which makes the contract a little misleading. If we want the guardrail, this lookup could join `Enrollment` and require `enrollments.section_id == section.id` while still not constraining `payments.section_id`. That would preserve the new cross-template redemption behavior but avoid returning a payment if a mismatched enrollment/section pair ever reaches this helper.

If that check is not needed, then changing this to `get_active_payment_for(enrollment_id)` would make the new behavior explicit.

gastonabella (approved):
Looks good! I just left one comment about either keeping a defensive section/enrollment check or removing the now-unused section parameter. I’ll leave it up to you whether you think that change is worth making here.

PR #6764, opened 2026-08-06
[ENHANCEMENT] [MER-5850] Add parameterized OAuth login Playwright tests

manelli:
## Summary

Adds fully parameterized Playwright smoke tests for Google OAuth login across Torus account classes and deployment environments.

The suite covers:

- Learner login through `/users/log_in`
- Author login through `/authors/log_in`
- Redirect to Google authorization
- Successful callback to Torus without an OAuth error
- Verification of the signed-in account using its email address
- Correct learner or author workspace
- Session persistence after reload
- Logout and return to the public/login state

## Implementation

- Adds one parameterized test case for each account class: `learner` and `author`.
- Reads only environment-specific values from the YAML selected by `PLAYWRIGHT_PARAMETER_CONFIG_URL`:
  - Torus base URL
  - Operation timeout
  - Dedicated learner credentials
  - Dedicated author credentials
  - Optional author account label
- Keeps stable Google selectors, Torus routes, and expected workspace copy in the test code.
- Verifies the exact signed-in email from the Torus account settings page.
- Uses resilient helpers for the cookie modal and LiveView account menu.
- Gives the complete test three times the configured per-operation timeout.
- Runs headed Chrome because Google may reject automated headless authentication.
- Disables tracing to avoid storing credentials, session cookies, and OAuth response data.
- Validates runtime configuration and rejects placeholder values before attempting login.
- Adds `npm run test-oauth-logins`, which selects the suite using the `@oauth` tag.

Full configuration documentation is in [`assets/automation/README.md`](assets/automation/README.md#live-oauth-login-smoke-tests).

## Runtime configuration

```yaml
target:
  base_url: https://torus.example.test
tests:
  oauth_logins:
    timeout_ms: 120000
    accounts:
      learner:
        email: REPLACE_WITH_LEARNER_GOOGLE_EMAIL
        password: REPLACE_WITH_LEARNER_GOOGLE_PASSWORD
      author:
        email: REPLACE_WITH_AUTHOR_GOOGLE_EMAIL
        password: REPLACE_WITH_AUTHOR_GOOGLE_PASSWORD
        account_label: Author
```

Do not commit real credentials or the runtime YAML.

## Prerequisites

Each target environment requires:

- Google OAuth enabled in Torus.
- One dedicated learner Google account.
- One dedicated author Google account.
- Accounts without MFA, CAPTCHA, recovery, or other interactive challenges.
- The Google identities linked to the correct Torus account class.
- A private YAML configuration containing the account credentials and target URL.

Install the automation dependencies once from the repository root:

```bash
cd assets/automation
npm install
npx playwright install --with-deps chrome
```

## Testing locally

1. Configure a Google OAuth web client with these authorized redirect URIs:

   ```text
   http://localhost/users/auth/google/callback
   http://localhost/authors/auth/google/callback
   ```

2. Start or restart Torus with `GOOGLE_CLIENT_ID` and `GOOGLE_CLIENT_SECRET` set.

3. Log in manually once with each dedicated account:

   - Learner: `http://localhost/users/log_in`
   - Author: `http://localhost/authors/log_in`

   This creates and links the corresponding Torus learner and author records.

4. Save the runtime configuration as `/tmp/oauth-local.yaml`, set the base URL to `http://localhost`, and replace all credential placeholders.

5. Serve the configuration from a separate terminal:

   ```bash
   python3 -m http.server 8787 --bind 127.0.0.1 --directory /tmp
   ```

6. From `assets/automation`, run:

   ```bash
   PLAYWRIGHT_PARAMETER_CONFIG_URL=http://127.0.0.1:8787/oauth-local.yaml npm run test-oauth-logins
   ```

## Testing Tokamak

Use dedicated Google accounts already linked to the correct account classes in Tokamak.

Set the YAML target to:

```yaml
target:
  base_url: https://tokamak.oli.cmu.edu
```

Then run from `assets/automation`:

```bash
PLAYWRIGHT_PARAMETER_CONFIG_URL=http://127.0.0.1:8787/oauth-tokamak.yaml npm run test-oauth-logins
```

## Testing Stellarator

Use dedicated Google accounts already linked to the correct account classes in Stellarator.

Set the YAML target to:

```yaml
target:
  base_url: https://stellarator.oli.cmu.edu
```

Then run from `assets/automation`:

```bash
PLAYWRIGHT_PARAMETER_CONFIG_URL=http://127.0.0.1:8787/oauth-stellarator.yaml npm run test-oauth-logins
```

The same procedure can target another staging or production deployment by changing `target.base_url` and supplying dedicated accounts linked in that environment. A real Chrome window opens and runs both account-class flows.

## Verification

- Prettier passes for the changed files.
- ESLint passes for the OAuth test files.
- Playwright discovers both parameterized cases:
  - Learner OAuth login
  - Author OAuth login

See: https://eliterate.atlassian.net/browse/MER-5850

### DIFF
--- a/assets/automation/README.md
+++ b/assets/automation/README.md
@@ -9,6 +9,7 @@ This platform contains the automated e2e tests.
 - Scenario seeding is authenticated with a default token of `my-token`; change this in the spec runtime config and in your Phoenix `PLAYWRIGHT_SCENARIO_TOKEN` if needed.
 - Browser auto-close behavior is controlled by runtime config in specs (defaults to keep the browser open between tests).
 - The nightly LTI spec reads `CANVAS_UI_EMAIL` and `CANVAS_UI_PASSWORD` from the environment instead of hardcoding credentials.
+- The live OAuth login spec reads its target, dedicated accounts, provider UI selectors, and expected Torus destinations from the remote YAML selected by `PLAYWRIGHT_PARAMETER_CONFIG_URL`.
 - GitHub Actions wiring for the nightly run lives in `.github/workflows/nightly-playwright.yml` and expects those values as environment secrets on the `nightly-ui` environment.
 
 ## 🧪 Configuration Tests & Report
@@ -35,6 +36,62 @@ Open the latest Playwright HTML report
 
 ```bash
 npm run show-report
+```
+
+### Live OAuth login smoke tests
+
+The OAuth suite exercises the complete provider redirect for one learner and one author account. It runs headed Chrome because Google rejects the automated headless sign-in browser. It does not save authenticated storage state, and tracing is disabled because OAuth traces can expose credentials, session cookies, or authorization response data.
+
+Host the following YAML in a secret-protected location and point `PLAYWRIGHT_PARAMETER_CONFIG_URL` at it. Do not commit real account credentials or this runtime configuration to the repository.
+
+```yaml
+target:
+  base_url: https://torus.example.test
+tests:
+  oauth_logins:
+    timeout_ms: 120000
+    provider:
+      name: Google
+      authorization_hostname: accounts.google.com
+      selectors:
+        email_input: 'input[name="identifier"], input#identifierId, input[aria-label="Email or phone"]'
+        email_submit: '#identifierNext'
+        password_input: 'input[type="password"]:not([aria-hidden="true"]):not([name="hiddenPassword"])'
+        password_submit: '#passwordNext'
+        consent_submit: 'button:has-text("Continue")'
+        error_message: 'text=/couldn.t find this account|wrong password|browser or app may not be secure|couldn.t sign you in/i'
+    accounts:
+      learner:
+        email: dedicated-learner@example.test
+        password: replace-at-runtime
+        torus_account_name: OAuth Learner
+        login_path: /users/log_in
+        authorization_path: /users/auth/google/new
+        callback_path: /users/auth/google/callback
+        landing_path: /workspaces/student
+        workspace_heading: Courses available
+        account_settings_link_name: Account Settings
+        account_settings_path: /users/settings
+        logout_path: /
+      author:
+        email: dedicated-author@example.test
+        password: replace-at-runtime
+        torus_account_name: OAuth Author
+        login_path: /authors/log_in
+        authorization_path: /authors/auth/google/new
+        callback_path: /authors/auth/google/callback
+        landing_path: /workspaces/course_author
+        workspace_heading: Course Author
+        account_settings_link_name: Edit Account
+        account_settings_path: /authors/settings
+        account_label: Author
+        logout_path: /authors/log_in
+```
+
+The dedicated provider accounts must already be linked to the corresponding Torus account class and configured without MFA, CAPTCHA, recovery, or other interactive challenges. Keep them least-privileged and isolated from production data. Omit `consent_submit` only when consent has already been granted and the provider never renders that step.
+
+```bash
+PLAYWRIGHT_PARAMETER_CONFIG_URL=https://config.example.test/oauth.yaml npm run test-oauth-logins
 ```
 
 ## 🤖 Automated Configurations
--- a/assets/automation/README.md
+++ b/assets/automation/README.md
@@ -9,7 +9,7 @@ This platform contains the automated e2e tests.
 - Scenario seeding is authenticated with a default token of `my-token`; change this in the spec runtime config and in your Phoenix `PLAYWRIGHT_SCENARIO_TOKEN` if needed.
 - Browser auto-close behavior is controlled by runtime config in specs (defaults to keep the browser open between tests).
 - The nightly LTI spec reads `CANVAS_UI_EMAIL` and `CANVAS_UI_PASSWORD` from the environment instead of hardcoding credentials.
-- The live OAuth login spec reads its target, dedicated accounts, provider UI selectors, and expected Torus destinations from the remote YAML selected by `PLAYWRIGHT_PARAMETER_CONFIG_URL`.
+- The live OAuth login spec reads its target, operation timeout, and dedicated accounts from the remote YAML selected by `PLAYWRIGHT_PARAMETER_CONFIG_URL`.
 - GitHub Actions wiring for the nightly run lives in `.github/workflows/nightly-playwright.yml` and expects those values as environment secrets on the `nightly-ui` environment.
 
 ## 🧪 Configuration Tests & Report
@@ -42,7 +42,9 @@ npm run show-report
 
 The OAuth suite exercises the complete provider redirect for one learner and one author account. It runs headed Chrome because Google rejects the automated headless sign-in browser. It does not save authenticated storage state, and tracing is disabled because OAuth traces can expose credentials, session cookies, or authorization response data.
 
-Host the following YAML in a secret-protected location and point `PLAYWRIGHT_PARAMETER_CONFIG_URL` at it. Do not commit real account credentials or this runtime configuration to the repository.
+The Google selectors and expected Torus routes and copy are stable test constants. The runtime YAML contains only values that vary by environment or account. `timeout_ms` is the timeout for each external operation; the complete test receives three times that budget.
+
+Host the YAML in a secret-protected location and point `PLAYWRIGHT_PARAMETER_CONFIG_URL` at it. Do not commit real account credentials or this runtime configuration to the repository.
 
 ```yaml
 target:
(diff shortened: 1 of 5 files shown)


assets/automation/tests/torus/oauth_login/support/oauthLoginConfig.ts
```diff
@@ -0,0 +1,144 @@
+/**
+ * Runtime parameter contracts and validation for the live OAuth login tests.
+ */
+
+export const OAUTH_ACCOUNT_CLASSES = ['learner', 'author'] as const;
+
+type OAuthAccountClass = (typeof OAUTH_ACCOUNT_CLASSES)[number];
+
+export type OAuthAccountParameters = {
```
nicocirio: The YAML config holds ~15 fields per account plus the full set of Google-selector strings, but only `base_url`, `email`, `password`, `torus_account_name`, and (optionally) `account_label` actually vary by environment/account — the Google selectors and the Torus paths/copy are constants of this codebase and of Google's login page, not of the deployment target (the PR's own Tokamak instructions only change `base_url`). Would it make sense to move those into the oauth test itself, keeping the YAML limited to what genuinely varies per environment/account? The trade-off: we'd lose the ability to point this test at an older deployed Torus version without an override — might be worth keeping that as an optional YAML override if that's a real need.

assets/automation/README.md
```diff
@@ -37,6 +38,62 @@ Open the latest Playwright HTML report
...
+        email_input: 'input[name="identifier"], input#identifierId, input[aria-label="Email or phone"]'
+        email_submit: '#identifierNext'
+        password_input: 'input[type="password"]:not([aria-hidden="true"]):not([name="hiddenPassword"])'
+        password_submit: '#passwordNext'
+        consent_submit: 'button:has-text("Continue")'
+        error_message: 'text=/couldn.t find this account|wrong password|browser or app may not be secure|couldn.t sign you in/i'
+    accounts:
+      learner:
+        email: dedicated-learner@example.test
+        password: replace-at-runtime
```
nicocirio: `requireRuntimeValue` only rejects strings starting with `REPLACE_WITH_`, but this example's password is `replace-at-runtime` (same for the author account below). If someone copies this verbatim, that guard never trips — it'd try logging in with the literal string instead of failing fast with a clear message. Maybe worth aligning it to something like `REPLACE_WITH_your_password`?

assets/automation/tests/torus/oauth_login/oauth-login.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+
+test.describe('OAuth login @oauth @smoke', () => {
+  OAUTH_ACCOUNT_CLASSES.forEach((accountClass) => {
+    test(`${accountClass} signs in through the configured OAuth provider`, async ({
+      page,
+    }, testInfo) => {
+      const account = parameters.accounts[accountClass];
+      const torusOrigin = new URL(baseUrl).origin;
+
+      testInfo.setTimeout(parameters.timeout_ms);
```
nicocirio: Since `testInfo.setTimeout(parameters.timeout_ms)` sets the whole test's budget to the same value used for each step's own `timeout` (lines 57, 70, 77, 90), a slow step early in the flow (e.g. Google's redirect) can eat into the budget the later steps assume they still have — the test would then fail on the overall timeout rather than a specific step. Would it make sense to give the whole test a larger multiple, like `parameters.timeout_ms * 3`?

assets/automation/package.json
```diff
@@ -7,6 +7,7 @@
     "test-config": "npm run pw config.spec.ts",
     "test-config:headed": "npm run test-config -- --headed",
     "test-dot-smoke": "npm run pw dot-chatbot.spec.ts",
+    "test-oauth-logins": "npm run pw -- oauth-login.spec.ts --headed --reporter=line",
```
nicocirio: `@oauth`/`@smoke` are metadata on the `describe` block, but no script or CI workflow filters by them — this selects the file by name instead. Since `--grep @nightly` already works this way for the LTI test in `nightly-playwright.yml`, could this become `npm run pw -- --grep @oauth --headed --reporter=line`? That'd give the `@oauth` tag an actual purpose, and any future OAuth spec tagged `@oauth` would get picked up automatically.

assets/automation/README.md
```diff
@@ -37,6 +38,62 @@ Open the latest Playwright HTML report
...
+        account_settings_link_name: Edit Account
+        account_settings_path: /authors/settings
+        account_label: Author
+        logout_path: /authors/log_in
+```
+
+The dedicated provider accounts must already be linked to the corresponding Torus account class and configured without MFA, CAPTCHA, recovery, or other interactive challenges. Keep them least-privileged and isolated from production data. Omit `consent_submit` only when consent has already been granted and the provider never renders that step.
+
+```bash
+PLAYWRIGHT_PARAMETER_CONFIG_URL=https://config.example.test/oauth.yaml npm run test-oauth-logins
```
nicocirio: Could we replace this generic command example with a concrete step-by-step walkthrough for an already-deployed environment? Would've saved me some back-and-forth the first time I ran this against Tokamak (working out account creation, hosting the YAML, and the working directory on my own).

````
```suggestion
### Running this suite against an already-deployed environment (example: Tokamak)

1. Log in once, manually, with each dedicated Google test account through that environment's real login pages (e.g. `https://tokamak.oli.cmu.edu/users/log_in` and `https://tokamak.oli.cmu.edu/authors/log_in`), so Torus creates the corresponding `learner`/`author` user there.
2. Build a YAML config with `target.base_url: https://tokamak.oli.cmu.edu` and those two accounts' real credentials — the rest of the example above can stay unchanged.
3. Host that YAML somewhere reachable over HTTP(S). For a quick run from your own machine, serve it from loopback:
   ```bash
   python3 -m http.server 8787 --bind 127.0.0.1 --directory /tmp
   ```
4. From the repo root, `cd assets/automation` — this is where `test-oauth-logins` is defined (`package.json`), and where the one-time setup (`npm i`, `npx playwright install --with-deps chrome`) happens.
5. Run:
   ```bash
   PLAYWRIGHT_PARAMETER_CONFIG_URL=http://127.0.0.1:8787/oauth-tokamak.yaml npm run test-oauth-logins
   ```
6. A real Chrome window opens on your machine and runs the full flow against that environment's `base_url`, for both the `learner` and `author` cases.

To target a different environment, repeat these steps changing only `base_url` and the account credentials.
```
````

File: assets/automation/tests/torus/oauth_login/oauth-login.spec.ts [hunk 6]
```diff
@@ -0,0 +1,148 @@
...
+  await expect(
+    page.getByRole('heading', { name: account.workspace_heading, exact: true }),
+  ).toBeVisible();
+  await expect(page.locator('.alert-danger, .alert-error')).toHaveCount(0);
+
+  const accountMenuButton = page.locator('#workspace-user-menu');
+
+  await expect(accountMenuButton).toHaveAttribute(
+    'aria-label',
+    `${account.torus_account_name} user account menu`,
```
nicocirio: If we want to simplify the YAML fields here, we might not need `torus_account_name` at all — we could assert the correct account is signed in by checking the account's own email instead, since it's already required for login anyway. Both `/users/settings` and `/authors/settings` render exactly one `input[type="email"]` with the current account's email as its `value`, so the same selector would work for both account classes without extra config.

Downside: this would mean navigating to `account_settings_path` as part of the flow (today we only check the settings link's `href`), which adds a step — worth keeping in mind given the shared timeout budget mentioned above. But it could save whoever sets up a new dedicated account from having to go look up that account's display name in its Google profile. Just an idea, not a strong ask.

assets/automation/tests/torus/oauth_login/support/oauthLogin.ts
```diff
@@ -0,0 +1,143 @@
+import { OAuthAccountParameters, OAuthLoginParameters } from './oauthLoginConfig';
+import { expect, Page, Response } from '@playwright/test';
```
nicocirio, line 2: Needed this for the fix below.

```suggestion
import { OAuthAccountParameters, OAuthLoginParameters } from './oauthLoginConfig';
import { expect, Page, Response } from '@playwright/test';
import { Utils } from '@core/Utils';
```

assets/automation/tests/torus/oauth_login/support/oauthLogin.ts
```diff
@@ -0,0 +1,143 @@
...
+  const acceptButton = cookieModal.locator('button:has-text("Accept")').first();
+  const appeared = await acceptButton
+    .waitFor({ state: 'visible', timeout: 3_000 })
+    .then(() => true)
+    .catch(() => false);
+
+  if (appeared) {
+    await acceptButton.click({ force: true });
+    await expect(cookieModal).toBeHidden({ timeout: 5_000 });
+  }
```
nicocirio, line 189: I ran this against a real deployed environment (Tokamak) and it consistently hung for 120s on both the `learner` and `author` cases, right after clicking "Sign in with Google" — the browser never left Torus's origin. Traced it to `acceptButton.click({ force: true })`: forcing the click can land it mid-animation, before the cookie modal's own Bootstrap show-transition settles, which races `CookieConsent.tsx`'s unmount-based dismiss and leaves its `.modal-backdrop` permanently orphaned in the DOM (confirmed live: 9 checks over 5s, always 1 element, never 0) — intercepting the next click. A manual, non-forced click never triggers this.

This suite already has a working recovery for exactly this (`Utils.modalDisappears`, `assets/automation/src/core/Utils.ts:89-97`). Reusing that instead of reimplementing it fixed it — re-ran against Tokamak afterward and this step now passes reliably for both account classes:

```suggestion
  if (appeared) {
    // A forced click can land mid-animation, before Bootstrap's own show transition settles;
    // a normal click waits for the button to be stable first, matching how a human would click.
    await acceptButton.click();
    await expect(cookieModal).toBeHidden({ timeout: 5_000 });

    // Dismissing the dialog unmounts its React tree directly, which can still race Bootstrap's
    // own async modal-hide transition and leave its `.modal-backdrop` overlay orphaned in the
    // DOM, intercepting subsequent clicks. Reuse the same recovery already used elsewhere in
    // this suite instead of duplicating it: wait for it to go away, reload as a last resort.
    await new Utils(page).modalDisappears();
  }
```

assets/automation/tests/torus/oauth_login/oauth-login.spec.ts
```diff
@@ -0,0 +1,148 @@
+import { loadParameterizedTestConfig } from '@core/parameterizedConfig';
+import { expect, Page, test } from '@playwright/test';
```
nicocirio: Needed this for the fixes below.

```suggestion
import { loadParameterizedTestConfig } from '@core/parameterizedConfig';
import { Utils } from '@core/Utils';
import { expect, Page, test } from '@playwright/test';
```

assets/automation/tests/torus/oauth_login/oauth-login.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+
+  await expect(accountMenuButton).toHaveAttribute(
+    'aria-label',
+    `${account.torus_account_name} user account menu`,
+  );
+  await accountMenuButton.click();
+
+  const accountMenu = page.locator('#workspace-user-menu-dropdown');
+
+  await expect(accountMenu).toBeVisible();
```
nicocirio: Ran this against Tokamak and the account menu didn't open reliably — after clicking `#workspace-user-menu`, the dropdown stayed hidden. The failure's accessibility snapshot showed the dropdown's `data-phx-id` changing between polls — LiveView replaces that node shortly after the click, resetting its client-only `JS.toggle` open state even when the click landed correctly. This suite already has a working retry for exactly this (`Utils.forceClick`, `assets/automation/src/core/Utils.ts:10-27`). Reusing it fixed it — re-ran against Tokamak and the menu now opens reliably for both account classes:

```suggestion
  await accountMenuButton.click();

  const accountMenu = page.locator('#workspace-user-menu-dropdown');

  // LiveView can still patch/replace this element shortly after the initial connect (visible
  // as a new `data-phx-id` on the same node), which resets the dropdown's client-only
  // JS.toggle open state back to hidden even after a successful click. Retry the click instead
  // of waiting for a single one to stick — same recovery already used elsewhere in this suite.
  await new Utils(page).forceClick(accountMenuButton, accountMenu);
```

assets/automation/tests/torus/oauth_login/oauth-login.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+
+  const accountSettingsLink = accountMenu.getByRole('link', {
+    name: account.account_settings_link_name,
+    exact: true,
+  });
```
nicocirio: Separately, the `learner` case's "Account Settings" assertion never matched, while `author`'s "Edit Account" passed. The accessibility snapshot showed the link was actually visible and interactive, but its computed accessible name was `"Account Settings "` (trailing space) — caused by the trailing chevron icon in the component used for "Account Settings"; the one used for "Edit Account" doesn't have that icon. Dropping `exact: true` for this specific locator fixed it, verified against Tokamak:

```suggestion
  // `exact: true` breaks for "Account Settings": its trailing chevron icon leaves an extra
  // trailing space in the computed accessible name ("Account Settings "), which doesn't match
  // exactly. "Edit Account" (author) uses a different markup pattern and isn't affected.
  const accountSettingsLink = accountMenu.getByRole('link', {
    name: account.account_settings_link_name,
    exact: false,
  });
```

nicocirio (commented):
I ran this suite against Tokamak using two of my own Google accounts (one per account class). After applying the fixes noted in the comments below, both the `learner` and `author` cases pass end-to-end.

Image

manelli:
Thanks @nicocirio for the reviews! I implemented them in https://github.com/Simon-Initiative/oli-torus/pull/6764/commits/6f991fec0200fcdba1e8f8b5ffcc686a541475cd

darrensiegel (approved):

PR #6771, opened 2026-08-13
[FEATURE] [MER-5795] Rework: Native AB testing intervention-scoped assignments and assessment-driven Thompson Sampling

eliknebel:
## Summary

Adds intervention-scoped assignment and assessment-driven Thompson Sampling to native Torus experiments. A decision point can now govern repeated placements of one experiment-controlled Alternatives Group while each learner receives a separate sticky assignment at every intervention and accepted assessment rewards update shared decision-point policy state.

## What changed

- Introduced explicit decision points, condition mappings, interventions, assessment bindings, intervention-scoped assignments, accepted-reward claims, and policy state.
- Added weighted-random and Beta-Bernoulli Thompson Sampling configuration with priors, warm-up, fixed-control, traffic-cap, and imbalance-monitoring guardrails.
- Connected finalized scored-page attempts to asynchronous, idempotent reward processing and posterior updates.
- Batched delivery assignment resolution by page and preserved stable assignment behavior across revisits, publication snapshots, and repeated group placements.
- Made Alternatives Group strategy authoritative, separated Learner Choice from Experiment-Controlled group management, and preserved legacy `upgrade_decision_point` compatibility at read boundaries.
- Prevented nested Alternatives insertion and delivery assignment while allowing placements inside ordinary containers.
- Added authoring configuration, lifecycle validation, posterior/assignment reporting, preview controls, telemetry, xAPI/ClickHouse evidence, and AppSignal operational guidance.
- Updated export/ingest and versioned schemas for repeated placements and canonical experiment-controlled group behavior.

## Why

The previous model coupled decision-point policy state with learner assignment identity. Reusing a decision point across pages therefore forced one course-wide condition, while creating a decision point per placement prevented earlier outcomes from improving later assignments. Reward discovery inside selected alternatives also could not represent a shared downstream assessment.

This change separates reusable group identity, placed intervention identity, sticky assignment identity, assessment binding, and shared policy state.

## User and developer impact

- Authors can configure multiple intervention instances under one decision point and bind a distinct scored assessment to each Thompson Sampling intervention.
- Learners may receive different conditions at different interventions, retain each assignment on revisit, and contribute one eligible binary reward per binding.
- Researchers and operators gain intervention, assessment, reward, assignment, posterior, lifecycle, and fallback evidence without exposing raw learner responses.
- Existing Alternatives content does not require rewriting, author re-save, forced republication, or a feature-specific backfill.

## Migrations and rollout

- Adds PostgreSQL persistence and delivery indexes plus an additive ClickHouse evidence schema.
- The policy-column migration is intentionally schema-only and does not convert legacy policy JSON.
- Legacy `upgrade_decision_point` group revisions remain supported as a read-compatible alias for canonical `experiment_controlled` behavior.
- Deployment and rollback guidance is documented in `docs/runbooks/appsignal/experiment-reward-handoff.md`.

## Validation

- `mix format`
- `mix compile --warnings-as-errors`
- Aggregate affected backend suites: 24 doctests and 296 tests passed
- Focused backend review-fix suites: 38 tests passed
- Affected Jest suites passed, including the Alternatives ancestry regression
- `yarn lint`
- `yarn format`
- PostgreSQL migration rollback/reapply rehearsal
- ClickHouse migration rollback/reapply rehearsal
- Requirements trace passed for FR-001 through FR-034 and AC-001 through AC-034
- Work-item validation and `git diff --check` passed

## Follow-ups

- Associate configuration errors directly with individual fields and improve narrow-screen experiment-heading layout.
- Convert reward-binding processing and broad activation resource resolution to set-oriented loading if production measurements cross the documented thresholds.


lib/oli_web/live/workspaces/course_author/experiments_live.ex
```diff
@@ -138,26 +126,37 @@ defmodule OliWeb.Workspaces.CourseAuthor.ExperimentsLive do
         </div>
       <% end %>
 
-      <section class="mt-10" aria-labelledby="decision-points-heading">
+      <section class="mt-10" aria-labelledby="experiment-alternatives-heading">
```
nicocirio, line 129: The PRD explicitly requires renaming the Experiments group-management surface from “Decision Points” to “Experiment-Controlled Alternatives” to avoid confusing reusable alternatives groups with policy bindings. This UI still renders “Decision Points” in the section heading, empty state, creation/edit/delete flows, validation copy, and the experiment-details selector; the LiveView tests also assert the old wording. Could we update the author-facing terminology consistently (while leaving internal compatibility names alone) and adjust the corresponding UI tests?

lib/oli_web/live/workspaces/course_author/experiment_details_live.ex
```diff
@@ -158,63 +198,429 @@ defmodule OliWeb.Workspaces.CourseAuthor.ExperimentDetailsLive do
...
+                        class="form-control"
+                        name={"configuration[conditions][#{condition_index}][active]"}
+                        disabled={@experiment.state in [:completed, :archived]}
+                        phx-change="change_condition_availability"
+                      >
+                        <option value="true" selected={condition.active}>Active</option>
+                        <option value="false" selected={not condition.active}>Inactive</option>
+                      </select>
+                    </div>
+                    <div>
```
nicocirio, line 328: Could we add brief help text to the `Weight` / `Warm-up weight` fields? The policy normalizes active weights, but the UI only shows a numeric value (defaulting to `1.0`), which can reasonably look like an absolute probability. Clarifying that `1 / 1` is an even split, `2 / 1` is approximately 2:1, and values do not need to sum to 1 would make the configuration self-explanatory.

lib/oli/delivery/attempts/page_lifecycle.ex
```diff
@@ -249,22 +251,42 @@ defmodule Oli.Delivery.Attempts.PageLifecycle do
...
+         } = finalization_summary,
+         resource_attempt_id,
+         section_id
+       }} ->
         if graded do
           Snapshots.queue_or_create_snapshot(part_attempt_guids, section_slug)
         end
 
+        if lifecycle_state == :evaluated do
+          case Experiments.RewardHandoffWorker.maybe_enqueue(resource_attempt_id, section_id) do
```
nicocirio: Could we clarify how this handoff is recovered if the process stops, or the enqueue fails, after `finalize/3` commits the evaluated attempt but before `maybe_enqueue/2` persists its job? A subsequent finalization returns `:already_submitted`, so I could not find a path that would enqueue the reward later; the result could remain absent from `AcceptedReward` and the Thompson posterior. If durable delivery is required here, would inserting the job within the evaluation transaction, or persisting an outbox/pending record with an idempotent reconciler, fit the intended failure semantics? Separately, the normal path could guard the call with `graded and lifecycle_state == :evaluated`: ungraded pages can be evaluated too, but cannot be valid Thompson assessment bindings, so there is no need to run the pre-enqueue relevance query for them.

nicocirio (commented):
Awesome feature and great work! 👏 👏

I’m sharing the findings I’ve identified so far so you can start evaluating them while I continue testing the feature locally. I may add follow-up observations as I work through the end-to-end flows.

One finding concerns regression coverage for delayed evaluation:

`apply_manual_scoring/3` and the auto-submit worker currently behave as intended: each calls `maybe_enqueue/2` once its resource-attempt finalization reaches `:evaluated`. The regular page-finalization path has a regression asserting that an evaluated resource attempt enqueues `RewardHandoffWorker`, but I could not find equivalent coverage for manual grading or auto-submit.

Could we add those regressions—ideally also asserting the resulting accepted reward/posterior update—so a future refactor cannot silently stop Thompson rewards for pages evaluated through either delayed path?

lib/oli_web/live/workspaces/course_author/experiments_live.ex
eliknebel: We decided to retain **Decision Points** as the author-facing terminology. Commit 176ae51aab updates the PRD, requirements, design, plan, and execution history to reflect that product direction, so the existing UI and its terminology tests remain intentional.

lib/oli_web/live/workspaces/course_author/experiment_details_live.ex
eliknebel: Added an accessible info-icon tooltip beside both **Weight** and **Warm-up weight** using the page existing GlobalTooltip pattern. It explains that weights are relative, need not sum to 1, and gives the 1/1 and 2/1 examples. Decimal weights remain supported intentionally, with a positive minimum and focused LiveView coverage.

lib/oli/delivery/attempts/page_lifecycle.ex
eliknebel: Closed the durability gap in 176ae51aab. The Oban handoff job is now inserted inside the same Ecto transaction that persists evaluation in regular finalization, manual grading, and auto-submit. Enqueue failure rolls back evaluation, leaving finalization retryable; the worker remains unique and idempotent. Regular finalization also limits the relevance check to `graded` attempts reaching `:evaluated`. The FDD, reward design, and operational runbook now document this guarantee.

eliknebel:
Added delayed-evaluation regression coverage in 176ae51aab for both manual grading and auto-submit. Each test drives the delayed path, asserts `RewardHandoffWorker` enqueue, executes the handoff, and verifies one accepted reward plus mutation of the assigned condition posterior. The focused attempt/LiveView suites pass (49 tests), the dedicated reward-handoff suites pass (18 tests), and `mix compile --warnings-as-errors` is clean.

nicocirio:
# Manual Test Summary

## Video


https://github.com/user-attachments/assets/5d0bf861-1d5f-4cd8-8f03-0264574d7249



## Environment

Local environment set up from scratch: PostgreSQL (transactional database) and ClickHouse (analytics, via Docker), running migrations for both engines and enabling the required flags/feature toggles for experiments, alternatives, and the ClickHouse admin panel.

## Flow Tested

An end-to-end scenario was built following the guiding example from the review document (Lucía/Sofía/Martín): an author creates a two-alternative group ("worked example" vs. "step-by-step hint"), inserts it into two different content pages, each linked to its own scored page, and configures an experiment with the **Thompson Sampling** policy over that group. The project was published into two sections — one participating, one not — and the experiment was activated.

With that in place, the full delivery cycle was walked through:

- A student (Sofía) opens the first page containing the help: she gets one alternative, and on reload keeps seeing the same one — assignment is stable per placement.
- Opening the second page (same group, different placement), she may get a different alternative without contradicting the first: each placement has its own independent assignment.
- After correctly completing the scored page linked to the first placement, exactly one accepted reward was generated and the policy state (the Beta distribution's posterior) updated for the alternative she had received.
- The same scored attempt was retried (once via a page reload, once via a deliberately failed second attempt): in neither case was a second reward created or the already-accepted state changed — the "only the first attempt counts" protection held.
- A second student (Martín), on his first visit, received a new assignment independent of Sofía's, using the state already updated by her evidence — without retroactively altering what Sofía had already seen.
- The Thompson Sampling policy report was refreshed in the authoring UI, and the displayed numbers (estimated probability, posterior α/β, counts) matched exactly what was persisted in the database.
- Evidence was confirmed to also reach ClickHouse asynchronously: exposures and the accepted reward show up correctly attributed to the experiment, intervention, and condition, each traceable back to its originating xAPI event.

## Negative Cases Tested

- **Non-participating section:** a third student, in a section not marked as participating in the experiment, saw the fallback alternative without generating any assignment or reward — completely outside the experimental flow despite sharing the same published content.
- **Duplicate binding:** attempting to configure, on a draft experiment, the same scored page for two different interventions was rejected on save with an explicit configuration error.
- **Invalid nesting (editor-side prevention):** attempting to insert an A/B Test inside the content of an already-existing one, the editor simply does not offer that option in the insert menu.

nicocirio (approved):
🚀

PR #6775, opened 2026-08-14
[ENHANCEMENT] [MER-5670] Queue automation project teardown with Oban

manelli:
## Summary

Moves full automation-project teardown to an asynchronous Oban job while keeping user and section cleanup synchronous.

## Why this is needed

Deleting a full imported course can take several minutes. In one observed run:

- Revision deletion took approximately **186 seconds**.
- Resource deletion took approximately **96 seconds**.
- Total project teardown took approximately **283 seconds**.

The slowdown comes primarily from PostgreSQL foreign-key validation and cascading behavior. Several tables that reference revisions and resources do not have supporting indexes on their foreign-key columns.

When a revision or resource is deleted, PostgreSQL must check the referencing tables for dependent rows. Without an index, this can require sequential table scans. Since a full course contains thousands of resources, those scans may be repeated many times during teardown.

The cost therefore depends not only on the course being deleted, but also on the overall size and state of the database. As shared test tables grow, the scans become longer, causing the same Playwright test to have significantly different teardown times across runs and environments.

Waiting for this work inside the HTTP request caused request timeouts, socket hang-ups, and otherwise successful Playwright runs to report teardown warnings.

## What changed

- User, learner, educator, and section cleanup remains synchronous.
- Full project cleanup is queued as an Oban job.
- The original complete deletion behavior is preserved, including publications, revisions, resources, analytics records, and imported media.
- Added a dedicated `automation_teardown` queue with concurrency `1` so multiple expensive teardown jobs do not overload PostgreSQL.
- Teardown jobs are unique per project while active and retry failures up to three times.
- Repeated execution after a project has already been deleted is treated as successful.
- Added timing logs for the request and background project cleanup.
- Extended the Cowboy idle timeout only for full-course imports.

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -741,7 +741,9 @@ if runtime_env == :prod do
       objectives: String.to_integer(System.get_env("OBAN_QUEUE_SIZE_OBJECTIVES", "3")),
       mailer: String.to_integer(System.get_env("OBAN_QUEUE_SIZE_MAILER", "10")),
       certificate_eligibility:
-        String.to_integer(System.get_env("OBAN_QUEUE_SIZE_CERTIFICATE_ELIGIBILITY", "10"))
+        String.to_integer(System.get_env("OBAN_QUEUE_SIZE_CERTIFICATE_ELIGIBILITY", "10")),
+      automation_teardown:
+        String.to_integer(System.get_env("OBAN_QUEUE_SIZE_AUTOMATION_TEARDOWN", "1"))
     ]
 
   config :oli, Oli.Vault,
--- a/lib/oli/automation_setup/project_teardown_worker.ex
+++ b/lib/oli/automation_setup/project_teardown_worker.ex
@@ -20,6 +20,11 @@ defmodule Oli.AutomationSetup.ProjectTeardownWorker do
 
   alias Oli.AutomationSetup
 
+  @non_retryable_messages [
+    "Can only delete projects with no authors",
+    "Project allows duplicates"
+  ]
+
   @impl Oban.Worker
   def perform(%Oban.Job{args: %{"project_slug" => project_slug}}) do
     started_at = System.monotonic_time(:millisecond)
@@ -37,6 +42,14 @@ defmodule Oli.AutomationSetup.ProjectTeardownWorker do
       %{success: false, message: "Project not found"} ->
         Logger.info("automation_project_teardown already_completed project=#{project_slug}")
         :ok
+
+      %{success: false, message: message} when message in @non_retryable_messages ->
+        Logger.error(
+          "automation_project_teardown cancelled project=#{project_slug} " <>
+            "duration_ms=#{System.monotonic_time(:millisecond) - started_at} message=#{message}"
+        )
+
+        {:cancel, message}
 
       %{success: false} = result ->
         message = Map.get(result, :message, "Could not delete project")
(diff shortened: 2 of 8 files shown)


config/runtime.exs
```diff
@@ -741,7 +741,9 @@ if runtime_env == :prod do
       objectives: String.to_integer(System.get_env("OBAN_QUEUE_SIZE_OBJECTIVES", "3")),
       mailer: String.to_integer(System.get_env("OBAN_QUEUE_SIZE_MAILER", "10")),
       certificate_eligibility:
-        String.to_integer(System.get_env("OBAN_QUEUE_SIZE_CERTIFICATE_ELIGIBILITY", "10"))
+        String.to_integer(System.get_env("OBAN_QUEUE_SIZE_CERTIFICATE_ELIGIBILITY", "10")),
+      automation_teardown:
+        String.to_integer(System.get_env("OBAN_QUEUE_SIZE_AUTOMATION_TEARDOWN", "1"))
     ]
```
gastonabella, line 747: Could we add `OBAN_QUEUE_SIZE_AUTOMATION_TEARDOWN` to the changelog/environment config notes? This PR introduces a new runtime env override in `config/runtime.exs` for the dedicated `automation_teardown` Oban queue, defaulting to `1`. Even though it is optional, documenting it would make the deployment/ops impact explicit and match the repo guidance for new environment configuration.

gastonabella (approved):
LGTM! Just left one suggestion

lib/oli/automation_setup/project_teardown_worker.ex
```diff
@@ -0,0 +1,52 @@
...
+
+      %{success: false} = result ->
+        message = Map.get(result, :message, "Could not delete project")
+
+        Logger.error(
+          "automation_project_teardown failed project=#{project_slug} " <>
+            "duration_ms=#{System.monotonic_time(:millisecond) - started_at} message=#{message}"
+        )
+
+        {:error, message}
```
nicocirio, line 62: Nit: the failure branch of `perform/1` (lines 41-49) doesn't have a test — only the "already deleted" idempotent path is covered. It'd be worth adding one that forces `teardown_project/1` to fail for a real reason (e.g. a project that still has authors) and asserts it maps to `{:error, message}`, so we know Oban actually gets what it needs to retry the job as intended.

nicocirio, line 62: Even with the precheck in `enqueue_project_teardown/1`, could we make `perform/1` explicit about which failures are worth retrying? Right now any `%{success: false}` other than "Project not found" maps to `{:error, message}` (lines 41-49), so Oban retries it 3 times regardless of cause. But "Can only delete projects with no authors" and "Project allows duplicates" come straight from `validate_project_for_teardown/1` — the same check that already ran right before the job was enqueued — so if either ever fires here, it's because something changed the project's state in between, and it won't change again just because we retry. Matching those specific messages to `:discard` (or `{:cancel, message}`) and reserving `{:error, message}` for the actually-transient case (the `Ecto.ConstraintError` already rescued in `teardown_project/1`) would make the intent explicit, at basically zero cost, without depending on the precheck staying airtight forever.

manelli: Agreed. Fixed in [c7d5098](https://github.com/Simon-Initiative/oli-torus/pull/6775/commits/c7d50983714e4ed822b22ed3f3976f3b29207a01)

manelli: Added it in [c7d5098](https://github.com/Simon-Initiative/oli-torus/pull/6775/commits/c7d50983714e4ed822b22ed3f3976f3b29207a01)

config/runtime.exs
manelli: Great catch, added in [c7d5098](https://github.com/Simon-Initiative/oli-torus/pull/6775/commits/c7d50983714e4ed822b22ed3f3976f3b29207a01)

PR #6785, opened 2026-08-19
[FEATURE] [MER-5855] Adds core support for LearningObjectiveCoverage Data Source

rgachuhi:
## Summary

- Added Oli.Authoring.ObjectiveCoverage, a read-only application-level data source for Learning Objective coverage.
- Implemented a compact working-publication projection with project scoping and deleted-revision filtering.
- Added normalized hierarchy, curriculum, page, activity, coverage, detail, and search indexes.
- Added embedded-activity-only coverage with banked activity exclusion.
- Added telemetry, malformed-data safeguards, cycle protection, and deterministic output.
- Added focused ExUnit coverage and Harness traceability artifacts.

## Scope

This PR covers the core data layer for the Objectives Editor epic. It does not include:

- Workspace UI integration
- Filters or CSV export
- Proficiency aggregation
- Activity-bank indexing
- Database migrations or caching

## Verification

- mix format --check-formatted
- mix compile --warnings-as-errors
- mix test test/oli/authoring/objective_coverage_test.exs
    - 6 tests, 0 failures

- Harness work-item validation passed
- Requirements implementation traceability passed

### DIFF
--- a/lib/oli/authoring/objective_coverage.ex
+++ b/lib/oli/authoring/objective_coverage.ex
@@ -1,0 +1,721 @@
+defmodule Oli.Authoring.ObjectiveCoverage do
+  @moduledoc """
+  Builds a read-only snapshot of the learning-objective coverage data for an
+  authoring project's working publication.
+
+  The snapshot is intentionally request-scoped. It is built from a compact
+  projection and does not load page or activity content, create a process, or
+  mutate authoring data. Coverage calculations and consumer-specific views are
+  layered on top of this index in later iterations.
+  """
+
+  import Ecto.Query, warn: false
+
+  alias Oli.Authoring.Course.Project
+  alias Oli.Publishing.Publications.Publication
+  alias Oli.Publishing.PublishedResource
+  alias Oli.Repo
+  alias Oli.Resources.Revision
+  alias Oli.Resources.ResourceType
+
+  @telemetry_prefix [:oli, :authoring, :objective_coverage]
+
+  @type row :: %{
+          required(:project_id) => pos_integer(),
+          required(:publication_id) => pos_integer(),
+          required(:revision_id) => pos_integer(),
+          required(:resource_id) => pos_integer(),
+          required(:resource_type_id) => pos_integer(),
+          optional(:slug) => String.t() | nil,
+          optional(:title) => String.t() | nil,
+          optional(:deleted) => boolean(),
+          optional(:objectives) => map() | nil,
+          optional(:children) => list() | nil,
+          optional(:graded) => boolean() | nil,
+          optional(:activity_refs) => list() | nil,
+          optional(:scope) => atom() | String.t() | nil,
+          optional(:activity_type_id) => pos_integer() | nil
+        }
+
+  @type t :: %{
+          project_id: pos_integer() | nil,
+          publication_id: pos_integer() | nil,
+          rows_by_type: %{pos_integer() => [row()]},
+          objectives_by_id: %{pos_integer() => map()},
+          parents_by_child: %{pos_integer() => [pos_integer()]},
+          children_by_parent: %{pos_integer() => [pos_integer()]},
+          pages_by_id: %{pos_integer() => map()},
+          activities_by_id: %{pos_integer() => map()},
+          curriculum_by_id: %{pos_integer() => map()},
+          top_level_objective_ids: [pos_integer()],
+          direct_page_ids_by_objective: %{pos_integer() => [pos_integer()]},
+          activity_ids_by_objective: %{pos_integer() => %{atom() => [pos_integer()]}},
+          coverage_by_objective: %{pos_integer() => map()},
+          details_by_objective: %{pos_integer() => %{atom() => [map()]}},
+          search_by_objective: %{pos_integer() => map()}
+        }
+
+  @projection_types [
+    ResourceType.id_for_objective(),
+    ResourceType.id_for_page(),
+    ResourceType.id_for_activity(),
+    ResourceType.id_for_container()
+  ]
+
+  @doc """
+  Loads the compact working-publication projection for a project slug or
+  project struct and builds its in-memory indexes.
+
+  An existing project without indexed objective/page/activity/container rows
+  returns an empty snapshot. Invalid project arguments return an error before
+  querying the database.
+  """
+  @spec load(Project.t() | %{required(:id) => pos_integer()} | String.t()) ::
+          {:ok, t()} | {:error, :invalid_project}
+  def load(project_or_slug) do
+    started_at = System.monotonic_time()
+
+    try do
+      result =
+        with {:ok, query} <- projection_query(project_or_slug) do
+          rows = Repo.all(query)
+          context = context_from_rows(rows)
+          {:ok, build(rows, context)}
+        end
+
+      emit_load(result, project_or_slug, started_at)
+      result
+    rescue
+      exception ->
+        safe_telemetry_execute(
+          @telemetry_prefix ++ [:load, :exception],
+          %{count: 1},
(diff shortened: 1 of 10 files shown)


lib/oli/authoring/objective_coverage.ex
```diff
@@ -0,0 +1,721 @@
...
+                    related_ids
+                    |> Enum.flat_map(fn related_id ->
+                      get_in(activity_occurrences, [related_id, bucket, page_id]) || []
+                    end)
+                    |> Enum.uniq()
+                    |> Enum.sort()
+
+                  activities = Enum.map(activity_ids, &Map.fetch!(activities_by_id, &1))
+
+                  if MapSet.member?(direct_page_ids, page_id) or activity_ids != [] do
```
eliknebel: Directly attached pages are included in both detail buckets because this condition does not compare page.graded with the current bucket. As a result, formative pages appear under Summative and graded pages appear under Formative; the current test assertions also encode that behavior. Could we require assessment_bucket(page.graded) == bucket for direct attachments, while still including pages containing activities in the selected bucket?

lib/oli/authoring/objective_coverage.ex
```diff
@@ -0,0 +1,721 @@
...
+  querying the database.
+  """
+  @spec load(Project.t() | %{required(:id) => pos_integer()} | String.t()) ::
+          {:ok, t()} | {:error, :invalid_project}
+  def load(project_or_slug) do
+    started_at = System.monotonic_time()
+
+    try do
+      result =
+        with {:ok, query} <- projection_query(project_or_slug) do
```
eliknebel, line 91: A zero-row query currently returns a successful model with nil project and publication IDs. This makes an unknown project, a missing working publication, and a valid empty publication indistinguishable, despite the FDD specifying tagged not-found errors. Could we preserve or verify the project/publication context and test these cases separately?

lib/oli/authoring/objective_coverage.ex
```diff
@@ -0,0 +1,721 @@
...
+      :revision_id,
+      :slug,
+      :title,
+      :objectives,
+      :scope,
+      :activity_type_id
+    ])
+  end
+
+  defp curriculum_summary(row) do
```
eliknebel, line 389: FR-003/AC-003 require deterministic curriculum descendant/path projections, but the model currently stores only immediate children and exposes descendant page IDs. There is no ancestry or path representation for filtering/export consumers. Could we add the required descendant/path indexes and nested-curriculum tests, or update the requirements if paths are intentionally deferred?

lib/oli/authoring/objective_coverage.ex
```diff
@@ -0,0 +1,721 @@
...
+    |> Enum.sort_by(fn objective_id ->
+      objective = Map.fetch!(objectives_by_id, objective_id)
+      {normalize_search_text(objective.title), objective_id}
+    end)
+  end
+
+  defp objective_scope_ids(objective_id, children_by_parent),
+    do: objective_scope_ids(objective_id, children_by_parent, MapSet.new())
+
+  defp objective_scope_ids(objective_id, children_by_parent, visited) do
```
eliknebel: The updated visited set is passed independently to each child, so sibling branches do not share traversal state. In a hierarchy with shared descendants, the same subtree is traversed once per path, potentially multiplying work substantially. Could we memoize descendant closures or thread one visited accumulator across the whole traversal?

Perhaps return a tuple `{visited, ids}` for each traversal?

lib/oli/authoring/objective_coverage.ex
```diff
@@ -0,0 +1,721 @@
...
+  end
+
+  defp curriculum_summary(row) do
+    Map.take(row, [:resource_id, :revision_id, :slug, :title, :children])
+  end
+
+  defp direct_page_ids_by_objective(pages) do
+    Enum.reduce(pages, %{}, fn page, acc ->
+      Enum.reduce(objective_ids(page.objectives), acc, fn objective_id, acc ->
+        Map.update(acc, objective_id, [page.resource_id], fn page_ids ->
```
eliknebel: This sorts and deduplicates the entire accumulated list after every insertion; the activity-occurrence index uses the same pattern. That makes index construction quadratic as attachments grow. Could we accumulate IDs in MapSets and convert/sort once after reduction?

test/oli/authoring/objective_coverage_test.exs
```diff
@@ -0,0 +1,210 @@
+defmodule Oli.Authoring.ObjectiveCoverageTest do
+  use Oli.DataCase
+
+  # Requirements proof: AC-001, AC-002, AC-003, AC-004, AC-005, AC-006,
```
eliknebel: This comment claims proof for AC-001 through AC-009, but it does not identify which test proves each criterion, and several criteria remain untested—notably project/publication isolation, deleted-row exclusion, and curriculum paths. Could we add per-criterion proof references and leave incomplete criteria marked accordingly?

eliknebel (changes requested):

lib/oli/authoring/objective_coverage.ex
rgachuhi: Changes:

  - Updated details_by_objective/6 to include directly attached pages only in their matching assessment bucket.
  - Preserved activity-driven inclusion for activities in the selected bucket.
  - Updated the test expectations:
      - Formative details now include only page 100.
      - Summative details now include only page 101.

lib/oli/authoring/objective_coverage.ex
rgachuhi: Changes:

  - Preserved project and working-publication context with a single left-joined projection.
  - Added distinct errors:
      - :project_not_found
      - :working_publication_not_found

  - Preserved successful empty models when a valid working publication has no indexed rows.
  - Added separate tests for all three cases.
  - Updated the load documentation and typespec.

lib/oli/authoring/objective_coverage.ex
rgachuhi: Changes:

  - Added curriculum indexes:
      - curriculum_parents_by_child
      - curriculum_children_by_parent
      - curriculum_descendants_by_id
      - curriculum_paths_by_id

  - Updated curriculum_pages/2 to use precomputed descendants.
  - Added nested curriculum tests for descendants, paths, ordering, and cycle protection.
  - Preserved deterministic output and safe handling of malformed/cyclic relationships.

lib/oli/authoring/objective_coverage.ex
rgachuhi: Changes:

  - Added a memoized objective_scope_by_id index.
  - Coverage, details, and search now reuse precomputed objective scopes.
  - Shared descendants are traversed once per build instead of once per objective path.
  - Preserved cycle protection, deduplication, and deterministic ordering.
  - Added a shared-descendant regression test.

lib/oli/authoring/objective_coverage.ex
rgachuhi: Changes:

  - direct_page_ids_by_objective/1 now accumulates IDs in MapSets and sorts once.
  - activity_occurrences_by_objective/2 now accumulates activity IDs in nested MapSets and normalizes once.
  - Preserved the existing sorted-list model shape and deduplication behavior.

test/oli/authoring/objective_coverage_test.exs
rgachuhi: Changes:

  - Replaced the blanket requirements comment with explicit AC-001–AC-009 test mappings.
  - Added project-isolation coverage.
  - Added deleted-revision exclusion coverage.
  - Added an explicit input-preservation assertion for read-only model construction.
  - Curriculum descendant/path tests are now explicitly referenced.

eliknebel (approved):

PR #6787, opened 2026-08-19
[FEATURE] [MER-5849] Add Playwright coverage for credential-based account flows

nicocirio:
## Jira

- Ticket: https://eliterate.atlassian.net/browse/MER-5849

## Summary

Adds end-to-end Playwright coverage for the full email/password account
lifecycle (registration, email confirmation, login, password reset,
old-password rejection, new-password login) across both Authoring (Author)
and Delivery (User) accounts, including provisioned system-admin and
instructor roles, plus the CI machinery to run this suite against a
dedicated ephemeral Torus instance on every pull request.

## What Was Added

- Guarded `/test/emails` mailbox JSON endpoints and `/test/scenario-yaml`
  scenario seeding, reusing the existing `enable_playwright_scenarios`
  compile-time flag and `PlaywrightAuth` token, backed by
  `Swoosh.Adapters.Local` (never mounted outside the dedicated
  `MIX_ENV=playwright` build).
- `config/playwright.exs`: the dedicated ephemeral environment configuration
  (Local mailer, deterministic reCAPTCHA, scenario/mailbox routes enabled).
- `CredentialAccountPO` + two Playwright specs: self-service Author /
  independent-learner registration→confirm→login→reset→login lifecycle, and
  provisioned system-admin Author / instructor User login→reset→login
  lifecycle (seeded via scenario YAML, since neither role has self-service
  registration).
- `.github/workflows/pr-playwright.yml`: a new CI job, separate from
  `nightly-playwright.yml`, that boots an ephemeral Postgres + Torus instance
  under `MIX_ENV=playwright` and runs every Playwright test tagged `@pr`,
  uploading Playwright artifacts and the server log on every run.
- Durable design record and decision log at
  `docs/exec-plans/current/epics/automated_testing/mer-5849/approach.md`.

## Validation

- `mix format --check-formatted`, `mix compile --warnings-as-errors`
  (`MIX_ENV=playwright`) pass.
- `mix test test/oli_web/controllers/playwright_mailbox_controller_test.exs
  test/oli_web/live/author_registration_live_test.exs` — 11 tests, 0 failures.
- `npx playwright test tests/torus/user_accounts/credential-account-flows.spec.ts
  tests/torus/user_accounts/credential-account-provisioned-roles.spec.ts` —
  passing consistently across 20+ clean local runs.
- ESLint/Prettier clean on all changed TypeScript/YAML files.
- `pr-playwright.yml` has not yet run in real GitHub Actions — opening this
  PR is the first real validation of that job; follow-up fixes may be needed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/pr-playwright.yml
+++ b/.github/workflows/pr-playwright.yml
@@ -69,6 +69,12 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v5
 
+      # Provides sane, non-secret defaults (e.g. CLOAK_VAULT_KEY) the same
+      # way nightly-scenarios does, so this job doesn't have to hand-list
+      # every variable the app reads at boot.
+      - name: Configure
+        run: cp oli.example.env oli.env
+
       - name: Setup BEAM toolchain
         uses: erlef/setup-beam@v1
         with:
@@ -110,7 +116,7 @@ jobs:
         run: mix deps.compile
 
       - name: Compile (MIX_ENV=playwright)
-        run: mix compile --warnings-as-errors
+        run: set -a && source oli.env && mix compile --warnings-as-errors
 
       - name: Create and migrate the ephemeral database
         run: |
@@ -126,7 +132,11 @@ jobs:
         run: npx playwright install --with-deps chrome
 
       - name: Start Torus (MIX_ENV=playwright)
-        run: nohup mix phx.server > /tmp/torus-server.log 2>&1 &
+        run: |
+          set -a
+          source oli.env
+          set +a
+          nohup mix phx.server > /tmp/torus-server.log 2>&1 &
 
       - name: Wait for Torus to be ready
         run: |
--- a/.github/workflows/pr-playwright.yml
+++ b/.github/workflows/pr-playwright.yml
@@ -1,0 +1,173 @@
+name: PR Playwright Suite
+
+# Runs Playwright tests against a dedicated Torus instance built and torn
+# down for this job only, with an isolated database, Swoosh.Adapters.Local
+# for email, and the /test/scenario-yaml and /test/emails endpoints enabled
+# (MIX_ENV=playwright). This is the counterpart to nightly-playwright.yml,
+# which targets a real, persistent Torus deployment and must never gain
+# access to those endpoints.
+#
+# To include a Playwright test here, tag its title (or its
+# test.describe title) with "@pr", the same way existing suites use
+# "@nightly" or "@smoke". The database starts out freshly created and empty
+# on every run - a test can (and, if it needs data, should) seed exactly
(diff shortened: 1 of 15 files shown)


config/playwright.exs
```diff
@@ -0,0 +1,21 @@
+import Config
+
+import_config "dev.exs"
+
+# This environment is reserved for the ephemeral Torus instance used by the
+# credential-account Playwright suite. It must never be used for a persistent
+# staging or production deployment.
+config :oli,
+  env: :playwright,
```
gastonabella, line 9: Could we rename the `playwright` Mix environment to something more purpose-oriented, like `ci_e2e` or `ephemeral_e2e` (or a better suggestion)? The current name couples the environment to Playwright as the tool, but the purpose seems broader: running isolated PR/CI validation against an ephemeral Torus instance. A more generic name would age better if future CI suites reuse this environment without necessarily being Playwright-specific.

.github/workflows/pr-playwright.yml
```diff
@@ -0,0 +1,199 @@
...
+          done
+          echo "Torus did not become ready in time." >&2
+          cat /tmp/torus-server.log
+          exit 1
+
+      - name: Run @pr-tagged Playwright tests
+        working-directory: assets/automation
+        run: npx playwright test --grep @pr --reporter=line
+
+      - name: Upload Playwright report
```
gastonabella, line 174: Should we upload these Playwright artifacts only when the job fails? The Jira ticket asks for traces, screenshots, videos, and server logs on failure, but this workflow uploads the report, test results, and server log on every run.

lib/oli_web/router.ex
```diff
@@ -354,6 +354,8 @@ defmodule OliWeb.Router do
       pipe_through [:api]
 
       post "/scenario-yaml", PlaywrightScenarioController, :run
+      get "/emails", PlaywrightMailboxController, :index
+      get "/emails/:id", PlaywrightMailboxController, :show
```
gastonabella, line 358: Could we tighten the gate for `/test/emails` so it is mounted only in the dedicated E2E environment? The Jira acceptance criteria say this test API must not be present in development, but `config/dev.exs` enables `enable_playwright_scenarios`, and the router mounts `/test/emails` whenever that flag is enabled. Since this endpoint exposes full email bodies with confirmation/reset links, it seems worth adding a narrower guard, such as a dedicated mailbox flag and/or an explicit E2E environment check, while keeping the existing token protection.

gastonabella (commented):
This is looking good overall! I left a couple of comments around tightening the test-only boundary a bit and making the CI diagnostics behavior line up more clearly with the Jira ticket.

darrensiegel (approved):

PR #6789, opened 2026-08-20
[BUG FIX] [MER-5885] Restore UpGrade data capture parity for native A/B experiments

eliknebel:
https://eliterate.atlassian.net/browse/MER-5885

> **Prerequisite work:** This PR is based on [PR #6786](https://github.com/Simon-Initiative/oli-torus/pull/6786) and [PR #6784](https://github.com/Simon-Initiative/oli-torus/pull/6784). Both prerequisites have been squash-merged into `hotfix-v0.34.1`. This branch has been rebuilt on those squash commits, the superseded prerequisite commits were removed, and the complete MER-5885 diff was reverified.

## Summary

Restores the experiment analytics data needed for UpGrade-style research and analysis while keeping raw learner outcomes independent from optional causal attribution.

### Experiment evidence

- restores focused experiment attribution for evaluated part, activity, and page attempts
- restores attribution for media statements using authoritative attempt or published-section content
- preserves weighted-random assignment scopes and existing Thompson Sampling outcome, reward, and policy-update semantics
- carries complete experiment, assignment, intervention, condition, enrollment, project, publication, algorithm, policy-version, revision, and attempt identities where applicable
- fails safely when assignment or selected-branch evidence cannot be established, without suppressing the underlying xAPI statement
- keeps explicit experiment rewards separate from attempt scores; missing `reward_value` remains `NULL`, while explicit values including `0.0` are preserved

### Initial condition assignments

- adds a dedicated `experiment_condition_assigned` xAPI statement and raw-event type
- emits exactly once when a new assignment is persisted and does not emit again for sticky reuse
- records the exact persisted `assigned_at` value and links the attribution to the dedicated statement hash
- supports weighted-random and Thompson Sampling assignments across intervention and section-enrollment scopes

### Raw analytics and ClickHouse parity

- propagates pseudonymous `enrollment_id` across applicable attempt, page-view, tutor-message, and video statements
- preserves Torus account IDs in `raw_events.user_id` without falling back to actor email/mbox values
- adds the nullable raw-event and experiment-attribution fields required for analysis and historical replay
- normalizes projection behavior across direct upload, Lambda transformation, and replay/backfill
- renames the attribution relationship column to `raw_event_type` for clarity before production release
- consolidates the unreleased ClickHouse experiment migrations into the original migration, consistent with the planned QA database reset
- adds a deterministic, project-scoped ClickHouse query that reconstructs the UpGrade v0.33.0 compatibility dataset from raw outcomes and assignment evidence

### Researcher and operator experience

- adds researcher documentation covering experiment workflow, terminology, S3 xAPI shape, ClickHouse fields, stable keys, event/attribution relationships, privacy boundaries, and example analysis queries
- displays the experiment UUID in experiment details and improves the responsive, accessible details layout
- constrains ClickHouse confirmation dialogs to the established modal sizing pattern

## Important design boundaries

- PostgreSQL remains the runtime source of truth; S3 xAPI remains the replay source; ClickHouse remains an analytical projection
- raw evaluated outcomes are captured whether or not experiment attribution succeeds
- condition-assignment delivery remains best-effort; a transactional outbox, reconciliation, and historical assignment backfill are deferred
- exact attempt-time publication provenance remains the separately tracked [MER-5889](https://eliterate.atlassian.net/browse/MER-5889) limitation
- S3 xAPI may contain sensitive learner information; ClickHouse projections use bounded identifiers, including linkable pseudonymous enrollment IDs, and remain subject to approved analytics access controls

## Verification

- 239 affected Elixir/LiveView tests passed, with 1 excluded
- 26 Lambda ETL tests passed
- `mix compile`
- `mix format --check-formatted`
- `git diff --check`
- requirements traceability passed for FDD, plan, and implementation evidence
- harness work-item validation passed
- security, performance, Elixir, UI, TypeScript, and requirements reviews completed with no remaining findings

## Reconciliation status

- [x] PR #6784 squash-merged as `1c14fc71c9`
- [x] PR #6786 squash-merged as `539b3c98d5`
- [x] Branch rebuilt on updated `hotfix-v0.34.1`
- [x] Superseded unsquashed prerequisite commits removed
- [x] `git range-diff` confirmed all retained MER-5885 commits were patch-equivalent
- [x] Final-base verification and reviews rerun successfully

[MER-5889]: https://eliterate.atlassian.net/browse/MER-5889?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


priv/clickhouse/queries/upgrade_v033_compatibility.sql
```diff
@@ -0,0 +1,66 @@
...
+        assumeNotNull(project_id) AS raw_project_id,
+        assumeNotNull(enrollment_id) AS participant_id,
+        timestamp,
+        score,
+        out_of,
+        event_hash
+    FROM raw_events
+    WHERE section_id = {section_id:UInt64}
+      AND event_type = 'activity_attempt'
+      AND verb_id = 'http://adlnet.gov/expapi/verbs/evaluated'
```
nicocirio: I think this query cannot currently reconstruct the compatibility dataset from the activity events we emit. It filters `verb_id = '.../evaluated'`, but `ActivityAttemptEvaluated.new/3` emits `.../completed` and both direct/Lambda project that as `activity_attempt`.

The producer test currently asserts the enrollment extension but not the emitted verb, while the compatibility-query test uses hand-built raw-event fixtures whose default verb is `.../evaluated`. Could we make these paths share one contract—align the query with the emitted verb and build the compatibility fixture from `ActivityAttemptEvaluated.new/3` (or assert the same canonical verb in both tests)—so this cannot drift again?

priv/clickhouse/queries/upgrade_v033_compatibility.sql
```diff
@@ -0,0 +1,66 @@
...
+      AND timestamp >= {from_timestamp:DateTime64(3)}
+      AND timestamp < {to_timestamp:DateTime64(3)}
+      AND enrollment_id IS NOT NULL
+      AND project_id IS NOT NULL
+    ORDER BY raw_section_id, raw_project_id, participant_id, timestamp
+) AS raw
+ASOF LEFT JOIN
+(
+    SELECT
```
nicocirio, line 45: If I understand correctly, a section may have active experiments at different Decision Points. This compatibility query only takes `section_id` as scope, unlike the query cookbook (`docs/user/experiments/queries.md`), which scopes experiment analysis by `experiment_id`. Could we either add an `experiment_id` parameter here or explicitly document that the query picks the most recent assignment for an enrollment regardless of experiment? Otherwise a researcher could silently attach a condition from a different concurrent experiment to an outcome.

lib/oli/delivery/experiments/attempt_attributions.ex
```diff
@@ -213,33 +255,161 @@ defmodule Oli.Delivery.Experiments.AttemptAttributions do
...
+          []
+        )
+      )
+
+    Enum.map(selected, fn selected ->
+      Map.put(assignment_match, :activity_resource_id, selected.activity_resource_id)
+    end)
+  end
+
+  defp hydrate_assignment_match(match) do
```
nicocirio: `hydrate_assignment_match/1` here is nearly identical to the one in `media_attributions.ex`, apart from `section_slug`. Would it make sense to extract the common hydration step so the two projections cannot drift the next time their selected fields change?

lib/oli/experiments/runtime_assignment.ex
```diff
@@ -31,7 +31,17 @@ defmodule Oli.Experiments.RuntimeAssignment do
         metadata
       )
 
-      result
+      case result do
```
nicocirio, line 34: Small consistency note, not a bug: the single-assignment path receives `{:ok, decision}` for sticky reuse and therefore never calls its emission callback, while the batch path reaches `ConditionAssignmentEmitter` for the same case and relies on its `reused?: true` clause to return `:ok`. Both paths are covered and correct today. Would it be worth converging on one place to own that guard when this flow is next touched?

priv/clickhouse/queries/upgrade_v033_compatibility.sql
eliknebel: Addressed in 743387a02d. The compatibility query now filters on the canonical `completed` verb emitted by `ActivityAttemptEvaluated.new/3`, and the contract coverage asserts the same producer verb while using it in the compatibility fixtures.

priv/clickhouse/queries/upgrade_v033_compatibility.sql
eliknebel: Addressed in 743387a02d. The query now requires `experiment_id` and scopes assignment evidence by it. The fixtures include a newer assignment from a concurrent experiment to prove that it cannot supply the selected condition.

lib/oli/delivery/experiments/attempt_attributions.ex
eliknebel: Addressed in 743387a02d. Attempt and media attribution now share `AssignmentMatch.hydrate/1`; the focused test also verifies that caller-specific projection fields such as `section_slug` are preserved.

lib/oli/experiments/runtime_assignment.ex
eliknebel: Agreed that the two paths currently place the reuse guard at different layers. Both are covered and produce the intended exactly-once behavior, so I’d prefer not to reshape the assignment result/event flow in this already broad fix. We’ll keep this consolidation in mind the next time the runtime assignment flow is changed.

nicocirio (approved):
🚀

PR #6792, opened 2026-08-21
[BUG FIX] [MER-5887] Learning Objectives Authoring Elements Corrections

gastonabella:
## Summary

This PR completes the authoring-side corrections for the Learning Objectives Introduction/Summary feature originally introduced in Simon-Initiative/oli-torus#6745.

It focuses on the authoring requirements from:
- [MER-5802](https://eliterate.atlassian.net/browse/MER-5802): authoring insert menu support for the Learning Objectives element
- [MER-5803](https://eliterate.atlassian.net/browse/MER-5803): authoring support for Learning Objective Introduction
- [MER-5804](https://eliterate.atlassian.net/browse/MER-5804): authoring support for Learning Objective Summary

## Changes

- Updated the Insert menu `Objectives` entry to match the intended icon, placement, hover/focus behavior, and root-only insertion behavior.
- Refined the authoring layout for Learning Objective Introduction and Summary, including descriptions, mode selector placement, panel structure, spacing, warning placement, and proficiency accordion.
- Updated LO cards to match the expected authoring structure, including `LO X` labels, parent/sub-objective grouping, recommendation selectors, chips, disabled add actions, and tooltip behavior.
- Added non-destructive remove/restore behavior for objectives, preserving recommendation config while toggling `enabled`.
- Improved the empty-LO warning banner, including warning icon, dismiss action, colors, and no-false-warning behavior while LOs are unresolved.
- Added the shared trash icon treatment for authoring delete buttons and LO remove actions.
- Added authoring dark mode color fixes for LO surfaces, cards, warnings, chips, controls, and proficiency blocks.
- Added focused Jest coverage for insert behavior, mode switching, hierarchy rendering, remove/restore preservation, recommendation actions, empty warning behavior, and unresolved LO state.

## Notes

The student/delivery corrections from [MER-5867](https://eliterate.atlassian.net/browse/MER-5867) are intentionally left for a follow-up PR covering [MER-5807](https://eliterate.atlassian.net/browse/MER-5807) and [MER-5808](https://eliterate.atlassian.net/browse/MER-5808).

https://github.com/user-attachments/assets/ed2a4234-7650-4efa-850d-b1943da5ecec

[MER-5802]: https://eliterate.atlassian.net/browse/MER-5802?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MER-5803]: https://eliterate.atlassian.net/browse/MER-5803?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MER-5804]: https://eliterate.atlassian.net/browse/MER-5804?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MER-5867]: https://eliterate.atlassian.net/browse/MER-5867?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


assets/src/components/resource/editors/LearningObjectivesEditor.tsx
```diff
@@ -43,9 +92,11 @@ export const LearningObjectivesEditor = ({
   onRemove,
   resourceContext,
 }: LearningObjectivesEditorProps) => {
+  const learningObjectivesResolved = resourceContext.learningObjectives !== undefined;
```
nicocirio: I've been tracing through the false "no learning objectives" warning fix, and as far as I can tell, the guard here might be checking the wrong signal — though I could easily be missing something.

My understanding is that `learningObjectivesResolved` (`!== undefined`) assumes `resourceContext.learningObjectives` can be in three states — `undefined` (still loading), `[]` (resolved, genuinely empty), and `[...]` (resolved, has data) — but if I followed the code correctly, it only ever ends up as `[]` or `[...]` in practice, never `undefined`.

Specifically: `learningObjectivesForPage` on the backend (`lib/oli/authoring/editing/page_editor.ex:286-295`) looks like it returns `[]` — not `nil` — when the page doesn't have a `learning_objectives` element yet, so on a fresh insert `resourceContext.learningObjectives` would start as `[]`. Then in `NonActivities.tsx`, `resolveLearningObjectives` only reuses that value when `.length > 0`, so it seems to fall through to an async fetch. If that's right, while the fetch is in flight `resourceContext.learningObjectives` would still be `[]` — the same value as "resolved and genuinely empty" — so the warning could show up even though the container actually has objectives.

If that's the case, and there's a window during the fetch where we end up showing the warning, maybe in that case we should show a loading spinner instead — that would better represent to the user what's actually happening.

gastonabella: Good callout! Pushed 👍

nicocirio (approved):
🚀

PR #6794, opened 2026-08-24
[FEATURE] [MER-5846] New Learning Proficiency model, core implementation.   PR 2 / 3

darrensiegel:
## Summary

  This PR is the core implementation for the new LKT-AOA learning proficiency model. It adds the
  delivery-side write path that calculates and persists:

  - LKT-AOA proficiency, stored as `learning_states.aoa`
  - confidence, stored as `learning_states.confidence`

  This is PR 2 of 3 for the learning model v2 work. The previous PR added the data-model foundation:
  semantic model versioning, typed revision parameters, and global LKT-AOA configuration. This PR uses
  those pieces to actually apply the new model when evaluated student attempts are processed.

  The key integration point is `Oli.Delivery.Snapshots.Worker`. Existing summary analytics and xAPI
  behavior are preserved. The new LKT-AOA calculation only runs for Sections whose
  `learning_model_version` is `:lkt_aoa`. Existing `:naive` Sections skip this new write path.

  ## Why this PR exists

  The existing proficiency model is not sufficient for the new Learning Proficiency Model. The new
  model needs to update learner state after evaluated activity parts, including bulk cases like graded
  pages or exams with many parts.

  A critical design goal is performance: the write path must not scan historical attempts and must not
  do one database read/write per learning objective. A learner with 2 attempts and a learner with
  2,000 attempts should require the same bounded online state update shape.

  This PR implements that by maintaining compact operational state instead of recomputing from
  history.

  ## Core performance contract

  The FDD’s central performance requirement is:

  > One batch uses a small fixed number of set-based database operations. Database round trips do not
  grow with attempt history, activity-part count, or learning-objective fan-out.

  Some in-memory work naturally grows with the number of submitted parts and targeted objectives, but
  database access should remain bounded by operation type.

  Reviewers should specifically check that this PR does not introduce:

  - a historical PartAttempt scan
  - a per-learning-objective database loop
  - a per-part database loop
  - writes back to the large `part_attempts` table
  - unbounded telemetry/logging cardinality
  - LKT-AOA work for `:naive` Sections

  ## New tables / schemas

  ### `learning_states`

  One row per:

```text
  section_id, user_id, learning_objective_id
```

  This stores the compact state needed to continue the LKT-AOA model without reading historical
  attempts.

  Important fields:

  - attempt_count
  - success_score
  - failure_score
  - recency_logit
  - aoa
  - unique_activity_part_count
  - confidence

  aoa is the persisted proficiency value. This is the display-oriented LKT-AOA value from the
  technical spec, not raw P(correct).

  confidence is also persisted because it is derived from distinct activity-part breadth.

  ### prior_activity_part_evidence

  One row per:

  section_id, user_id, activity_id, part_id

  This table answers: “Has this learner already encountered this activity part in this Section?”

  It is intentionally not keyed by learning objective. One activity part may target multiple learning
  objectives. If the evidence insert succeeds, that new part can increase confidence breadth for each
  targeted learner/LO state. If the insert conflicts, the learner has already seen that activity part,
  so proficiency still updates, but confidence does not increase.

  ### learning_model_attempt_applications

  One row per applied PartAttempt:

  part_attempt_id, learning_model_version, applied_at

  This is the idempotency table.

  It exists so Oban retries, duplicate GUIDs, and downstream summary/xAPI failures do not double-apply
  proficiency or confidence. It is deliberately separate from part_attempts because part_attempts is
  very large in production and should not receive additional update churn.

  This table has no surrogate ID and no standard inserted_at / updated_at.

  ## Key modules

  ### Oli.LearningModel

  Public dispatch boundary.

  It decides what to do based on the loaded Section’s learning_model_version:

  - :naive → skip LKT-AOA
  - :lkt_aoa → apply the new model
  - unsupported value → error

  ### Oli.LearningModel.LktAoa.Application

  Transactional application service.

  This is the core write path. It:

  1. Claims eligible evaluated PartAttempts.
  2. Filters to newly claimed attempts only.
  3. Resolves published LO beta parameters.
  4. Initializes and locks affected learning_states.
  5. Inserts prior activity-part evidence with conflict semantics.
  6. Replays proficiency transitions in deterministic order.
  7. Applies confidence increments only from newly inserted evidence.
  8. Writes final learner state.
  9. Commits all of that atomically.

  If any step fails, the whole batch rolls back.

  ### Oli.LearningModel.LktAoa.Contribution

  Normalizes evaluated PartAttempts into model contributions.

  This handles:

  - part-to-LO objective mappings
  - activity-part beta extraction
  - LO beta extraction
  - binary correctness
  - evidence keys
  - state keys

  ### Oli.LearningModel.LktAoa.Transition

  Pure model math.

  This applies:

  - pre-response P(correct)
  - running AOA update
  - recency state update
  - confidence calculation

  This module should be reviewable independently from Ecto/database behavior.

  ### Oli.Delivery.Snapshots.Worker

  Integration point.

  Snapshot Worker now builds the evaluated attempt group once, then:

  1. applies LKT-AOA only if the Section is :lkt_aoa
  2. runs the existing summary analytics pipeline
  3. emits existing xAPI

  This order is intentional. If LKT-AOA succeeds but summary/xAPI later fails, retrying the job
  becomes safe because the attempt application table turns the second LKT pass into a no-op.

  ## How to review this PR

  Recommended review order:

  ### 1. Start with the docs

  Read:

  - docs/exec-plans/current/epics/learning_model_v2/core_impl/fdd.md
  - docs/exec-plans/current/epics/learning_model_v2/core_impl/informal.md

  Focus on the performance and idempotency requirements. The code should match those constraints.

  ### 2. Review the migration and schemas

  Review:

  - operational table migration
  - LearningState
  - PriorActivityPartEvidence
  - AttemptApplication

  Check:

  - composite keys are correct
  - no unnecessary indexes were added
  - learning_model_attempt_applications has exactly the intended fields
  - cascade behavior from part_attempts is correct
  - constraints enforce non-negative/probability fields where expected

  ### 3. Review pure model math

  Review:

  - Oli.LearningModel.LktAoa.Transition
  - transition tests

  Check:

  - P(correct) is calculated from pre-response state
  - AOA is updated before recency state changes
  - binary correctness is score == out_of
  - confidence is based on distinct activity-part count
  - no Repo/database access exists in pure transition code

  ### 4. Review contribution normalization

  Review:

  - Oli.LearningModel.LktAoa.Contribution

  Check:

  - activity-part beta comes from the exact PartAttempt activity Revision
  - LO beta comes from the Section-published LO Revision
  - missing beta parameters resolve to 0.0
  - invalid parameter shapes fail clearly
  - multi-LO parts produce the right state/evidence keys

  ### 5. Review the transaction

  Review:

  - Oli.LearningModel.LktAoa.Application

  This is the most important file.

  Check the transaction sequence carefully:

  - claim attempts first
  - only returned claimed IDs contribute
  - initialize then lock learner states
  - evidence insert uses ON CONFLICT DO NOTHING RETURNING
  - confidence increments only from returned evidence rows
  - final states are written in bulk
  - all of this is one transaction
  - there is no per-LO database loop
  - LO Revision fetch is narrow and does not load full Revision content

  ### 6. Review Snapshot Worker integration

  Review:

  - Oli.Delivery.Snapshots.Worker
  - worker tests

  Check:

  - :naive Sections do not run LKT-AOA calculation or telemetry
  - :lkt_aoa Sections do run it
  - summary analytics still run
  - xAPI behavior is preserved
  - downstream retry does not double-apply LKT-AOA state

  ### 7. Review performance tests

  Review:

  - query-count tests
  - concurrency tests
  - telemetry tests

  Check:

  - one attempt and large batch use the same operational query count
  - duplicate retries short-circuit after claim
  - concurrent workers do not double-count evidence or lose attempts
  - telemetry contains aggregate counts only, not learner IDs, GUIDs, responses, SQL binds, or
    parameter payloads

  ## What this PR intentionally does not do

  This PR does not:

  - change the default model for new Projects or Sections
  - expose model selection in the UI
  - implement usage/read-side proficiency displays
  - aggregate parent/page/container/course proficiency
  - train or upload model parameters
  - backfill historical learner state
  - replace xAPI, S3, ClickHouse, or existing summary analytics

  Those are separate pieces of the overall learning model v2 effort.

  ## Verification

  Key test coverage includes:

  - pure transition tests
  - contribution normalization tests
  - operational storage tests
  - transaction/idempotency tests
  - concurrency tests
  - Snapshot Worker integration tests
  - performance/query-count tests
  - telemetry privacy tests

  The most important behavioral assertions are:

  - :naive Sections create no LKT-AOA operational rows
  - :lkt_aoa Sections update learning_states
  - retries do not double-apply attempts
  - repeated activity parts update proficiency but not confidence
  - multi-part and multi-LO batches are handled in one collection-oriented operation
  - transaction failures roll back claims, evidence, and state together


lib/oli/delivery/snapshots/worker.ex
```diff
@@ -61,38 +72,52 @@ defmodule Oli.Delivery.Snapshots.Worker do
...
-          |> Oli.Analytics.Common.to_jsonlines()
+    attempt_group = AttemptGroup.from_attempt_summary(results, project_id, host_name())
+
+    # LKT-AOA is applied before summary/xAPI writes so a downstream failure can
+    # safely retry through the exact attempt-claim projection instead of double
+    # applying learner state. The AttemptGroup is intentionally built once and
+    # reused so the learning model, summaries, and xAPI share the same evaluated
+    # attempt set.
+    with {:ok, _learning_model_result} <-
+           LearningModel.apply_evaluated_attempts(section, attempt_group),
```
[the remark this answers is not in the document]
darrensiegel: This isn't possible.

lib/oli_web/live/manual_grading/manual_grading_view.ex
```diff
@@ -769,6 +770,8 @@ defmodule OliWeb.ManualGrading.ManualGradingView do
 
       case ManualGrading.apply_manual_scoring(section, attempt, score_feedbacks) do
         {:ok, finalized_part_attempt_guids} ->
+          Snapshots.queue_or_create_snapshot(finalized_part_attempt_guids, section.slug)
```
nicocirio: I think this ends up queuing the Snapshot Worker twice for the same manual-grading submission: `ManualGrading.apply_manual_scoring/3` already routes through `ApplyClientEvaluation.apply/5`, which internally calls `Snapshots.maybe_create_snapshot/3` (`apply_client_evaluation.ex:117`) for these same part attempts. On top of that, `Worker` isn't declared with a `unique:` Oban option, so nothing collapses the two resulting jobs into one. Is this intentional? If not, this line might be safe to remove. If it is intentional, it'd be great to have a test pinning down how many jobs/analytics writes a single manual-grading submission produces.

lib/oli/delivery/snapshots/worker.ex
```diff
@@ -61,38 +73,62 @@ defmodule Oli.Delivery.Snapshots.Worker do
...
-      {:ok, %Pipeline{data: attempt_group}} ->
-        body =
-          StatementFactory.to_statements(attempt_group)
-          |> Oli.Analytics.Common.to_jsonlines()
+  defp apply_learning_model(_section, nil), do: {:ok, BatchResult.new(:noop)}
 
-        bundle_id = create_bundle_id(attempt_group)
+  defp apply_learning_model(_section, %AttemptGroup{} = attempt_group) do
+    {:ok,
+     BatchResult.new(:skipped, input_attempt_count: length(attempt_group.part_attempts || []))}
```
nicocirio: `apply_learning_model/2` ends up re-implementing the `:naive`/`:lkt_aoa`/`nil` dispatch that `Oli.LearningModel.apply_evaluated_attempts/2` already owns (its moduledoc calls it "the public delivery-side learning-model application boundary"). Since this only delegates to `LearningModel` for `:lkt_aoa`, its `:naive` and "unsupported version" clauses never actually get exercised from this call site. Would it make sense to simplify this to always call `LearningModel.apply_evaluated_attempts(section, attempt_group)` and let that module own the full dispatch? That way, if a third model version shows up down the line, there's only one place to update.

test/oli/learning_model/lkt_aoa/concurrency_test.exs
```diff
@@ -0,0 +1,107 @@
+defmodule Oli.LearningModel.LktAoa.ConcurrencyTest do
+  use Oli.DataCase
+
+  alias Oli.LearningModel
+  alias Oli.LearningModel.{AttemptApplication, LearningState, PriorActivityPartEvidence}
+  alias Oli.LearningModel.LktAoaFixtures
```
nicocirio, line 6: A question about these tests: they don't declare `async: true`, so by default (`data_case.ex:32-40`) they run under `Sandbox.mode(Oli.Repo, {:shared, self()})` — the "concurrent" `Task`s share the test's own DB connection, so I'm not fully sure this exercises real cross-transaction locking, or whether that's an intentional pattern here. Could you confirm? Separately, all three tests in this file start from a state that doesn't exist yet in the DB — would it be worth adding a case that first commits a state row and then races two concurrent updates against that already-existing row? That's a scenario the FDD calls out explicitly that I didn't see covered.

test/oli/learning_model/lkt_aoa/performance_test.exs
```diff
@@ -0,0 +1,112 @@
...
+        assert {:ok,
+                %{
+                  status: :applied,
+                  input_attempt_count: @large_batch_size,
+                  claimed_attempt_count: @large_batch_size
+                }} = LearningModel.apply_evaluated_attempts(large_section, large_group)
+      end)
+
+    assert single_count == large_count
+    assert large_count <= 6
```
nicocirio: The FDD asks for query-count verification with batches of 500-1000 PartAttempt GUIDs, and `@large_batch_size` here only reaches 500 — could we bump it to 1000, or add a second case at that size? Also noticed `assert large_count <= 6` is a one-sided bound; if the count ever dropped unexpectedly (e.g. a step got skipped), this assertion wouldn't catch it. Might be worth asserting an exact value instead?

test/oli/learning_model/lkt_aoa/concurrency_test.exs
darrensiegel: Confirmed: these tasks currently use the shared SQL Sandbox connection, so they do not exercise independent database transactions. I left the existing suite intact in this follow-up rather than add a misleading test; a real cross-transaction test needs dedicated fixture setup outside DataCase’s rollback sandbox. I will cover that with the appropriate isolation harness separately.

nicocirio (approved):

Francisco-Castro:
@darrensiegel Late to the party here 🙂 — I started reviewing this PR but didn't get the chance to post my findings before the merge landed, so leaving them here for the follow-up:

**1. Learner answers can end up stored in Oban's error records.**
When a batch fails validation, the fallback error is `{:invalid_part_attempt, part_attempt}` — the whole struct. The worker returns it as-is, and Oban writes `inspect(reason)` into `oban_jobs.errors`. So one malformed row (I reproduced it with an evaluated attempt whose score was nil) puts a student's `response`, `feedback`, and `hints` as plain text into a table nobody treats as student data. TThe telemetry side of this is done really well — [FDD §12](https://github.com/Simon-Initiative/oli-torus/blob/9464af00c035391104b0a0ff6fd92bdf32589a1f/docs/exec-plans/current/epics/learning_model_v2/core_impl/fdd.md#12-security--privacy)'s aggregate-only rule is enforced with tests, key by key. The error-return path is just a second channel to the same kind of storage that slipped past that same discipline.

**2. "Apply update" can silently corrupt model state on a live section.**
Imagine a pilot section where question Q is tagged to LO A; students answer it and build state. The author realizes Q really teaches LO B, republishes, and the instructor clicks apply update — a completely routine action. From then on, A keeps proficiency and confidence it no longer deserves, and B never gets confidence credit for Q because the evidence row (keyed without objective) already exists. No error anywhere; the reports just go quietly wrong. [FDD §16](https://github.com/Simon-Initiative/oli-torus/blob/9464af00c035391104b0a0ff6fd92bdf32589a1f/docs/exec-plans/current/epics/learning_model_v2/core_impl/fdd.md#16-open-questions--follow-ups) makes a reconciliation workflow a precondition for exactly this, but `apply_publication_update/2` has no learning-model guard. Either guard the update path for `:lkt_aoa` sections, or write "pilots don't receive publication updates" into the rollout rules — just not the current open, silent door.

**3. Objective resolution fails hard on incomplete publication data — and takes the old analytics down with it.**
`resolve_objective_revisions/3` looks up every objective in one `context.publication_id`. That publication is chosen correctly for normal traffic (it comes from the attempted page's own project), so ordinary and remixed sections resolve fine. The concern is what happens when an objective's `published_resources` mapping is absent or dangling: the batch fails with `missing_published_objective_revisions`, and because LKT runs first in the worker, the retries exhaust and that batch loses proficiency, summary analytics, *and* xAPI — data the legacy path processed without complaint. Two adjacent things surfaced while verifying this: `pub_id_for_section_project(section_id, nil)` matches `[{pub_id}]` against a scalar select, so that clause crashes if ever reached; and the resolve query doesn't filter `revision.deleted`, so soft-deleted objectives still resolve and quietly accumulate LKT state. Suggested first step: a remixed-page LKT test to pin the working behavior, plus a decision on whether a missing mapping should stay a hard error, and fixes for the two side items.

**4. The batch counts are in the wrong telemetry compartment.**
[FDD §11](https://github.com/Simon-Initiative/oli-torus/blob/9464af00c035391104b0a0ff6fd92bdf32589a1f/docs/exec-plans/current/epics/learning_model_v2/core_impl/fdd.md#11-observability) puts duration *and* the five batch counts in measurements, with metadata limited to bounded labels. The code emits only duration as a measurement and all five counts as metadata — and `telemetry_test.exs` pins that inverted shape. To be fair about impact: reporters can still reach metadata (`Telemetry.Metrics` allows it), and metadata doesn't become tags unless configured that way — so this is spec conformance and metric ergonomics rather than a live incident. The reason to fix it now is timing: the counts move to measurements with a few lines today, while after rollout dashboards are configured against the current shape, each of them moves too.

PR #6806, opened 2026-08-28
[ENHANCEMENT] [MER-5790] Populate page level learning objectives based on activities

manelli:
## Summary

- Include learning objectives from activities that enrolled students have actually answered, including dynamically selected activity-bank questions that are not statically attached to a page.
- Union those realized objectives with the existing statically contained objectives for course and unit/module scopes.
- Preserve the exact page containment relationship so a shared bank candidate appears only beneath the unit where a learner encountered it.
- Read from `ResourceSummary`, `ResourcePartResponse`, and `StudentResponse` instead of traversing the attempt hierarchy.

## Why

Activity-bank selections store selection criteria rather than every activity that may eventually be chosen. Static containment therefore cannot know which banked activity appeared on a particular page.

Once an enrolled learner answers an activity, the analytics summary records provide the realized learner/page/activity relationship. Instructor responses are excluded, and activity objectives are resolved from the current publication.

## Verification

The regression scenario covers:

- Static objectives are available before any learner response.
- An instructor response does not add a realized bank objective.
- An enrolled learner response adds the bank objective to Unit 1 and the entire course.
- The same eligible bank activity does not leak into Unit 2 when it was not answered there.

## Ticket

[MER-5790](https://eliterate.atlassian.net/browse/MER-5790)

[MER-5790]: https://eliterate.atlassian.net/browse/MER-5790?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli/delivery/sections.ex
+++ b/lib/oli/delivery/sections.ex
@@ -55,7 +55,14 @@ defmodule Oli.Delivery.Sections do
   alias OliWeb.Common.FormatDateTime
   alias Oli.Delivery.PreviousNextIndex
   alias Ecto.Multi
-  alias Oli.Delivery.Attempts.Core.{ResourceAccess, ResourceAttempt}
+
+  alias Oli.Delivery.Attempts.Core.{
+    ActivityAttempt,
+    PartAttempt,
+    ResourceAccess,
+    ResourceAttempt
+  }
+
   alias Oli.Delivery.Metrics
   alias Oli.Delivery.Paywall
   alias Oli.Delivery.Sections.PostProcessing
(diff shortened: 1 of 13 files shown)


darrensiegel (dismissed):
As a mechanism to determine "which activities have actually been answered by students", the approach taken in function `realized_objective_container_pairs` will certainly face performance challenges on Proton, given that it queries over the attempt hierarchy.

Instead we can get the same result in a much more efficient manner by querying over the `ResourceSummary` records.

manelli:
@darrensiegel Addressed in commit 42678e7c72. The realized-objective lookup now starts from per-student ResourceSummary records instead of traversing the attempt hierarchy, while the response summary joins preserve the exact student/page/activity relationship. The regression scenario also verifies that an answered bank activity contributes its objective only to the unit where it was answered, without leaking into another eligible unit.

manelli:
Regarding the course-scope concern: no filtering change is needed. A nil container_id is the deliberate course-root sentinel in both ContainedObjective and ContainedPage, so course-level lookup includes the root containment records rather than returning an empty scope. The regression scenario exercises and passes the course-level assertion. Skipping containment filtering for nil would instead broaden the result to objectives outside the course containment tree.

manelli:
Regarding the historical-publications concern: SectionsProjectsPublications represents the current publication for each section/project pair, enforced by its unique section_id/project_id index; it is not a history of every publication version. The PublishedResource join additionally requires the exact activity resource_id, and the final objective/container pairs are deduplicated. The query therefore does not expand each response across historical publications.

eliknebel (approved):

nicocirio (approved):

PR #6813, opened 2026-09-01
[BUG FIX] [MER-5863] Allow for restoring activities on practice page after customizations removed all

andersweinstein:
## Problem

[Ticket](https://eliterate.atlassian.net/browse/MER-5863) When instructor customizations remove all activities from an ungraded practice page, the learner's next attempt is created from customized, activity-free page content. If the instructor later restores any of those activities, the learner remains pinned to the existing empty attempt, so the restored activities do not appear.

Normally a learner can create a fresh practice attempt with the Reset Answers button. In this case, however, the realized attempt has zero activities, so the button was not rendered. The learner therefore had no UI path to discard the empty attempt and pick up restored activities.

## Fix

Preserve the existing learner-driven reset workflow and make its action available when either the current attempt has realized activities or the authored page contains an embedded activity reference or activity-bank selection.

When realized activities are present, the action remains Reset Answers. When instructor customizations have reduced the realized activity count to zero but the authored page has activity opportunities, the action is labeled Refresh Activities. Activating it finalizes the empty practice attempt; the subsequent visit creates a fresh attempt using the current instructor customizations, making restored activities available.

The authored-content check is shared by the LiveView and legacy delivery paths. True content-only pages continue to omit the action.

## Summary

- detect embedded activity references and activity-bank selections in authored page content
- expose Refresh Activities when customizations reduce a practice attempt to zero activities
- retain Reset Answers for normal practice attempts with realized activities
- cover embedded activities, bank selections, content-only pages, and explicit refresh after restoration

## Verification

- mix test test/oli/resources/page_content_test.exs test/oli_web/live/delivery/student/utils_test.exs test/oli_web/live/delivery/student/lesson_live_test.exs test/scenarios/instructor_customizations/instructor_customizations_test.exs
- 159 tests, 0 failures
- scenario schema validation passed
- mix format on changed Elixir, HEEx, and test files
- git diff --check


darrensiegel (changes requested):
This approach is too invasive for a bug currently classified as low priority. It modifies the core ungraded page-attempt lifecycle, introduces row locking and new concurrency behavior, automatically supersedes attempts during page visits, and adds repeated activity realization to the delivery hot path. Those changes carry considerably more correctness and performance risk than the original issue warrants.

A substantially lower-risk solution is to preserve the existing recovery workflow and ensure the learner can always access the “Reset Answers” / “Reset all activities” action when the page’s authored revision contains an activity opportunity (even when instructor customization has reduced the realized activity count to zero).

Suggested technical approach:

1. Keep the lightweight PageContent.contains_activity_opportunity?/1 helper introduced by this PR. It should detect both embedded "activity-reference" nodes and activity-bank "selection" nodes.
2. Change the reset-button condition from effectively:
```
@activity_count > 0
```

to:

```
     @activity_count > 0 or
       PageContent.contains_activity_opportunity?(@page_context.page.content)
```

The check must use the authored page revision content, not the customized/transformed resource_attempt.content.
3. Apply the equivalent condition to both the LiveView component and the legacy page-delivery template.
4. Remove the automatic empty-attempt recovery, ResourceAccess FOR UPDATE locking, latest-attempt refetching, and related hierarchy lifecycle changes.

andersweinstein:
> A substantially lower-risk solution is to preserve the existing recovery workflow and ensure the learner can always access the “Reset Answers” / “Reset all activities” action when the page’s authored revision contains an activity opportunity (even when instructor customization has reduced the realized activity count to zero).

OK. One concern with this approach is that if the page has no activities, a student will never discern any reason to click "Reset Answers"  -- for there are no answers visible to reset. They might still be told to do this by an instructor, but to mitigate this, I have adjusted the text in the case of zero activities to read "Refresh Activities" rather than "Reset Answers". Other even more explanatory text would be possible too, but this seems adequate.

andersweinstein:
[Codex:] Addressed in fe5d2a0b4a. I removed the automatic empty-attempt recovery, ResourceAccess locking/refetching, and hierarchy lifecycle changes. The existing reset/finalize workflow is now available when authored content contains an embedded activity reference or bank selection even if the realized activity count is zero, in both LiveView and legacy delivery. The zero-activity action is labeled Refresh Activities; normal attempts retain Reset Answers. The scenario now verifies explicit refresh after restoring both embedded and bank-selected activities.

darrensiegel (approved):

PR #6815, opened 2026-09-01
[FEATURE] [MER-5846] Learning model v2, PR 3/3

darrensiegel:
Introduces the usage of the new learning proficiency model, through a uniform interface `Oli.Delivery.Proficiency` which dispatches to two concrete implementations in `Oli.Delivery.Proficiency.Naive` and `Oli.Delivery.Proficiency.LktAoa`.    Several calling spots were updated, but mainly in the Metrics module and in the oracles for the learning dashboard. 

This work had to ensure that we properly had SectionResource records capturing the learning objective hierarchy.  It was discovered that not all paths that created and updated section resource records were setting the `children` attribute of parent objectives.  This work adds a new JIT section resource migration, and formalizes that JIT migration to be based on a persisted migrated version in the Section table.


lib/oli/delivery/sections/section_resource_migration.ex
```diff
@@ -1,13 +1,165 @@
...
+
+  @doc """
+  Upgrades a legacy Section immediately before its depot table is loaded.
+
+  The row lock serializes concurrent first access. The version is re-read under
+  that lock because the value observed before waiting may already be stale.
+  """
+  @spec ensure_current(integer()) ::
+          {:ok, :current | :migrated} | {:error, term()}
+  def ensure_current(section_id) do
```
gastonabella, line 38: `ensure_current/1` locks `sections` before writing `section_resources`, but `project_current/1` writes `section_resources` first and only locks `sections` at the very end (in `set_version`). If a JIT migration (depot miss) and an ordinary post-processing run (remix/blueprint/publication update) hit the same section concurrently, this is a textbook ABBA lock-order inversion — Postgres will deadlock-abort one of them, and today that raises unhandled instead of returning `{:error, reason}`. Could `project_current/1` take the same `SELECT ... FOR UPDATE` on `sections` as its first statement, mirroring `ensure_current/1`?

lib/oli/learning_model/lkt_aoa/application.ex
gastonabella, line 250: This changes what happens when an objective's published revision can't be resolved during attempt application — previously the whole claim rolled back (hard, retryable failure), now unresolved objectives are silently dropped and the attempt is permanently claimed as applied. That's a write-path policy change, not a read-path addition, and it sits outside this PR's stated read-only scope for `learning_states`. Could you confirm this was intentional? A cheaper middle ground than reverting everything: only hard-fail when a contribution loses *all* its objectives (total evidence loss, permanent since the attempt gets claimed either way) — partial loss (one objective among several) seems fine to skip as-is. At minimum, this drop should be logged/counted so it's not silent.

lib/oli/scenarios/directives/assert/proficiency_assertion.ex
```diff
@@ -131,14 +125,9 @@ defmodule Oli.Scenarios.Directives.Assert.ProficiencyAssertion do
...
-        calculate_student_proficiency(section, objective, state, proficiency)
-
-      # Average proficiency across all students
-      true ->
-        calculate_average_proficiency(section, objective, state, proficiency)
+    case proficiency.student do
+      nil -> calculate_average_proficiency(section, objective, state, proficiency)
+      _student -> calculate_student_proficiency(section, objective, state, proficiency)
     end
   end
```
gastonabella: `calculate_proficiency/4` only branches on `proficiency.student` (student vs. class average) — it never reads `proficiency.page`/`proficiency.container`, even though the schema and parser accept them. Neither `calculate_student_proficiency/4` nor `calculate_average_proficiency/4` below call the scope-aware `Proficiency.estimates_for_scopes/4` / `scope_aggregates/3` — only the objective-scoped APIs. A scenario with `page: "..."` today silently evaluates the whole course instead. Could we wire `page`/`container` through to the scope APIs here, or reject them at parse time until supported?

lib/oli/delivery/metrics.ex
```diff
@@ -768,113 +770,24 @@ defmodule Oli.Delivery.Metrics do
...
-      }
-    )
-    |> Repo.all()
-    |> Map.new()
-  end
+  defp estimate_label(nil), do: "Not enough data"
+  defp estimate_label(%{label: :low}), do: "Low"
+  defp estimate_label(%{label: :medium}), do: "Medium"
+  defp estimate_label(%{label: :high}), do: "High"
+  defp estimate_label(_estimate), do: "Not enough data"
```
gastonabella, line 790: `:unavailable` and `:not_enough_information` are distinct labels on purpose (per the FDD: "unavailable... never [maps] to a naive result"), but the catch-all on line 790 routes `:unavailable` through the same path as `:not_enough_information` → `"Not enough data"`. If the LKT-AOA provider degrades for a section, students with real evidence would silently look like brand-new students instead of surfacing an outage. The identical gap exists in `ObjectivesProficiency.label/1` (lib/oli/instructor_dashboard/oracles/objectives_proficiency.ex:85-88) — worth fixing both, ideally by adding an explicit `:unavailable` clause in each, or exposing one canonical converter on the `Proficiency` boundary so they can't drift again.

lib/oli/scenarios/directives/assert/proficiency_assertion.ex
gastonabella, line 132: `calculate_proficiency/4` only branches on `proficiency.student` (student vs. class average) — it never reads `proficiency.page`/`proficiency.container`, even though the schema and parser accept them. Neither `calculate_student_proficiency/4` nor `calculate_average_proficiency/4` below call the scope-aware `Proficiency.estimates_for_scopes/4` / `scope_aggregates/3` — only the objective-scoped APIs. A scenario with `page: "..."` today silently evaluates the whole course instead. Could we wire `page`/`container` through to the scope APIs here, or reject them at parse time until supported?

lib/oli/learning_model/lkt_aoa/application.ex
darrensiegel: Intentional.

lib/oli/delivery/metrics.ex
darrensiegel: "Unavailable" is not an actual UX requirement

gastonabella (approved):

PR #6816, opened 2026-09-01
[ENHANCEMENT] [MER-5888]  LOs Delivery Corrections

gastonabella:
## Summary

This PR delivers the student-facing and instructor-preview corrections for the Learning Objectives Introduction and Summary elements originally shipped in #6745.

It covers the delivery requirements from:
- [MER-5807](https://eliterate.atlassian.net/browse/MER-5807): Student Learning Objective Introduction element
- [MER-5808](https://eliterate.atlassian.net/browse/MER-5808): Student Learning Objective Summary element

## Changes

- Restructured the Summary to group objectives into two sections: *Learning Objectives You're Applying* (strong proficiency) and *Recommended Review* (beginning/growing proficiency). Each section is omitted when empty. Objectives with no proficiency data are excluded entirely since there is no learning signal to act on yet.
- Added *Show next steps* / *Hide next steps* expand/collapse behavior for review objectives with configured resources. The panel shows Revisit and Practice links only when resources are available; the toggle is hidden otherwise. Strong-proficiency objectives never show next steps.
- Added smooth open/close animation for the next-steps panel, with no animation for users who prefer reduced motion.
- Scoped recommendation page candidates to pages that exist in the authoring hierarchy, preventing authors from selecting pages that would silently disappear in student delivery after publication.
- Added a search input to the recommendation page selector modal so authors can filter candidates by title.
- Added a DOT Explain card inside the expanded next-steps panel when the assistant is available and AI Activation Points are enabled for the section. When both pages and the Explain card are absent, the toggle is suppressed entirely.
- Fixed student recommendation links that were causing 500 errors by resolving through revision slugs when available.
- Fixed instructor preview: the LO element previously rendered empty or errored for authors; preview context now precomputes the same learning objectives payload used in student delivery.
- Updated proficiency icons and section icons to match the design.
- Removed visible *Sub-Objective* label text from student-facing output.
- Fixed dark mode for summary cards, section styling, and the recommendation page-picker modal.
- Added test coverage for summary grouping, next-step availability, DOT card toggling, strong-proficiency suppression, recommendation page scoping, and preview context.

## Notes

Mobile responsive layout for the Summary and Introduction elements is intentionally deferred to a follow-up PR.

https://github.com/user-attachments/assets/d4619b64-af1c-4c2e-b519-5b7317e4d8b8

[MER-5807]: https://eliterate.atlassian.net/browse/MER-5807?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MER-5808]: https://eliterate.atlassian.net/browse/MER-5808?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gastonabella:
**Review comment follow-up**

**Implemented in this PR:**
- `SelectModal`: `clearAriaLabel` prop added (default `"Clear selection"`); `LearningObjectivesEditor` passes `"Clear selected page"` — fixes the hardcoded page-specific label flagged by the UI and TypeScript bots
- `SelectModal`: removed duplicate `btn-close` class (Bootstrap background-image ×) from the modal close button — the Font Awesome icon already provides the glyph
- `LearningObjectivesEditor.scss`: added `@media (prefers-reduced-motion: reduce)` block disabling the `::details-content` and chevron transitions
- `explain_objective_button.ts`: capped the exponential backoff at 60 s (`Math.min(delay * 2, MAX_DELAY)`)
- `explain_objective_button.ts`: `.catch` on `Trigger.invoke` — re-enables the button and resets delay on failure

**Deferred to follow-up:**
- **Combobox keyboard navigation** (`SelectModal.tsx`): valid, but a cross-cutting effort (affects all `SelectModal` consumers) — better handled in a dedicated accessibility ticket
- **Base64 PNG in proficiency cards** (`learning_objectives.ex`): pre-existing before this PR; moving it to a static asset is a standalone cleanup
- **Full hierarchy loading in `resource_controller.ex`**: also pre-existing; a tailored query would be an optimization ticket on its own

lib/oli_web/delivery/instructor/preview_page_context.ex
```diff
@@ -466,6 +467,17 @@ defmodule OliWeb.Delivery.Instructor.PreviewPageContext do
...
+    render_context = %Context{
+      render_context
+      | learning_objectives:
+          LearningObjectivesPageElement.prepare_render_payload(
+            section,
+            revision.resource_id,
+            content,
+            nil
+          )
+    }
```
nicocirio, line 479: This fixes a real bug (`learning_objectives` was `nil` in preview render context), but I don't see a test anywhere that exercises the instructor-preview path with a `learning_objectives` element — the existing coverage in `learning_objectives_test.exs` only goes through student delivery. Could we add a small regression test here (either unit-level on `PreviewPageContext` or through the preview LiveView) so this doesn't silently regress?

lib/oli/rendering/content/learning_objectives.ex
```diff
@@ -8,11 +8,10 @@ defmodule Oli.Rendering.Content.LearningObjectives do
   alias Oli.Delivery.Sections.SectionResourceDepot
   alias Oli.Rendering.Context
   alias Oli.Rendering.Content.UrlHelpers
-  alias OliWeb.Icons
   alias Phoenix.HTML
-  alias Phoenix.HTML.Safe
 
   @page_type_id Oli.Resources.ResourceType.id_for_page()
+  @default_proficiency ProficiencyDisplay.default_label()
```
nicocirio, line 14: It took me a bit to understand why this renders as concatenated HTML string lists instead of a HEEx component, and why the styling here uses named CSS classes in `app.css` instead of inline Tailwind like the HEEx components I'm used to in this codebase. Digging into it, my best guess is: `learning_objectives` is just another block in the same flat page-content list as `activity-reference` (see `Oli.Rendering.Elements.render/3`), and MER-5807's AC requires that authors be able to place this component anywhere on the page — including between two activities. Since the whole page content gets flattened into one HTML string and injected as a single `phx-update="ignore"` island (LiveView never diffs it), pulling this out into a standalone HEEx component isn't really an option, and without HEEx, named CSS classes are more maintainable here than repeating long Tailwind strings across every `~s|...|`. If that's indeed the reasoning, would it be worth a short note in the `@moduledoc` capturing it, so the next person reading this file doesn't have to rediscover it?

nicocirio (approved):

lib/oli_web/delivery/instructor/preview_page_context.ex
gastonabella: Good catch. Added a test in `PreviewLessonLiveTest` that mounts the instructor preview LiveView with a page whose content includes a `learning_objectives` element and asserts the rendered output contains expected objective text. Covers the nil-payload path that was broken.

lib/oli/rendering/content/learning_objectives.ex
gastonabella: Your reading is exactly right. Added a `@moduledoc` with two sections: one explaining why this renders as flat HTML strings (the `phx-update="ignore"` island model makes HEEx incompatible), and one explaining why named CSS classes are preferred over inline Tailwind here — embedding long utility chains across many `~s|...|` string literals hurts readability, so the styles live in `app.css` instead.

PR #6817, opened 2026-09-01
[FEATURE] [MER-5821] Aggregate child Learning Objective proficiency into parent Learning Objectives

nicocirio:
## Jira

- Ticket: https://eliterate.atlassian.net/browse/MER-5821

## Summary

Parent Learning Objectives now show an aggregated proficiency that combines their own
directly-tagged evidence (when present) with all their Sub-LOs' evidence, instead of showing
"not enough data" whenever a parent has no direct evidence of its own. The exact aggregation
rule ("Option B": always combine, accepting that an activity tagged to both a parent and a
Sub-LO contributes to both) was confirmed with the ticket owner after implementation surfaced a
genuine ambiguity in the original acceptance criteria — see
`docs/exec-plans/current/epics/learning_model_v2/lo_aggregation/parent_evidence_aggregation_options.md`
for the alternatives considered.

## What Was Added

- `Metrics.aggregate_weighted_proficiency/1`: a shared, algorithm-agnostic weighted-average
  aggregation function.
- `Metrics.evidence_resource_ids/2` and `Metrics.proficiency_bucket_for_student/3`: the shared
  always-combine rule and bucketing, now used identically by all three real call sites — student
  delivery pages, the Instructor Dashboard / per-student view / CSV export, and the
  "learning_objectives" page element.
- Consolidated the pre-existing student-facing proficiency calculation onto this shared
  pipeline, removing the dead code it replaced.
- A bug fix in `Oli.Scenarios.Builder`: a scenario-built parent objective's cached revision
  went stale once its Sub-LOs were attached, found while adding test coverage below.
- New `Oli.Scenarios`-based tests that exercise the real attempt-submission pipeline instead of
  factory-inserted `ResourceSummary` rows, including a LiveView/controller suite that checks the
  actual rendered Instructor Dashboard tab, per-student view, and CSV export.

## Validation

- `mix format`, `mix compile --warnings-as-errors` clean.
- `yarn format`, `yarn check-types`, `yarn deploy` clean.
- Automated test suite green (unit, Instructor Dashboard/student LiveView, and the full
  `Oli.Scenarios` suite — 750 tests, 0 failures).
- Manual QA and the Jira status update are still open — this PR covers implementation and
  automated test coverage.

### DIFF
--- a/docs/exec-plans/current/epics/learning_model_v2/lo_aggregation/parent_evidence_aggregation_options.md
+++ b/docs/exec-plans/current/epics/learning_model_v2/lo_aggregation/parent_evidence_aggregation_options.md
@@ -1,0 +1,125 @@
+# Parent LO Proficiency: How Should a Parent's Own Directly-Tagged Evidence Be Combined With Its Sub-LOs?
+
+## Context
+
+When a parent Learning Objective has one or more Sub-LOs, how should proficiency be aggregated at
+the parent level? Depending on how we choose to weigh the effect of tagging an activity directly to
+the parent (rather than to one of its Sub-LOs), we arrive at different possible results — and each
+approach carries its own difficulties and/or trade-offs, laid out below.
+
+This question specifically matters for **existing** course content. A separate, already-filed
+ticket will restrict authors from tagging activities directly to a parent LO once it has Sub-LOs,
+for newly authored course packages going forward. That upcoming restriction does not apply
+retroactively, so any course that has already tagged activities directly to a parent LO will keep
+that tagging as-is. This discussion is about how proficiency should be aggregated for that existing
+content, not about the tagging rules for new courses.
+
+All four options below rely on `Oli.Analytics.Summary.ResourceSummary`, which stores one row per
+`(project_id, section_id, user_id, resource_id, resource_type_id, part_id)`, with running totals
+(`num_attempts`, `num_correct`, `num_first_attempts`, `num_first_attempts_correct`) incremented as
+attempts occur. Critically, **this row does not retain which specific activity contributed to a
+given increment** — counts from every activity ever tagged to that `resource_id` are folded
+together at write time. As the options below show, this is the central constraint driving the
+difficulty and trade-offs of each approach.
+
+## Option A — Exact reconciliation (activity-level exclusion)
+
+**Logic:** For the parent, determine which activities are tagged *only* to the parent (not to any
+of its Sub-LOs), and compute the parent's own contribution using only attempts on those activities.
+Any activity tagged to both the parent and a Sub-LO is attributed entirely to the Sub-LO and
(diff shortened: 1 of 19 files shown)


nicocirio:
Re: **AI Review — performance** (proficiency_bucket_for_student/3, pairs list allocation)

`resource_ids` here is bounded (a parent plus its handful of Sub-LOs, not hundreds), and this runs once per (objective, student) cell over already-fetched data, not in a query loop. `aggregate_weighted_proficiency/1` is intentionally a separate, independently-tested shared function (single algorithm-agnostic aggregator used by all three call sites) — inlining it here would duplicate that logic for a negligible GC gain on a tiny list. Deferring this.

nicocirio:
Re: **AI Review — requirements** (AC-005, aggregate_weighted_proficiency_test.exs:44)

This test is deliberately narrow — see its own comment: a zero-count child can't mathematically skew a weighted average, so this documents that trivial fact, not AC-005's guarantee. The test right above it, `"does not silently drop a Sub-LO with insufficient (but nonzero) evidence"` (line 26), is the one that proves AC-005: `{1.0, 10}` vs. `{1.0, 10}, {0.0, 1}` produce different results, showing thin-but-real evidence isn't dropped. No change needed here.

nicocirio:
Re: **AI Review — elixir** (LKT-AOA v2 evidence weight, metrics.ex:1598)

Agreed in the abstract, and already anticipated: the `@doc` right above `aggregate_weighted_proficiency/1` explicitly documents this exact naive-vs-LKT-AOA count distinction as a caller-responsibility contract. There's no LKT-AOA/v2 caller anywhere in this codebase yet — the only real caller feeds naive `num_first_attempts` counts exclusively. Adding an `evidence_count` rename and a v2 test now would mean inventing a caller with no corresponding production code. Deferring until an actual v2 caller exists to test against.

docs/exec-plans/current/epics/learning_model_v2/lo_aggregation/parent_evidence_aggregation_options.md
```diff
@@ -0,0 +1,125 @@
+# Parent LO Proficiency: How Should a Parent's Own Directly-Tagged Evidence Be Combined With Its Sub-LOs?
```
gastonabella, line 1: Nice doc 👏

gastonabella (approved):
Looks good! Really thorough documentation and traceability on the aggregation decisions here.

One small ask before merging: since the "Option B" behavior (always combine parent + Sub-LO evidence, accepting double-counting) was confirmed over Slack rather than in a Jira comment, could we add a short comment on MER-5821 linking to `parent_evidence_aggregation_options.md` and summarizing the final decision? The Jira AC text still describes equal-weighting and no-double-counting, which contradicts what ships here — future readers of the ticket (without cloning the repo) would get the wrong idea.

PR #6820, opened 2026-09-02
[ENHANCEMENT] [MER-5797] Improve Learning Objectives search and toolbar layout

rgachuhi:
## Summary

Improves the authoring Learning Objectives page by supporting search across the full objective hierarchy and reorganizing toolbar actions.

## Changes

- Search across:
    - Learning Objective titles
    - Sub-Objective titles
    - Page titles
    - Activity titles

- Automatically expand matching objective and sub-objective hierarchy levels.
- Highlight matching search terms.
- Add an accessible empty state when no results are found.
- Group search and sort controls separately from Download CSV and New Objective actions.
- Preserve existing sorting, URL state, CSV parameters, and objective mutation behavior.
- Use the existing ObjectiveCoverage in-memory model without additional database queries.


lib/oli_web/live/workspaces/course_author/objectives_live.ex
```diff
@@ -96,66 +133,70 @@ defmodule OliWeb.Workspaces.CourseAuthor.ObjectivesLive do
...
+      <div class="flex flex-wrap items-center justify-between gap-3">
+        <div class="flex min-w-0 flex-1 flex-wrap items-center gap-3">
+          <div class="w-full sm:max-w-md">
+            <Filter.render
+              change="change_search"
+              reset="reset_search"
+              apply="apply_search"
+              query={@query}
+              apply_icon={true}
+            />
```
nicocirio: Now that this PR wires up real hierarchical auto-expand, I think it's worth reconsidering the input itself: `Filter.render(apply_icon: true)` requires an explicit click on the magnifying glass for anything (filtering *and* auto-expand) to happen — typing, blurring the field, and even pressing Enter all do nothing (there's no wrapping `<form>`/`phx-submit`, so Enter has no bound action). It's also not quite what the Figma shows (no inline "×" there). `OliWeb.Common.SearchInput` (already used in ~20 other views, including the analogous delivery-side Learning Objectives search) ships with `phx-debounce="300"` built in and matches the design more closely, so we'd get real search-as-you-type, and if the debounced change handler does what `apply_search` does today, it also naturally fixes the "clearing via typing vs. the × button" inconsistency, since there'd be one code path instead of three. I tried this locally and it works well:

```elixir
<.form for={%{}} id="objectives-search-form" phx-change="apply_search">
  <SearchInput.render
    id="objectives-search"
    name="query"
    text={@query}
    placeholder="Search..."
    aria_label="Search learning objectives, sub-objectives, pages, and activities"
    class="w-56"
  />
</.form>
```

with `apply_search` folding in what it already computes today (`matching_ids` + `expand_search_results/3`), and the empty-query branch of `expand_search_results/3` pruning `search_expanded_objective_slugs` instead of no-oping, so clearing the field this way is handled identically to before.

### Before

https://github.com/user-attachments/assets/fbe6b1db-4a09-4d0e-984b-6fe1b5e1642a


### After

https://github.com/user-attachments/assets/cea7defb-35fd-470f-b7c7-50d1ec088e84

lib/oli_web/live/workspaces/course_author/objectives_live.ex
```diff
@@ -293,11 +338,97 @@ defmodule OliWeb.Workspaces.CourseAuthor.ObjectivesLive do
...
+        else
+          auto_expanded =
+            socket.assigns.objectives
+            |> Enum.flat_map(fn objective ->
+              child_ids =
+                objective.children
+                |> Enum.reject(&is_nil/1)
+                |> Enum.map(& &1.resource_id)
+
+              if objective.resource_id in matching_ids or
```
nicocirio: Small UX question, not a blocker: the ticket only asks to auto-expand a Learning Objective when needed to reveal a matching Sub-Objective/page/activity underneath it. Right now `expand_search_results/3` also expands the objective when *only* its own title matches (nothing nested matches), showing its full Sub-Objective/page/activity list even though none of it matched the term. Might be worth only expanding when there's an actual nested match — but happy to leave it as-is if that's the intended behavior.

Image

lib/oli_web/live/workspaces/course_author/objectives_live.ex
```diff
@@ -96,66 +133,70 @@ defmodule OliWeb.Workspaces.CourseAuthor.ObjectivesLive do
...
+              class="h-9 rounded-[3px] border border-Border-border-default bg-Background-bg-primary px-2 text-sm font-semibold text-Text-text-high focus-visible:outline focus-visible:outline-2 focus-visible:outline-offset-2 focus-visible:outline-Fill-Buttons-fill-primary"
+            >
+              <%= for column_spec <- @table_model.column_specs do %>
+                <%= if column_spec.name != :action do %>
+                  <option value={column_spec.name} selected={@table_model.sort_by_spec == column_spec}>
+                    {column_spec.label}
+                  </option>
+                <% end %>
+              <% end %>
+            </select>
```
nicocirio: Minor visual bug: selecting "# of Activity Attachments" in the sort dropdown makes the label text overlap the sort-direction chevron next to it — it's the longest of the four options and the `<select>` doesn't reserve enough width/padding for it. Could we give it a bit more room (e.g. a `min-w-` on the select) so the longest label doesn't crowd the chevron?
Screenshot 2026-09-03 at 5 45 22 PM

lib/oli_web/live/workspaces/course_author/objectives_live.ex
rgachuhi: Implemented recommended approach

nicocirio (approved):

nicocirio:
The **"Typescript build and test"** workflow was failing at the `Configure` step because `master` now uses `node-version-file: .node-version` (introduced in #6829, which bumped Node to `24.20.0`), but this branch didn't have that file yet since it wasn't up to date with `master`.

I merged `master` into this branch to bring in `.node-version` and the related workflow changes, so CI can resolve the Node version correctly. No conflicts.

lib/oli_web/live/workspaces/course_author/objectives_live.ex
rgachuhi: Applied the behaviour to reveal sub-objective only on match

lib/oli_web/live/workspaces/course_author/objectives_live.ex
rgachuhi: Addressed the minor visual bug.

PR #6821, opened 2026-09-02
[BUG FIX] [MER-5885] Rework - Make page attempt snapshot reward evidence authoritative

eliknebel:
## Summary

- make the normal page-attempt snapshot the single authoritative xAPI statement for Thompson Sampling assessment rewards
- attach persisted `AcceptedReward` evidence to that statement instead of emitting a second page-attempt statement
- preserve the pre-native-A/B snapshot and core attempt flow for sections not participating in an active Thompson Sampling experiment
- process Thompson rewards synchronously within learner-attempt finalization, gated by a lightweight active-experiment query
- remove reward mutation from the analytics-only snapshot worker and remove the obsolete evidence-dispatch compatibility worker
- add AppSignal metrics and structured logging for synchronous reward-processing failures

## Problem identified by QA

During manual QA of a Thompson Sampling experiment, ClickHouse appeared to contain duplicate `page_attempt` events for the same learner attempt. One event contained no experiment attribution, while the other contained the Thompson Sampling assessment-reward attribution.

Because ClickHouse identifies raw events by a hash of the complete JSON statement, the attributed and unattributed versions received different event hashes and were retained as separate rows even though they represented the same page-attempt GUID.

## Root cause

Two independent asynchronous paths emitted a `PageAttemptEvaluated` xAPI statement for the same resource attempt:

1. The normal snapshot pipeline emitted the canonical page-attempt statement along with its activity and part statements.
2. After accepting a Thompson Sampling assessment reward, `EvidenceDispatchWorker` constructed and emitted a second page-attempt statement carrying the reward attribution.

The normal snapshot and reward workers were independently scheduled, so the snapshot commonly ran without the accepted reward evidence and emitted an unattributed statement. The reward-evidence worker subsequently emitted the attributed copy. Since these statements had different serialized content, ClickHouse deduplication could not collapse them.

## Final processing model

The standard snapshot pipeline remains responsible only for analytics statement construction and emission. It produces one page-attempt statement and reads any committed `AcceptedReward` evidence when the statement factory builds experiment attributions.

Reward processing now occurs in the same database transaction that evaluates the resource attempt for automatic finalization, manual grading, and auto-submit. Before doing reward work, each path performs a lightweight indexed check for an active Thompson Sampling experiment associated with the section. Sections without an active Thompson experiment retain the pre-native-A/B behavior apart from that gate query.

For participating sections, an accepted-reward claim and Thompson policy-state update commit atomically with attempt evaluation. A reward-processing error fails and rolls back the submission. The accepted-reward uniqueness constraint keeps retries idempotent.

## Production observability

Synchronous failures emit the AppSignal counter `oli.experiments.reward_handoff.failure` with a bounded `reason` tag. They also produce a structured `Thompson reward processing failed` log containing the bounded classification plus section and resource-attempt identifiers for investigation. Those identifiers are deliberately excluded from metric tags.

Unexpected exceptions are reported and then re-raised unchanged, preserving transaction rollback and AppSignal exception capture. The AppSignal runbook documents the dashboard, alert, and investigation workflow.

## Verification

- `mix test test/oli/delivery/attempts/page_lifecycle_test.exs test/oli/delivery/attempts/manual_grading_test.exs test/oli/delivery/attempts/auto_submit/worker_test.exs test/oli/delivery/experiments/telemetry_test.exs test/oli/delivery/experiments/assessment_reward_handoff_test.exs`
- 47 tests, 0 failures
- `mix format`
- `git diff --check`

## Review notes

- Security: attribution and reward data are derived from persisted server-side assignment, binding, reward, and attempt records; no new client input or learner-response exposure
- Performance: non-participating sections execute only the active-Thompson existence query; policy-row locking and reward processing are limited to participating Thompson sections
- Correctness: snapshots remain analytics-only, reward processing is transactional with attempt evaluation, and only one page-attempt statement is emitted
- Operations: failures are countable by bounded reason in AppSignal and individually traceable through structured logs and existing Phoenix/Oban exception instrumentation


lib/oli/delivery/snapshots/worker.ex
```diff
@@ -57,6 +58,12 @@ defmodule Oli.Delivery.Snapshots.Worker do
       )
       |> Repo.all()
 
+    with :ok <- record_assessment_rewards(results) do
+      create_and_emit_snapshot(results, section_slug)
+    end
+  end
```
nicocirio: `record_assessment_rewards/1` now runs inline before we build the snapshot, so any uncaught exception inside `RewardHandoff.record_evaluated_resource_attempt/1` would abort the whole `perform_now/3` call, not just reward processing — losing the "core" `PageAttemptEvaluated` statement for that attempt, not only its attribution. I didn't find a concrete way to trigger that today (PolicyState/AcceptedReward both have DB-level uniqueness and PolicyState is created atomically with its Assignment), so this isn't a live bug, but it's a change in blast radius: before this PR, the same kind of failure only took down the isolated `RewardHandoffWorker` job. `AttemptAttributions.for_attempt_group/1` already established the pattern for this exact situation via its `rescue`/`empty_attributions()` fallback — would it be worth applying the same degrade-gracefully approach here as a preventive measure?

lib/oli/delivery/snapshots/worker.ex
```diff
@@ -95,6 +102,19 @@ defmodule Oli.Delivery.Snapshots.Worker do
...
+      [] ->
+        :ok
+
+      [{_, _, resource_attempt, _, _}] ->
+        RewardHandoff.record_evaluated_resource_attempt(resource_attempt.id)
+
+      _multiple_resource_attempts ->
+        {:error, :snapshot_spans_multiple_resource_attempts}
+    end
+  end
```
nicocirio: `RewardHandoffWorker` is still enqueued independently from `page_lifecycle.ex:295`, `manual_grading.ex:502`, and `auto_submit/worker.ex:65` — the last one is the clearest case, since it calls `RewardHandoffWorker.maybe_enqueue/2` and then, eight lines later, `Snapshots.Worker.perform_now/2` directly, both for the same `resource_attempt_id` in the same transaction. Since both paths end up calling this same idempotent `record_evaluated_resource_attempt/1`, every relevant Thompson Sampling submission now emits reward-outcome telemetry twice (one `:accepted` + one `:duplicate`), which would double-count the `outcome` AppSignal metric this subsystem's runbook documents. Now that the snapshot worker handles this inline, is `RewardHandoffWorker`'s independent enqueue still needed, or should it be removed as part of this same cleanup?

nicocirio: The existing "makes accepted reward evidence available to the authoritative page snapshot" test builds the `AttemptGroup` by hand and calls `StatementFactory.to_statements/1` directly, so it doesn't exercise `Worker.perform_now/3`'s new `record_assessment_rewards/1` step at all. A test that goes through `perform_now/3` end-to-end for an attempt with an active Thompson Sampling binding would give this integration more solid regression coverage, rather than only the statement-building helper.

nicocirio (approved):
This looks good to me — the fix addresses the duplicate page_attempt rows QA reported, and the core mechanism (processing the Thompson Sampling reward synchronously before building the snapshot, so the single emitted statement can carry its attribution) is sound and idempotent on retry.

I left three comments that I think would make this more robust against future bugs, but none of them block this hotfix for me — they're optional hardening, not requirements:
- A defensive fallback around the reward-processing step, so a future bug in the experiments subsystem can't take the core snapshot emission down with it.
- RewardHandoffWorker still runs in parallel with the new synchronous call, which double-counts reward telemetry — worth checking if it's still needed.
- Some end-to-end regression coverage for the new perform_now/3 integration path.

lib/oli/delivery/snapshots/worker.ex
eliknebel: We’re intentionally keeping reward processing as a prerequisite for emitting the authoritative snapshot. Falling back to an unattributed statement would permanently lose experiment evidence while making the snapshot job appear successful. A reward failure therefore fails the snapshot job and lets Oban retry the idempotent handoff. Durable reconciliation/outbox behavior remains explicitly out of scope for MER-5885.

lib/oli/delivery/snapshots/worker.ex
eliknebel: Addressed in c849c9f7d6. Snapshot processing is now the sole owner of assessment reward handoff, and the redundant RewardHandoffWorker plus lifecycle enqueues were removed. Manual grading defers snapshot creation until the outer transaction has finalized the page attempt, preserving the same authoritative path.

eliknebel: Addressed in c849c9f7d6. Added an end-to-end SnapshotWorker.perform_now test that verifies exactly one accepted reward, one posterior update, and one emitted page_attempt statement containing the reward attribution.

eliknebel: Follow-up correction in 28a54cf112: after comparing against v0.34.1, we restored the original manual-grading and ApplyClientEvaluation snapshot flow rather than relocating snapshot creation. We also removed the now-unused reward-worker scaffolding from persistence and page finalization. Reward processing remains layered inside the existing snapshot worker, and the manual-grading test explicitly models the worker’s normal post-commit production timing.

PR #6824, opened 2026-09-02
[FEATURE] [MER-5861] Restrict learning objectives by content type

andersweinstein:
## Jira

[MER-5861: Restrict learning objectives by content type](https://eliterate.atlassian.net/browse/MER-5861)

## Summary

Adds a persisted compatibility state that allows well-formed projects to enforce the intended learning-objective hierarchy while preserving permissive behavior for legacy projects.

## Behavior

- Adds nullable `projects.lo_well_formed`; existing projects remain unclassified and new projects default to `true`.
- Lazily classifies unclassified projects from their working publication when Basic Page, Adaptive Page, or Activity Bank authoring is opened.
- Treats a project as well formed when pages attach only top-level objectives and activities attach only sub-objectives.
- Keeps an existing `true` or `false` classification sticky.
- Preserves the exact `true`, `false`, or `nil` state when cloning.
- Preserves the state through project export and ingest; legacy archives with a missing or null value remain unclassified.
- For well-formed projects, page selectors offer only top-level objectives and continue to support just-in-time top-level objective creation.
- For well-formed projects, activity selectors display top-level objectives as disabled and allow only sub-objective attachment.
- Activity selectors with creation disabled behave as search fields, with clear no-results messaging and transient text cleared on blur.
- Legacy projects retain the existing permissive selector behavior.

## Testing

- Relevant backend suite: 143 tests passing.
- Classifier suite: 9 tests passing.
- Focused objective-selector Jest suites: 10 tests passing.
- `yarn check-types`
- Targeted ESLint and formatting checks.

### DIFF
--- a/lib/oli/authoring/learning_objectives/project_classifier.ex
+++ b/lib/oli/authoring/learning_objectives/project_classifier.ex
@@ -1,0 +1,113 @@
+defmodule Oli.Authoring.LearningObjectives.ProjectClassifier do
+  @moduledoc """
+  Classifies whether a project follows the supported learning-objective attachment structure.
+
+  A well-formed project attaches only top-level objectives to pages and only sub-objectives
+  to activities. Existing classifications are sticky; only projects whose classification is
+  `nil` are inspected and updated.
+  """
+
+  import Ecto.Query, warn: false
+
+  alias Oli.Authoring.Course.Project
+  alias Oli.Publishing.PublishedResource
+  alias Oli.Repo
+  alias Oli.Resources.{ResourceType, Revision}
+
+  @objective_type_id ResourceType.id_for_objective()
+  @page_type_id ResourceType.id_for_page()
+  @activity_type_id ResourceType.id_for_activity()
+  @classified_resource_type_ids [@objective_type_id, @page_type_id, @activity_type_id]
+
+  @doc """
+  Returns a project's existing classification, or computes and persists it when currently nil.
+  """
+  def ensure_classified(%Project{lo_well_formed: value}, _publication_id)
+      when is_boolean(value),
+      do: {:ok, value}
+
+  def ensure_classified(%Project{id: project_id}, publication_id) do
+    value = publication_id |> classification_data() |> well_formed?()
+
+    case persist_if_unclassified(project_id, value) do
+      :updated -> {:ok, value}
+      :already_classified -> current_classification(project_id)
+    end
+  end
+
+  @doc """
+  Determines whether classification data follows the learning-objective attachment rules.
+
+  References to objectives absent from the current publication make the project malformed,
+  because they cannot be verified as valid for either attachment type.
+  """
+  def well_formed?(rows) do
+    objective_rows = Enum.filter(rows, &(&1.resource_type_id == @objective_type_id))
+
+    objective_ids = objective_rows |> Enum.map(& &1.resource_id) |> MapSet.new()
+
+    sub_objective_ids =
+      objective_rows
+      |> Enum.flat_map(&(&1.children || []))
+      |> MapSet.new()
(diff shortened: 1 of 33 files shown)


andersweinstein:
[codex:]

- **Performance — classification scan/stampede:** Classification is a one-time, projected query for previously unclassified projects; subsequent loads use the persisted value. The concurrent-first-load case is narrow and unmeasured, so locking/background-job complexity is deferred pending performance evidence.
- **TypeScript/UI — hierarchy flag forwarding:** Both activity selectors already forward `loWellFormed` through `{...props}`; the explicitly listed props do not override it.
- **Elixir — absent archive attribute:** This is the archive-specific creation path. A missing or null value intentionally becomes `nil` so legacy imports are lazily classified; ordinary project creation retains the `true` default.
- **Elixir — missing migration:** The migration exists at `priv/repo/migrations/20260902160000_add_lo_well_formed_to_projects.exs`; it preserves existing rows as null and sets the default for new rows to true.
- **Security — imported compatibility state:** This field controls authoring compatibility behavior, not authorization or tenant access. Importers can already define the imported objective structure, and preserving the boolean supports export/ingest round trips; this does not expand their authority.

lib/oli/interop/ingest/processor/project.ex
```diff
@@ -48,7 +48,8 @@ defmodule Oli.Interop.Ingest.Processor.Project do
           legacy_svn_root: Map.get(project_details, "svnRoot"),
           attributes: Map.get(project_details, "attributes"),
           welcome_title: Map.get(project_details, "welcomeTitle"),
-          encouraging_subtitle: Map.get(project_details, "encouragingSubtitle")
+          encouraging_subtitle: Map.get(project_details, "encouragingSubtitle"),
+          lo_well_formed: Map.get(project_details, "loWellFormed")
```
manelli, line 52: Should we validate `loWellFormed` before passing it to the trusted changeset? This value comes directly from archive JSON: Ecto coerces values such as `"false"`, while objects or arrays produce an invalid changeset and cause the success-only match above to raise. We could explicitly accept only `true`, `false`, and `nil`, returning a descriptive ingest error otherwise.
Let me know what you think

lib/oli/authoring/learning_objectives/project_classifier.ex
```diff
@@ -0,0 +1,124 @@
...
+
+  def ensure_classified(
+        %Project{id: project_id},
+        %Publication{id: publication_id, project_id: project_id}
+      ) do
+    value = publication_id |> classification_data() |> well_formed?()
+
+    case persist_if_unclassified(project_id, value) do
+      :updated -> {:ok, value}
+      :already_classified -> current_classification(project_id)
```
manelli, line 43: Is this classification refreshed somewhere when objectives are edited?

manelli (commented):
Looks good, just left some questions.

lib/oli/authoring/learning_objectives/project_classifier.ex
andersweinstein: Looked to me like the design was explicitly for one-time "sticky" classification: Once set to well-formed, the authoring changes in the PR should prevent any LO attachment that breaks well-formedness, while legacy projects that are non-well-formed get flagged as such and remain permissive.

lib/oli/interop/ingest/processor/project.ex
andersweinstein: That's a nicety that would be easy to do, but I didn't think it was important. I think ingestion of malformed archives often throws rather than validate everything to produce nice error message. Ingestion is an admin function, if the archive has been messed up, you take what you get. But maybe this has changed over time and it has been hardened.

manelli (approved):
Looks great! 🚀

PR #6825, opened 2026-09-03
[BUG FIX] [MER-5823] Student LO proficiency view shows deleted objectives

simonchoxx:
[MER-5823](https://eliterate.atlassian.net/browse/MER-5823)

## Summary

This PR fixes delivery learning objective filtering so deleted or no-longer-contained objectives are not shown in current course views.

**Behavior change:** the criterion is containment, not the `deleted` flag. A live objective that is not attached to any page or activity is also hidden. This affects the CSV export and the instructor dashboard.

### Changes

- Centralized filtering in `Sections.get_objectives_and_subobjectives/2`.
- Excludes objectives that still have a `section_resource` but are no longer actively contained by pages or activities.
- Preserves a parent objective when at least one of its subobjectives is still contained.
- Removed duplicate instructor dashboard filtering.
- Subobjective rows are filtered at expansion time rather than by pruning `children` on the row, so callers that resolve subobjectives by id from that field keep working.
- Added coverage for delivery context behavior, learning objectives CSV export, and the individual student dashboard.

### Tests

```
mix test test/oli/delivery/sections_test.exs \
  test/oli_web/controllers/delivery_controller_test.exs \
  test/oli_web/live/delivery/student_dashboard/components/learning_objectives_tab_test.exs \
  test/oli_web/components/delivery/learning_objectives/ \
  test/oli_web/live/delivery/instructor_dashboard/insights/
```

Result: **309 tests, 0 failures**

### How to verify

Open both views of an affected section and compare: Insights → Objectives and Overview → Students → *(student)* → Objectives should show the same set. They currently do not match.

This only reproduces in sections created or rebuilt after the content was deleted, because `create_nonstructural_section_resources` does not filter `deleted` revisions. That is why it does not happen in every course.

[MER-5823]: https://eliterate.atlassian.net/browse/MER-5823?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/oli/delivery/sections.ex
+++ b/lib/oli/delivery/sections.ex
@@ -5973,13 +5973,15 @@ defmodule Oli.Delivery.Sections do
         }
       end)
 
+    # Only the two containment fields are needed, and both consumers of `container_ids` are
+    # membership checks, so the grouped lists are prepended and never reversed.
     objective_to_container_ids_map =
       from(co in ContainedObjective)
       |> where([co], co.section_id == ^section.id)
-      |> select([co], co)
+      |> select([co], {co.objective_id, co.container_id})
       |> Repo.all()
-      |> Enum.reduce(%{}, fn co, acc ->
-        Map.update(acc, co.objective_id, [co.container_id], &(&1 ++ [co.container_id]))
+      |> Enum.reduce(%{}, fn {objective_id, container_id}, acc ->
+        Map.update(acc, objective_id, [container_id], &[container_id | &1])
       end)
 
     # Section resources outlive the content that referenced them, so containment - not the
--- a/lib/oli/delivery/sections.ex
+++ b/lib/oli/delivery/sections.ex
@@ -5905,6 +5905,9 @@ defmodule Oli.Delivery.Sections do
   top level objectives, they appear with their proficiency for only those activities that
   directly attached to them.
 
+  Only objectives contained by the currently delivered course content are returned, so that the
+  instructor dashboard, the student dashboard and the CSV export report on the same objective set.
+
   ## Options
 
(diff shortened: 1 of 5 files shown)


simonchoxx:
> ### AI Review — ui
> ### Nested objectives now leak into the dashboard
> file: lib/oli_web/live/delivery/instructor_dashboard/instructor_dashboard_live.ex line: 123 Description: Removing `filter_root_contained_objectives/1` causes objectives contained only in nested resources to be displayed alongside root-contained objectives, potentially producing duplicate or misleading dashboard rows. Suggestion: Restore the filtering step, or apply equivalent root-container filtering in the data-loading context before assigning the objectives.

The filter wasn't removed, it was moved into Sections.get_objectives_and_subobjectives/2 — which is what this suggestion asks for. It now applies to all three consumers (instructor dashboard, student dashboard, CSV export); previously only the dashboard had it.

Dashboard behavior is unchanged: with this view's exact options our version returns 8 rows vs master's 9, the extra one being an objective with no active containment.

Francisco-Castro (approved):
Approve. Suite green, fix reproduced on all three surfaces.

### Two non-blocking notes:

1. Expanded panel (`expanded_objective_view.ex:312`) still shows uncontained sub-objectives, while the table, CSV and student tab hide them. MER-5788 pinned the panel behaviour on purpose, so worth checking with Eli how to handle the mismatch.

Image

2. The section-resource cleanup half from Eli's comment isn't in this PR and has no ticket. Could you open one and link it to MER-5823?

PR #6827, opened 2026-09-03
[ENHANCEMENT] [MER-5800] Adds Course Content Filter to Learning Objectives Authoring

rgachuhi:
## Summary

Implements MER-5800, adding a Course content filter to the Learning Objectives authoring page.

Authors can select Units, Modules, Sections, or Pages from a hierarchical checklist to filter objectives and sub-objectives by their associated course content.

## Changes

- Added Course content toolbar filter with:
  - Hierarchical curriculum checklist
  - Multiple selection with OR semantics
  - Descendant inclusion for selected parents
  - Active selection count
  - Clear-all behavior
  - Keyboard and ARIA support
  - Title truncation and tooltips
- Added URL persistence through the `course_content` query parameter.
- Preserved search, sorting, Coverage Issues, pagination, and expansion state.
- Extended `ObjectiveCoverage` with:
  - Curriculum selection normalization
  - Page-to-objective indexes
  - Embedded activity objective matching
  - Objective ancestor scope resolution
  - Deterministic curriculum nodes
- Added cycle-safe curriculum rendering for malformed hierarchy data.
- Added explicit tree expansion accessibility metadata.
- Added unit and LiveView coverage for filtering, selection state, URL restoration, composition, and malformed data handling.


lib/oli_web/live/workspaces/course_author/objectives_live.ex
gastonabella, line 216: The "Download CSV" link now carries `course_content` in its query string, but `CsvExport.filtered_and_sorted_objectives/3` never reads that param — only `query`/`sort_by`/`sort_order` are applied. Right now an author who filters by course content and downloads the CSV will silently get the *unfiltered* export. This is called out as an intentional non-goal in the FDD, but since the UI now passes the param through, it reads as if the export respects the filter. Could we either (a) apply the course-content filter to the CSV export in this PR/a fast follow, or (b) confirm with product that this gap is acceptable and consider not passing `course_content` to the download link (or adding a tooltip/note) so authors aren't misled?

lib/oli_web/live/workspaces/course_author/objectives_live.ex
```diff
@@ -343,30 +372,117 @@ defmodule OliWeb.Workspaces.CourseAuthor.ObjectivesLive do
...
+  def before_table_params(params, socket) do
+    case socket.assigns.coverage_model do
+      nil ->
+        assign(socket, course_content_selection: nil, course_content_scope_ids: [])
+
+      model ->
+        selection =
+          ObjectiveCoverage.normalize_curriculum_selection(model, params["course_content"])
+
+        scope_ids = ObjectiveCoverage.objective_scope_for_pages(model, selection.page_ids)
```
gastonabella: `course_content_scope_ids` is computed on every `handle_params/3` call (any sort/page/search event, not just content-filter changes) but I couldn't find it read anywhere — `filter_content_rows/2` uses `selection.objective_ids` instead. Is this intended for a future auto-expand-matching-parents feature (like `search_expansion_ids`), or is it dead code we can drop along with the `objective_scope_for_pages/2` call to avoid the extra ancestor-walk on every table interaction?

rgachuhi: Confirmed it was dead LiveView state. Removed

lib/oli_web/live/workspaces/course_author/objectives_live.ex
rgachuhi: After consulting with product, I applied option a. i.e. course-content filter is applied to CSV export

gastonabella (approved):

PR #6835, opened 2026-09-08
[FEATURE] [MER-5799] Add Learning Objectives coverage issues filter

nicocirio:
## Jira

- Ticket: https://eliterate.atlassian.net/browse/MER-5799
- Includes requested rework from [MER-5794, comment 56877](https://eliterate.atlassian.net/browse/MER-5794?focusedCommentId=56877)

## Summary

Adds a configurable Coverage Issues filter to the Learning Objectives authoring workspace so authors can quickly identify objectives and sub-objectives with insufficient formative or summative activity coverage.

This PR also includes the [rework requested in MER-5794](https://eliterate.atlassian.net/browse/MER-5794?focusedCommentId=56877) to align the summative assignment icon color with the orange treatment defined in Figma.

## What Was Added

- Added project-level formative and summative coverage thresholds, with recommended defaults of three.
- Added coverage issue classification, including propagation of sub-objective issues to parent objectives.
- Added toolbar filtering, affected-objective counts, warning states, and contextual coverage messages.
- Added accessible threshold settings with increment, decrement, persistence, and restore-default actions.
- Included the active coverage filter in CSV exports and preserved its composition with search and sort.
- Made the Learning Objectives toolbar responsive to its container width.
- Corrected the summative assignment icon from blue to orange, addressing the rework requested in MER-5794.
- Added automated coverage for classification, persistence, filtering, UI states, accessibility, and CSV behavior.

## Validation

- [x] `mix format`
- [x] `mix compile`
- [x] `yarn --cwd assets run format`
- [x] `yarn --cwd assets run check-types`
- [x] `yarn --cwd assets run deploy`
- [x] Full `mix test` run completed: 26 doctests and 8,821 tests. Two failures outside this change passed immediately when rerun alone:
  - `mix test test/oli_web/live/delivery/student/learn_live_test.exs:1739`
  - `mix test test/oli/analytics/xapi/clickhouse_uploader_test.exs:305`

## Video

https://github.com/user-attachments/assets/37aabe27-1105-476b-a773-497c22cddc53


nicocirio:
Regarding "Thresholds are not fully bounded": MER-5799 defines recommended defaults of 3 and supports increment/decrement, but it does not specify an upper limit. The lower bound is enforced at 0 in both the UI and server validation. I updated the work-item wording from "bounded" to "non-negative" to match the ticket. If product later wants an upper limit, we can address that in a follow-up ticket instead of introducing an arbitrary cap here.

Addressed in bb214b365c.

nicocirio:
Regarding "Issue state is conveyed by border color alone": the governed Figma design uses warning icons inside the formative/summative pills for direct issues and the red card/row outline for the rolled-up any-issue state; it does not include an additional header warning. I kept this PR aligned with Figma. If product/design wants an additional descendant-only header indicator, we can address that in a follow-up ticket.

The other UI findings in this review—popover keyboard/focus handling and narrow-toolbar reflow—were addressed in bb214b365c.

lib/oli/authoring/course.ex
```diff
@@ -1186,6 +1186,25 @@ defmodule Oli.Authoring.Course do
...
+  their schema defaults.
+  """
+  @spec update_project_attributes(%Project{}, map()) ::
+          {:ok, %Project{}} | {:error, Ecto.Changeset.t()}
+  def update_project_attributes(%Project{} = project, attrs) do
+    attributes = project.attributes || %ProjectAttributes{}
+
+    project
+    |> Ecto.Changeset.change()
+    |> Ecto.Changeset.put_embed(:attributes, ProjectAttributes.changeset(attributes, attrs))
```
simonchoxx, line 1204: I think this covers the case where saving attributes would wipe the other fields, but the Overview form still saves through `Course.update_project/2` without sending the two new thresholds. I tried it locally on a project set to 7/9 and after that save they came back as 3/3. How do you see it?

lib/oli_web/live/workspaces/course_author/objectives_live.ex
```diff
@@ -148,90 +153,186 @@ defmodule OliWeb.Workspaces.CourseAuthor.ObjectivesLive do
...
+              <form
+                id="sort"
+                phx-change="sort"
+                class="flex h-[38px] w-full min-w-0 items-center gap-2 @[520px]:w-auto @[520px]:shrink-0"
+              >
+                <label for="select_sort" class="sr-only">Sort objectives</label>
+                <select
+                  name="sort_by"
+                  id="select_sort"
+                  class="h-[38px] min-w-0 flex-1 rounded-md border border-Border-border-default bg-Background-bg-primary px-[11px] text-[13px] font-semibold leading-[19.5px] text-Text-text-high focus-visible:outline focus-visible:outline-2 focus-visible:outline-offset-2 focus-visible:outline-Fill-Buttons-fill-primary @[520px]:min-w-[210px]"
```
simonchoxx: I think the master merge might have dropped the styling #6827 added to this select (the custom arrow and appearance-none) — this file came up as a conflict in 55ed9e23d5. Could you check whether that was intentional?

lib/oli_web/live/workspaces/course_author/objectives/coverage_settings_popover.ex
```diff
@@ -0,0 +1,190 @@
...
+    """
+  end
+
+  @doc """
+  Builds the `phx-click` JS command that opens the settings popover.
+
+  Also updates the trigger's expanded state and moves focus into the popover.
+  """
+  @spec open_js(String.t(), String.t()) :: JS.t()
+  def open_js(popover_id, trigger_id) do
```
simonchoxx, line 169: Two small doubts about the open/close cycle: a second click on the gear doesn't seem to close the popover (it's inside the click-away wrapper and this only opens), and the Escape handler on line 30 stays active while the popover is hidden, so pressing Escape elsewhere might pull focus onto the gear, if I'm not mistaken.

simonchoxx (commented):
Really nice work 👏
I left three observations — let me know what you think. I could be missing something on any of them, so tell me how you see it.

lib/oli/authoring/objective_coverage/csv_export.ex
rgachuhi, line 151: From Codex local review:
direct_ids contains integer objective IDs, but objective.children contains child objective maps. Therefore these checks compare incompatible values:

MapSet.member?(direct_ids, %{resource_id: 42, ...})

They always return false for child objectives.

Example:

- Parent objective ID: 10
- Child objective ID: 42
- Course-content selection resolves to objective_ids = MapSet.new([42])

The parent is removed because:

MapSet.member?(direct_ids, 10) # false
Enum.any?(children, &MapSet.member?(direct_ids, &1)) # false

As a result, the CSV contains no rows for the selected child. The LiveView path already handles this correctly by checking child.resource_id.

The fix is:

Enum.any?(objective.children, fn child ->
  MapSet.member?(direct_ids, child.resource_id)
end)

and:

Enum.filter(children, fn child ->
  MapSet.member?(direct_ids, child.resource_id)
end)

lib/oli/authoring/course.ex
nicocirio: Nice catch! 🙌

lib/oli/authoring/objective_coverage/csv_export.ex
nicocirio: Thanks for flagging this, but I don't think it applies to this module — `objective.children` here isn't the same shape as in `objectives_live.ex`.

`filter_course_content/3` operates on objectives from `ObjectiveCoverage.objectives(model)`, which returns `model.objectives_by_id` entries built by `normalize_row/1`:

```elixir
children: normalize_ids(Map.get(row, :children)),
```

and `normalize_ids/1` returns a plain list of **integer** resource ids:

```elixir
defp normalize_ids(value) when is_list(value),
  do: value |> Enum.filter(&is_integer/1) |> Enum.uniq() |> Enum.sort()
```

So `Enum.any?(objective.children, &MapSet.member?(direct_ids, &1))` is comparing integers to integers, which is correct. This is different from the LiveView, where `build_objectives/1` enriches each child into a map (`Map.merge(child, base_coverage_fields())`) before filtering — that's presumably what the review compared against.

There's also an existing test covering exactly this parent/child scenario, and it passes on current `master`/this branch:

```
test/oli/authoring/objective_coverage/csv_export_test.exs:138
"applies the course content selection before exporting"
```

it selects a page that belongs to a child objective and asserts the parent row is exported with the child's title.

Applying the suggested fix as-is would actually break this: `child` here is an integer, so `child.resource_id` would raise at runtime, not silently return `false`.

Happy to add a more explicit regression test if it'd help close this out, but I don't think a code change is needed here.

simonchoxx (approved):
LGTM 🚀🚀

PR #6837, opened 2026-09-08
[FEATURE] [MER-5845] Learning proficiency UI

darrensiegel:
**NOTE: It probably makes sense to allow https://github.com/Simon-Initiative/oli-torus/pull/6834 to merge first before reviewing and merging this one.  Then we can update this branch so that the reviewer only sees the true differences (right now you are seeing that PRs changes and this ones**

Implements the UI that allows an Admin user to move existing projects forward so that they can begin using the new LKT_AOA framework. 

A key change had to be made relative to the description of the ticket: I realized that we also needed to edit the `learning_model_version` attribute on **all** templates of a project in addition to the `learning_model_version` of the project.  If we did not, pre-existing templates would be permanently stuck on `:naive`.

### DIFF
--- a/lib/oli/authoring/course.ex
+++ b/lib/oli/authoring/course.ex
@@ -24,6 +24,7 @@ defmodule Oli.Authoring.Course do
   alias Oli.Repo.{Paging, Sorting}
   alias Oli.Resources.{ResourceType, Revision, ScoringStrategy}
   alias Oli.Delivery.Sections.SectionsProjectsPublications
+  alias Oli.Delivery.Sections.Section
   alias Oli.Publishing.PublishedResource
 
   def create_project_resource(attrs) do
@@ -1062,9 +1063,11 @@ defmodule Oli.Authoring.Course do
     create_and_attach_resource(project, attrs)
   end
 
+  @doc "Creates a new project using the latest learning proficiency framework."
   def create_project(attrs) do
     %Project{}
     |> Project.changeset(attrs)
+    |> Project.trusted_learning_model_changeset(%{learning_model_version: :lkt_aoa})
     |> Repo.insert()
   end
 
@@ -1085,8 +1088,9 @@ defmodule Oli.Authoring.Course do
     |> Repo.insert()
   end
 
+  @doc "Creates a new project and its initial content using LKT-AOA by default."
   def create_project(title, author, additional_attrs \\ %{}) do
-    do_create_project(title, author, additional_attrs, nil, :use_default)
+    do_create_project(title, author, additional_attrs, :lkt_aoa, :use_default)
   end
 
   @doc """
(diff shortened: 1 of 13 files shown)


lib/oli/authoring/course.ex
```diff
@@ -1186,6 +1190,50 @@ defmodule Oli.Authoring.Course do
...
+
+  Only project authors and administrators with project access may upgrade. The
+  conditional update prevents stale requests from changing an already upgraded
+  project. Templates are updated in one bulk query in the same transaction.
+  Existing course sections, content, and learner proficiency remain unchanged.
+  """
+  @spec upgrade_learning_model(%Project{}, %Author{}) ::
+          {:ok, %Project{}} | {:error, :not_authorized | :not_upgradable}
+  def upgrade_learning_model(%Project{} = project, %Author{} = author) do
+    case Accounts.can_access?(author, project) do
```
nicocirio, line 1204: One thing to keep in mind is that this check allows admins and project authors to change this configuration. In the current model, project authors include both project owners and collaborators. I’m calling this out because the acceptance criteria do not clarify whether collaborators should also have this permission.

lib/oli/learning_model/parameter_csv.ex
```diff
@@ -0,0 +1,420 @@
...
+  alias Oli.LearningModel.{Parameters, PartIds}
+  alias Oli.LearningModel.V2.{ActivityParameters, LearningObjectiveParameters, PartParameters}
+  alias Oli.Publishing
+  alias Oli.Publishing.{PublishedResource, Publications.Publication}
+  alias Oli.Repo
+  alias Oli.Resources.{ResourceType, Revision}
+
+  @batch_size 250
+
+  @headers ["title", "resource_id", "children_ids", "beta_lo", "beta_difficulty"]
```
nicocirio, line 23: While an activity uses `beta_difficulty` and an LO uses `beta_lo`, it may still be helpful to add a column that explicitly identifies the resource type (`activity` or `learning objective`). That would make the CSV easier to interpret when editing it outside the application.

In particular, the project I tested had several activities titled “Multiple Choice”, so it was not easy to tell which specific MCQ I was setting `beta_difficulty` values for. I am not sure what the best solution is here, but this seems like a difficulty users may face when working with the CSV.

This becomes even more important for activities with multiple independently scored parts. For example, Likert items and Multi Input blanks are represented by opaque part IDs in `beta_difficulty`, without enough context to identify which row, question, or input each value applies to. For example, a Likert activity may export `{"1070794608":0.0,"1163375187":0.0,"4039622412":0.0}`, while a Multi Input activity may export `{"1":0.0,"3660145108":0.0}`.

image


As follow-up ideas, it may be worth including some additional read-only context, such as the activity type, the page location(s) where the activity is referenced, a short activity/prompt preview and, where applicable, labels or previews for the individual scored parts, or a link/slug that lets the author open the corresponding activity in authoring. Since an activity may be referenced from more than one page, any page-location field should support multiple values.

lib/oli/authoring/course.ex
```diff
@@ -1186,6 +1190,50 @@ defmodule Oli.Authoring.Course do
...
+
+        Repo.transaction(fn ->
+          updates = [
+            learning_model_version: :lkt_aoa,
+            updated_at: DateTime.utc_now() |> DateTime.truncate(:second)
+          ]
+
+          case Repo.update_all(query, set: updates) do
+            {1, [updated]} ->
+              from(s in Section,
```
nicocirio, line 1223: Regarding the bot comment, I agree that this race is possible, but I think the likelihood is very low in practice: it requires one author to be creating a blueprint while another author upgrades the framework for the same project.

The interleaving would be:

1. Author A starts creating a blueprint and reads the project while it is still `:naive`.
2. Author B confirms the framework upgrade, which updates the project and all blueprints that already exist to `:lkt_aoa`.
3. Author A’s creation continues and persists the new blueprint with the previously read `:naive` value.

This would leave a legacy blueprint in an upgraded project. I’m calling it out so you can judge whether this concurrency window is likely enough in practice to warrant addressing now.

lib/oli/learning_model/parameter_csv.ex
```diff
@@ -0,0 +1,420 @@
...
+        values = Map.new(parts, fn {id, parameters} -> {id, parameters.beta_difficulty} end)
+        [safe_title(record.title), record.resource_id, "", "", Jason.encode!(values)]
+    end
+  end
+
+  defp import_batch(project, publication, author, batch, acc) do
+    {rows, seen} = Enum.map_reduce(batch, acc.seen, &parse_row/2)
+    resource_ids = Enum.map(rows, & &1.resource_id)
+
+    # Acquire mapping locks in resource order before reading parameter projections.
```
nicocirio, line 240: Regarding the bot comment, I agree that this lock ordering can theoretically deadlock, although it seems very unlikely in practice. It would require two content admins to upload CSVs for the same project at nearly the same time, with more than one batch of overlapping resources arranged in opposing batch orders. Since the download is already ordered by `resource_id`, this would also require at least one of them to have reordered the CSV rows before uploading it.

Given how narrow this concurrency window is, I don’t think we need to add coordination logic to handle this case at this time.

nicocirio (approved):
I tested this locally and the main flows behaved as expected: upgrading the framework migrated the project and its associated templates, and sections subsequently created from those sources inherited the new framework correctly. I also exported and imported the parameter CSV successfully; intentionally corrupting the CSV before import correctly produced an error.

I did find some difficulty identifying which specific activity or learning objective I was editing from the CSV information alone, so I left a few follow-up ideas inline.

I’m approving this, feel free to merge it as is.

PR #6838, opened 2026-09-08
[ENHANCEMENT] [MER-5859] Retain legacy project descriptions

manelli:
- Enforces the 300-character description limit in authoring forms for new projects, sections, and templates.
- Grandfathers existing descriptions longer than 300 characters so they remain editable.
- Re-enables the limit once a grandfathered description is shortened to 300 characters or fewer.
- Removes schema-level validation to prevent long legacy descriptions from breaking ingestion, section creation, and template duplication.


andersweinstein (approved):
Looks good, thanks!

PR #6841, opened 2026-09-09
[ENHANCEMENT] [MER-5928] Fix LO Summary/Introduction Mobile Layout

gastonabella:
## Summary

Closes out the mobile responsive work left after #6792 and #6816 for the LO Summary/Introduction components: aligns the shared proficiency 2×2 grid, the Explain card layout (icon size, title/subtitle spacing, button width/alignment), and desktop-only Explain card padding with the Figma mobile designs.

## Changes

- Proficiency cards (`What is proficiency and how is it estimated?`) now render as a 2×2 grid on mobile instead of stacking in a single column.
- Explain card on mobile: smaller AI icon (top-left), title/subtitle in a right column with proper spacing, `Explain` button left-aligned at 50% width instead of full-width.
- Explain card padding is horizontal-only on non-mobile (vertical padding removed so the card no longer inherits extra height from the 72px icon).
- AI icon reduced from 72px to 62px on non-mobile to better match Figma.
- Revisit/Practice recommendation links reduced to 14px on mobile only (was 16px everywhere).

https://github.com/user-attachments/assets/d9118b98-7db1-4001-a791-6dc4c7d294b5

### DIFF
--- a/assets/css/app.css
+++ b/assets/css/app.css
@@ -496,7 +496,7 @@ https://tailwindcss.com/docs/utility-first
   align-items: center;
   gap: 12px;
   min-width: 0;
-  padding: 12px;
+  padding: 0 12px;
   border-radius: 12px;
   box-shadow: 0 2px 10px 0 rgba(0, 50, 99, 0.05);
 }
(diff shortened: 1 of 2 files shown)


simonchoxx (commented):
I think the order of the cards in the bottom row of `"What is proficiency and how is it estimated?`" doesn't match what Figma shows. 

Should we check with Jess to know which order we should keep?

gastonabella:
> I think the order of the cards in the bottom row of `"What is proficiency and how is it estimated?`" doesn't match what Figma shows.
> 
> Should we check with Jess to know which order we should keep?

@simonchoxx Good catch! I think we should stick to what is included in the Figma designs. I've just pushed the changes.

simonchoxx (approved):
LGTM 🚀

PR #6849, opened 2026-09-10
[FEATURE] [MER-5814] Replace expanded LO chart with HEEx Student Distribution matrix (PR 1/2)

nicocirio:
## Jira

- Ticket: https://eliterate.atlassian.net/browse/MER-5814

## Summary

Replaces the one-dimensional proficiency chart in the expanded Learning Objective row of the instructor dashboard with a two-dimensional Student Distribution matrix (proficiency x activity completion), so instructors can see at a glance which students need support, are excelling, or have limited activity for a given LO.

This is PR 1 of 2 for this ticket. This PR ships the chart and group-selection state only (no student table, no selection/email flow yet).

## What Was Added

- New `Oli.Delivery.Metrics.StudentDistributionGroup` domain module classifying students into `needs_support` / `excelling` / `limited_activity` groups.
- Widened linked-activity denominator in `ExpandedObjectiveView` via `SectionResourceDepot.objectives_with_effective_children_for/2`, so activity completion reflects all activities linked to the objective (not just a single one).
- New stateless `StudentDistributionMatrix` component: a pure HEEx/SVG chart (no React, no charting library) replacing the old `DotDistributionChart` React component (deleted, along with its Jest test).
- Renamed LiveView events to `select_student_group` / `deselect_student_group`, with click and keyboard (Enter/Space) support on each region.
- New `StudentDistributionMatrixLabels` JS hook that fades a region's count label when it overlaps student dots under the cursor.
- New `Graph-*` design tokens (dot colors per proficiency level x active/inactive, region backgrounds) with light/dark values, sourced from Figma.

## Validation

- `mix compile --warnings-as-errors`: clean
- `mix test`: passing (one unrelated, pre-existing flaky test in `discussions_live_test.exs` also fails on isolated reruns unrelated to this change)
- `yarn --cwd assets run check-types`: passing
- Manual visual comparison against Figma (colors, spacing, typography, dot placement, badges)

## Demo

https://github.com/user-attachments/assets/9743ad48-fa37-4c7b-9027-a25bee20b821

> Note: the data shown in the chart is dummy data seeded in the DB for visual QA purposes, so it won't match the proficiency bar shown in the collapsed/expanded LO row — that's expected for this demo, not a bug.
>
> The count label above each of the three zones (Needs Support / Excelling / Limited Activity) turns semi-transparent on hover **only when it overlaps a student dot behind it**, so you can still see the dot underneath. Labels with no dots behind them are unaffected.
>
> All three zones are keyboard-accessible: tab into a zone and press Enter to select/deselect it, same as clicking.


nicocirio:
Replying to the [performance review comment](https://github.com/Simon-Initiative/oli-torus/pull/6849#issuecomment-5624476494) re: one SVG node per student.

This isn't something the ticket requires. During refinement we discussed that if usage in production shows this need, we'll evaluate alternatives for rendering overlapping dots at that point.

nicocirio:
Replying to the [ui review comment](https://github.com/Simon-Initiative/oli-torus/pull/6849#issuecomment-5624478507) (both the overlap-cache-on-resize and the unthrottled-mousemove findings).

Responsiveness will be addressed in a future PR, once the table is added next to the chart — that's when we'll tackle resizing behavior (including overlap-cache invalidation on resize) end to end.

nicocirio:
Replying to the [typescript review comment](https://github.com/Simon-Initiative/oli-torus/pull/6849#issuecomment-5624478621) (same root cause as the ui review's overlap-cache finding).

Responsiveness will be addressed in a future PR, once the table is added next to the chart — that's when we'll tackle resizing behavior (including overlap-cache invalidation on resize) end to end.

nicocirio:
Replying to the [requirements review comment](https://github.com/Simon-Initiative/oli-torus/pull/6849#issuecomment-5624481220).

- Space-key activation: fixed, added a test in 9682957953 mirroring the existing Enter test.
- Execution record vs. FDD telemetry contradiction: fixed in the same commit — the early scope/implementation bullets in `phase_1_execution_record.md` were stale from before telemetry was removed; the record's "Telemetry Removed" section already had the correct account.

assets/src/hooks/student_distribution_matrix_labels.ts
```diff
@@ -0,0 +1,116 @@
+import type { Hook } from 'phoenix_live_view/assets/js/types/view_hook';
```
gastonabella, line 1: Now that `DotDistributionChart_test.tsx` is gone, this hook (`assets/src/hooks/student_distribution_matrix_labels.ts`) is the only client-side behavior left for the matrix. Could we add a Jest test for it, mocking `getBoundingClientRect` to cover mount-time overlap detection, hover fade/clear, `updated()` recomputation vs. skip, and listener cleanup on `destroyed()`? `assets/test/hooks/preview_alternatives_tabs_test.ts` looks like a reasonable pattern to follow.

lib/oli_web/components/delivery/learning_objectives/expanded_objective_view.ex
```diff
@@ -171,12 +165,26 @@ defmodule OliWeb.Components.Delivery.LearningObjectives.ExpandedObjectiveView do
     end
   end
 
-  def handle_event("show_students_list", %{"proficiency_level" => proficiency_level}, socket) do
-    {:noreply, assign(socket, selected_proficiency_level: proficiency_level)}
+  # Fired by StudentDistributionMatrix's region phx-click/phx-keydown bindings. A phx-keydown
+  # payload also carries a "key" field; only Enter and Space should activate the region so
+  # unrelated keydown events (e.g. Tab, arrow keys) while a region has focus are ignored.
+  def handle_event("select_student_group", %{"group" => group} = params, socket) do
+    with true <- Map.get(params, "key") in [nil, "Enter", " "],
```
gastonabella, line 172: Small nit, not blocking: the region's `phx-keydown` has no key filter, so any keydown while it's focused (Tab, arrows, etc.) still round-trips to the server — `handle_event` already ignores anything but Enter/Space, so this is just extra socket traffic, not a correctness issue. If it's easy, filtering client-side (e.g. two `phx-key` bindings, or a tiny hook) would cut that down.

gastonabella (commented):
Great work! Tested this locally with a much larger cohort (~180 students, synthetic data) to see how the chart holds up at scale, and two things came up:

1. **Dot density/overlap:** once there are enough students, dots start overlapping enough that individual students become hard to tell apart, and hovering a specific one to get its tooltip gets fiddly too. This is probably a product call (density cap, clustering, virtualization, etc.) rather than something to fix in this PR, but wanted to flag it in case it's worth its own follow-up ticket.
2. **Cursor flicker:** while interacting with a region (clicking it, hovering over dots), the cursor sometimes stops showing the pointer/hand icon after a couple of seconds and falls back to the default arrow, then goes back to pointer later. I don't see a clear consistent trigger yet. Attaching a video below.

https://github.com/user-attachments/assets/8c061940-7e05-49d7-a53f-deeac40c04c4

nicocirio:
> 1. **Dot density/overlap:** once there are enough students, dots start overlapping enough that individual students become hard to tell apart, and hovering a specific one to get its tooltip gets fiddly too. This is probably a product call (density cap, clustering, virtualization, etc.) rather than something to fix in this PR, but wanted to flag it in case it's worth its own follow-up ticket.

Agreed this is a real improvement we'd want to tackle eventually, and agreed it should be its own follow-up ticket. For what it's worth, the current implementation already follows the option Figma specifies for this density scenario ([node 374-7117](https://www.figma.com/design/iVKgFJwC1iKP7jmJBGILOK/Learning-Objectives-Updates?node-id=374-7117&m=dev))

gastonabella (approved):
LGTM! Thanks for making the changes

PR #6853, opened 2026-09-11
[FEATURE] [MER-5812] Instructor: Learning Objective tooltips & filter updates

gastonabella:
## Summary

Updates the Learning Objectives table (Instructor Dashboard → Insights) per MER-5812:

- **Student Proficiency tooltip**: updated copy to match Figma (per-level definitions), added keyboard focus support.
- **Proficiency Distribution tooltip**: restyled to match Figma (color dots, spacing, labels), fixed hover trigger scope (was accidentally activating on row hover instead of just the chart).
- **Proficiency filter**: added the missing "Not Enough Data" option.
- **Confidence column & filter** (new, only shown for `lkt_aoa` sections): icon + info tooltip, positioned between Student Proficiency and Proficiency Distribution; filter placed next to the Proficiency one. Backend surfaces a class-level confidence estimate per objective via `Metrics`/`Proficiency`.
- **Bug fixes**: both the Proficiency and Confidence dropdown filters could show a parent objective that didn't actually match the selected level, because a differently-valued sub-objective incorrectly promoted its parent into view. Both now match only the value actually displayed on the row.
- Accessibility: keyboard focus + ARIA wiring for the info-icon tooltips and the new filter.

https://github.com/user-attachments/assets/03e2603d-14db-43d3-8490-daa122f2aeb4


gastonabella:
Re: [AI Review — performance] `Avoid materializing all per-student confidence values` (`lib/oli/delivery/sections.ex`)

Deferring this one. The class-wide confidence computation intentionally mirrors the existing `proficiency_dist_for_objectives` pattern a few lines above it (also an in-memory per-student aggregation over an already-fetched, single-query result set), so it's consistent with the established approach in this function rather than a query-in-a-loop issue. Pushing this into a SQL/data-layer aggregate would mean touching the `Oli.Delivery.Proficiency` provider boundary, which is deliberately storage-agnostic — a good follow-up, but better done for both proficiency and confidence together rather than making just one of the two inconsistent with the other in this PR.

gastonabella:
Re: [AI Review — elixir] `Not Enough Data confidence filter cannot match` (`lib/oli_web/components/delivery/learning_objectives/learning_objectives.ex`)

This one doesn't apply — the Confidence filter (`@confidence_options`) only has 3 options: Low/Medium/High. There's no "Not Enough Data" option for Confidence (unlike Proficiency, which does have one per the ticket's AC), so there's no id-4/nil case to map here. Looks like this got cross-wired with the Proficiency filter's id 4. No change needed.

lib/oli_web/components/delivery/learning_objectives/proficiency.ex
```diff
@@ -25,6 +25,38 @@ defmodule OliWeb.Delivery.LearningObjectives.Proficiency do
...
+  def confidence_bar_classes(level) when level in ["Low", "Medium", "High"] do
+    filled = "fill-" <> color_token(level)
+    inactive = "fill-Icon-icon-default"
+
+    case level do
+      "High" -> {filled, filled, filled}
+      "Medium" -> {filled, filled, inactive}
+      "Low" -> {filled, inactive, inactive}
+    end
+  end
```
nicocirio: Interesting approach 🚀  
I would have used pattern matching to select the specific Icons.confidence_bars based on the confidence level value, instead of building a "flexible" icon based on bar classes.

lib/oli_web/live/delivery/instructor_dashboard/html_components.ex
```diff
@@ -281,23 +281,46 @@ defmodule OliWeb.Delivery.InstructorDashboard.HTMLComponents do
...
+
   def render_label(assigns) do
+    assigns = Map.put(assigns, :tooltip_id, assigns[:id] || tooltip_id(assigns.title))
+
     ~H"""
     <div class="flex items-center gap-x-2 bg-transparent">
-      <span class="group relative flex cursor-pointer items-center text-Text-text-high">
+      <span
+        class="group relative flex cursor-pointer items-center text-Text-text-high rounded focus:outline-none focus-visible:ring-2 focus-visible:ring-blue-500 dark:focus-visible:ring-blue-600 focus-visible:ring-offset-2"
+        tabindex={if assigns[:info_tooltip], do: "0"}
```
nicocirio, line 296: Just keep in mind that these additional focus classes affect other labels from OliWeb.Components.Delivery.Pages.PagesTableModel, such as `students_completion`, `total_attempts` and `avg_score`.
I think this is the right approach, but I am just pointing this out in case you find a reason to avoid implementing the change in those cases as well.

nicocirio (approved):
Nice work! 🚀 
There is a failing test that needs to be fixed; then the PR will be ready to be merged

lib/oli_web/components/delivery/learning_objectives/proficiency.ex
gastonabella: Good call, this was overkill for just 3 fixed states. Simplified it — `confidence_bars/1` now pattern-matches on `level` directly and renders each fully-baked SVG, no more computing a flexible bar-classes tuple elsewhere. Thanks!

PR #6863, opened 2026-09-16
[BUG FIX] [MER-5922] Prevent responsive image resize crashes

kjsegovi:
Resizing images on the canvas with Responsive Layout enabled could crash Advanced Author. Disable those resize handles and show a little blurb on the component side if responsive layout is enabled. This way the author knows that they have it enabled and the platform isn't just denying them the ability to resize the image. 

Replace the zero resize-grid increment with a valid value, preserve vertical resizing for other responsive components, and reject invalid dimensions before saving. Fixed-layout image resizing remains unchanged.


kjsegovi:
not doing the ai reviewer sugggestions as it seemingly defeats the purpose of this fix. We do not want authors to be able to edit the size of the image via drag and drop. We want them to use the component tab with various css calls to ensure accessible lessons.

manelli (approved):

